PR #150, opened 2023-02-16
Change for equality query to use array_contains for atomics

### LINKED ISSUE #107: Change equality query to use `array_contains` for atomics once shredder includes them
After #106 is implemented `array_contains` field will contain values of atomic fields as well as array elements.
If so, equality queries (for everything other that document id, `_id`) should start using that field, to match both atomic and array values similar to how Mongo storage works wrt find operations

See #109 for specification; needs #106 completed before implementing.

maheshrajamani:
**What this PR does**:
Change for equality query to use array_contains for atomics

Fixes #107 

**Checklist**
- [x] Changes manually tested


src/test/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/CountOperationTest.java
```diff
@@ -65,10 +66,12 @@ public void countWithNoFilter() throws Exception {
     @Test
     public void countWithDynamic() throws Exception {
       String collectionReadCql =
-          "SELECT COUNT(key) AS count FROM \"%s\".\"%s\" WHERE query_text_values[?] = ?"
+          "SELECT COUNT(key) AS count FROM \"%s\".\"%s\" WHERE array_contains CONTAINS ?"
```
tatu-at-datastax, line 69: Would it make sense to refer to that `DATA_CONTAINS` constant for `array_contains` (add in a constants class)?

tatu-at-datastax (approved):

src/test/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/CountOperationTest.java
maheshrajamani: Renaming the column in database from array_contains to data_contains?

tatu-at-datastax: sorry no, not yet, but to use constant reference.

But we can do that at a later point.

Issue #153, opened 2023-02-18
Support nested path ("dot notation") for `$unset` operator


tatu-at-datastax:
(note: sub-task of #74 )

Although there are common pieces wrt "dot notation" paths for update operations, there are some differences so I'll first tackle simplest case of `$unset` and probably create separate issues for other operators too -- this helps keep PRs from growing too big.

tatu-at-datastax:
PR #154 ready for review.

PR #176, opened 2023-02-23
Support dot notation for $set

### LINKED ISSUE #166: Support nested path ("dot notation") for $set operator
(note: sub-task of https://github.com/stargate/jsonapi/issues/74 )

With #153 implemented can tackle more complex case of `$set` (where patch conflict failures may play a role).
The additional complexity comes from the need to potentially create "missing" path segments, and possibility of conflicts (trying to create a property in Array (or any non-Object) value).

tatu-at-datastax:
**What this PR does**:

Adds support for "dot notation" (sub-documents/arrays) for "$set" operator (similar to how "$unset" supported via #153)

Fixes #166

**Checklist**
- [ ] Changes manually tested


src/test/java/io/stargate/sgv2/jsonapi/api/v1/FindAndUpdateIntegrationTest.java
```diff
@@ -557,91 +557,6 @@ public void findByIdAndUnset() {
           .body("data.docs[0]", jsonEquals(expected));
     }
 
-    @Test
```
tatu-at-datastax, line 560: Just moved to the next nested section; not removed.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/update/SetOperation.java
```diff
@@ -31,8 +31,10 @@ public static SetOperation construct(ObjectNode args) {
   public boolean updateDocument(ObjectNode doc, UpdateTargetLocator targetLocator) {
     boolean modified = false;
     for (SetAction addition : additions) {
-      JsonNode prev = doc.replace(addition.path, addition.value);
-      modified |= !Objects.equals(prev, addition.value);
+      UpdateTarget target = targetLocator.findOrCreate(doc, addition.path());
+      JsonNode newValue = addition.value();
+      JsonNode oldValue = target.replaceValue(newValue);
+      modified |= !Objects.equals(newValue, oldValue);
```
maheshrajamani, line 37: Does equals handles the order of element? if not need to check the order of the elements incase the set done for object type?

maheshrajamani (approved):

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/update/SetOperation.java
tatu-at-datastax: `JsonNode.equals()` does not consider order, true. But individual `$set` operation never changes order of existing properties so I think this is still safe. To change order, I think removal plus addition (or addition of non-existing) property is needed.

tatu-at-datastax: Actually, I think I should refactor things slightly, to extract comparison method just in case this needs to be reconsidered -- I will create an issue.
I think I see the potential problem: replacing an existing JSON Object but with different field order.

tatu-at-datastax: Created #190 for future follow-up work.

PR #177, opened 2023-02-23
Added Request & Response to API Spec

### LINKED ISSUE #175: SPEC - confirm the general response format and what each command returns
This is a bit of a mess, it is in the HTTP spec when it should prob be in the API spec and each command needs to lock in what it returns.

amorton:
Moved from the HTTP network spec so we can reference in the API spec, needed to lock down the responses

resolves #175

**What this PR does**:

Moved from the HTTP network spec so we can reference in the API spec, needed to lock down the responses

Fixes #175

**Checklist**
- [NA] Changes manually tested
- [ NA] Automated Tests added/updated


docs/jsonapi-spec.textile
```diff
@@ -52,6 +52,104 @@ bc(sample). SELECT sample_usage FROM cql;
...
+
+p. _Syntax:_
+
+bc(syntax).. 
+<response-message> ::=  (errors (<error>)+)?,    
+                        (status (<command-status>)+)?, 
+                        (data <response-data>)?,
+
+<<error> ::= message <ascii-string>, 
+             errorCode <ascii-string>,
```
ivansenic, line 93: Not sure about the error code currently, but I think we don't have it implemented at the moment.

docs/jsonapi-spec.textile
```diff
@@ -363,6 +448,15 @@ h4(#commandFindOptions). find Command Options
 | @pageState@ | ASCII String | The page state of the previous page, when supplied the @find@ command will return the next page from the result set. If unspecified, null, or an empty string the first page is returned.|
 | @skip@ | Positive Integer | Skips the specified number of documents, in sort order, before returning any documents.|
 
+h4(#commandFindResponse). find Command Response
+
+|_. Response Element |_. Description |
+| @data@ | Present with fields : @data@, @count@ and @nextPageState@ which will be @null@ if no further data is available. |
```
ivansenic, line 454: * first should `@data@` be `@docs@` here?
* second, `nextPageState` will be null if there are no more documents, but count and docs should never be null.. These should be empty array and zero in case nothing is found?

docs/jsonapi-spec.textile
```diff
@@ -408,12 +502,33 @@ h4(#commandFindOneAndUpdateOptions). findOneAndUpdate Command Options
 | @returnDocument@ | String Enum | Specifies which document to perform the projection on. If @"before"@ the projection is performed on the document before the update is applied, if @"after"@ the document projection is from the document after the update. Defaults to @"before"@.|
 | @upsert@ | Boolean | When @true@ if no documents match the @filter@ clause the command will create a new _empty_ document and apply the @update@ clause to the empty document. If the @_id@ field is included in the @filter@ the new document will use this @_id@, otherwise a random value will be used see "Upsert Considerations":#considerationsUpsert for details. When false the command will only update a document if one matches the filter. Defaults to @false@.|
 
+h4(#commandFindOneAndUpdateResponse). findOneAndUpdate Command Response
+
+|_. Response Element |_. Description |
+| @data@ | Present with fields : @data@ only |
```
ivansenic, line 508: again `@docs@`? what doc is returned, unclear?

docs/jsonapi-spec.textile
```diff
@@ -408,12 +502,33 @@ h4(#commandFindOneAndUpdateOptions). findOneAndUpdate Command Options
...
 h3(#commandFindOne). findOne Command
 
 h4(#commandFindOneOptions). findOne Command Options
 
 The @findOne@ command does not support any options.
 
+h4(#commandFindOneResponse). findOne Command Response
+
+|_. Response Element |_. Description |
+| @data@ | Present with fields : @docs@ only |
```
ivansenic, line 525: so find one has no count? Is it ok that find commands have different responses?

docs/jsonapi-spec.textile
```diff
@@ -438,6 +571,15 @@ h4(#commandUpdateManyOptions). updateMany Command Options
 |_. Option |_. Type |_. Description |
 | @upsert@ | Boolean | When @true@ if no documents match the @filter@ clause the command will create a new _empty_ document and apply the @update@ clause to the empty document. If the @_id@ field is included in the @filter@ the new document will use this @_id@, otherwise a random value will be used see "Upsert Considerations":#considerationsUpsert for details. When false the command will only update a document if one matches the filter. Defaults to @false@.|
 
+h4(#commandUpdateManyResponse). updateMany Command Response
+
+|_. Response Element |_. Description |
+| @data@ | Not present. |
+| @status@ |Present with fields: @matchedCount@ count of documents that matched the @filter@, @modifiedCount@ count of document updated (may be less than match count), @upsertedId@ if a document was upserted (when @matchCount@ is zero) not present otherwise.|
```
ivansenic, line 578: As I understood idea is to insert a single doc with upsert in case match count is zero? Seems fine :+1:

ivansenic (approved):
Looks good.. I am missing a option that defines if updates/writes in many commands are executed sequentially or not (ordered option or something like this).

docs/jsonapi-spec.textile
amorton: I got the errorCode from examples provided by @tatu-at-datastax

docs/jsonapi-spec.textile
amorton: good catch on the data, and bad writing from my. Intention was only to say @nextPageState@ is null . Fixing.

docs/jsonapi-spec.textile
amorton: fixed docs, the projection is controlled by the @returnDocument@ in the options, will mention that

docs/jsonapi-spec.textile
amorton: We decided to only include fields when there is useful information, so dropped count. Honestly not sure we need count anywhere, it is only the size of the `docs` array.

docs/jsonapi-spec.textile
```diff
@@ -447,6 +589,15 @@ h4(#commandUpdateOneOptions). updateOne Command Options
 |_. Option |_. Type |_. Description |
 | @upsert@ | Boolean | When @true@ if no documents match the @filter@ clause the command will create a new _empty_ document and apply the @update@ clause to the empty document. If the @_id@ field is included in the @filter@ the new document will use this @_id@, otherwise a random value will be used see "Upsert Considerations":#considerationsUpsert for details. When false the command will only update a document if one matches the filter. Defaults to @false@.|
 
+h4(#commandUpdateOneResponse). updateOne Command Response
+
+|_. Response Element |_. Description |
+| @data@ | Not present. |
+| @status@ |Present with fields: @matchedCount@ count of documents that matched the @filter@ (no more than 1), @modifiedCount@ count of document updated (may be less than match count, no more than 1), @upsertedId@ if a document was upserted (when @matchCount@ is zero) otherwise not present. |
```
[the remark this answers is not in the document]
amorton: In a case where the update operations have no work to do, such as removing a field with `$unset` that does not exist in the document , or `$max` to set a field to the max of a existing value or supplied value when existing value is already greater.

docs/jsonapi-spec.textile
tatu-at-datastax: Yes I think it is returned and integration tests verify it, along with `message`.

amorton:
> Looks good.. I am missing a option that defines if updates/writes in many commands are executed sequentially or not (ordered option or something like this).

should be done today (monday my  time)

docs/jsonapi-spec.textile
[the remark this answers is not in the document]
ivansenic: Cool, thanks for clarification. resolving.

PR #180, opened 2023-02-23
run CI on pull request only on code-related changes

ivansenic:
**What this PR does**:
Noticed in https://github.com/stargate/jsonapi/pull/179 that the main CI runs even if no source code changes are made. Fixing that to run only:
* if `src` had changes
* if `pom.xml` had changes (deps updates)
* workflow changed itself

### DIFF
--- a/.github/workflows/continuous-integration.yaml
+++ b/.github/workflows/continuous-integration.yaml
@@ -2,13 +2,19 @@
 name: Continuous Integration
 
 # runs on
-# * pushes and pull requests on the "main"
+# * pushes and pull requests on the "main" (pull request only for specific paths)
 # * manual trigger
 on:
   push:
     branches: [ "main" ]
+
   pull_request:
     branches: [ "main" ]
+    paths:
+      - 'src/**'
+      - 'pom.xml'
+      - '.github/workflows/continuous-integration.yaml'
+
   workflow_dispatch:
 
 


.github/workflows/continuous-integration.yaml
```diff
@@ -2,13 +2,19 @@
 name: Continuous Integration
 
 # runs on
-# * pushes and pull requests on the "main"
+# * pushes and pull requests on the "main" (pull request only for specific paths)
```
jeffreyscarpenter, line 5: why wouldn't we apply the path filters to the push trigger as well?

ivansenic: We could, I just like that there is a clear check-mark on each commit to main.. Should I change?

jeffreyscarpenter: Ok so with this configuration, the workflow would run on each PR that has main as its target (for each commit on the branch of the PR, I guess), and again when the PR is merged to main? If so I think that is fine.

jeffreyscarpenter (approved):

.github/workflows/continuous-integration.yaml
ivansenic: * runs always we push something on main (basically merge a PR)
* runs in PRs for commits that have source code changes

OK?

jeffreyscarpenter: yes, I approved PR

PR #188, opened 2023-02-24
release workflow fixes

### LINKED ISSUE #189: Release workflow does not run
Release workflow did not trigger automatically when I pushed the `v1.0.0-ALPHA-1` tag. When I triggered it manually, there were errors:

```
curl https://www.datastax.com/favicon.ico -o ./src/main/resources/META-INF/branding/favicon.ico
Warning: Failed to create the file 
Warning: ./src/main/resources/META-INF/branding/favicon.ico: No such file or 
Warning: directory
curl: (23) Failure writing output to destination
```

jeffreyscarpenter:
Release workflow is failing because the directory `src/main/resources/META-INF/branding` does not exist. Adding a favicon file so the workflow will run correctly

Fixes #189 

**Checklist**
- [ ] Changes manually tested


.github/workflows/release.yaml
```diff
@@ -8,7 +8,7 @@ name: Release Workflow
 on:
   push:
     tags:
-      - 'v.*'
+      - 'v*'
```
jeffreyscarpenter, line 11: The previous match was wrong, it would match on tags starting with `v.`. Read this: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-including-branches-and-tags.

tatu-at-datastax: So difference b/w regexps and file globs?

jeffreyscarpenter: right

Issue #169, opened 2023-02-22
SPEC - document moreData behaviour for deleteMany and updateMany


amorton:
If there is more rows to delete or update than the server can support we return the number of updates and a `moreData` : `true` status field. 

See #145

Issue #168, opened 2023-02-22
updateOne and updateMany Response fields


kathirsvn:
Currently, updateOne and updateMany Response fields are missing some of the counts required.

Here's a sample response:
`{"status":{"updatedIds":["7505ffec-47a0-4346-8ee3-72fd70ad09e5"]}}`

But it's supposed to have below fields:

| Field |  Description |
| -----|--------------|
| matchedCount | `updateOne` - always 1 in the case of updateOne and a document is found for the given filter <br/>`updateMany` - its the number of docs that matched the criteria |
| modifiedCount | `updateOne` - 1 in the case when a document is found and the modification is applied and 0 if any modification was not needed/applied<br/>`updateMany` - Number of docs the modification has been applied on.|
|upsertedId|  When the `upsert` option is present and set to `true` in the request and no doc is found for the given filter and a new doc is inserted -`the _id of the new doc inserted`<br/><br/>When either the `upsert` option is not present in the request or if it's set to `false`, this field should not be included in the response or `undefined` in other words.<br/><br/>This is applicable for both `updateOne` and `updateMany`|

Aaron Morton - Please use the description of the response data in the specification, not the table above https://github.com/stargate/jsonapi/blob/main/docs/jsonapi-spec.textile#commands

maheshrajamani:
@kathirsvn @amorton Can you confirm on the upsertedId? I thought we discussed this field to have document id if we created a new document, because upsert flag set to true in request and no document found in DB for the given filter condition.

maheshrajamani:
> `updateMany` - its the number of docs that matched the criteria

@amorton Does this count going to be number of records read from the database based on the filter, if limit is 20 we read max of 20 records. If we need documents for the matched criteria we need to run count query separately.

kathirsvn:
> @kathirsvn @amorton Can you confirm on the upsertedId? I thought we discussed this field to have document id if we created a new document, because upsert flag set to true in request and no document found in DB for the given filter condition.

@maheshrajamani yes you are correct. Let me fix the description of `upsertedId`.

amorton:
I have added a note to the description that the source of truth is the API spec, all commands now how a response section see https://github.com/stargate/jsonapi/blob/main/docs/jsonapi-spec.textile#commands

> @amorton Does this count going to be number of records read from the database based on the filter, if limit is 20 we read max of 20 records. If we need documents for the matched criteria we need to run count query separately.

[...]

PR #192, opened 2023-02-24
Update Many command implementation

### LINKED ISSUE #145: Limit number of documents for deleteMany and updateMany commands
We need some sort of limit, set it to 10 for now and make it configurable. 

When done, pass the ticket back and we will update the spec.

maheshrajamani:
**What this PR does**:
Update many command implementation
Response change for UpdateOne and FindOneAndUpdate
Fixes #145 #168 and #169 

**Checklist**
- [X ] Changes manually tested
- [X ] Automated Tests added/updated

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/ReadAndUpdateOperation.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/ReadAndUpdateOperation.java
@@ -70,7 +70,7 @@ public record ReadAndUpdateOperation(
                 findResponse -> {
                   final List<ReadDocument> docs = findResponse.docs();
                   if (upsert() && docs.size() == 0 && matchedCount.get() == 0) {
-                    return Multi.createFrom().item(readOperation().getEmptyDocuments());
+                    return Multi.createFrom().item(readOperation().getNewDocument());
                   } else {
                     // Below conditionality is because we read up to deleteLimit +1 record.
                     if (matchedCount.get() + docs.size() <= updateLimit) {
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/ReadAndUpdateOperation.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/ReadAndUpdateOperation.java
@@ -21,6 +21,18 @@ import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Supplier;
 
+/**
+ * This operation method is used for 3 commands findOneAndUpdate, updateOne and updateMany
+ *
+ * @param commandContext
+ * @param readOperation
+ * @param documentUpdater
+ * @param returnDocumentInResponse
+ * @param returnUpdatedDocument
+ * @param upsert
+ * @param shredder
+ * @param updateLimit
+ */
 public record ReadAndUpdateOperation(
     CommandContext commandContext,
     ReadOperation readOperation,
@@ -41,9 +53,9 @@ public record ReadAndUpdateOperation(
             .uni(
                 () -> new AtomicReference<String>(null),
                 stateRef -> {
-                  Uni<ReadOperation.FindResponse> docsToDelete =
+                  Uni<ReadOperation.FindResponse> docsToUpdate =
                       readOperation().getDocuments(queryExecutor, stateRef.get());
-                  return docsToDelete
+                  return docsToUpdate
                       .onItem()
                       .invoke(findResponse -> stateRef.set(findResponse.pagingState()));
                 })
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/ReadAndUpdateOperation.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/ReadAndUpdateOperation.java
@@ -34,7 +34,7 @@ public record ReadAndUpdateOperation(
 
   @Override
   public Uni<Supplier<CommandResult>> execute(QueryExecutor queryExecutor) {
-    final AtomicBoolean boolenFlag = new AtomicBoolean(false);
+    final AtomicBoolean moreDataFlag = new AtomicBoolean(false);
     final Multi<ReadOperation.FindResponse> findResponses =
         Multi.createBy()
             .repeating()
@@ -68,7 +68,7 @@ public record ReadAndUpdateOperation(
                       int needed = updateLimit - matchedCount.get();
                       matchedCount.addAndGet(needed);
 
-                      boolenFlag.set(true);
+                      moreDataFlag.set(true);
                       return Multi.createFrom()
                           .items(findResponse.docs().subList(0, needed).stream());
                     }
@@ -111,7 +111,11 @@ public record ReadAndUpdateOperation(
         .transform(
             updates ->
                 new UpdateOperationPage(
-                    matchedCount.get(), modifiedCount.get(), updates, returnDocumentInResponse()));
+                    matchedCount.get(),
+                    modifiedCount.get(),
+                    updates,
+                    returnDocumentInResponse(),
+                    moreDataFlag.get()));
   }
 
   private Uni<DocumentId> updatedDocument(
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/ReadAndUpdateOperation.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/ReadAndUpdateOperation.java
@@ -16,8 +16,10 @@ import io.stargate.sgv2.jsonapi.service.shredding.model.DocumentId;
 import io.stargate.sgv2.jsonapi.service.shredding.model.WritableShreddedDocument;
 import io.stargate.sgv2.jsonapi.service.updater.DocumentUpdater;
 import java.util.List;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Supplier;
-import java.util.stream.Stream;
 
 public record ReadAndUpdateOperation(
     CommandContext commandContext,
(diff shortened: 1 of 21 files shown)


src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/ReadAndUpdateOperation.java
```diff
@@ -26,47 +28,94 @@ public record ReadAndUpdateOperation(
...
   public Uni<Supplier<CommandResult>> execute(QueryExecutor queryExecutor) {
-    Uni<ReadOperation.FindResponse> docsToUpate = readOperation().getDocuments(queryExecutor, null);
+    final AtomicBoolean moreDataFlag = new AtomicBoolean(false);
+    final Multi<ReadOperation.FindResponse> findResponses =
+        Multi.createBy()
+            .repeating()
+            .uni(
+                () -> new AtomicReference<String>(null),
+                stateRef -> {
+                  Uni<ReadOperation.FindResponse> docsToDelete =
```
tatu-at-datastax, line 44: Should it be named `docsToDelete` if this is used for other operations (update) too?
Unless this is only for deletions.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/ReadAndUpdateOperation.java
```diff
@@ -16,8 +16,10 @@
 import io.stargate.sgv2.jsonapi.service.shredding.model.WritableShreddedDocument;
 import io.stargate.sgv2.jsonapi.service.updater.DocumentUpdater;
 import java.util.List;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Supplier;
-import java.util.stream.Stream;
 
 public record ReadAndUpdateOperation(
```
tatu-at-datastax, line 24: I think it'd be good to have Javadoc here to explain which operations this is used for: I am guessing it is for update and delete ones but would be good to clarify.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/ReadAndUpdateOperation.java
```diff
@@ -26,47 +28,94 @@ public record ReadAndUpdateOperation(
...
+        findResponses
             .onItem()
             .transformToMulti(
                 findResponse -> {
-                  if (findResponse.docs().isEmpty()) {
-                    if (upsert) {
-                      return Multi.createFrom().item(readOperation().getEmptyDocuments());
+                  final List<ReadDocument> docs = findResponse.docs();
+                  if (upsert() && docs.size() == 0 && matchedCount.get() == 0) {
+                    return Multi.createFrom().item(readOperation().getEmptyDocuments());
```
tatu-at-datastax, line 61: It sounds like this would be resulting in document upsert? But mention of "empty document" would suggest something else -- or is that just a document with `_id` field and nothing else?

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/ReadAndUpdateOperation.java
maheshrajamani: This operation method is used for 3 commands findOneAndUpdate, updateOne and updateMany, Will add javadoc for this

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/UpdateOperationPage.java
```diff
@@ -5,28 +5,36 @@
...
+    final DocumentId[] upsertedId = new DocumentId[1];
     List<JsonNode> updatedDocs = new ArrayList<>(updatedDocuments().size());
     updatedDocuments.forEach(
         update -> {
-          updatedIds.add(update.id());
-          updatedDocs.add(update.document());
+          if (update.upserted()) upsertedId[0] = update.id();
+          if (returnDocs) updatedDocs.add(update.document());
         });
+    Map<CommandStatus, Object> updateStatus = new HashMap<>();
```
tatu-at-datastax, line 29: Could probably be `EnumMap` as key is enum?

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/ReadAndUpdateOperation.java
maheshrajamani: This class is for update, will rename to docsToUpdate. My bad, copy & paste mistake.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/ReadAndUpdateOperation.java
maheshrajamani: The method just creates an empty document, if _id filter is part of filter it adds it too.

src/main/java/io/stargate/sgv2/jsonapi/service/updater/DocumentUpdater.java
```diff
@@ -13,10 +13,14 @@ public static DocumentUpdater construct(UpdateClause updateDef) {
...
 
-  public JsonNode applyUpdates(JsonNode readDocument) {
+  public DocumentUpdaterResponse applyUpdates(JsonNode readDocument) {
     UpdateTargetLocator targetLocator = new UpdateTargetLocator();
     ObjectNode docToUpdate = (ObjectNode) readDocument;
-    updateOperations.forEach(u -> u.updateDocument(docToUpdate, targetLocator));
-    return readDocument;
+    boolean modified = false;
+    for (UpdateOperation updateOperation : updateOperations)
+      modified |= updateOperation.updateDocument(docToUpdate, targetLocator);
```
tatu-at-datastax, line 21: Is this safe? If `modified` is `true`, would this not short-circuit and skip call to `updateOperation.upateDocument(...)` completely?

src/test/java/io/stargate/sgv2/jsonapi/api/v1/UpdateManyIntegrationTest.java
```diff
@@ -0,0 +1,349 @@
...
+          .header(HttpConstants.AUTHENTICATION_TOKEN_HEADER_NAME, getAuthToken())
+          .contentType(ContentType.JSON)
+          .body(json)
+          .when()
+          .post(CollectionResource.BASE_PATH, keyspaceId.asInternal(), collectionName)
+          .then()
+          .statusCode(200)
+          .body("data.docs[0]", jsonEquals(expected));
+
+      cleanUpData();
```
tatu-at-datastax, line 96: Instead of relying every test to clean up after themselves, would it make sense to call this before each test?
Otherwise failing tests could leave garbage for other test methods?

tatu-at-datastax (changes requested):
I think it all looks good, but I am unsure if that calculation of `updated` could lead to operations skipped -- I think it does?

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/ReadAndUpdateOperation.java
maheshrajamani: Will rename to newDocument()

src/main/java/io/stargate/sgv2/jsonapi/service/updater/DocumentUpdater.java
maheshrajamani: Yes, it runs the method before evaluating applying or condition.
The below code should print the numbers 0, 1 ... 9.
`public class Test{
  public static boolean update(int i){
        System.out.println(i);
        return i%2 ==0;
    }
    public static void main(String args[]){
        boolean modified = false;
        for(int i = 0; i < 10; i++){
            modified |= update(i);
       }
  }
}`

src/test/java/io/stargate/sgv2/jsonapi/api/v1/UpdateManyIntegrationTest.java
maheshrajamani: Make sense, Will add @AfterEach to the cleanUp(), so the test doesn't need to call this.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/UpdateOperationPage.java
maheshrajamani: Done

tatu-at-datastax (approved):

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/ReadAndUpdateOperation.java
tatu-at-datastax: np.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/UpdateManyIntegrationTest.java
```diff
@@ -0,0 +1,342 @@
...
+
+    @Test
+    @Order(6)
+    public void updateManyUpsert() {
+      insert(5);
+      String json =
+          """
+                                    {
+                                      "updateMany": {
+                                        "filter" : {"_id": "doc6"},
```
vkarpov15, line 239: Another detail worth confirming: do additional fields in the filter get added to the resulting document on upsert? For example, if this is `"filter" : {"_id": "doc6", "answer": 42}`, does the resulting document have an `answer` property?

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/ReadAndUpdateOperation.java
```diff
@@ -16,57 +16,118 @@
...
     final Uni<List<UpdatedDocument>> updatedDocuments =
-        docsToUpate
+        findResponses
             .onItem()
             .transformToMulti(
                 findResponse -> {
-                  if (findResponse.docs().isEmpty()) {
-                    if (upsert) {
-                      return Multi.createFrom().item(readOperation().getEmptyDocuments());
+                  final List<ReadDocument> docs = findResponse.docs();
```
ivansenic, line 71: :warning: Same issue as for the `deleteMany`.

maheshrajamani: I have explained this in deleteMany PR. Once it's resolved will resolve here.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/UpdateManyIntegrationTest.java
tatu-at-datastax: Good question: I think method "getEmptyDocument(s)" (or renamed as "getNewDocument") was only inserting document id?

maheshrajamani: @vkarpov15 During upsert the getNewDocument() method creates document populated with _id if it is part of the  filter condition. So the resulting document will not have the answer property populated based on the filter condition. I will add a test case for this in a separate PR.

vkarpov15: @maheshrajamani this is a somewhat important detail. Upserting should apply any fields from the filter that don't have a non- `$eq` query operator. Below is some output from the mongodb shell that shows this behavior.

Mongoose does use this behavior for setting schema defaults on upsert.

```
MongoDB Enterprise > db.tests.countDocuments({})
0
MongoDB Enterprise > db.tests.updateOne({ _id: 'foo', answer: 42 }, { $set: { bar: 'baz' } }, { upsert: true })
{
	"acknowledged" : true,
	"matchedCount" : 0,
	"modifiedCount" : 0,
	"upsertedId" : "foo"
}
MongoDB Enterprise > db.tests.findOne()
{ "_id" : "foo", "answer" : 42, "bar" : "baz" }
MongoDB Enterprise > 
MongoDB Enterprise > db.tests.updateOne({ _id: 'foo', answer: { $eq: 42 }, count: { $gte: 42 } }, { $set: { bar: 'baz' } }, { upsert: true })
{
	"acknowledged" : true,
	"matchedCount" : 0,
	"modifiedCount" : 0,
	"upsertedId" : "foo"
}
MongoDB Enterprise > db.tests.findOne()
{ "_id" : "foo", "answer" : 42, "bar" : "baz" }
MongoDB Enterprise > 

```

maheshrajamani: @vkarpov15 Got it. Will add equals condition filter fields to the newly created document and will submit a new PR.

maheshrajamani: Resolved in PR #200

PR #198, opened 2023-02-28
fixes for docker compose

ivansenic:
**What this PR does**:
Fixes docker compose setup:

* add default value for env `JSONTAG`
* removed `--wait` as it's docker compose v2 option and not all developers have this installed (both @tatu-at-datastax and me can not run it)


tatu-at-datastax (approved):

docker-compose/README.md
```diff
@@ -4,6 +4,11 @@ This directory provides two ways to start the JSON API and Stargate coordinator
 
 ## Stargate JSON API with 3-node DSE 6.8 cluster
 
+> **PREREQUISITES:** Please build the latest JSONAPI docker image first by running: 
+> ```bash
+> ./mvnw clean package -Dquarkus.container-image.build=true -DskipTests
+> ```
+
```
kathirsvn, line 11: Good improvement. This was missing.

docker-compose/start_dse_68.sh
```diff
@@ -37,4 +37,4 @@ export JSONTAG
 
 echo "Running with DSE $DSETAG, Stargate $SGTAG, JSON API $JSONTAG"
 
-docker-compose up -d --wait
```
kathirsvn, line 40: This has both its advantages and disadvantages. The advantage is that this can be simply used in any work flow (for it to wait before moving to next step) but the disadvantage is to have the latest docker version.
My take will be to remove '--wait', so we will be compatible with old versions too and also we can work it around in any work flow by adding something like below (this is some thing I added in a Github action post docker compose)

```
- name: Wait for services
  run: |
     while ! nc -z localhost 8080; do sleep 1; done
     while ! nc -z localhost 8081; do sleep 1; done
```

kathirsvn (approved):

docker-compose/start_dse_68.sh
tatu-at-datastax: I noticed, FWTW, that Docker desktop had "enable docker compose v2" so at least I was able to upgrade easily.

ivansenic: I would remove it for now, we can add later on if needed and if used in any of the workflows.

ivansenic: cc @jeffreyscarpenter

PR #200, opened 2023-02-28
Add Equals DBFilters to newly created document

### LINKED ISSUE #199: Upsert - Add all equals filter data to document
Upsert to add all equals filter condition fields to the new document. 
For details check Val comments in #192 
Eg:
{
 "updateOne" :
   {
     "filter" : { "_id": "foo", "answer": 42 },
     "options" : {"upsert" : true},
     "update" : {"$set" : {"col1" : "boo"}}
   }
}

The newly upserted document should be like,

{
 "_id" : "foo",
  "answer" : 42,
  "col1" : "boo"
}

maheshrajamani:
**What this PR does**:
Add Equals DBFilters to newly created document

Fixes #199 

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DBFilterBase.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DBFilterBase.java
@@ -1,5 +1,9 @@
 package io.stargate.sgv2.jsonapi.service.operation.model.impl;
 
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.node.ArrayNode;
+import com.fasterxml.jackson.databind.node.JsonNodeFactory;
+import com.fasterxml.jackson.databind.node.ObjectNode;
 import io.stargate.bridge.grpc.Values;
 import io.stargate.bridge.proto.QueryOuterClass;
 import io.stargate.sgv2.api.common.cql.builder.BuiltCondition;
@@ -17,6 +21,18 @@ import java.util.function.Supplier;
 
 /** Base for the DB filters / conditions that we want to use with queries */
 public abstract class DBFilterBase implements Supplier<BuiltCondition> {
+  private final String path;
+
+  protected DBFilterBase(String path) {
+    this.path = path;
+  }
+
+  abstract JsonNode asJson(JsonNodeFactory nodeFactory);
+
+  protected String getPath() {
+    return path;
+  }
+
   /** Filter for the map columns we have in the super shredding table. */
   public abstract static class MapFilterBase<T> extends DBFilterBase {
 
@@ -47,6 +63,7 @@ public abstract class DBFilterBase implements Supplier<BuiltCondition> {
 
     protected MapFilterBase(
         String columnName, String key, MapFilterBase.Operator operator, T value) {
+      super(key);
       this.columnName = columnName;
       this.key = key;
       this.operator = operator;
@@ -92,22 +109,46 @@ public abstract class DBFilterBase implements Supplier<BuiltCondition> {
 
   /** Filters db documents based on a text field value */
   public static class TextFilter extends MapFilterBase<String> {
+    private final String strValue;
+
     public TextFilter(String path, Operator operator, String value) {
       super("query_text_values", path, operator, value);
+      this.strValue = value;
+    }
+
+    @Override
+    JsonNode asJson(JsonNodeFactory nodeFactory) {
+      return nodeFactory.textNode(strValue);
     }
   }
 
   /** Filters db documents based on a boolean field value */
   public static class BoolFilter extends MapFilterBase<Boolean> {
+    private final boolean boolValue;
+
     public BoolFilter(String path, Operator operator, Boolean value) {
       super("query_bool_values", path, operator, value);
+      this.boolValue = value;
+    }
+
+    @Override
+    JsonNode asJson(JsonNodeFactory nodeFactory) {
+      return nodeFactory.booleanNode(boolValue);
     }
   }
 
   /** Filters db documents based on a numeric field value */
   public static class NumberFilter extends MapFilterBase<BigDecimal> {
+    private final BigDecimal numberValue;
+
     public NumberFilter(String path, Operator operator, BigDecimal value) {
       super("query_dbl_values", path, operator, value);
+      this.numberValue = value;
+    }
+
+    @Override
+    JsonNode asJson(JsonNodeFactory nodeFactory) {
+      return nodeFactory.numberNode(numberValue);
     }
   }
 
@@ -121,6 +162,7 @@ public abstract class DBFilterBase implements Supplier<BuiltCondition> {
     protected final DocumentId value;
 
     public IDFilter(IDFilter.Operator operator, DocumentId value) {
+      super("_id");
       this.operator = operator;
       this.value = value;
     }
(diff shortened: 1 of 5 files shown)


src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DBFilterBase.java
```diff
@@ -17,6 +21,37 @@
...
+   * Returns filter condition element path.
+   *
+   * @return
+   */
+  protected String getPath() {
+    return path;
+  }
+
+  /**
+   * Returns `true` if the filter condition can be added to upsert row
```
tatu-at-datastax, line 49: Should perhaps say "should be added" instead of "can be added" -- since this is what happens, right?
(or needs to be added)

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DBFilterBase.java
```diff
@@ -17,6 +21,37 @@
...
+    this.path = path;
+  }
+
+  /**
+   * Get JsonNode for the representing filter condition value.
+   *
+   * @param nodeFactory
+   * @return
+   */
+  abstract JsonNode asJson(JsonNodeFactory nodeFactory);
```
tatu-at-datastax, line 37: This works; but if it was easier to access/pass, `ObjectMapper` also works (objectMapper.getNodeFactory() gives factory needed).

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DBFilterBase.java
```diff
@@ -121,6 +196,7 @@ public enum Operator {
     protected final DocumentId value;
 
     public IDFilter(IDFilter.Operator operator, DocumentId value) {
+      super("_id");
```
tatu-at-datastax, line 199: I think there is constants for this (DocumentConstants.Fields.DOC_ID), fwtw

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DBFilterBase.java
```diff
@@ -216,14 +314,37 @@ public IsNullFilter(String path) {
    */
   public static class ExistsFilter extends SetFilterBase<String> {
     public ExistsFilter(String path, boolean existFlag) {
-      super("exist_keys", path, Operator.CONTAINS);
+      super("exist_keys", path, path, Operator.CONTAINS);
+    }
+
+    @Override
+    JsonNode asJson(JsonNodeFactory nodeFactory) {
+      return null;
```
tatu-at-datastax, line 322: Should this be `nodeFactory.nullNode()` instead? Or are we not inserting explicit `null`s?

Would probably be good to test against Mongo implementation; I think (but could be wrong) that Mongo tends to add explicit `null`s.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DBFilterBase.java
```diff
@@ -216,14 +314,37 @@ public IsNullFilter(String path) {
...
+    }
+
+    @Override
+    boolean canAddField() {
+      return false;
```
tatu-at-datastax, line 327: I guess this is tied to above, where "field $eq null" should insert `null` value on creation or not.

tatu-at-datastax (approved):
Very nice, looks good! Added some comments but nothing too important.

The only thing I think matters would be that test for explicit `null` property on insert: while sort of small issue, null handling is one area where it's not immediately obvious how things should go without checking what Mongo does.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/FindAndUpdateIntegrationTest.java
```diff
@@ -455,6 +455,53 @@ public void findByIdUpsert() {
           .body("data.docs[0]", jsonEquals(expected));
     }
 
+    @Test
+    @Order(2)
+    public void findByIdAndColumnUpsert() {
```
tatu-at-datastax, line 460: Based on an earlier comment; would be good to test vs explicit `null` equality: I _think_ (but may be wrong) that explicit `null` property would also be added on insert (when upserting). But even if not we would want to verify expected behavior.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DBFilterBase.java
maheshrajamani: For values that need to be added as null to the document I have returned nodeFactory.nullNode(). The filters for which we shouldn't be adding data I have this method return null.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DBFilterBase.java
maheshrajamani: This is $exists filter, we won't be adding data data to new row. So canAddField() return false.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DBFilterBase.java
```diff
@@ -264,6 +421,39 @@ private static QueryOuterClass.Value getGrpcValue(Object value) {
...
   }
 
+  /**
+   * Return JsonNode for a filter conditions value, used to set in new document created for upsert.
+   *
+   * @param nodeFactory
+   * @param value
+   * @return
+   */
+  private static JsonNode getJsonNode(JsonNodeFactory nodeFactory, Object value) {
```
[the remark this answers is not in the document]
maheshrajamani: Not sure of another place logic to change Object to JsonNode is used. Can you let me know I will move this into util method and create another PR.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/FindAndUpdateIntegrationTest.java
maheshrajamani: Those `return null` is for filters like $exists, $size etc for which data not added to document on upsert. I have added null check filter to the use case to show, null JsonNode value is correctly added to the document.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DBFilterBase.java
```diff
@@ -17,6 +22,37 @@
...
+
+  protected DBFilterBase(String path) {
+    this.path = path;
+  }
+
+  /**
+   * Get JsonNode for the representing filter condition value.
+   *
+   * @param nodeFactory
+   * @return
```
tatu-at-datastax, line 36: Forgot to suggest that Javadoc should explain what returning `null` means here (not to be included in.... default value sets?)

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DBFilterBase.java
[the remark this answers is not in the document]
tatu-at-datastax: @maheshrajamani Ok actually no need to move -- I think I saw something similar earlier but can't remember it now.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DBFilterBase.java
tatu-at-datastax: Ah ok.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DBFilterBase.java
tatu-at-datastax: Ok; logic for returning/receiving `null` should probably be explained in Javadocs of `asJson()`.
Although isn't there some overlap between this and `canAddFied()`? (as in either one can prevent addition, I think)

Issue #178, opened 2023-02-23
Confirm all response messages against spec


amorton:
Issue #175 has updates to the API spec that locks in the response messages from a meeting today.

The API spec now has a general description of the request and response, and each command specifies the data is should return. 

This is a catchall ticket to ensure those changes to the spec and the code are in sync. There may be some  changes that have already happened like removing the updatedId's list from the status. 

I think the spec changes are good based on the meeting today, and we can probably start implementing. 

@maheshrajamani  and @ivansenic  this also includes the response data when there are errors. Still have changes for #160 in  my git stash, but this describes what the response should be.

PR #202, opened 2023-02-28
relates to #178: adapt countDocuments per spec, improve tests

ivansenic:
**What this PR does**:

Adaptation to spec for the `countDocuments`, including test improvements.

* [x] adapted status to `count` as per spec, from `counted_documents` (@kathirsvn fyi)
* [x] removed option class from command as per spec no options exists
* improved tests:
   * [x] added resolver test
   * [ ] added error case in operation tests
      * assert no errors & data
      * few tests were missing `@Test` annotation
      * improved readability
      * [x] remove test order - not needed
* [x] fixed swagger examples

Relates to #178.

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/StargateJsonApi.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/StargateJsonApi.java
@@ -274,7 +274,7 @@ import org.eclipse.microprofile.openapi.annotations.tags.Tag;
                       """
                                     {
                                       "status": {
-                                        "counted_documents": 2
+                                        "count": 2
                                       }
                                     }
                                     """),
--- a/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/CommandStatus.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/CommandStatus.java
@@ -5,7 +5,7 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 /** Enum with it's json property name which is returned in api response inside status */
 public enum CommandStatus {
   /** The element has the count of document */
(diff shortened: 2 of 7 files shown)


maheshrajamani:
@vkarpov15 Can you confirm if countDocuments response element has to be count or counted_documents? Remember seeing counted_documents response name in some documentation.

maheshrajamani (approved):

vkarpov15 (approved):

tatu-at-datastax:
I was trying to figure out Mongo equivalent, but `countDocuments()` documentation mostly referes to mongosh command; like:

https://www.geeksforgeeks.org/mongodb-countdocuments-method/

but regardless seems to just return raw numeric count, not a BSON Object.
That is not a good response for various reasons, but I was hoping to find better definition of what Mongoose would expect if not plain number.

ivansenic:
@tatu-at-datastax @maheshrajamani I am not sure if I can merge this now? Will you provide your reviews?

vkarpov15:
@tatu-at-datastax Mongoose expects `countDocuments()` to return a number. However, if it makes more sense for Stargate to return `{ count: 2 }` as opposed to just `2`, that's fine, I think it isn't a big deal for the stargate-mongoose driver to transform this result.

ivansenic:
OK I am merging this when I fix remaining tasks.

tatu-at-datastax:
Sorry, I am fine with the change, just wanted to verify I understood if it's different from Mongo handling; and if so if it is intentional and not accidental. Since it's intentional that's fine.

tatu-at-datastax (approved):

ivansenic:
Will merge this, and add the error test case when updating SG to `v2.0.9`.

PR #212, opened 2023-03-02
adapt the update parent workflow

ivansenic:
**What this PR does**:

### DIFF
--- a/.github/workflows/update-parent-version.yaml
+++ b/.github/workflows/update-parent-version.yaml
@@ -1,5 +1,5 @@
 # @author Ivan Senic
-name: Update parent version
+name: Update Parent Version
 
 # runs on
 # * stargate v2 release event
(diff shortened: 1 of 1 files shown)


.github/workflows/update-parent-version.yaml
```diff
@@ -1,5 +1,5 @@
 # @author Ivan Senic
-name: Update parent version
+name: Update Parent Version
```
tatu-at-datastax, line 2: Should name include "Main" somewhere, as it applies to main branch? (since this is shown in UI)

So like

    name: Update Parent Version (main)

tatu-at-datastax (approved):

.github/workflows/update-parent-version.yaml
ivansenic: But this is jsonapi, there is only main :)

tatu-at-datastax: Ah. Right, got confused. :)

PR #215, opened 2023-03-02
add failure mode test for countDocuments

ivansenic:
**What this PR does**:
Ensure count operation propagate received errors, as per spec.


src/test/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/CountOperationTest.java
```diff
@@ -168,7 +168,37 @@ public void countWithDynamicNoMatch() {
 
     @Test
     public void error() {
-      // TODO with stargate v2.0.9
+      // failures are propaged down
```
tatu-at-datastax, line 171: nit: propaged -> propagated

tatu-at-datastax (approved):

PR #216, opened 2023-03-02
relates to #178: insert commands per spec, added oerdered/unordered, better tests

ivansenic:
**What this PR does**:

Adaptation to spec for the `insertOne` & `insertMany`, including test improvements.

* [x] removed option class from `insertOne`  command as per spec no options exists
* [x] added `ordered` option for the `insertMany`
* [x] implemented ordered/unordered insertion
* [x] implemented fail fast for order version
* [x] implemented fail silently for unordered
* improved tests:
   * [x] extended operation tests to ensure all failure modes are tested
   * [x] created/updated tests for command resolvers 
   * [x] created tests for commands validation 
      * add unordered tests
      * simulate possible failure modes (using duplicated entries in insert many)
      * assert no errors & data
      * improved readability
* [x] fixed swagger examples

Relates to #178.

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/InsertOperation.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/InsertOperation.java
@@ -14,10 +14,8 @@ import io.stargate.sgv2.jsonapi.service.bridge.serializer.CustomValueSerializers
 import io.stargate.sgv2.jsonapi.service.operation.model.ModifyOperation;
 import io.stargate.sgv2.jsonapi.service.shredding.model.DocumentId;
 import io.stargate.sgv2.jsonapi.service.shredding.model.WritableShreddedDocument;
-import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
-import java.util.Objects;
 import java.util.function.Supplier;
 
 /**
@@ -61,7 +59,7 @@ public record InsertOperation(
 
                     // if we fail to insert propagate the FailFastInsertException
                     .onFailure()
-                    .transform(t -> new FailFastInsertException(doc.id(), t)))
+                    .transform(t -> new FailFastInsertException(doc, t)))
 
         // if no failures reduce to the op page
         .collect()
@@ -77,20 +75,19 @@ public record InsertOperation(
             e -> {
               // safe to cast, asserted class in onFailure
               FailFastInsertException failFastInsertException = (FailFastInsertException) e;
-              DocumentId failedId = failFastInsertException.documentId;
+              WritableShreddedDocument failed = failFastInsertException.document;
 
-              // collect inserted, since it's sequential iterate until failed id
-              List<DocumentId> insertedId = new ArrayList<>();
-              for (WritableShreddedDocument document : documents()) {
-                if (Objects.equals(document.id(), failedId)) {
-                  break;
-                }
-                insertedId.add(document.id());
-              }
+              // collect inserted, since it's sequential iterate until failed index
+              int failedIndex = documents().lastIndexOf(failed);
+              List<DocumentId> insertedId =
+                  documents().stream()
+                      .limit(failedIndex)
+                      .map(WritableShreddedDocument::id)
+                      .toList();
 
               return new InsertOperationPage(
                   insertedId,
-                  Collections.singletonMap(failedId, failFastInsertException.getCause()));
+                  Collections.singletonMap(failed.id(), failFastInsertException.getCause()));
             })
 
         // use object identity to resolve to Supplier<CommandResult>
@@ -181,11 +178,11 @@ public record InsertOperation(
   // simple exception to propagate fail fast
   private static class FailFastInsertException extends Exception {
 
-    private final DocumentId documentId;
+    private final WritableShreddedDocument document;
 
-    public FailFastInsertException(DocumentId documentId, Throwable cause) {
+    public FailFastInsertException(WritableShreddedDocument document, Throwable cause) {
       super(cause);
-      this.documentId = documentId;
+      this.document = document;
     }
   }
 }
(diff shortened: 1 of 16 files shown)


src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/InsertOperationPage.java
```diff
@@ -2,21 +2,73 @@
...
+  }
+
+  /** {@inheritDoc} */
   @Override
   public CommandResult get() {
+    // if we have errors, transform
+    if (null != failedIds && !failedIds().isEmpty()) {
+
+      List<CommandResult.Error> errors = new ArrayList<>(failedIds.size());
+      failedIds.forEach((documentId, throwable) -> errors.add(getError(documentId, throwable)));
```
ivansenic, line 40: currently for each failed document, I return an error, should we aggregate? if yes how?

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/InsertOperation.java
```diff
@@ -13,61 +14,148 @@
...
+                      .map(WritableShreddedDocument::id)
+                      .toList();
+
+              return new InsertOperationPage(
+                  insertedId,
+                  Collections.singletonMap(failed.id(), failFastInsertException.getCause()));
+            })
+
+        // use object identity to resolve to Supplier<CommandResult>
+        .map(i -> i);
```
ivansenic, line 94: can be removed

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/InsertOperation.java
```diff
@@ -13,61 +14,148 @@
...
+            doc ->
+                insertDocument(queryExecutor, query, doc)
+
+                    // handle errors fail silent mode
+                    .onItemOrFailure()
+                    .transform((id, t) -> Tuple2.of(doc, t)))
+
+        // then reduce here
+        .collect()
+        .in(InsertOperationPage::new, (agg, in) -> agg.aggregate(in.getItem1().id(), in.getItem2()))
```
ivansenic, line 117: this is not thread safe

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/InsertOperation.java
```diff
@@ -13,61 +14,148 @@
...
+        // ensure to construct correctly the information about inserted documents
+        .onFailure(FailFastInsertException.class)
+        .recoverWithItem(
+            e -> {
+              // safe to cast, asserted class in onFailure
+              FailFastInsertException failFastInsertException = (FailFastInsertException) e;
+              WritableShreddedDocument failed = failFastInsertException.document;
+
+              // collect inserted, since it's sequential iterate until failed index
+              int failedIndex = documents().lastIndexOf(failed);
```
maheshrajamani, line 81: Do we need this logic, since we use concatenate is already ordered and status can be captured with invoke() similar to unordered.

InsertOperationPage response = new InsertOperationPage();
`insertDocument(queryExecutor, query, doc)
                    .onItemOrFailure()
                    .invoke(
                        (documentId, error) -> {
                          response.aggregate(doc.id(), error);
                        }))`

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/InsertOperation.java
```diff
@@ -13,61 +14,148 @@
...
+                    // handle errors fail silent mode
+                    .onItemOrFailure()
+                    .transform((id, t) -> Tuple2.of(doc, t)))
+
+        // then reduce here
+        .collect()
+        .in(InsertOperationPage::new, (agg, in) -> agg.aggregate(in.getItem1().id(), in.getItem2()))
+
+        // use object identity to resolve to Supplier<CommandResult>
+        .map(i -> i);
```
maheshrajamani, line 120: Why is this needed in() would have already return Uni< InsertOperationPage>

maheshrajamani (approved):

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/InsertOperation.java
ivansenic: Hmm but how can I reduce? Where is the `response` coming from?

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/InsertOperation.java
ivansenic: Does not compile. The method signature must be 

`Uni<? extends Supplier<CommandResult>>` 

so that we can return the `Uni<InsertOperationPage>`.

I wanted to refactor, but it's not a small change, should be changed in all sub-classes and tests.. Do you think it's worth it?

ivansenic:
@maheshrajamani can you answer to the questions I posted?

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/InsertOperation.java
maheshrajamani: @ivansenic 
`InsertOperationPage response = new InsertOperationPage();
    return Multi.createFrom()
        .items(documents.stream())
        .onItem()
        .transformToUniAndConcatenate(
            doc ->
                insertDocument(queryExecutor, query, doc)
                    .onItemOrFailure()
                    .invoke(
                        (documentId, error) -> {
                          response.aggregate(doc.id(), error);
                        }))
        .collect()
        .last()
        .onItemOrFailure()
        .transform(
            (id, error) -> {
              return response;
            });`

ivansenic: Hmm, I don't like those side effects, especially when I am using a non-thread-safe structure.. Let me see if I can optimize this..

ivansenic: @maheshrajamani this is the bug from the demo, I can not use last index on a document, as if first fails, I would assume that the first was written.. Need to change this :)

PR #218, opened 2023-03-02
Delete - LWT failure retries

### LINKED ISSUE #217: Delete - LWT transaction failure retries
Handle LWT failure in DeleteOperation by retrying for configured number of attempt.

maheshrajamani:
**What this PR does**:
Delete - Handling LWT failure retries 
Fixes #217 

**Checklist**
- [X] Changes manually tested


maheshrajamani:
@ivansenic @tatu-at-datastax I have done IT for this PR. Not sure how to generate LWT failure for IT test case.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/FindOperation.java
```diff
@@ -83,11 +84,14 @@ public ReadDocument getNewDocument() {
     return doc;
   }
 
-  private QueryOuterClass.Query buildSelectQuery() {
+  private QueryOuterClass.Query buildSelectQuery(DBFilterBase.IDFilter idOverride) {
```
maheshrajamani, line 87: Added _id filter with other conditions instead of filtering the data in memory. When the query to cassandra has partition key filter, it will only route the query to that specific range that has the partition. Since it's not a not run range query this should be faster to filter.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DeleteOperation.java
```diff
@@ -68,7 +71,34 @@ public Uni<Supplier<CommandResult>> execute(QueryExecutor queryExecutor) {
             .concatenate()
             .onItem()
             .transformToUniAndConcatenate(
-                readDocument -> deleteDocument(queryExecutor, delete, readDocument))
+                readDocument -> {
+                  AtomicInteger attempt = new AtomicInteger(0);
+                  return Multi.createBy()
+                      .repeating()
+                      .uni(() -> deleteDocument(queryExecutor, delete, readDocument, attempt))
```
maheshrajamani, line 78: Can't use onFailure().retry() because deleteDocument() is not idempotent, it needs to be passed appropriate flag toto be read document again.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DeleteOperation.java
```diff
@@ -68,7 +71,34 @@ public Uni<Supplier<CommandResult>> execute(QueryExecutor queryExecutor) {
...
+                            switch (respVal) {
+                              case DELETED:
+                                return true;
+                              case MODIFIED_BY_CONCURRENT_PROCESS:
+                                return false;
+                              case CONCURRENCY_FAILURE:
+                              default:
+                                throw new JsonApiException(
+                                    ErrorCode.CONCURRENCY_FAILURE,
+                                    "Delete failed for %s because of concurrent transaction"
```
tatu-at-datastax, line 97: for %s -> for Document with id %s

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DeleteOperation.java
```diff
@@ -133,4 +210,17 @@ private static QueryOuterClass.Query bindDeleteQuery(
...
+  public enum DeleteResponse {
+    /** Successfully deleted a document */
+    DELETED,
+    /**
+     * Document modified by concurrent process and doesn't match the condition Could have changed
+     * value or deleted
+     */
+    MODIFIED_BY_CONCURRENT_PROCESS,
+
+    /** Failed because of concurrent process */
```
tatu-at-datastax, line 223: concurrent process -> concurrent transaction ?

tatu-at-datastax (approved):

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/FindOperation.java
```diff
@@ -83,11 +84,14 @@ public ReadDocument getNewDocument() {
...
     List<BuiltCondition> conditions = new ArrayList<>(filters.size());
     for (DBFilterBase filter : filters) {
       conditions.add(filter.get());
     }
+    if (idOverride != null) {
+      conditions.add(idOverride.get());
```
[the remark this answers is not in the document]
maheshrajamani: Just checked the trace for changing the order in the where clause of partition key makes no difference.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DeleteOperation.java
maheshrajamani: Yes that's the reason for reading again.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DeleteOperation.java
```diff
@@ -68,7 +71,34 @@ public Uni<Supplier<CommandResult>> execute(QueryExecutor queryExecutor) {
...
+                              case MODIFIED_BY_CONCURRENT_TRANSACTION:
+                                return false;
+                              case CONCURRENCY_FAILURE:
+                              default:
+                                throw new JsonApiException(
+                                    ErrorCode.CONCURRENCY_FAILURE,
+                                    "Delete failed for document with id %s because of concurrent transaction"
+                                        .formatted(readDocument.id().toString()));
+                            }
+                          });
```
ivansenic, line 100: this should not be done like this, retries should be done with the Multiny support.. No need to have your own counter and everything, see: https://smallrye.io/smallrye-mutiny/2.1.0/tutorials/retrying/

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DeleteOperation.java
```diff
@@ -107,20 +137,67 @@ private QueryOuterClass.Query buildDeleteQuery() {
...
+                    .executeWrite(boundQuery)
+                    .onItem()
+                    .transform(
+                        result -> {
+                          if (result.getRows(0).getValues(0).getBoolean()) {
+                            return DeleteResponse.DELETED;
+                          } else {
+                            return DeleteResponse.CONCURRENCY_FAILURE;
+                          }
+                        });
```
ivansenic, line 178: lwt is send back as a failure or? I see this here: https://github.com/stargate/stargate/blob/8c1208a461e68e3068fadb576f63ac812d78a1bf/coordinator/bridge/src/main/java/io/stargate/bridge/service/ExceptionHandler.java#L267-L279

Or is this something different?

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DeleteOperation.java
```diff
@@ -27,7 +29,8 @@ public record DeleteOperation(
      * Added parameter to pass number of document to be deleted, this is needed because read
      * documents limit changed to deleteLimit + 1
      */
-    int deleteLimit)
+    int deleteLimit,
+    int retryLimit)
     implements ModifyOperation {
   @Override
   public Uni<Supplier<CommandResult>> execute(QueryExecutor queryExecutor) {
```
ivansenic, line 36: Would be great if you can start adding some improvements to readbility of reactive seq creation:

* add inline comments that clearly split the steps, like here f.e.: https://github.com/stargate/jsonapi/blob/46aa965df35ff6ec7f520712298ae0a2abbf4d17/src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/InsertOperation.java#L46-L94
* avoid doing variable assignment..for example:

```
FROM

  private Uni<? extends ReadDocument> readDocumentAgain(
      QueryExecutor queryExecutor, ReadDocument prevReadDoc) {
    // Read again if retry flag is `true`
    final Uni<ReadOperation.FindResponse> findResponse =
        readOperation()
            .getDocuments(
                queryExecutor,
                null,
                new DBFilterBase.IDFilter(DBFilterBase.IDFilter.Operator.EQ, prevReadDoc.id()));
    return findResponse
        .onItem()
        .transform(
            response -> {
              if (!response.docs().isEmpty()) {
                return response.docs().get(0);
              } else {
                // If data changed and doesn't satisfy filter conditions
                return null;
              }
            });
  }

TO

  private Uni<? extends ReadDocument> readDocumentAgain(
      QueryExecutor queryExecutor, ReadDocument prevReadDoc) {
    // Read again if retry flag is `true`
    return readOperation()
            .getDocuments(
                queryExecutor,
                null,
                new DBFilterBase.IDFilter(DBFilterBase.IDFilter.Operator.EQ, prevReadDoc.id()))

        // then map
        .onItem()
        .transform(
            response -> {
              if (!response.docs().isEmpty()) {
                return response.docs().get(0);
              } else {
                // If data changed and doesn't satisfy filter conditions
                return null;
              }
            });
  }

```

src/test/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DeleteOperationTest.java
```diff
@@ -38,6 +40,7 @@ public class DeleteOperationTest extends AbstractValidatingStargateBridgeTest {
 
   @Inject QueryExecutor queryExecutor;
   @Inject ObjectMapper objectMapper;
+  @Inject DocumentConfig documentConfig;
```
ivansenic, line 43: I think we don't need this can control in the tests ourself

src/test/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DeleteOperationTest.java
```diff
@@ -235,6 +244,314 @@ public void deleteWithDynamic() {
       assertThat(result.status()).hasSize(1).containsEntry(CommandStatus.DELETED_COUNT, 1);
     }
 
+    @Test
```
ivansenic, line 247: missing the test case where you would hit the maximum of retries and abort because of that

ivansenic (changes requested):
Some comments, plus missing integration test.. Should be done with multiple update threads and one delete thread.. Or one update and one delete thread, but then the test must run many times to see both can succeed..

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/FindOperation.java
maheshrajamani: We are reading document based on the filter in came in request and additional id filter. I don't see adding more conditions to this with the use cases we have. Not required to be collection unless we have use case for it. 
Renamed it the variable as `additionalIdFilter`.

src/main/java/io/stargate/sgv2/jsonapi/service/bridge/config/DocumentConfig.java
```diff
@@ -64,4 +64,10 @@ public interface DocumentConfig {
   @Positive
   @WithDefault("20")
   int maxDocumentUpdateCount();
+
+  /** @return Defines the maximum retry for lwt failure <code>3</code>. */
+  @Max(5)
+  @Positive
+  @WithDefault("3")
+  int maxLWTFailureRetry();
```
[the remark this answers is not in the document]
maheshrajamani: Ok will do.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/ReadOperation.java
```diff
@@ -118,7 +119,8 @@ default Uni<CountResponse> countDocuments(
    * @param queryExecutor
    * @return
    */
-  Uni<FindResponse> getDocuments(QueryExecutor queryExecutor, String pagingState);
+  Uni<FindResponse> getDocuments(
+      QueryExecutor queryExecutor, String pagingState, DBFilterBase.IDFilter idOverride);
```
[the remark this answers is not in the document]
maheshrajamani: Ok will do

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DeleteOperation.java
maheshrajamani: Did the retry logic in a previous commit before using onFailure().retry(), then changed it this logic because this will be easier to capture errors if case we want to return both deleted count and error document ids.
Will change it back to onFailure().retry().

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DeleteOperation.java
maheshrajamani: Ok will do

src/test/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DeleteOperationTest.java
maheshrajamani: Not sure why can't be used. Anyway removed it as suggested.

src/test/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DeleteOperationTest.java
maheshrajamani: It is there deleteWithDynamicRetryFailure, to show if maximum retries done it aborts and return error

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DeleteOperation.java
maheshrajamani: Not sure what this is. But incase of LWT failure it returns false in result as expected.

maheshrajamani:
> Should be done with multiple update threads and one delete thread.. Or one update and one delete thread, but then the test must run many times to see both can succeed..

@ivansenic None of these guarantee delete retry logic is executed. Can't assert for the use case with above testing.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/FindOperation.java
ivansenic: So it's not an override in fact? What happens if the original query already had id condition?

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DeleteOperation.java
ivansenic: But if there is no document, what would be returned? The same? Then it's OK, next read would fail..

src/test/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DeleteOperationTest.java
ivansenic: Because test depends not on it, you can add retry values yourself in the test.. What s the point of `@Injecting` this?

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DeleteOperation.java
maheshrajamani: Yes the same logic will handle, if no document and we try to delete with txid = 'xxxx' it will return false.
`cassandra@cqlsh:purchase_database> select * from myt where pk = '1';

 pk | col1 | col2
----+------+------

cassandra@cqlsh:purchase_database> delete from myt where pk = '1' if col1 = 'aaa';

 [applied]
-----------
     False
`

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/FindOperation.java
maheshrajamani: Ah, this will be a problem, fixed it.

tatu-at-datastax:
> > Should be done with multiple update threads and one delete thread.. Or one update and one delete thread, but then the test must run many times to see both can succeed..
> 
> @ivansenic None of these guarantee delete retry logic is executed. Can't assert for the use case with above testing.

We discussed this with Aaron, Mahesh and myself and considered it very difficult if not impossible to have a reproducible reliable test to verify this, as an IT. But that later on it could be tested with load test. Unit test can verify other aspects.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DeleteOperation.java
```diff
@@ -107,20 +124,77 @@ private QueryOuterClass.Query buildDeleteQuery() {
...
-      QueryExecutor queryExecutor, QueryOuterClass.Query query, ReadDocument doc) {
-    query = bindDeleteQuery(query, doc);
-    return queryExecutor
-        .executeWrite(query)
+  private Uni<Boolean> deleteDocument(
+      QueryExecutor queryExecutor,
+      QueryOuterClass.Query query,
+      ReadDocument doc,
+      AtomicInteger retryAttempt)
+      throws JsonApiException {
```
ivansenic, line 136: this `throws` is suspicious and should not be there for sure

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DeleteOperation.java
```diff
@@ -27,58 +29,73 @@ public record DeleteOperation(
...
+            document -> {
+              AtomicInteger retryAttempt = new AtomicInteger(0);
+              return deleteDocument(queryExecutor, delete, document, retryAttempt)
+
+                  // Retry `retryLimit` times in case of LWT failure
+                  .onFailure()
+                  .retry()
+                  .until(
+                      error ->
+                          error instanceof JsonApiException && retryAttempt.get() < retryLimit);
```
ivansenic, line 87: we should add here the error code assertion, you can not sure know this is LWt failure here?

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DeleteOperation.java
```diff
@@ -27,58 +29,73 @@ public record DeleteOperation(
...
+              }
+            })
+        .concatenate()
+
+        // Run delete for selected documents and retry in case of
+        .onItem()
+        .transformToUniAndConcatenate(
+            document -> {
+              AtomicInteger retryAttempt = new AtomicInteger(0);
+              return deleteDocument(queryExecutor, delete, document, retryAttempt)
```
ivansenic, line 80: this is not good imo.. The delete document should not do the re-read, it should be seomthing like:

```
deleteDocument(..)

   .onFailure(jwtPredicate)
   .switchTo(() -> {
        readDocumentAgain(queryExecutor, document)

           .flatMap(document -> deleteDocument(..))

           .onFailure(jwtPredicate)
           .retry().max(maxLwtRetries - 1)
    })
```

You need to avoid using Atomic stuff, and side effects in general.. Plus the suggested code is way simpler and smaller..

ivansenic (approved):
I approve, but there are still some issue that should be fixed:

* simplify reactive sequence
* assert JWT failure in the `JsonApiException`
* etc

ivansenic:
> > > Should be done with multiple update threads and one delete thread.. Or one update and one delete thread, but then the test must run many times to see both can succeed..
> > 
> > 
> > @ivansenic None of these guarantee delete retry logic is executed. Can't assert for the use case with above testing.
> 
> We discussed this with Aaron, Mahesh and myself and considered it very difficult if not impossible to have a reproducible reliable test to verify this, as an IT. But that later on it could be tested with load test. Unit test can verify other aspects.

But verify what? I mean it's fairly easy to do an update and delete in parallel. They both have to succeed. Run this `N` amount of times in a `@RepeatableTest`, and you get the assurance that retries work. Yes I agree, it's not directly assuring that retry was executed, but if you write this test and apply to the current state, it would fail actually. So if you ask me that's pretty good, way better than not having it.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DeleteOperation.java
maheshrajamani: Make sense. Removed it.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DeleteOperation.java
maheshrajamani: Fixed. Handled now with inherited exception class to handle this.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DeleteOperation.java
ivansenic: Sorry, mixed the frameworks.. Here it's

```
.onFailure(..).recoverWithUni(..)
```

PR #221, opened 2023-03-02
Fix #164: order individual update operation actions as per spec

### LINKED ISSUE #164: Enforce execution of field update operations is in alphabetic order
(from #132)

Currently update operations on fields are executed in the order specified by input document, so for example:

```
{"$set": {"foo" : 1, "bar" : 2}}
```

would first set value of `"foo"` and then `"bar"`. Specification mandates ordering to be the opposite.
The most obvious change resulting from this is that in case of adding (setting) new properties, insertion order is changed; for most other aspects it should not significantly change behavior.

tatu-at-datastax:
**What this PR does**:

Fixes #<issue number>

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/update/ActionWithPath.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/update/ActionWithPath.java
@@ -1,0 +1,5 @@
+package io.stargate.sgv2.jsonapi.api.model.command.clause.update;
+
+public interface ActionWithPath {
+  String path();
+}
--- a/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/update/AddToSetOperation.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/update/AddToSetOperation.java
@@ -19,7 +19,7 @@ public class AddToSetOperation extends UpdateOperation {
   private List<AddToSetAction> actions;
 
   private AddToSetOperation(List<AddToSetAction> actions) {
-    this.actions = actions;
+    this.actions = sortByPath(actions);
   }
 
   public static AddToSetOperation construct(ObjectNode args) {
@@ -153,5 +153,6 @@ public class AddToSetOperation extends UpdateOperation {
   }
 
   /** Value class for per-field update operations. */
-  private record AddToSetAction(String path, JsonNode value, boolean each) {}
+  private record AddToSetAction(String path, JsonNode value, boolean each)
+      implements ActionWithPath {}
 }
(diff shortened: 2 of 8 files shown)


src/test/java/io/stargate/sgv2/jsonapi/service/operation/model/command/clause/update/PushOperationTest.java
```diff
@@ -83,6 +83,26 @@ public void testPushToNonExistingNested() {
...
+      UpdateOperation oper =
+          UpdateOperator.PUSH.resolveOperation(
+              objectFromJson("{ \"subdoc.newArray\" : \"value\", \"array\": 3 }"));
+      assertThat(oper).isInstanceOf(PushOperation.class);
+      ObjectNode doc = objectFromJson("{ }");
+      assertThat(oper.updateDocument(doc, targetLocator)).isTrue();
+      ObjectNode expected =
+          objectFromJson(
+              """
+                      { "array": [ 3 ], "subdoc" : { "newArray": [ "value" ] } }
```
vkarpov15, line 100: Looking at this, it does seem a little strange that the keys are created in alphabetical order versus the order in which they were specified in the `$push` operation. I don't think it would cause any concrete issues in Mongoose, but just bringing this up for potential further discussion. Is there a reason to prefer alphabetical?

tatu-at-datastax: @vkarpov15 This is as per Mongo's own definitions from https://www.mongodb.com/docs/v6.0/reference/operator/update/#std-label-update-operators-processing-order

"Starting in MongoDB 5.0, update operators process document fields with string-based names in lexicographic order. "

I agree that if it was up to me, I'd have strictly followed order in which operations are defined in message -- that seems more like caller's intent.

vkarpov15: That's fair, I wasn't aware of MongoDB's ordering. I guess in most cases this ordering doesn't matter, because document updates are atomic in MongoDB and you can't update the same property using 2 different update operators. You can disregard my comment.

maheshrajamani (approved):
LGTM

src/test/java/io/stargate/sgv2/jsonapi/service/operation/model/command/clause/update/PushOperationTest.java
tatu-at-datastax: Right, I just happened to notice it (based on Aaron's comments). And I agree that in almost all cases it is irrelevant -- the main case I see is the insertion order of fields that depends on ordering.

PR #236, opened 2023-03-09
relates to #178: update commands per spec, improved tests

ivansenic:
**What this PR does**:

Adaptation to spec for the `updateOne` & `updateMany`, including test improvements.

* [x] make `update` clause mandatory I see no reason to allow this empty or?
    * [ ] if so we should ensure that `updateOperationDefs` are also not empty and at least one given 
* improved tests:
   * [x] created command tests for validation
   * [x] created/updated tests for command resolvers, added tests for `upsert=true`
   * [x] refactored operation tests, added some new cases 
   * [ ] integration tests assertions
* [ ] fixed swagger examples

Relates to #178.


maheshrajamani (approved):

ivansenic:
But I did no finish this PR, there were some tasks left? OK can do separate PR next week for that..

PR #237, opened 2023-03-09
Failure modes for update and delete operations

### LINKED ISSUE #214: Implement failure modes for update and delete operations
See spec for more information. Failure should be asserted in the `DeleteOperationTest`.

maheshrajamani:
**What this PR does**:
Delete and Update operation - fail silently

Fixes #214 

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DeleteOperationPage.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DeleteOperationPage.java
@@ -5,11 +5,10 @@ import com.google.common.collect.Multimap;
 import io.smallrye.mutiny.tuples.Tuple3;
 import io.stargate.sgv2.jsonapi.api.model.command.CommandResult;
 import io.stargate.sgv2.jsonapi.api.model.command.CommandStatus;
-import io.stargate.sgv2.jsonapi.exception.JsonApiException;
 import io.stargate.sgv2.jsonapi.service.shredding.model.DocumentId;
+import io.stargate.sgv2.jsonapi.util.ExceptionUtil;
 import java.util.ArrayList;
 import java.util.Collection;
-import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.function.Supplier;
@@ -26,14 +25,14 @@ import java.util.stream.Collectors;
 public record DeleteOperationPage(
     List<Tuple3<Boolean, Throwable, DocumentId>> deletedInformation, boolean moreData)
     implements Supplier<CommandResult> {
+  private static final String ERROR = "Failed to delete documents with _id %s: %s";
+
   @Override
   public CommandResult get() {
     int deletedCount =
         (int)
             deletedInformation.stream()
-                .filter(
-                    deletedDocument ->
-                        deletedDocument.getItem1() != null && deletedDocument.getItem1())
+                .filter(deletedDocument -> Boolean.TRUE.equals(deletedDocument.getItem1()))
                 .count();
 
     // aggregate the errors by error code or error class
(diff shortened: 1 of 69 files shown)


src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DeleteOperationPage.java
```diff
@@ -1,23 +1,91 @@
...
+    List<Tuple3<Boolean, Throwable, DocumentId>> deletedInformation, boolean moreData)
     implements Supplier<CommandResult> {
   @Override
   public CommandResult get() {
+    int deletedCount =
+        (int)
+            deletedInformation.stream()
+                .filter(
+                    deletedDocument ->
+                        deletedDocument.getItem1() != null && deletedDocument.getItem1())
```
tatu-at-datastax, line 36: Also works: 

```
Boolean.TRUE.equals(deletedDocument.getItem1())
````

File: src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/UpdateOperationPage.java [hunk 2]
```diff
@@ -46,9 +71,9 @@ public CommandResult get() {
     }
   }
 
-  private CommandResult.Error getError(DocumentId documentId, Throwable throwable) {
+  private CommandResult.Error getError(List<DocumentId> documentIds, Throwable throwable) {
     String message =
-        "Failed to update document with _id %s: %s".formatted(documentId, throwable.getMessage());
+        "Failed to update documents with _id %s: %s".formatted(documentIds, throwable.getMessage());
```
tatu-at-datastax, line 76: I think that similar to exception key, formatting of documentIds probably should be a shared utility method since we may want to add quoting/escaping at some point -- start with just `List.toString()` for now, but I have found that it is good to distinguish f.ex between number `1` and String `"1"`... especially as String keys can have all kind of content.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/UpdateOperationPage.java
```diff
@@ -23,13 +27,34 @@ public record UpdateOperationPage(
...
     List<JsonNode> updatedDocs = new ArrayList<>(updatedDocuments().size());
-    List<CommandResult.Error> errors = new ArrayList<>();
+
+    // aggregate the errors by error code or error class
+    Multimap<String, ReadAndUpdateOperation.UpdatedDocument> groupedErrorUpdates =
+        ArrayListMultimap.create();
     updatedDocuments.forEach(
         update -> {
           if (update.upserted()) upsertedId[0] = update.id();
           if (returnDocs) updatedDocs.add(update.document());
```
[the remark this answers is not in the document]
maheshrajamani: Yes. Upserted document will be available only in case of findOneAndUpdate command and returnDocument option set to "after".

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/UpdateOperationPage.java
tatu-at-datastax: Come to think of this, since the only value where simple "value.toString()" may not be sufficient is `String` (where we'd want to surround value in quotes), we could easily change `toString()` method for `DocumentId.StringId` and the message would be changed?

Downside is that there are lots of tests that compare exact exception message.

I can do a PR after merging this one so as not to extend scope here.

tatu-at-datastax (approved):
LGTM

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/UpdateOperationPage.java
```diff
@@ -23,13 +27,34 @@ public record UpdateOperationPage(
...
-          if (update.error() != null) errors.add(getError(update.id(), update.error()));
+          //
+          if (update.error() != null) {
+            String key = update.error().getClass().getSimpleName();
```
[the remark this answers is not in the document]
maheshrajamani: Done

PR #241, opened 2023-03-09
Fix #240: surround String-valued DocumentId in single quotes

### LINKED ISSUE #240: Enclose `String`-backed `DocumentId` values in quotes (for Exception messages)
Document ids (`_id`) can have any of scalar types (String, Number, Boolean, null). When including id in things like Exception messages, we simply include value using it's `toString()` method. This means that ids of number `1` and String `"1"` are both include as `1`. It would make sense to enclose String values in quotes to make distinction between types.

In addition, since String values can have any content, non-enclosed ids may lead to difficult to interpret messages when trouble-shooting: adding surrounding quotes helps.

We have a choice of single or double-quotes (since these are for human-readability): I think single quotes make sense wrt Cassandra String literals, as well as not needing escaping for Java tests.

tatu-at-datastax:
**What this PR does**:

Enclose `String`-valued `DocumentId` in single quotes, for use in Exception messages -- avoids confusion wrt id type, contents.

Also adds a separate method to use for constructing DB primary keys, instead of `Document.toString()`.

Fixes #240

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/bridge/serializer/CustomValueSerializers.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/bridge/serializer/CustomValueSerializers.java
@@ -64,7 +64,7 @@ public class CustomValueSerializers {
   public static List<QueryOuterClass.Value> getDocumentIdValue(DocumentId documentId) {
     // Temporary implementation until we convert it to Tuple in DB
     List<QueryOuterClass.Value> tupleValues =
-        List.of(Values.of(documentId.typeId()), Values.of(documentId.toString()));
+        List.of(Values.of(documentId.typeId()), Values.of(documentId.asDBKey()));
     return tupleValues;
   }
 }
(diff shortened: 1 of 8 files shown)


src/main/java/io/stargate/sgv2/jsonapi/service/bridge/serializer/CustomValueSerializers.java
```diff
@@ -64,7 +64,7 @@ public static Map<QueryOuterClass.Value, QueryOuterClass.Value> getDoubleMapValu
   public static List<QueryOuterClass.Value> getDocumentIdValue(DocumentId documentId) {
     // Temporary implementation until we convert it to Tuple in DB
     List<QueryOuterClass.Value> tupleValues =
-        List.of(Values.of(documentId.typeId()), Values.of(documentId.toString()));
+        List.of(Values.of(documentId.typeId()), Values.of(documentId.asDBKey()));
```
tatu-at-datastax, line 67: Necessary to avoid single-quotes from being added wrt `DocumentId.toString()` change: also semantically good to have separate method for use case.

maheshrajamani (approved):
LGTM

vkarpov15:
LGTM

PR #252, opened 2023-03-10
Fix #224: implement `$mul` update operator

### LINKED ISSUE #224: Implement $mul update operator
One more in-scope update operator is `$mul`:

https://www.mongodb.com/docs/v6.0/reference/operator/update/mul/

and it only works on Numeric values so somewhat easier to implement (may choose for development just for simplicity).

tatu-at-datastax:
**What this PR does**:

Adds `$mul` update operator implementation

Fixes #224

**Checklist**
- [ ] Changes manually tested


vkarpov15 (approved):
LGTM :+1:

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/update/MulOperation.java
```diff
@@ -0,0 +1,95 @@
...
+
+    return modified;
+  }
+
+  private JsonNode multiply(ObjectNode doc, JsonNode oldValue, JsonNode multiplierValue) {
+    BigDecimal multiplier = multiplierValue.decimalValue();
+    BigDecimal old = oldValue.decimalValue();
+
+    // Short-cuts to avoid calculation
+    if (BigDecimal.ONE.equals(multiplier) || BigDecimal.ZERO.equals(old)) {
```
maheshrajamani, line 87: Do we need this, cpu cost to multiplying by 1 or 0 vs running if condition every time.

maheshrajamani (approved):

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/update/MulOperation.java
tatu-at-datastax: Probably not a huge difference either way, but doing check avoids document modification and keeps `JsonNode` comparison as identity comparison. But I guess it assumes that case of multiplier `1` or old value of `0` is common.

Actually, come to think of this, I think latter seems much more likely; so what I should do is remove first check, leave second.

tatu-at-datastax: Changed to only check for old value being 0 (seems like common enough case), don't check on multiply by 1.

PR #254, opened 2023-03-13
relates to #178: int test improvements for update commands

ivansenic:
**What this PR does**:

Improves integration tests for the update commands. Left over from the #236 

* [x] correct split in integration tests between `findOneAndUpdate`, `updateOne` and `updateMany`
* [x] added assertions for the `errors`
* [x] added more `status` assertions
* [x] added more checks in `updateMany` 
* [x] added concurrent update tests, that assert no failures on concurrent updates 
* [x] improved readability  
* [x] removed unnecessary annotations & order, fixed delete docs from collection on after each 
* [x] validation cases 

Plus:
* [x] fixed lwt integration tests
* [x] added concurrency test for delete ops
* [x] more swagger examples to showcase diff between upsert and no upsert responses


ivansenic:
@amorton The spec here does not mention anything about the `status.moreData`:

https://github.com/stargate/jsonapi/blob/main/docs/jsonapi-spec.textile?plain=1#L615-L661

Currently:
* the `updateOne` never returns `moreData` flag
* the `updateMany` returns `moreData=true` in case there is more data that matches the filter

maheshrajamani (approved):

maheshrajamani:
LGTM

PR #255, opened 2023-03-13
update docker compose readme

### LINKED ISSUE #208: Intermittent DEADLINE_EXCEEDED error
While running `insertMany` command, getting below error intermittently.
```json
{
    "errors": [
        {
            "exceptionClass": "StatusRuntimeException",
            "message": "DEADLINE_EXCEEDED: Operation timed out - received only 0 responses."
        }
    ]
}
```

Sample request

``` json
{
    "insertMany": {
        "documents": [
            {
                "__v": 0,
                "_id": "63ff496c0694802df9117c82",
                "name": "Wilson Tennis Racket",
                "price": 225
            },
            {
                "__v": 0,
                "_id": "63ff496c0694802df9117c83",
                "name": "Hilton Water Bottle",
                "price": 12
            },
            {
                "__v": 0,
                "_id": "63ff496c0694802df9117c84",
                "name": "Beats Earphone Classic",
                "price": 145
            }
        ],
        "options": {}
    }
}
```

ivansenic:
**What this PR does**:

Small update that come out from working on #208.

Fixes #208 as can not reproduce.

### DIFF
--- a/docker-compose/README.md
+++ b/docker-compose/README.md
@@ -5,9 +5,12 @@ This directory provides two ways to start the JSON API and Stargate coordinator 
 ## Stargate JSON API with 3-node DSE 6.8 cluster
 
 > **PREREQUISITES:** Please build the latest JSONAPI docker image first by running: 
+> 
 > ```bash
 > ./mvnw clean package -Dquarkus.container-image.build=true -DskipTests
 > ```
+> 
+> You can control the platform using the `-Dquarkus.docker.buildx.platform=linux/amd64,linux/arm64` property.
 
 You can start a simple Stargate configuration with the following command:
 
--- a/docker-compose/README.md
+++ b/docker-compose/README.md
@@ -12,7 +12,7 @@ This directory provides two ways to start the JSON API and Stargate coordinator 
 You can start a simple Stargate configuration with the following command:
 
 ```
-docker compose up -d
+docker-compose up -d
 ``` 
 
 This brings up the configuration described in the `docker-compose.yml` file. The configuration includes health criteria for each container that is used to ensure the containers come up in the correct order.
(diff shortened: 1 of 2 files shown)


docker-compose/README.md
```diff
@@ -5,14 +5,17 @@ This directory provides two ways to start the JSON API and Stargate coordinator
...
 > ./mvnw clean package -Dquarkus.container-image.build=true -DskipTests
 > ```
+> 
+> You can control the platform using the `-Dquarkus.docker.buildx.platform=linux/amd64,linux/arm64` property.
 
 You can start a simple Stargate configuration with the following command:
 
 ```
-docker compose up -d
+docker-compose up -d
```
tatu-at-datastax, line 18: Question on this: both work, I think, but I was not 100% sure which one is recommended approach. F.ex:

https://stackoverflow.com/questions/66514436/difference-between-docker-compose-and-docker-compose

seems to suggest `docker compose` is the "new" way and `docker-compose` being deprecated.

So maybe it should be left as-is? I'll ask on our Slack channel just so we know which one to go with.

tatu-at-datastax (changes requested):
Looks good but I want to know if we should document `docker compose` or `docker-compose`.

docker-compose/README.md
ivansenic: That's again that v1 vs v2 story.. Let's do whatever you want, I am fine with both.. I did remove that `--wait` before to make it work with v1, so I wanted to make this as well..

tatu-at-datastax: Ok let's see if we can figure out one to use. Not a huge thing don't want to make it one. Will approve in the meantime.

tatu-at-datastax (approved):

Issue #232, opened 2023-03-08
Check that different update operators do not target same field (conflict)


tatu-at-datastax:
Although I have not seen this mentioned in general (but will see if I can find), update operators can NOT target same field in same operation. There is an explicit mention of "$set" and "$unset" not allowing this but it appears general limitation; for example trying:

```
db.arrays.findOneAndUpdate({ _id : 1 }, { "$max" : {"x":false }, "$min" : {"x":true} })
```

gives error

```
Updating the path 'x' would create a conflict at 'x'
```

possibly because order in which operators are executed does not appear to be strongly defined (unlike per-field ordering that is).

We should similarly cross-check paths across different operators (within operators JSON does not allow duplicates) and fail update upon found conflict.

vkarpov15:
An additional caveat is that an update operator cannot target subpaths of a path that another update operator is targetting. For example, the following throws a `"Updating the path 'coordinates.lat' would create a conflict at 'coordinates'"` error.

```
> db.test.insertOne({ coordinates: { lat: 37, lng: -132 } })
> db.test.updateOne({}, { $set: { coordinates: { lat: 37.14, lng: -132.33 } }, $mul: { 'coordinates.lat': 1.1 } })
```

tatu-at-datastax:
@vkarpov15 yes, I was worried this comes up. Straight up comparison is much easier than prefixes.

But... well, I'll see whether it makes sense to first cover simple(r) case & then do follow-up, or if I just want to tackle the full case right on. Implementation-wise latter can be done, I think, by sorting the path & comparing adjacent paths to see if first one is a parent-prefix of the second one (suffix starting with `.`). So not necessarily much more complicated, but a bit.

vkarpov15:
What we often end up doing in Mongoose is creating a `Set` of all the updated paths and their parents. So something like:

```javascript
const updatedPaths = new Set();
for (const key of Object.keys(update)) {
  if (key.indexOf('.') === -1) {
    updatedPaths.add(key);
    continue;
  }
  const pieces = key.split('.');
  let cur = pieces[0];
  updatedPaths.add(cur);
  for (let i = 1; i < pieces.length; ++i) {
    cur += '.' + pieces[i];
    updatedPaths.add(cur);
  }
}
```

Add a check for duplicates and you're done. Should work in this case, unless you think it is too slow.

tatu-at-datastax:
[...]

PR #262, opened 2023-03-14
UpdateTargetLocator->ActionTargetLocator, refactor part 2

### LINKED ISSUE #257: Refactor `UpdateTargetLocator` for use with in-memory sorting, projection
As implemented, `UpdateTargetLocator` works for Update operation use case (where typically updates done for just one Document), but inefficient where same locator would be used over a larger set of documents (such as sorting, projection).

So let's refactor it a bit to separate preparation (splitting of path) from matching methods.

tatu-at-datastax:
**What this PR does**:

Further refactoring of existing `UpdateTarget[Locator]` into `ActionTarget[Locator]` for more efficient handling, reuse.

Specifically now decoding of target locator for `UpdateOperation`s occurs during construction, and not evaluation: this reduces processing for "updateMany()" use case. Similarly added method (ActionTargetLocator.findValueIn()) is planned to be usable for in-memory Sorting (and later on hopefully with Projection as well).

Also pushes handling of `Action`s (per-field part of full Update Operation) to base class to allow centralized path-conflict checking (to solve #232 in near future)

Fixes #257

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/update/ActionTarget.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/update/ActionTarget.java
@@ -6,10 +6,12 @@ import com.fasterxml.jackson.databind.node.ObjectNode;
 import java.util.Objects;
 
 /**
- * Definition of a target for an {@link UpdateOperation}; built from "dot path" and document to
- * update, by {@link UpdateTargetLocator}.
+ * Definition of a target for an action, such as ones contained by {@link UpdateOperation}, or
+ * performed as part of Projection or in-memory Sort processing. Instances are created by {@link
+ * ActionTargetLocator} based on locator's configuration (decoded from "Dotted path") and JSON
+ * document to evalute against.
  */
-public record UpdateTarget(
+public record ActionTarget(
     /** Full path to target for error reporting purposes */
     String fullPath,
     /**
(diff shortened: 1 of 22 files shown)


src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/update/ActionTarget.java
```diff
@@ -6,10 +6,12 @@
...
 /**
- * Definition of a target for an {@link UpdateOperation}; built from "dot path" and document to
- * update, by {@link UpdateTargetLocator}.
+ * Definition of a target for an action, such as ones contained by {@link UpdateOperation}, or
+ * performed as part of Projection or in-memory Sort processing. Instances are created by {@link
+ * ActionTargetLocator} based on locator's configuration (decoded from "Dotted path") and JSON
+ * document to evalute against.
  */
-public record UpdateTarget(
+public record ActionTarget(
```
maheshrajamani, line 14: Same as what is in UpdateTargetLocator.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/update/ActionTargetLocator.java
```diff
@@ -7,8 +7,17 @@
...
+ * Helper classes built from "dotted paths" and evaluated in context of a JSON document to produce
+ * one of:
+ *
+ * <ul>
+ *   <li>{@link ActionTarget} instances for Document updates
+ *   <li>{@link JsonNode} when extracting value using {@link #findValueIn} (used for Sorting and
+ *       possibly Projection)
+ * </ul>
+ */
+public class ActionTargetLocator {
```
maheshrajamani, line 20: Can we have this class named as just TargetLocator and move it to util package? That way we can use it for projection/sorting.

maheshrajamani (changes requested):
@tatu-at-datastax Can you see if we can move the 2 class to util package. Also there is a unit test failure.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/update/ActionTargetLocator.java
tatu-at-datastax: @maheshrajamani Yes, I will do that today. Was wondering about it since it's not really update (or update action) specific.

maheshrajamani: Are you going to do it as part of this PR?

tatu-at-datastax:
Renamed "UpdateTarget[Locator]" to "PathMatch[Locator]", moved to util. Did other related renaming, hopefully didn't miss too many references.

maheshrajamani (approved):
LGTM

PR #263, opened 2023-03-15
closes #207: add spec for

### LINKED ISSUE #207: Define `status.moreData` in the spec
Currently our code sets the `status.moreData` in case of a `deleteMany|updateMany` commands. I guess other many commands, like update, have the same behavior.

However, there are several unclear things that should be documented and that will help me continue with the #178:

* is the `status.moreData` set to false in case there is no more? currently we don't send this status back at all in case there is no data
* is the `status.moreData` set in case of `deleteOne` command? if so, we would need to change the current logic

ivansenic:
**What this PR does**:
Documents `status.moreData`.

Fixes #207

### DIFF
--- a/docs/jsonapi-spec.textile
+++ b/docs/jsonapi-spec.textile
@@ -389,7 +389,7 @@ h4(#commandDeleteManyResponse). deleteMany Command Response
 
 |_. Response Element |_. Description |
 | @data@ | Not present. |
-| @status@ | Preset with fields: @deletedCount: <zero-or-positive-integer>@ |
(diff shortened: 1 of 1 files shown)


jeffreyscarpenter (approved):

docs/jsonapi-spec.textile
```diff
@@ -389,7 +389,7 @@ h4(#commandDeleteManyResponse). deleteMany Command Response
 
 |_. Response Element |_. Description |
 | @data@ | Not present. |
-| @status@ | Preset with fields: @deletedCount: <zero-or-positive-integer>@ |
+| @status@ | Preset with fields: @deletedCount: <zero-or-positive-integer>@, @moreData@ with value @true@ if there are more documents that match the delete @filter@, but were not delete since the limit of documents to delete in the single operation was hit. |
```
tatu-at-datastax, line 392: "were not delete" -> "were not deleted"

PR #264, opened 2023-03-15
cancel running ci workflows for pr branches

ivansenic:
**What this PR does**:
Cancel existing workflows for PR branches on update. Meaning once you push a new commit to a PR branch, we will cancel the workflow that might still be running. This ensures even more saving of minutes and cancelling runs of unnecessary CI.

Port of https://github.com/stargate/stargate/issues/2493

### DIFF
--- a/.github/workflows/continuous-integration.yaml
+++ b/.github/workflows/continuous-integration.yaml
@@ -17,6 +17,11 @@ on:
 
   workflow_dispatch:
 
+# cancel same workflows in progress for pull request branches
+concurrency:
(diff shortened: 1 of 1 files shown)


.github/workflows/continuous-integration.yaml
```diff
@@ -35,6 +40,9 @@ jobs:
     name: Unit tests
     runs-on: ubuntu-latest
 
+    # max run time 5 minutes
+    timeout-minutes: 5
```
ivansenic, line 44: added this as well.. sometimes jobs could hang and we want to avoid this..

PR #265, opened 2023-03-15
closes #178: findOne and findOneUpdate per spec, improved tests

### LINKED ISSUE #178: Confirm all response messages against spec

ivansenic:
**What this PR does**:

Adaptation to spec for the `findOne` & `findOneAndUpdate`, including test improvements.

* [x] removed option class from `findOne` command as per spec no options exists
* [x] validation and documentation of options in the `findOneAndUpdate`
* improved tests:
   * [x] added command validation tests
   * [x] refactored, extended resolver tests, test all options 
   * [x] refactored operation tests
   * [x] added error case in operation tests
      * added cases with not matching and missing fields as well
      * assert no errors & data
      * improved readability
      * remove test order - not needed
* [x] fixed swagger examples

Closes to #178. (final task)


ivansenic:
@vkarpov15 I need your help quickly. I specified in the `updateOne`, `updateMany` and `findOneAndUpdate` that the `update` clause is mandatory. Does this make any sense or should I revert?

I was thinking there is no sense in not having the `update` clause, but since there is `upsert` option, commands like this do make sense for example:

```
{
  "findOneAndUpdate": {
    "filter" : {"key": "value"},
    "options": {
       "upsert": true
    }
  }
}
``` 

I think I should revert and make `update` optional?

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/FindOneAndUpdateCommand.java
```diff
@@ -17,15 +18,26 @@
...
               message = "returnDocument value can only be 'before' or 'after'")
+          @Schema(
+              description =
+                  "Specifies which document to perform the projection on. If `before` the projection is performed on the document before the update is applied, if `after` the document projection is from the document after the update.",
+              defaultValue = "before")
           String returnDocument,
-      boolean upsert) {}
+      @Schema(
+              description =
+                  "When `true`, if no documents match the `filter` clause the command will create a new _empty_ document and apply the `update` clause to the empty document.",
```
tatu-at-datastax, line 40: Strictly speaking it's not exactly empty I think, as some values are initialized from filter clause.

@maheshrajamani Is this correct? And if so, how are default values from filter clause selected/applied (only equality ones)

maheshrajamani: Yes, When new document is created all equals filter's are added to the document.

ivansenic: Will fix. Should be fixed in update commands as well

vkarpov15: @maheshrajamani is right about this one, any equals filters are added to the document before updates are applied.

vkarpov15 (approved):
One quick comment but otherwise LGTM

ivansenic:
> @vkarpov15 I need your help quickly. I specified in the `updateOne`, `updateMany` and `findOneAndUpdate` that the `update` clause is mandatory. Does this make any sense or should I revert?
> 
> I was thinking there is no sense in not having the `update` clause, but since there is `upsert` option, commands like this do make sense for example:
> 
> ```
> {
>   "findOneAndUpdate": {
>     "filter" : {"key": "value"},
>     "options": {
>        "upsert": true
>     }
>   }
> }
> ```
> 
> I think I should revert and make `update` optional?

@vkarpov15 about about this one? :arrow_up:

maheshrajamani (approved):
LGTM

PR #267, opened 2023-03-16
Add checks to make sure Update Operators do not use conflicting/overlapping paths

### LINKED ISSUE #232: Check that different update operators do not target same field (conflict)

tatu-at-datastax:
**What this PR does**:

Improves checking of dotted paths used with update operator to ensure overlapping paths are not allowed (same as with other mongoose backends).

Fixes #232

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/update/UpdateClause.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/update/UpdateClause.java
@@ -74,6 +74,16 @@ public record UpdateClause(EnumMap<UpdateOperator, ObjectNode> updateOperationDe
 
       while (it.hasNext()) {
         var curr = it.next();
+        PathMatchLocator prevLoc = prev.getKey();
+        PathMatchLocator currLoc = curr.getKey();
+        // So, if parent/child, parent always first so this check is enough
+        if (currLoc.isSubPathOf(prevLoc)) {
+          throw new JsonApiException(
+              ErrorCode.UNSUPPORTED_UPDATE_OPERATION_PARAM,
+              "Update operator path conflict due to overlap: '%s' (%s) vs '%s' (%s)"
+                  .formatted(
+                      prevLoc, prev.getValue().operator(), currLoc, curr.getValue().operator()));
+        }
       }
     }
     actionMap.entrySet().forEach(e -> {});
--- a/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/update/UpdateClause.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/update/UpdateClause.java
@@ -86,7 +86,6 @@ public record UpdateClause(EnumMap<UpdateOperator, ObjectNode> updateOperationDe
         }
       }
     }
-    actionMap.entrySet().forEach(e -> {});
 
     return new ArrayList<>(operationMap.values());
   }
--- a/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/update/UpdateClause.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/update/UpdateClause.java
@@ -5,10 +5,12 @@ import com.fasterxml.jackson.databind.node.ObjectNode;
 import io.stargate.sgv2.jsonapi.api.model.command.deserializers.UpdateClauseDeserializer;
 import io.stargate.sgv2.jsonapi.exception.ErrorCode;
 import io.stargate.sgv2.jsonapi.exception.JsonApiException;
+import io.stargate.sgv2.jsonapi.util.PathMatchLocator;
 import java.util.ArrayList;
 import java.util.EnumMap;
 import java.util.List;
 import java.util.Set;
+import java.util.TreeMap;
 import java.util.stream.Collectors;
 import org.eclipse.microprofile.openapi.annotations.enums.SchemaType;
 import org.eclipse.microprofile.openapi.annotations.media.Schema;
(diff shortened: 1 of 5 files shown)


src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/update/UpdateClause.java
```diff
@@ -37,29 +39,54 @@ public record UpdateClause(EnumMap<UpdateOperator, ObjectNode> updateOperationDe
...
+              }
+            });
 
-    if ((setOp != null) && (unsetOp != null)) {
-      Set<String> paths = getPaths(setOp);
-      paths.retainAll(getPaths(unsetOp));
+    // Second: check for sub-paths -- efficiently done by using alphabetic-sorted paths
+    // and comparing adjacent ones; for each pair see if later one is longer one and
+    // starts with dot.
+    // For example: "path.x" vs "path.x.some.more" we notice latter has suffix ".some.more"
```
vkarpov15, line 69: I don't remember off the top of my head which characters we allow in field names, but I don't think this will work if we allow spaces, `$`, or `+` in field names. Because "path.x$foo" comes before "path.x.foo" in lexicographical order.

Is there any way we can use a different approach here? I'd rather not add code that blocks us from supporting "$" in field names.

src/test/java/io/stargate/sgv2/jsonapi/service/updater/DocumentUpdaterTest.java
```diff
@@ -235,10 +235,42 @@ public void invalidSetAndUnsetSameField() throws Exception {
...
+    public void invalidMulAndIncSameFieldNested() {
+      Throwable t =
+          catchThrowable(
+              () ->
+                  DocumentUpdater.construct(
+                      DocumentUpdaterUtils.updateClause(
+                          UpdateOperator.INC,
+                          (ObjectNode) objectMapper.readTree("{\"root.x\":-7, \"root.inc\":-3}"),
+                          UpdateOperator.MUL,
+                          (ObjectNode) objectMapper.readTree("{\"root.mul\":3, \"root.x\":2}"))));
```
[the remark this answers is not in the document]
tatu-at-datastax: Yes. I should probably add case to verify.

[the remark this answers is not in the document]
tatu-at-datastax: Added another IT that checks `nested.old` ($unset) and `nested.new` ($set) (same logic between all operators).

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/update/UpdateClause.java
vkarpov15: @tatu-at-datastax the issue is your alphabetical sorting assumption. The assumption here is that `path.x.some.more` would _immediately_ follow `path.x` in the sort order, unless there's some other conflicting path. But in ASCII, `$` is less than `.`. So conventional string sorting would put fields with dollar paths _before_ conflicting paths. Below is an example in JavaScript

```
> ['path.x', 'path.x.some.more', 'path.x$foo'].sort()
[ 'path.x', 'path.x$foo', 'path.x.some.more' ]
```

Not sure if that applies here, just figured I'd call that out.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/update/UpdateClause.java
```diff
@@ -37,29 +39,54 @@ public record UpdateClause(EnumMap<UpdateOperator, ObjectNode> updateOperationDe
...
+        if (currLoc.isSubPathOf(prevLoc)) {
+          throw new JsonApiException(
+              ErrorCode.UNSUPPORTED_UPDATE_OPERATION_PARAM,
+              "Update operator path conflict due to overlap: '%s' (%s) vs '%s' (%s)"
+                  .formatted(
+                      prevLoc, prev.getValue().operator(), currLoc, curr.getValue().operator()));
+        }
       }
     }
+    actionMap.entrySet().forEach(e -> {});
```
maheshrajamani, line 89: Do we need this statement?

maheshrajamani (approved):

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/update/UpdateClause.java
tatu-at-datastax: @vkarpov15 Ah. Ok, let me think this over. Good catch.

tatu-at-datastax: I can make sorting work with some more work. Two obvious extensions:

1. Use modified String comparison where `.` is sorted like `null` byte (value 0)
2. Sort split segments (which Locator already has)

both would work; I may go with (2)

tatu-at-datastax: Added segment-aware sorting, testing to verify functioning.

vkarpov15: Nice, thanks. It looks like the reason why this works though is because you've included a loop over path segments in the `compare()` method - any worries about this being potentially slow? You'll have to iterate over path segments O(nlogn) times, creating a map of all parent paths will likely be faster but more memory intensive. I don't think this is a big deal either way, just food for thought.

tatu-at-datastax: @vkarpov15 that's a reasonable question. I guess it all depends on depth of document. My thinking was that for deeper documents most differences are at first segment(s) anyway so that it evens out, compared to String comparison. But my main consideration was, I guess, between straight String comparison (just changing sort value of `.`) vs segmented, where I think performance is similar. I did not consider possibility of building lookup for covered paths: in that case it does come back to both memory usage added and (more importantly), construction of new String instances to reconstruct parent paths. I guess it can go either way wrt overall performance, in the end.

Fundamentally I guess it also comes down to how many updates do we usually have: I suspect differences are not meaningful until into thousands of operations.

Since this logic is fairly isolated, I think I'll keep the question in mind and can reconsider re-implementation if we find a hotspot.

PR #271, opened 2023-03-16
Add `$rename` update operator

### LINKED ISSUE #165: Implement general update operator `$rename`
Another general-purpose array operator to implement is `$rename`. Its semantics are similar to `$unset` followed by `$set` (if that was allowed), so that:

1. If target path exists, the value is overwritten
2. If source path does not exist, operation does nothing (but does not fail)

Format is simple: 

```
{ "$rename" : { "field1" : "newName1", .... , "fieldN", "newNamen" } }
```

Following error modes exist:

1. `_id` either as source or target: cannot modify immutable property "_id"
2. Same source and target (must differ)
3. Attempt to move source to incompatible path

The last case occurs, for example when an attempt is madeto set a property on Array value: that is, target path points to something that is not possible to add: for example with

```
{ "x" : 3, "array" : [ 1 ] }
```

attempt to apply operation

```
{ "$rename" : { "x" : "array.x" }  }
```

would fail. Same is true for paths "through" atomic fields; only (sub)documents have properties.

tatu-at-datastax:
**What this PR does**:

Adds `$rename` general update operator

Fixes #165

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/api/configuration/ObjectMapperConfiguration.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/api/configuration/ObjectMapperConfiguration.java
@@ -1,6 +1,5 @@
 package io.stargate.sgv2.jsonapi.api.configuration;
 
-import com.fasterxml.jackson.core.JsonGenerator;
 import com.fasterxml.jackson.core.StreamReadFeature;
 import com.fasterxml.jackson.core.StreamWriteFeature;
 import com.fasterxml.jackson.databind.DeserializationFeature;
--- a/src/main/java/io/stargate/sgv2/jsonapi/api/configuration/ObjectMapperConfiguration.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/api/configuration/ObjectMapperConfiguration.java
@@ -1,6 +1,8 @@
 package io.stargate.sgv2.jsonapi.api.configuration;
 
 import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.core.StreamReadFeature;
+import com.fasterxml.jackson.core.StreamWriteFeature;
 import com.fasterxml.jackson.databind.DeserializationFeature;
 import com.fasterxml.jackson.databind.MapperFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
@@ -33,8 +35,10 @@ public class ObjectMapperConfiguration {
         .enable(DeserializationFeature.USE_BIG_DECIMAL_FOR_FLOATS)
         // case insensitive enums, so "before" will match to "BEFORE" in an enum
         .enable(MapperFeature.ACCEPT_CASE_INSENSITIVE_ENUMS)
+        // Verify uniqueness of JSON Object properties
+        .enable(StreamReadFeature.STRICT_DUPLICATE_DETECTION)
         // Prevent use of Engineering Notation with trailing zeroes:
-        .enable(JsonGenerator.Feature.WRITE_BIGDECIMAL_AS_PLAIN)
+        .enable(StreamWriteFeature.WRITE_BIGDECIMAL_AS_PLAIN)
         .build();
   }
 }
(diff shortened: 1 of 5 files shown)


src/main/java/io/stargate/sgv2/jsonapi/api/configuration/ObjectMapperConfiguration.java
```diff
@@ -33,8 +34,10 @@ private ObjectMapper createMapper() {
         .enable(DeserializationFeature.USE_BIG_DECIMAL_FOR_FLOATS)
         // case insensitive enums, so "before" will match to "BEFORE" in an enum
         .enable(MapperFeature.ACCEPT_CASE_INSENSITIVE_ENUMS)
+        // Verify uniqueness of JSON Object properties
+        .enable(StreamReadFeature.STRICT_DUPLICATE_DETECTION)
         // Prevent use of Engineering Notation with trailing zeroes:
-        .enable(JsonGenerator.Feature.WRITE_BIGDECIMAL_AS_PLAIN)
+        .enable(StreamWriteFeature.WRITE_BIGDECIMAL_AS_PLAIN)
```
tatu-at-datastax, line 40: StreamWriteFeature/JsonWriteFeature will replace JsonGenerator.Feature; same behavior in Jackson 2.x but latter removed from 3.0

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/update/RenameOperation.java
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  @Override
+  public boolean updateDocument(ObjectNode doc) {
+    boolean modified = false;
+    for (Action action : actions) {
+      PathMatch src = action.sourceLocator().findIfExists(doc);
+      JsonNode value = src.removeValue();
+      if (value != null) {
+        // Mongo does not allow "renaming" of Array values
```
maheshrajamani, line 45: @tatu-at-datastax Aaron asked to not refer to Mongo in the code base, can you have this removed

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/update/RenameOperation.java
```diff
@@ -0,0 +1,85 @@
...
+              ErrorCode.UNSUPPORTED_UPDATE_OPERATION_PATH.getMessage()
+                  + ": $rename does not allow ARRAY field as source ('"
+                  + action.sourceLocator()
+                  + "')");
+        }
+
+        // If there is a value will be a modification (since source value will
+        // disappear), regardless of whether target changes
+        modified = true;
+        PathMatch dst = action.targetLocator().findOrCreate(doc);
```
maheshrajamani, line 58: There is document which has address sub-document like {"address" : {"address line" : "123 main st", "city" : "NYC", "state": "NY"}} 
If the update request is to rename city to town, the sub-document fields (city and state) order will change as {"address" : {"address line" : "123 main st", "state": "NY", "city" : "NYC"}}? Is this the expected behavior?

maheshrajamani (approved):

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/update/RenameOperation.java
tatu-at-datastax: @maheshrajamani As far as I can see, mongo docs state (and Atlas seem to implement it like so) that `$rename` is implemented as "$unset followed by $set"; and $set appends at the end.

So I think it would indeed be expected that field "moves" the way you describe; `city` removed, `town` appended.

tatu-at-datastax:
One thing I may want to do is to check the case of 'src == dst` -- looks like that needs to become an error too.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/update/RenameOperation.java
tatu-at-datastax: Will change yes.

PR #281, opened 2023-03-21
improve readbility and flow in the ReadAndUpdateOperation#processUpdate

ivansenic:
**What this PR does**:

Improves readability and the flow in the `ReadAndUpdateOperation#processUpdate`..

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/ReadAndUpdateOperation.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/ReadAndUpdateOperation.java
@@ -138,42 +138,52 @@ public record ReadAndUpdateOperation(
       ReadDocument document, QueryExecutor queryExecutor, AtomicInteger modifiedCount) {
     return Uni.createFrom()
         .item(document)
-        // Perform update operation and save only if data is modified.
-        .onItem()
-        .transformToUni(
+
+        // perform update operation and save only if data is modified.
+        .flatMap(
             readDocument -> {
+
+              // if there is no document return null item
               if (readDocument == null) {
                 return Uni.createFrom().nullItem();
               }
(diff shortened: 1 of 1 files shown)


src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/ReadAndUpdateOperation.java
```diff
@@ -138,42 +138,52 @@ private Uni<UpdatedDocument> processUpdate(
...
-              Uni<DocumentId> updated = Uni.createFrom().nullItem();
-              final WritableShreddedDocument writableShreddedDocument;
-              if (documentUpdaterResponse.modified()) {
-                writableShreddedDocument = shredder().shred(updatedDocument, readDocument.txnId());
-                updated = updatedDocument(queryExecutor, writableShreddedDocument);
-              } else {
-                writableShreddedDocument = null;
+                  documentUpdater().applyUpdates(readDocument.document().deepCopy(), upsert);
+              if (!documentUpdaterResponse.modified()) {
+                return Uni.createFrom().nullItem();
```
maheshrajamani, line 160: We need to do null check on responses collected as list and passed to UpdateOperationPage

maheshrajamani (approved):

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/ReadAndUpdateOperation.java
maheshrajamani: Apologies, didn't realize its going into collect() that should handle null value.

ivansenic: I did not change at all the previous logic :)

PR #282, opened 2023-03-22
separate nosqlbench execution in spaces to have more connections

ivansenic:
**What this PR does**:
Enable nosqlbench test to send requests over multiple HTTP2 connections, thus allowing high number of threads and custom configurability. We can now specify how many threads to run and how many connections to use, for example:

* `threads=30 connections=30` - each request has it's own connections
* `threads=30 connections=3` - group of 10 request would use same connection (sequence `1, 4, 7, etc`)
* `threads=30 connections=1` - all requests executed over a single connection (multiplex and limited by the max http2 concurrent streams)

### DIFF
--- a/.github/workflows/performance-testing.yaml
+++ b/.github/workflows/performance-testing.yaml
@@ -10,7 +10,12 @@ on:
         description: 'Number of documents to use for each test'
         required: true
         type: string
-        default: '20000'
+        default: '1000000'
+      connections:
+        description: 'Number of HTTP connections to use for each test'
+        required: true
+        type: string
+        default: '20'
 
 # global env vars, available in all jobs and steps
 env:
@@ -92,7 +97,7 @@ jobs:
         run: |
           cd nosqlbench
           export AUTH_TOKEN=$(curl -s -X POST 'http://localhost:8081/v1/auth' -H 'Content-Type: application/json' --data-raw '{"username": "cassandra", "password": "cassandra"}' | jq -r '.authToken')
-          ../nb5 -v --report-csv-to logs --log-histograms logs/hdr_metrics.log ${{ matrix.test }} jsonapi_host=localhost auth_token=$AUTH_TOKEN docscount=${{ inputs.docscount }}
+          ../nb5 -v --report-csv-to logs --log-histograms logs/hdr_metrics.log ${{ matrix.test }} jsonapi_host=localhost auth_token=$AUTH_TOKEN docscount=${{ inputs.docscount }} spaces=${{ inputs.connections }}
 
       - name: Save Results
         uses: actions/upload-artifact@v3
--- a/.github/workflows/performance-testing.yaml
+++ b/.github/workflows/performance-testing.yaml
@@ -97,7 +97,7 @@ jobs:
         run: |
           cd nosqlbench
           export AUTH_TOKEN=$(curl -s -X POST 'http://localhost:8081/v1/auth' -H 'Content-Type: application/json' --data-raw '{"username": "cassandra", "password": "cassandra"}' | jq -r '.authToken')
-          ../nb5 -v --report-csv-to logs --log-histograms logs/hdr_metrics.log ${{ matrix.test }} jsonapi_host=localhost auth_token=$AUTH_TOKEN docscount=${{ inputs.docscount }} spaces=${{ inputs.connections }}
+          ../nb5 -v --report-csv-to logs --log-histograms logs/hdr_metrics.log ${{ matrix.test }} jsonapi_host=localhost auth_token=$AUTH_TOKEN docscount=${{ inputs.docscount }} connections=${{ inputs.connections }}
 
       - name: Save Results
         uses: actions/upload-artifact@v3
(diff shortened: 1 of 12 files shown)


nosqlbench/http-jsonapi-crud-basic.md
```diff
@@ -58,6 +58,7 @@ Note that error handling is set to `errors=timer,warn`, which means that in case
 ## Workload Parameters
 
 - `docscount` - the number of documents to process in each step of a scenario (default: `10_000_000`)
+- `connections` - number of HTTP2 connections to be shared between the threads (default: `20`)
```
jeffreyscarpenter, line 61: how does this relate to the threads? If I set threads=10 and connections=20 what happens? Should we provide guidance on whether these need to be tuned together?

Would it make sense to reframe this as a number of connections per thread, or perhaps I misunderstand how this works?

nosqlbench/http-jsonapi-crud-basic.yaml
```diff
@@ -91,6 +95,7 @@ blocks:
     ops:
       # aka insertOne
       write-document:
+        space: "{space}"
```
jeffreyscarpenter, line 98: Do we have to add this to every instruction because of a shortcoming in NoSQLBench?

nosqlbench/http-jsonapi-crud-basic.md
ivansenic: * `threads=10 and connections=20` - each thread own connection..
* connections per thread would be great, but I don't know how to do that? do you know?

nosqlbench/http-jsonapi-crud-basic.yaml
ivansenic: Yes, see: https://github.com/nosqlbench/nosqlbench/issues/1182

.github/workflows/performance-testing.yaml
```diff
@@ -92,10 +97,11 @@ jobs:
         run: |
           cd nosqlbench
           export AUTH_TOKEN=$(curl -s -X POST 'http://localhost:8081/v1/auth' -H 'Content-Type: application/json' --data-raw '{"username": "cassandra", "password": "cassandra"}' | jq -r '.authToken')
-          ../nb5 -v --report-csv-to logs --log-histograms logs/hdr_metrics.log ${{ matrix.test }} jsonapi_host=localhost auth_token=$AUTH_TOKEN docscount=${{ inputs.docscount }}
+          ../nb5 -v --report-csv-to logs --log-histograms logs/hdr_metrics.log ${{ matrix.test }} jsonapi_host=localhost auth_token=$AUTH_TOKEN docscount=${{ inputs.docscount }} connections=${{ inputs.connections }}
```
jeffreyscarpenter, line 100: @ivansenic added option to workflow

docker-compose/start_dse_68.sh
```diff
@@ -47,4 +47,11 @@ export JSONIMAGE
 
 echo "Running with DSE $DSETAG, Stargate $SGTAG, JSON API $JSONIMAGE:$JSONTAG"
 
-docker-compose up -d
+COMPOSE_ARGS=("-d")
```
jeffreyscarpenter, line 50: Adding the `--wait` option to docker compose (if available). Lack of this option appears to be the cause of the failure in this workflow: https://github.com/stargate/jsonapi/actions/runs/4494697370. The schema creation command(s) failed due to the JSON API not being all the way up, causing following calls in the NB workload to fail. (This parallels errors seen in automated testing of Postman scripts against Docker compose-based configuration when `--wait` flag was not used.)

nosqlbench/http-jsonapi-crud-basic.md
jeffreyscarpenter: Let's not worry about this now.

nosqlbench/http-jsonapi-crud-basic.yaml
```diff
@@ -152,12 +158,16 @@ blocks:
...
         method: POST
         uri: <<protocol:http>>://{weighted_hosts}:<<jsonapi_port:8080>><<path_prefix:>>/v1/<<namespace:jsonapi_crud_basic>>/<<collection:docs_collection>>
         Accept: "application/json"
         X-Cassandra-Request-Id: "{request_id}"
         X-Cassandra-Token: "<<auth_token:my_auth_token>>"
         Content-Type: "application/json"
+        # because this is not an upsert, modified count could technically be 0 or 1,
+        # but since we are fixing the _ids to be sequential over the docscount range during the write phase,
+        # every update during this phase should update exactly 1 document.
         ok-body: ".*\"modifiedCount\":1.*"
```
jeffreyscarpenter, line 171: @ivansenic added clarifying comment

docker-compose/start_dse_68.sh
ivansenic: OK cool..

ivansenic:
@jeffreyscarpenter merging this, thanks for all the contributions

PR #283, opened 2023-03-22
Refactor FilterableResolver to address #201

### LINKED ISSUE #201: Separate concerns in the `ReadOperation`
Currently the [ReadOperation](https://github.com/stargate/jsonapi/blob/main/src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/ReadOperation.java) serves as a base class for `FindOperation` and `CountOperation`. However, the count operation does not implement any of the abstract methods:

```
  @Override
  public Uni<FindResponse> getDocuments(QueryExecutor queryExecutor, String pagingState) {
    return Uni.createFrom().failure(new JsonApiException(ErrorCode.UNSUPPORTED_OPERATION));
  }

  @Override
  public ReadDocument getNewDocument() {
    throw new JsonApiException(ErrorCode.UNSUPPORTED_OPERATION);
  }
```

In addition the count related method `countDocuments` is used only by the `CountOperation`, thus is not needed at all in this interface. This all looks like a bad design.

Tasks:
* Count operation does not need to extend the read operation
   * Move count related method to the count operation
* Same can be done for Find operation, effectivelly leaving Read operation as empty interface (then count can continue using it)
* In order for this to work the `FilterableResolver` class should be transitioned to `FilterableResolverService` and resolver can directly use this service to resolve the filter clause

@amorton @amorton  Please review and approve issue.

maheshrajamani:
**What this PR does**:
Refactored FilterableResolver to resolve to DBFilters.
Change the update and delete method to use FindOperation instead of ReadOperation
Removed getDocuments and getNewDocument from ReadOperation interface.
Refactored the test cases
Fixes #201 

**Checklist**
- [x] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/ReadType.java
```diff
@@ -13,7 +13,5 @@ public enum ReadType {
    * Return only document id and transaction id of documents which satisfies the filter conditions
    * as part of response
    */
-  KEY,
-  /** Return only count of documents which satisfies the filter condition */
-  COUNT
+  KEY
```
maheshrajamani, line 16: Holding on to this Enum, will be adding SORT_DOCUMENT type in the next PR for in memory sory.

tatu-at-datastax (approved):
LGTM

PR #286, opened 2023-03-22
Fix #158: support empty Projection clause; add full decoding of projection definition (but not yet processing)

### LINKED ISSUE #158: Support missing or empty projection clause for commands
All commands that support a `projection` clause must support the following: 

- `projection` field not present in the command JSON object
- `projection` field present, but an empty JSON object. 

If the `projection` field is present, but is not a JSON object (e.g. `{"projection" : "yes please :)"}`) the service should return an error. 

 See this usecase form the client https://github.com/stargate/stargate-mongoose/discussions/18#discussioncomment-5027700

tatu-at-datastax:
**What this PR does**:

Adds support for optional empty "projection" parameter, validation that if passed its JSON Object.
Serves as the base for actual projection implementation

Fixes #158

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/ReadAndUpdateOperation.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/ReadAndUpdateOperation.java
@@ -12,7 +12,6 @@ import io.stargate.sgv2.jsonapi.service.bridge.executor.QueryExecutor;
 import io.stargate.sgv2.jsonapi.service.bridge.serializer.CustomValueSerializers;
 import io.stargate.sgv2.jsonapi.service.operation.model.ModifyOperation;
 import io.stargate.sgv2.jsonapi.service.operation.model.ReadOperation;
-import io.stargate.sgv2.jsonapi.service.projection.DocumentProjector;
 import io.stargate.sgv2.jsonapi.service.shredding.Shredder;
 import io.stargate.sgv2.jsonapi.service.shredding.model.DocumentId;
 import io.stargate.sgv2.jsonapi.service.shredding.model.WritableShreddedDocument;
@@ -182,10 +181,11 @@ public record ReadAndUpdateOperation(
                         if (returnDocumentInResponse) {
                           documentToReturn =
                               returnUpdatedDocument ? updatedDocument : originalDocument;
-                          // In this case, we'll apply projection on before/after document (was
-                          // not applied during find() phase
-                          DocumentProjector projector = findOperation().projection();
-                          projector.applyProjection(documentToReturn);
+                          // 24-Mar-2023, tatu: Will need to add projection in follow-up PR;
+                          //   one we get here is identity-projection which does nothing.
+                          //
+                          // DocumentProjector projector = findOperation().projection();
+                          // projector.applyProjection(documentToReturn);
                         }
                         return new UpdatedDocument(
                             writableShreddedDocument.id(), upsert, documentToReturn, null);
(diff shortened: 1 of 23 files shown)


tatu-at-datastax:
Right now this will only support empty "projection", fail on non-empty. I will create a follow-up for completing handling but want to get this merged so that we can reconcile work for sorting and keep PRs from getting huge.

vkarpov15:
On a related note @tatu-at-datastax , what do you think about adding support for empty `options` object as well? Right now, Mongoose always sends empty `options` to `deleteOne()`, for example, which always fails. We can strip that out in the driver layer, but I think it would be better for the jsonapi to handle that.

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/model/impl/FindOneAndUpdateCommandResolver.java
```diff
@@ -67,6 +67,14 @@ public Operation resolveCommand(CommandContext commandContext, FindOneAndUpdateC
   private FindOperation getFindOperation(
       CommandContext commandContext, FindOneAndUpdateCommand command) {
     List<DBFilterBase> filters = resolve(commandContext, command);
-    return new FindOperation(commandContext, filters, null, 1, 1, ReadType.DOCUMENT, objectMapper);
+    return new FindOperation(
+        commandContext,
+        filters,
+        command.buildProjector(),
```
maheshrajamani, line 73: Not sure this is correct, the FindOperation inside FindOneAndUpdate should be passed identityProjector() because the document is merged with updates. Need to apply projection on top document that is returned [here] (https://github.com/stargate/jsonapi/blob/b6d4f73be517cf2e90fb50384e8e60f411754f99/src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/ReadAndUpdateOperation.java#L183).  This needs Projection passed into ReadAndUpdateOperation

maheshrajamani (changes requested):

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/model/impl/FindOneAndUpdateCommandResolver.java
tatu-at-datastax: Makes sense, will change.

tatu-at-datastax: Changed in 2 commits.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/ReadAndUpdateOperation.java
```diff
@@ -181,6 +182,10 @@ private Uni<UpdatedDocument> processUpdate(
                         if (returnDocumentInResponse) {
                           documentToReturn =
                               returnUpdatedDocument ? updatedDocument : originalDocument;
+                          // In this case, we'll apply projection on before/after document (was
+                          // not applied during find() phase
+                          DocumentProjector projector = findOperation().projection();
```
maheshrajamani, line 187: Think ReadAndUpdateOperation get the projection in it's constructor? As stated in your comment is it part of different commit?

tatu-at-datastax: Right, this will not work; it's the identity one.

maheshrajamani: Approved.

tatu-at-datastax: Discussed offline, yes, this would not work. Commented out and will resolve in follow-up PR; merging to avoid blocking.

tatu-at-datastax:
Merged; had to re-run cancelled CI once.

PR #290, opened 2023-03-23
Set serial consistency for write operations

### LINKED ISSUE #238: Setting Serial consistency for write operation
All the write operation are run as LWT. This needs Serial consistency to be set.

By default set serial consistency to SERIAL. In case the write consistency is set as LOCAL_QUORUM set the serial consistency to LOCAL_SERIAL.

maheshrajamani:
**What this PR does**:
Set serial consistency for write as per cql [documentation](https://docs.datastax.com/en/cassandra-oss/3.0/cassandra/dml/dmlConfigSerialConsistency.html).
Fixes #238 

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/.github/workflows/continuous-integration.yaml
+++ b/.github/workflows/continuous-integration.yaml
@@ -61,7 +61,7 @@ jobs:
   int-tests:
     name: Integration tests
     needs: [ build ]
-    runs-on: ubuntu-latest-4-cores
+    runs-on: sg-ubuntu-latest-4-cores
 
     # max run time 20 minutes
     timeout-minutes: 20
--- a/.github/workflows/continuous-integration.yaml
+++ b/.github/workflows/continuous-integration.yaml
@@ -38,7 +38,7 @@ jobs:
   # runs unit tests
   build:
     name: Unit tests
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-latest.04-16core
 
     # max run time 5 minutes
     timeout-minutes: 5
--- a/.github/workflows/continuous-integration.yaml
+++ b/.github/workflows/continuous-integration.yaml
@@ -38,7 +38,7 @@ jobs:
   # runs unit tests
   build:
     name: Unit tests
-    runs-on: ubuntu-latest-4core
+    runs-on: ubuntu-latest
 
     # max run time 5 minutes
     timeout-minutes: 5
@@ -61,7 +61,7 @@ jobs:
   int-tests:
     name: Integration tests
     needs: [ build ]
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04-4core
 
     # max run time 20 minutes
     timeout-minutes: 20
--- a/.github/workflows/continuous-integration.yaml
+++ b/.github/workflows/continuous-integration.yaml
@@ -61,7 +61,7 @@ jobs:
   int-tests:
     name: Integration tests
     needs: [ build ]
-    runs-on: ubuntu-20.04-4core
+    runs-on: ubuntu-latest-4-cores
 
     # max run time 20 minutes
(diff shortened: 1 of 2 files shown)


.github/workflows/continuous-integration.yaml
```diff
@@ -61,7 +61,7 @@ jobs:
   int-tests:
     name: Integration tests
     needs: [ build ]
-    runs-on: ubuntu-latest
+    runs-on: sg-ubuntu-latest-4-cores
```
ivansenic, line 64: is this really making any changes?

src/main/java/io/stargate/sgv2/jsonapi/service/bridge/executor/QueryExecutor.java
```diff
@@ -67,8 +67,13 @@ public Uni<QueryOuterClass.ResultSet> executeWrite(QueryOuterClass.Query query)
...
     QueryOuterClass.ConsistencyValue.Builder consistencyValue =
         QueryOuterClass.ConsistencyValue.newBuilder().setValue(consistency);
+    QueryOuterClass.Consistency serialConsistency = queriesConfig.serialConsistency();
+    QueryOuterClass.ConsistencyValue.Builder serialConsistencyValue =
+        QueryOuterClass.ConsistencyValue.newBuilder().setValue(serialConsistency);
     QueryOuterClass.QueryParameters.Builder params =
-        QueryOuterClass.QueryParameters.newBuilder().setConsistency(consistencyValue);
+        QueryOuterClass.QueryParameters.newBuilder()
+            .setConsistency(consistencyValue)
+            .setSerialConsistency(serialConsistencyValue);
```
ivansenic, line 76: Please extend unit test to confirm that configured serial consistency is set on writes..

ivansenic (commented):
I am still wandering why do we need bigger runner? I would rather say this is not and should not be solution..

.github/workflows/continuous-integration.yaml
maheshrajamani: Yes. After making this change there were no timeout errors.

src/main/java/io/stargate/sgv2/jsonapi/service/bridge/executor/QueryExecutor.java
ivansenic: No I meant the uni test of the `QueryExecutor` class? Yes it's fine, the configuration is tested, but there is not test that configuration is being used in the `QueryExecutor`, we need this.. The main point of such tests is that if somebody tomorrow changes these lines above we would have a failing test.. Otherwise, deleting this would result in green CI, "showcasing" that is all fine.

ivansenic:
> > I am still wandering why do we need bigger runner? I would rather say this is not and should not be solution..
> 
> This was decided after discussion on Friday standup. The error returned was from dse cassandra because query time out for lwt transaction even before entering the lwt logic. This error can happen even without setting serial consistency as reported in Issue #208 . The other option is to run tests with increased timeout value for cas, this will again don't gurantee the tests will finish successfully.

I don't really understand this.. There is a "single user - single op" being run on a machine and there is a C* timeout related to LTW. And it was decided that we simply need a bigger runner? How's that, I mean how is this not bring more concern and action, than simply increasing the hardware size? Does this mean we can not even have a single user on such hardware?

ivansenic (changes requested):
Blocking this until we resolve the actual problem.

ivansenic:
@maheshrajamani do we need to change the consistency (not serial-consistency) along side these changes? Currently for all read/write/schema-change we have `LOCAL_QUORUM`..

maheshrajamani:
Closing for a new PR

PR #296, opened 2023-03-27
In memory sort - merged

### LINKED ISSUE #272: Implement in-memory sort
1. Read index columns for the sort fields. Along with it read the doc_json Example with sort field name run the query as 
           select key, doc_json, query_bool_values[‘sort field name’], query_text_values[‘sort field name’], query_dbl_values[‘sort field name’], query_null_values[‘sort field name’] from <table> where <conditions in filter> 
	  Example:
           select key, doc_json, query_text_values['amount'], query_dbl_values['amount'], query_bool_values['amount'], query_null_values['amount'], query_text_values['data'],  query_null_values['data'] from purchase;
2. Read up to max_read_limit + 1 keys to sort, if there are (max_read_limit + 1) keys returned throw errors suggesting to add more filters, else proceed.
3. Deserialize only the sort fields. Use JsonNodeComparator to the sort keys, use MinMaxPriorityQueue to evict record outside the required limit of keys.
4. Deserialize the doc_jon value only for the required keys and return response

maheshrajamani:
**What this PR does**:
In memory sort - merged
Fixes #272 

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCode.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCode.java
@@ -8,7 +8,7 @@ public enum ErrorCode {
 
   CONCURRENCY_FAILURE("Unable to complete transaction due to concurrent transactions"),
 
-  DATASET_TOO_BIG("Response data set too big, add more filters"),
+  DATASET_TOO_BIG("Response data set too big to be sorted, add more filters"),
 
   DOCUMENT_ALREADY_EXISTS("Document already exists with the given _id"),
 
--- a/src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCode.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCode.java
@@ -8,6 +8,8 @@ public enum ErrorCode {
 
   CONCURRENCY_FAILURE("Unable to complete transaction due to concurrent transactions"),
 
+  DATASET_TOO_BIG("Response data set too big, add more filters"),
+
   DOCUMENT_ALREADY_EXISTS("Document already exists with the given _id"),
 
   DOCUMENT_UNPARSEABLE("Unable to parse the document"),
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/bridge/config/DocumentConfig.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/bridge/config/DocumentConfig.java
@@ -26,17 +26,17 @@ import javax.validation.constraints.Positive;
 @ConfigMapping(prefix = "stargate.document")
 public interface DocumentConfig {
 
-  /** @return Defines the maximum document page size, defaults to <code>500</code>. */
-  @Max(500)
-  @Positive
-  @WithDefault("100")
-  int maxPageSize();
-
   /** @return Defines the default document page size, defaults to <code>20</code>. */
   @Max(500)
   @Positive
   @WithDefault("20")
   int defaultPageSize();
+
+  /** @return Defines the default document page size, defaults to <code>100</code>. */
+  @Max(500)
+  @Positive
+  @WithDefault("100")
+  int defaultSortPageSize();
 
   /**
    * @return Defines the maximum limit of document that can be returned for a request, defaults to
@@ -46,6 +46,15 @@ public interface DocumentConfig {
   @Positive
   @WithDefault("1000")
   int maxLimit();
+
+  /**
+   * @return Defines the maximum limit of document read to perform in memory sorting <code>10000
+   *     </code>.
+   */
+  @Max(10000)
+  @Positive
+  @WithDefault("10000")
+  int maxSortReadLimit();
 
   /**
    * @return Defines the maximum limit of document that can be deleted for a request, defaults to
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/bridge/config/DocumentConfig.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/bridge/config/DocumentConfig.java
@@ -32,7 +32,10 @@ public interface DocumentConfig {
   @WithDefault("20")
   int defaultPageSize();
 
-  /** @return Defines the default document page size, defaults to <code>100</code>. */
+  /**
+   * @return Defines the default document page size for sorting, having separate config because sort
+   *     will more rows in per page, defaults to <code>100</code>.
+   */
   @Max(500)
   @Positive
   @WithDefault("100")
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/ChainedComparator.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/ChainedComparator.java
@@ -1,0 +1,27 @@
+package io.stargate.sgv2.jsonapi.service.operation.model;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import io.stargate.sgv2.jsonapi.service.operation.model.impl.FindOperation;
+import io.stargate.sgv2.jsonapi.service.operation.model.impl.ReadDocument;
+import io.stargate.sgv2.jsonapi.util.JsonNodeComparator;
+import java.util.Comparator;
+import java.util.List;
+
+public record ChainedComparator(List<FindOperation.OrderBy> sortColumns)
+    implements Comparator<ReadDocument> {
+  @Override
+  public int compare(ReadDocument o1, ReadDocument o2) {
+    int i = 0;
+    final List<JsonNode> sortValuesO1 = o1.sortColumns();
+    final List<JsonNode> sortValuesO2 = o2.sortColumns();
+    for (FindOperation.OrderBy sortColumn : sortColumns()) {
+      int compareValue =
+          sortColumn.ascending()
+              ? JsonNodeComparator.ascending().compare(sortValuesO1.get(i), sortValuesO2.get(i))
+              : JsonNodeComparator.descending().compare(sortValuesO1.get(i), sortValuesO2.get(i));
+      if (compareValue != 0) return compareValue;
+      i++;
+    }
+    return 0;
+  }
+}
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/ReadOperation.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/ReadOperation.java
@@ -34,11 +34,11 @@ public interface ReadOperation extends Operation {
   String[] documentKeyColumns = {"key", "tx_id"};
   String[] sortedDataColumns = {"key", "doc_json"};
   int SORTED_DATA_COLUMNS = sortedDataColumns.length;
-  List<String> sortColumns =
+  List<String> sortIndexColumns =
       List.of(
           "query_text_values['%s']",
           "query_dbl_values['%s']", "query_bool_values['%s']", "query_null_values['%s']");
-  int SORT_INDEX_COLUMNS_SIZE = sortColumns.size();
+  int SORT_INDEX_COLUMNS_SIZE = sortIndexColumns.size();
   /**
    * Default implementation to query and parse the result set
    *
@@ -77,7 +77,7 @@ public interface ReadOperation extends Operation {
                     projection.applyProjection(root);
                   }
                   document =
-                      new ReadDocument(
+                      ReadDocument.from(
                           getDocumentId(row.getValues(0)), // key
                           Values.uuid(row.getValues(1)), // tx_id
                           root);
@@ -175,7 +175,7 @@ public interface ReadOperation extends Operation {
                 // Create ReadDocument with document id, grpc value for doc json and list of sort
                 // values
                 document =
-                    new ReadDocument(
+                    ReadDocument.from(
                         getDocumentId(row.getValues(0)), // key
                         row.getValues(1), // Deserialized value of doc_json
                         sortValues);
@@ -218,7 +218,7 @@ public interface ReadOperation extends Operation {
                             try {
                               final JsonNode jsonNode =
                                   objectMapper.readTree(Values.string(readDoc.docJsonValue()));
-                              return new ReadDocument(readDoc.id(), readDoc.txnId(), jsonNode);
+                              return ReadDocument.from(readDoc.id(), readDoc.txnId(), jsonNode);
                               // This error should never happen because we are creating Json object
                               // out of the validated json stored in DB.
                             } catch (JsonProcessingException e) {
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/ReadOperation.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/ReadOperation.java
@@ -3,6 +3,9 @@ package io.stargate.sgv2.jsonapi.service.operation.model;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.JsonNodeFactory;
+import com.google.common.collect.MinMaxPriorityQueue;
+import io.smallrye.mutiny.Multi;
 import io.smallrye.mutiny.Uni;
 import io.stargate.bridge.grpc.BytesValues;
 import io.stargate.bridge.grpc.Values;
@@ -14,9 +17,13 @@ import io.stargate.sgv2.jsonapi.service.operation.model.impl.ReadDocument;
 import io.stargate.sgv2.jsonapi.service.projection.DocumentProjector;
 import io.stargate.sgv2.jsonapi.service.shredding.model.DocumentId;
 import java.util.ArrayList;
+import java.util.Comparator;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Optional;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.stream.Collectors;
 
 /**
  * ReadOperation interface which all find command operations will use. It also provides the
@@ -25,7 +32,13 @@ import java.util.Optional;
 public interface ReadOperation extends Operation {
   String[] documentColumns = {"key", "tx_id", "doc_json"};
   String[] documentKeyColumns = {"key", "tx_id"};
-
+  String[] sortedDataColumns = {"key", "doc_json"};
+  int SORTED_DATA_COLUMNS = sortedDataColumns.length;
+  List<String> sortColumns =
+      List.of(
+          "query_text_values['%s']",
+          "query_dbl_values['%s']", "query_bool_values['%s']", "query_null_values['%s']");
+  int SORT_INDEX_COLUMNS_SIZE = sortColumns.size();
   /**
    * Default implementation to query and parse the result set
    *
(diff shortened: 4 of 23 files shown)


src/test/java/io/stargate/sgv2/jsonapi/api/v1/FindIntegrationTest.java
```diff
@@ -772,5 +780,139 @@ public void findByBooleanColumn() {
...
+      given()
+          .header(HttpConstants.AUTHENTICATION_TOKEN_HEADER_NAME, getAuthToken())
+          .contentType(ContentType.JSON)
+          .body(json)
+          .when()
+          .post(CollectionResource.BASE_PATH, keyspaceId.asInternal(), collectionName)
+          .then()
+          .statusCode(200)
+          .body("data.count", is(20))
+          .body("data.docs", jsonEquals(arrayNode.toString()));
```
maheshrajamani, line 816: Is there better way to check if the documents present in sorted order?

src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCode.java
```diff
@@ -8,6 +8,8 @@ public enum ErrorCode {
 
   CONCURRENCY_FAILURE("Unable to complete transaction due to concurrent transactions"),
 
+  DATASET_TOO_BIG("Response data set too big, add more filters"),
```
tatu-at-datastax, line 11: Should this refer to sorting ("too big to be sorted"?), as it won't happen with regular Find command?

src/main/java/io/stargate/sgv2/jsonapi/service/bridge/config/DocumentConfig.java
```diff
@@ -26,18 +26,18 @@
...
-  @WithDefault("100")
-  int maxPageSize();
-
   /** @return Defines the default document page size, defaults to <code>20</code>. */
   @Max(500)
   @Positive
   @WithDefault("20")
   int defaultPageSize();
 
+  /** @return Defines the default document page size, defaults to <code>100</code>. */
```
tatu-at-datastax, line 35: Maybe mention it refers only to cases where sorting is used.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/ReadOperation.java
```diff
@@ -25,7 +32,13 @@
 public interface ReadOperation extends Operation {
   String[] documentColumns = {"key", "tx_id", "doc_json"};
   String[] documentKeyColumns = {"key", "tx_id"};
-
+  String[] sortedDataColumns = {"key", "doc_json"};
+  int SORTED_DATA_COLUMNS = sortedDataColumns.length;
+  List<String> sortColumns =
```
tatu-at-datastax, line 37: sortColumns -> sortIndexColumns?

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/ReadOperation.java
```diff
@@ -77,6 +90,146 @@ default Uni<FindResponse> findDocument(
...
+                  }
+                  columnCounter++;
+                  value = row.getValues(columnCounter);
+                  if (!value.hasNull()) {
+                    sortValues.add(nodeFactory.booleanNode(Values.bool(value)));
+                    continue;
+                  }
+                  columnCounter++;
+                  value = row.getValues(columnCounter);
+                  if (!value.hasNull()) {
```
tatu-at-datastax, line 168: Since `query_null_values` is a set, not map, what kind of value will be queried from it?

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/ReadDocument.java
```diff
@@ -10,5 +12,21 @@
...
+ * @param sortColumns List<JsonNode> Serialized sort column value
+ * @param docJsonValue Grpc column value for doc_json.
  */
-public record ReadDocument(DocumentId id, UUID txnId, JsonNode document) {}
+public record ReadDocument(
+    DocumentId id,
+    UUID txnId,
+    JsonNode document,
+    List<JsonNode> sortColumns,
+    QueryOuterClass.Value docJsonValue) {
```
tatu-at-datastax, line 23: Is there a reason for not passing `String` but raw gRPC value here? (perhaps we don't always need it so less conversions?)

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/ReadDocument.java
```diff
@@ -10,5 +12,21 @@
...
+  public ReadDocument(DocumentId id, UUID txnId, JsonNode document) {
```
tatu-at-datastax, line 24: Instead of separate constructors, maybe create 2 factory methods so name can describe kind of document created -- since we are now creating 2 kinds of `ReadDocument`s? They can both use one canonical constructor.

ivansenic:
@maheshrajamani I will have a look on this tomorrow..

src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCode.java
maheshrajamani: Will change

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/model/impl/FindCommandResolver.java
```diff
@@ -38,20 +40,46 @@ public Class<FindCommand> getCommandClass() {
...
+    ReadType readType = ReadType.DOCUMENT;
+    // If sort request
+    if (sortClause != null && !sortClause.sortExpressions().isEmpty()) {
+      orderBy =
+          command.sortClause().sortExpressions().stream()
+              .map(
+                  sortExpression ->
+                      new FindOperation.OrderBy(sortExpression.path(), sortExpression.ascending()))
+              .collect(Collectors.toList());
+      skip =
```
tatu-at-datastax, line 60: Should we limit maximum skip value that can be defined? So it cannot be used to increase number of documents read beyond maximum limit? (and basically validation combination of skip+limit not to exceed maximum)

src/test/java/io/stargate/sgv2/jsonapi/api/v1/FindIntegrationTest.java
tatu-at-datastax: I can't think of anything obvious. I guess it is possible to build all kinds of helper methods to filter out other fields and so on but I think Rest-assured mostly focuses on equality comparisons and not ordering.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/FindIntegrationTest.java
```diff
@@ -772,5 +780,139 @@ public void findByBooleanColumn() {
...
+    }
+
+    private void insert(Map<String, String> documents) {
+      documents.values().forEach(doc -> insertDoc(doc));
+    }
+
+    private Map<String, String> getDocuments(int countOfDocuments, boolean asc) {
+      String json = "{\"_id\":\"doc%s\", \"username\":\"user%s\", \"active_user\":true}";
+      Map<String, String> data =
```
tatu-at-datastax, line 904: Is the idea here to keep the order same, or opposite? If we wanted to avoid something that is not sorted (to verify sorting), could just use basic `HashMap`. It should still produce repeatable runs, I think; order would be arbitrary but stable (based on hashCode() of `String`).

src/test/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/FindOperationTest.java
```diff
@@ -830,6 +863,484 @@ public void failurePropagated() {
       // then result
       assertThat(failure).isEqualTo(exception);
     }
+
```
tatu-at-datastax, line 866: Would it make sense to separate these tests as something like `FindOperationWithSortTest`?

tatu-at-datastax (approved):
Looks good: I added some suggestions but no blockers.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/ReadOperation.java
maheshrajamani: In case of set or list type cassandra will return the entry name came in request if it contains the value

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/ReadDocument.java
maheshrajamani: Idea is to deserialize the doc_json only for the selected documents that are returned in response.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/ReadDocument.java
maheshrajamani: Will make the change

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/model/impl/FindCommandResolver.java
maheshrajamani: The maxSortReadLimit is set as limit in cql so the read will stop once the limit is hit.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/FindIntegrationTest.java
maheshrajamani: Wanted to keep the data sorted in this response. This ordered value is used to check against the find command response to confirm sort.

src/test/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/FindOperationTest.java
maheshrajamani: Will Do

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/ReadDocument.java
tatu-at-datastax: Ok that makes sense then, thanks!

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/model/impl/FindCommandResolver.java
maheshrajamani: @tatu-at-datastax But if skip > maxSortReadLimit think we can short circuit and return. In this case does it has to be a error or empty data set response.

tatu-at-datastax: Hmmh. Good question; I can see either case. Maybe worth asking @amorton on opinion.

src/main/java/io/stargate/sgv2/jsonapi/service/bridge/config/DocumentConfig.java
```diff
@@ -26,18 +26,21 @@
...
 
+  /**
+   * @return Defines the default document page size for sorting, having separate config because sort
+   *     will more rows in per page, defaults to <code>100</code>.
+   */
+  @Max(500)
+  @Positive
+  @WithDefault("100")
+  int defaultSortPageSize();
+
```
ivansenic, line 43: Let's create a ticket for documenting all config options?

src/test/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DeleteOperationTest.java
```diff
@@ -219,7 +225,10 @@ public void deleteWithDynamic() {
               1,
               1,
               ReadType.KEY,
-              objectMapper);
+              objectMapper,
+              null,
+              0,
+              0);
```
ivansenic, line 231: we could have created a shortcut constructors.. or static method providers..

src/test/java/io/stargate/sgv2/jsonapi/api/v1/FindOperationWithSortIntegrationTest.java
```diff
@@ -0,0 +1,168 @@
...
+public class FindOperationWithSortIntegrationTest extends CollectionResourceBaseIntegrationTest {
+  @Nested
+  @TestMethodOrder(MethodOrderer.OrderAnnotation.class)
+  class FindOperationWithSort {
+    private ObjectMapper objectMapper = new ObjectMapper();
+
+    @Test
+    @Order(1)
+    public void sort() {
+      deleteAllDocuments();
```
ivansenic, line 37: `deleteAllDocuments();` could be extracted as before all

src/test/java/io/stargate/sgv2/jsonapi/api/v1/FindOperationWithSortIntegrationTest.java
```diff
@@ -0,0 +1,168 @@
...
+import java.util.TreeMap;
+import org.junit.jupiter.api.MethodOrderer;
+import org.junit.jupiter.api.Nested;
+import org.junit.jupiter.api.Order;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestMethodOrder;
+
+@QuarkusIntegrationTest
+@QuarkusTestResource(DseTestResource.class)
+public class FindOperationWithSortIntegrationTest extends CollectionResourceBaseIntegrationTest {
```
ivansenic, line 28: missing cases:

* sort on non-existing column
* sort on multiple columns
* sort on numbers
* sort on boolean
* sort on null/non-null values

src/test/java/io/stargate/sgv2/jsonapi/api/v1/FindIntegrationTest.java
```diff
@@ -29,80 +28,80 @@ class Find {
...
-                                  "insertOne": {
-                                    "document": {
-                                      "_id": "doc1",
-                                      "username": "user1",
-                                      "active_user" : true
-                                    }
-                                  }
-                                }
-                                """;
+            {
```
ivansenic, line 31: thx for doing this :heart:

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/model/impl/FindCommandResolver.java
```diff
@@ -38,20 +40,46 @@ public Class<FindCommand> getCommandClass() {
...
     List<DBFilterBase> filters = resolve(commandContext, command);
+    final FindCommand.Options options = command.options();
     int limit =
-        command.options() != null && command.options().limit() != null
-            ? command.options().limit()
-            : documentConfig.maxLimit();
+        options != null && options.limit() != null ? options.limit() : documentConfig.maxLimit();
     int pageSize = documentConfig.defaultPageSize();
     String pagingState = command.options() != null ? command.options().pagingState() : null;
+    final SortClause sortClause = command.sortClause();
```
ivansenic, line 48: this should have been extracted to separate method, now it all mixed up

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/ReadDocument.java
ivansenic: Then we should have used the `Supplier<JsonNode>` or what ever the generic typing would be.. I also very much dislike that we have the grpc here..

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/FindOperation.java
```diff
@@ -132,4 +151,50 @@ private QueryOuterClass.Query buildSelectQuery(DBFilterBase.IDFilter additionalI
...
+  private QueryOuterClass.Query buildSortedSelectQuery(DBFilterBase.IDFilter additionalIdFilter) {
+    List<BuiltCondition> conditions = new ArrayList<>(filters.size());
+    for (DBFilterBase filter : filters) {
+      if (additionalIdFilter == null
+          || (additionalIdFilter != null && !(filter instanceof DBFilterBase.IDFilter)))
+        conditions.add(filter.get());
+    }
+    if (additionalIdFilter != null) {
+      conditions.add(additionalIdFilter.get());
+    }
```
ivansenic, line 164: we should extract condition construction from the `buildSelectQuery` and use it here as well..

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/FindOperation.java
```diff
@@ -132,4 +151,50 @@ private QueryOuterClass.Query buildSelectQuery(DBFilterBase.IDFilter additionalI
...
+    String[] columns = sortedDataColumns;
+
+    if (orderBy() != null) {
+      List<String> sortColumns = Lists.newArrayList(columns);
+      orderBy().forEach(order -> sortColumns.addAll(order.getOrderingColumn()));
```
ivansenic, line 169: addAll on a single column? or is `getOrderingColumn` returning a list? if so you would have duplicated values possible here

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/ReadOperation.java
```diff
@@ -77,6 +90,146 @@ default Uni<FindResponse> findDocument(
...
+                    sortValues.add(nodeFactory.nullNode());
```
ivansenic, line 169: can we get any explanation on what is going on here?

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/ReadOperation.java
```diff
@@ -77,6 +90,146 @@ default Uni<FindResponse> findDocument(
...
+      int pageSize,
+      ObjectMapper objectMapper,
+      Comparator<ReadDocument> comparator,
+      int numberOfOrderByColumn,
+      int skip,
+      int limit,
+      int errorLimit) {
+    final AtomicInteger documentCounter = new AtomicInteger(0);
+    MinMaxPriorityQueue<ReadDocument> sortedData =
+        MinMaxPriorityQueue.orderedBy(comparator).maximumSize(skip + limit).create();
```
ivansenic, line 121: but this is now not thread safe, as you are not using it in a reduce way? you have a merge below and this really worries me :warning:

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/ReadOperation.java
```diff
@@ -77,6 +90,146 @@ default Uni<FindResponse> findDocument(
...
+                  sortValues.add(nodeFactory.missingNode());
+                }
+                // Create ReadDocument with document id, grpc value for doc json and list of sort
+                // values
+                document =
+                    ReadDocument.from(
+                        getDocumentId(row.getValues(0)), // key
+                        row.getValues(1), // Deserialized value of doc_json
+                        sortValues);
+                sortedData.add(document);
```
ivansenic, line 182: non thread safe access

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/ReadOperation.java
```diff
@@ -77,6 +90,146 @@ default Uni<FindResponse> findDocument(
...
+              }
+              return Uni.createFrom().item(true);
+            })
+        .collect()
+        .last()
+        .onItem()
+        .transform(
+            flag -> {
+              if (documentCounter.get() == errorLimit)
+                throw new JsonApiException(ErrorCode.DATASET_TOO_BIG);
```
ivansenic, line 192: this is too late, you already have all the documents fetched..

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/ChainedComparator.java
```diff
@@ -0,0 +1,27 @@
...
+import io.stargate.sgv2.jsonapi.util.JsonNodeComparator;
+import java.util.Comparator;
+import java.util.List;
+
+public record ChainedComparator(List<FindOperation.OrderBy> sortColumns)
+    implements Comparator<ReadDocument> {
+  @Override
+  public int compare(ReadDocument o1, ReadDocument o2) {
+    int i = 0;
+    final List<JsonNode> sortValuesO1 = o1.sortColumns();
```
ivansenic, line 15: is this what ReadOperation would create for each document right?

ivansenic (commented):
This is merged so I can just submit my comments.. Definitely a lot of things to look at..

src/main/java/io/stargate/sgv2/jsonapi/service/bridge/config/DocumentConfig.java
maheshrajamani: Yes will do it.

src/test/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DeleteOperationTest.java
maheshrajamani: Will refactor code as separate PR.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/FindOperationWithSortIntegrationTest.java
maheshrajamani: Will do

src/test/java/io/stargate/sgv2/jsonapi/api/v1/FindOperationWithSortIntegrationTest.java
maheshrajamani: Will do

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/ReadOperation.java
maheshrajamani: Will add comments

src/test/java/io/stargate/sgv2/jsonapi/api/v1/FindOperationWithSortIntegrationTest.java
vkarpov15: Also want to see:

1. sort with filter
2. sort with `findOne()` and `findOneAndUpdate()` (if in scope)

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/ReadOperation.java
maheshrajamani: The logic doesn't read and write at the same time. sortedData.pool() is done only after the collect() is done. Not sure what the concern here is.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/ReadOperation.java
maheshrajamani: There is no other way to check it, unless we run count query. And we will fetch only (max allowed + 1) document

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/ChainedComparator.java
maheshrajamani: Yes.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/ReadDocument.java
tatu-at-datastax: I think simply extracting `String` would not add much overhead; it's parsing `String`/`byte[]` to `JsonNode` that is heavy. Could do supplier too if that's cleaner of course. Either way, avoiding constructing `JsonNode` makes sense, and so does not passing gRPC type.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/ReadOperation.java
ivansenic: Talked about this yesterday use `collect().in()` approach..

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/ReadOperation.java
ivansenic: Don't understand this? Yes you can count how many you already fetched in the first fetching loop.. Just have the state that counts saves next page state and the documents already fetched, if you are over the maximum, push exception down immediately..

src/test/java/io/stargate/sgv2/jsonapi/api/v1/FindOperationWithSortIntegrationTest.java
maheshrajamani: @vkarpov15 

- Added sort with filter test case in #308 
- sorting for findOne and findOneAndUpdate sort not supported, is this needed?
- Projection is work in progress, will add a case once it's completed.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/ReadOperation.java
maheshrajamani: Yes. Implemented in #308

src/test/java/io/stargate/sgv2/jsonapi/api/v1/FindOperationWithSortIntegrationTest.java
tatu-at-datastax: Quick note: I do not think Mongo actually supports "sort" on `findOne()` (or `findOneAndUpdate()`).
Atlas shell allows it for find:

    db.arrays.find().sort({ "x":1 })

but it fails for

    db.arrays.findOne().sort({ "x":1 })

with error

```
Atlas atlas-v8vini-shard-0 [primary] examples> db.arrays.findOne().sort({ "x":-1 })
TypeError: db.arrays.findOne().sort is not a function
```

tatu-at-datastax: @maheshrajamani Could you add a link here? One doc I saw listed driver-level "find" options which were not specific to "findOne()", but I may have missed something.

maheshrajamani: https://www.mongodb.com/docs/manual/reference/method/db.collection.findOneAndUpdate/#:~:text=projection%3A%20%3Cdocument%3E%2C-,sort%3A%20%3Cdocument%3E%2C,-maxTimeMS%3A%20%3Cnumber%3E%2C

tatu-at-datastax: @maheshrajamani Ah. Yes, you are right! Oddly, findOne -- https://www.mongodb.com/docs/manual/reference/method/db.collection.findOne/ -- has different layout for settings.
But even that may accept `sort` (possibly) as it refers to "findOptions" which lists a big set of options, without indicating that some are only applicable to "full" `find()`.

And playing with Atlas console a bit more, I think this does actually work:

   db.arrays.findOne({}, {}, { "sort" : { "x":-1 }})

So I think I agree now that "sort" is actually supported, even if `findOne()` documentation sort of hides it.

PR #297, opened 2023-03-27
fix bumping of parent version script

ivansenic:
**What this PR does**:

Unfortunately the existing way of bumping the parent version works only if parent exists on the maven repository.. However, it ususally takes 15-20 minutes for new version to be available on the maven repository.. And since we use the script in the automated way, once the release is done it could actually resolve the new version to be not available and thus not updating the pom..

This is also the reason why automated workflow so far did not create any PR.. Simply there we no changes in the working three so you could not create a PR..

The update command sets the parent directly..

### DIFF
--- a/scripts/bump_stargate.sh
+++ b/scripts/bump_stargate.sh
@@ -15,5 +15,5 @@ else
    exit 1
 fi
 
-# Note: brackets around version needed for parent (version "range") but not for child modules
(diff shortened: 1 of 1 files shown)


ivansenic:
@kathirsvn same here, can you test if this works now locally for you if you are on Mac, just paste me the output of:

```
./scripts/bump_stargate.sh v2.0.13 && git diff
```

PR #299, opened 2023-03-27
closes #298: count down latch to be decrease on assertion errors

### LINKED ISSUE #298: Decrease count-down latch on failing concurrency tests

ivansenic:
**What this PR does**:
Fixes the count down latch now being decreased on failed asserts.. This will never result in a block..

But the exception would be ignored.. Thus I am thinking what the options are:

* [x] catch any exception, set to the `AtomicReference`, after CTL goes to zero iterate over and re-throw any exception?
* [ ] better ideas?

Fixes #298

**WARNING: remove serial consistency commit before merging**

### DIFF
--- a/src/test/java/io/stargate/sgv2/jsonapi/api/v1/DeleteManyIntegrationTest.java
+++ b/src/test/java/io/stargate/sgv2/jsonapi/api/v1/DeleteManyIntegrationTest.java
@@ -300,28 +300,33 @@ public class DeleteManyIntegrationTest extends CollectionResourceBaseIntegration
       for (int i = 0; i < threads; i++) {
         new Thread(
                 () -> {
-                  Integer deletedCount =
-                      given()
-                          .header(HttpConstants.AUTHENTICATION_TOKEN_HEADER_NAME, getAuthToken())
-                          .contentType(ContentType.JSON)
-                          .body(deleteJson)
-                          .when()
-                          .post(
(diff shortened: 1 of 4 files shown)


src/main/java/io/stargate/sgv2/jsonapi/service/bridge/executor/QueryExecutor.java
```diff
@@ -67,8 +67,13 @@ public Uni<QueryOuterClass.ResultSet> executeWrite(QueryOuterClass.Query query)
     QueryOuterClass.Consistency consistency = queriesConfig.consistency().writes();
     QueryOuterClass.ConsistencyValue.Builder consistencyValue =
         QueryOuterClass.ConsistencyValue.newBuilder().setValue(consistency);
+    QueryOuterClass.Consistency serialConsistency = queriesConfig.serialConsistency();
+    QueryOuterClass.ConsistencyValue.Builder serialConsistencyValue =
+        QueryOuterClass.ConsistencyValue.newBuilder().setValue(serialConsistency);
     QueryOuterClass.QueryParameters.Builder params =
-        QueryOuterClass.QueryParameters.newBuilder().setConsistency(consistencyValue);
+        QueryOuterClass.QueryParameters.newBuilder()
```
maheshrajamani, line 74: Do you want to revert the serial consistency change, so it can be done unit test case as suggested in #290 ?

maheshrajamani (approved):

PR #301, opened 2023-03-27
Fix remaining gap in #158, connect findOneAndUpdate() to projection

tatu-at-datastax:
**What this PR does**:

Fixes minor issue in earlier #158 to connect findOneAndUpdate() with projection too (actual projection implementing to follow)

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/FindOperation.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/FindOperation.java
@@ -25,7 +25,11 @@ import java.util.function.Supplier;
 public record FindOperation(
     CommandContext commandContext,
     List<DBFilterBase> filters,
-    DocumentProjector projection,
+    /**
+     * Projection used on input document; if no changes desired, identity projection. Only used for
+     * "pure" read operations: for updates (like {@code findOneAndUpdate}) is passed differently.
+     */
+    DocumentProjector inputProjection,
     String pagingState,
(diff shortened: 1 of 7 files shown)


src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/FindOperation.java
```diff
@@ -25,7 +25,11 @@
 public record FindOperation(
     CommandContext commandContext,
     List<DBFilterBase> filters,
-    DocumentProjector projection,
+    /**
+     * Projection used on input document; if no changes desired, identity projection. Only used for
+     * "pure" read operations: for updates (like {@code findOneAndUpdate}) is passed differently.
+     */
+    DocumentProjector inputProjection,
```
tatu-at-datastax, line 32: Renamed here to reflect different application

maheshrajamani: This will be result projection for Find and FindOne, think resultProjection sounds better in this case too.

maheshrajamani (approved):

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/FindOperation.java
tatu-at-datastax: I guess that is true. Will change as suggested.

PR #309, opened 2023-03-29
closes #261: grpc retries and timeouts to fix availability failures

### LINKED ISSUE #261: Implement retries for availability faults
Spec defines the [Availability Faults](https://github.com/stargate/jsonapi/blob/main/docs/jsonapi-spec.textile#multi-document-failure-considerations). We need to add implementation for this. 

Currently the Stargate V2 allows retries based on the gRPC status codes, we need to see if this is enough. It could be that we need to extend the bridge, so that we can recognize client-side and server-side timeouts. Or client side timeouts can be disabled for now.

ivansenic:
**What this PR does**:
Sets correct gRPC client timeouts and client retries for the jsonapi.

**This issue depends on the Stargate v2.0.11**.

Fixes #261 

**Checklist**
- [x] update to Stargate `2.0.11`
- [x] create integration test for confirming policy is used

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/grpc/retries/impl/JsonApiGrpcRetryPolicy.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/grpc/retries/impl/JsonApiGrpcRetryPolicy.java
@@ -5,14 +5,13 @@ import io.grpc.Status;
 import io.grpc.StatusRuntimeException;
 import io.grpc.protobuf.ProtoUtils;
 import io.stargate.bridge.proto.QueryOuterClass;
+import io.stargate.sgv2.api.common.grpc.retries.GrpcRetryPredicate;
 import java.util.Objects;
-import java.util.function.Predicate;
 import javax.enterprise.context.ApplicationScoped;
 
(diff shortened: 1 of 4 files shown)


src/main/java/io/stargate/sgv2/jsonapi/grpc/retries/impl/JsonApiGrpcRetryPolicy.java
```diff
@@ -0,0 +1,53 @@
...
+import io.grpc.Metadata;
+import io.grpc.Status;
+import io.grpc.StatusRuntimeException;
+import io.grpc.protobuf.ProtoUtils;
+import io.stargate.bridge.proto.QueryOuterClass;
+import io.stargate.sgv2.api.common.grpc.retries.GrpcRetryPredicate;
+import java.util.Objects;
+import javax.enterprise.context.ApplicationScoped;
+
+/** Default gRPC retry policy used in the project. */
```
tatu-at-datastax, line 12: Maybe add a brief description of logic ("Retry UNAVAILABLE and true Server-side read/write timeouts")

tatu-at-datastax (approved):

PR #317, opened 2023-03-30
(WIP) Add @JsonIgnoreProperties for optional "options" property

tatu-at-datastax:
**What this PR does**:

Adds `@JsonIgnoreProperties("options")` in `Command` to make "options" truly optional

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/Command.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/Command.java
@@ -1,5 +1,6 @@
 package io.stargate.sgv2.jsonapi.api.model.command;
 
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonSubTypes;
 import com.fasterxml.jackson.annotation.JsonTypeInfo;
 import io.stargate.sgv2.jsonapi.api.model.command.impl.CountDocumentsCommands;
@@ -51,5 +52,10 @@ import io.stargate.sgv2.jsonapi.service.resolver.model.CommandResolver;
   @JsonSubTypes.Type(value = InsertManyCommand.class),
   @JsonSubTypes.Type(value = UpdateManyCommand.class),
   @JsonSubTypes.Type(value = UpdateOneCommand.class),
+})
+@JsonIgnoreProperties({
+  // Conditional ignoral so command impls may or may not have "options" property but
+  // even if they don't, no error if caller passes such property.
+  "options"
 })
 public interface Command {}


tatu-at-datastax:
@ivansenic so my understanding of `@JsonIgnoreProperties` was wrong: it does indicate hard removal, so unless sub-class overrides it, `options` won't be included. So this is NOT the way to do what we want.

We could still consider only disabling reporting for `Command` types by using

    @JsonIgnoreProperties(ignoreUnknown = true)
    public interface Command { }

and not configure it globally. But I think `Command` implementations are also where we sort of would want to validate if properties are known.

I am also ok with just changing it globally as suggested, not a big deal. Just trying to figure out optimal way to have validation where we want, but allowing optional "options" for Commands

tatu-at-datastax:
Think I figured out a simple enough way that does work. At least unit tests pass fine (which do verify binding of "options" where needed); exception ITs to pass too but we'll see.

tatu-at-datastax:
Ok all tests pass so should be safe.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/Command.java
```diff
@@ -52,4 +56,13 @@
...
 })
-public interface Command {}
+public interface Command {
+  /**
+   * Default method if (and only if) there is no "options" property. Used for allowing all Commands
+   * to take in empty Options without having to add Record field if none accepted.
+   */
+  @JsonProperty("options")
+  default void options(Map<String, Object> optionsIgnored) {
+    // Should we fail for non-empty options?
```
vkarpov15, line 66: I think individual commands should decide which options are valid or not. It's just empty options should be fine for all commands.

vkarpov15 (approved):

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/Command.java
tatu-at-datastax: @vkarpov15 Agreed. This only applies to `Command`s that do not declare that they take `options` parameter -- which is the default "only empty accepted" case. Ones that do take it would not call this method.

So the question is that if `Command` is not marked/defined to take Options, then should we verify that only missing/empty JSON Object options are allowed and give error for others.
I'm fine with no error, it's a philosophical choice of leniency of arguments.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/Command.java
```diff
@@ -33,6 +36,7 @@
  *
  * <p>Each command should validate itself using the <i>javax.validation</i> framework.
  */
+@RegisterForReflection
```
tatu-at-datastax, line 39: Interestingly enough, without this Native build failed "allowEmptyOptions" tests. Not 100% sure what is missing (I'd have thought this class and subtypes referenced would be reachable) but adding annotation resolved the issue.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/Command.java
```diff
@@ -52,4 +56,13 @@
...
   @JsonSubTypes.Type(value = UpdateOneCommand.class),
 })
-public interface Command {}
+public interface Command {
+  /**
+   * Default method if (and only if) there is no "options" property. Used for allowing all Commands
+   * to take in empty Options without having to add Record field if none accepted.
+   */
+  @JsonProperty("options")
+  default void options(Map<String, Object> optionsIgnored) {
```
ivansenic, line 65: I am not sure I like this option.. This looks very confusing and I dont even understand what does it mean.. now all sub classes have this method and looks strange, especially in the classes that do have the options field.. I am not really pro this change..

ivansenic: But the classes that do have options are not overriding this method because they are records and don't even have this setter method.. and even if there would be a setter method, it would still not override as signatures would be different

src/test/java/io/stargate/sgv2/jsonapi/service/resolver/model/impl/FindOneAndReplaceCommandResolverTest.java
```diff
@@ -234,7 +234,7 @@ public void idFilterConditionWithOptions() throws Exception {
               "findOneAndReplace": {
                 "filter" : {"_id" : "id"},
                 "replacement" : {"col1" : "val1", "col2" : "val2"},
-                "options" : {"returnDocument" : "after", "upsert": true }
```
tatu-at-datastax, line 237: @ivansenic @maheshrajamani Interestingly enough this minor bug (I think copy-pasted) was hidden by "ignore unknown properties"), `Options` for "findOneAndReplace" do not have it (`findOneAndUpdate` does)

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/Command.java
```diff
@@ -35,6 +36,7 @@
  *
  * <p>Each command should validate itself using the <i>javax.validation</i> framework.
  */
+@RegisterForReflection
```
ivansenic, line 39: We can remove this now or?

src/test/java/io/stargate/sgv2/jsonapi/service/resolver/model/impl/FindOneAndReplaceCommandResolverTest.java
```diff
@@ -234,7 +234,7 @@ public void idFilterConditionWithOptions() throws Exception {
           "findOneAndReplace": {
             "filter" : {"_id" : "id"},
             "replacement" : {"col1" : "val1", "col2" : "val2"},
-            "options" : {"returnDocument" : "after", "upsert": true }
+            "options" : {"returnDocument" : "after" }
```
ivansenic, line 237: why this removed?

ivansenic (approved):
Good, one small comment only..

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/Command.java
tatu-at-datastax: I will try and make sure it won't break anything but probably yes.

src/test/java/io/stargate/sgv2/jsonapi/service/resolver/model/impl/FindOneAndReplaceCommandResolverTest.java
tatu-at-datastax: As per my comment (not sure if it got removed?), `findOneAndReplace` does not support this option.
With "ignore all unknown" it was let through, but removing that tests flagged it as unknown.

So it's a fix to bug hidden by ignoring unknown properties.

ivansenic: Ahaaa good!

PR #331, opened 2023-04-03
closes #305: reorganize and document config options

### LINKED ISSUE #305: Document all json document config options
Add some details to spec as well.

ivansenic:
**What this PR does**:
Tries to organize and document all configuration options we have. Currently those are spread into two classes and there is no clear path to follow when adding a new config options.. Also spec says nothing about these options..
 
Fixes #305

**Checklist**
- [x] use instance injection instead of `@ConfigProperty`
- [x] add updates to spec
- [x] correctly organize into config classes
- [ ] optionally add tests where needed (currently completely not tested)


CONFIGURATION.md
```diff
@@ -11,6 +11,19 @@ Here are some Stargate-relevant property groups that are necessary for correct s
...
 
+
+## Document limits configuration
+*Configuration for document limits, defined by [DocumentLimitsConfig.java](src/main/java/io/stargate/sgv2/jsonapi/config/DocumentLimitsConfig.java).*
+
+| Property                                                 | Type  | Default     | Description                                                                                                 |
+|----------------------------------------------------------|-------|-------------|-------------------------------------------------------------------------------------------------------------|
+| `stargate.jsonapi.document.limits.max-size`              | `int` | `1_000_000` | The maximum size of a single document in characters. Defaults to 1 million characters or approximately 1MB. |
+| `stargate.jsonapi.document.limits.max-depth`             | `int` | `8`         | The maximum document depth (nesting).                                                                       |
+| `stargate.jsonapi.document.limits.max-name-length`       | `int` | `48`        | The maximum length of property names in a document for an individual segment.                               |
```
ivansenic, line 22: @tatu-at-datastax this is a bit confusing for me, should we go for one of:

1. `stargate.jsonapi.document.limits.max-property-length`
2. `stargate.jsonapi.document.limits.max-promerty-name-length`

src/test/java/io/stargate/sgv2/jsonapi/service/shredding/ShredderDocLimitsTest.java
```diff
@@ -48,7 +48,7 @@ public void catchTooBigDoc() {
           .isInstanceOf(JsonApiException.class)
```
ivansenic, line 48: @tatu-at-datastax see this, this test is still green, although I completely changed the properties.. usually when writing tests that depend on the config you should assert that the changes in the config are applied in the tests and would break it.. For example, by explicitly setting values in the test profile, like here:

```
@QuarkusTest
@TestProfile(InsertManyCommandTest.Profile.class)
class InsertManyCommandTest {

  public static class Profile implements NoGlobalResourcesTestProfile {

    @Override
    public Map<String, String> getConfigOverrides() {
      return Map.of("stargate.jsonapi.doc-limits.max-insert-many-documents", "2");
    }
  }
```

The `InsertManyCommandTest` fails now, signaling that there was a breaking change.

src/main/java/io/stargate/sgv2/jsonapi/config/LwtConfig.java
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * Configuration setup for the Light-weigh transactions.
+ */
+@ConfigMapping(prefix = "stargate.jsonapi.lwt")
+public interface LwtConfig {
+
+    /** @return Defines the maximum retry for lwt failure <code>3</code>. */
+    @Max(5)
+    @PositiveOrZero
```
ivansenic, line 18: @maheshrajamani changed `@Positive` to `@PositiveOrZero` so one can disable retires, does this make sense?

CONFIGURATION.md
```diff
@@ -11,6 +11,33 @@ Here are some Stargate-relevant property groups that are necessary for correct s
...
+| `stargate.jsonapi.operations.max-sort-read-limit`       | `int` | `10000` | The maximum amount of documents that could be sorted using the in-memory sorting.                                                                                                                   |
+| `stargate.jsonapi.operations.max-document-insert-count` | `int` | `20`    | The maximum amount of documents that can be inserted in a single operation. The request will fail fast without inserts if limit is broken.                                                          |
+| `stargate.jsonapi.operations.max-document-update-count` | `int` | `20`    | The maximum amount of documents that can be updated in a single operation. In case there are more documents that could be updated, the operation will set the `moreData` response status to `true`. |
+| `stargate.jsonapi.operations.max-document-delete-count` | `int` | `20`    | The maximum amount of documents that can be deleted in a single operation. In case there are more documents that could be delete, the operation will set the `moreData` response status to `true`.  |
```
ivansenic, line 38: @maheshrajamani could you check this, especially the `max-limit`?

CONFIGURATION.md
```diff
@@ -11,6 +11,33 @@ Here are some Stargate-relevant property groups that are necessary for correct s
...
+| `stargate.jsonapi.operations.default-page-size`         | `int` | `20`    | The default Cassandra page size used for read queries.                                                                                                                                              |
+| `stargate.jsonapi.operations.default-sort-page-size`    | `int` | `100`   | The default Cassandra page size used for read queries that are used for sorting purposes.                                                                                                           |
+| `stargate.jsonapi.operations.max-limit`                 | `int` | `1000`  | // TODO from code this seems like default, not maximum???                                                                                                                                           |
+| `stargate.jsonapi.operations.max-sort-read-limit`       | `int` | `10000` | The maximum amount of documents that could be sorted using the in-memory sorting.                                                                                                                   |
```
ivansenic, line 35: @maheshrajamani should we call this `stargate.jsonapi.operations.max-document-sort-count`, I could also add explanation that command fails if the limit is reached/broken?

CONFIGURATION.md
```diff
@@ -11,6 +11,33 @@ Here are some Stargate-relevant property groups that are necessary for correct s
...
 * `quarkus.cache.caffeine.keyspace-cache` - property group  for defining the keyspace cache used by [SchemaManager](../sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/schema/SchemaManager.java) (see [Caffeine cache configuration](https://quarkus.io/guides/cache#caffeine-configuration-properties) for all options)
 
+
+## Document limits configuration
+*Configuration for document limits, defined by [DocumentLimitsConfig.java](src/main/java/io/stargate/sgv2/jsonapi/config/DocumentLimitsConfig.java).*
+
+| Property                                                 | Type  | Default     | Description                                                                                                 |
+|----------------------------------------------------------|-------|-------------|-------------------------------------------------------------------------------------------------------------|
+| `stargate.jsonapi.document.limits.max-size`              | `int` | `1_000_000` | The maximum size of a single document in characters. Defaults to 1 million characters or approximately 1MB. |
```
ivansenic, line 20: hmm not sure if this should be `stargate.jsonapi.document.limits`, maybe if we would also have `stargate.jsonapi.operations.limits`? Otherwise, lets just call this `stargate.jsonapi.document`?

CONFIGURATION.md
tatu-at-datastax: Yes I'd like "name" to be included (in case we also have property value lengths) so (2)

`stargate.jsonapi.document.limits.max-property-name-length`

sounds good to me.

CONFIGURATION.md
tatu-at-datastax: To me reference to "document" makes sense as it limits size of the stored document? Not sure why there should be reference to "operations".

src/test/java/io/stargate/sgv2/jsonapi/service/shredding/ShredderDocLimitsTest.java
tatu-at-datastax: @ivansenic I think this is @maheshrajamani 's limit; I worked only on document limits. I agree that testing of breaking limits (and yes included changes to defaults) should be tested.

src/main/java/io/stargate/sgv2/jsonapi/config/DocumentLimitsConfig.java
```diff
@@ -46,11 +46,4 @@ public interface DocumentLimitsConfig {
   @Positive
   @WithDefault("100")
   int maxArrayLength();
-
```
tatu-at-datastax, line 49: Ah, good catch. Was misplaced.

CONFIGURATION.md
ivansenic: Done.

src/test/java/io/stargate/sgv2/jsonapi/service/shredding/ShredderDocLimitsTest.java
ivansenic: Resolving.

CONFIGURATION.md
ivansenic: Done.

ivansenic:
@maheshrajamani @tatu-at-datastax Can you help me out resolve the remaining comments? I would like to merge this asap..

src/main/java/io/stargate/sgv2/jsonapi/config/LwtConfig.java
tatu-at-datastax: @ivansenic I think that makes sense, +1.

maheshrajamani (approved):

CONFIGURATION.md
tatu-at-datastax: @ivansenic @maheshrajamani I think Mahesh needs to confirm this one.

CONFIGURATION.md
tatu-at-datastax: I think all similar limit settings should use same naming pattern: if this is true, all 3 look fine. (2) would be bit more compact so that sounds good if you want to change them.

CONFIGURATION.md
maheshrajamani: Basically this is the maximum limit of document we will return for a query. It's redundant with in memory sort (max doc read for sort is 10k) and limit option. We can evaluate the use case and remove it.

src/main/java/io/stargate/sgv2/jsonapi/config/LwtConfig.java
ivansenic: @maheshrajamani what do you mean it's coded to retry at least once on LWT failure? What does this mean? Wouldn't setting this to `0` allow anybody to say don't retry on the LWT failure? Would it not work?

vkarpov15 (approved):

ivansenic:
On hold until @maheshrajamani is not back from the holiday, so we can resolve the `max-limit` issue.

maheshrajamani (approved):

jeffreyscarpenter:
@maheshrajamani should we go ahead and merge?

src/main/java/io/stargate/sgv2/jsonapi/config/LwtConfig.java
maheshrajamani: Yes. This it will retry at least once in case of LWT.

CONFIGURATION.md
ivansenic: OK I will move this to the `stargate.jsonapi.operations.max-read-limit` and try to explain it..

src/main/java/io/stargate/sgv2/jsonapi/config/LwtConfig.java
ivansenic: Done.

CONFIGURATION.md
ivansenic: Done. Removed `maxLimit` as agreed with @maheshrajamani

ivansenic:
@johnsmartco added some spec updates in the https://github.com/stargate/jsonapi/pull/331/commits/7bc71da32cf14dfeb8a4b6804ab294f36465ed40.. If needed feel free to adapt ones I merge this..

PR #335, opened 2023-04-04
closes #321: dropNamespace command

### LINKED ISSUE #321: Implement dropNamespace command

ivansenic:
**What this PR does**:
Implementation of the `dropNamespace` command. I have one issue and that's deletion of the system namespaces? 

I tried to exclude `system` from the keyspaces that could not be deleted, but here are actually multiple built-in keyspaces.. How to deal with these.. Any advice would be appreciated.. 

Fixes #321 

**Checklist**
- [x] Swagger
- [ ] Spec
- [x] Test drop with existing collection works
- [x] Resolve issue about system keyspaces

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/DropNamespaceCommand.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/DropNamespaceCommand.java
@@ -1,0 +1,21 @@
+package io.stargate.sgv2.jsonapi.api.model.command.impl;
+
+import com.fasterxml.jackson.annotation.JsonTypeName;
+import io.stargate.sgv2.jsonapi.api.model.command.GeneralCommand;
+import javax.validation.constraints.NotNull;
+import javax.validation.constraints.Pattern;
+import javax.validation.constraints.Size;
+import org.eclipse.microprofile.openapi.annotations.media.Schema;
+
+@Schema(description = "Command that deletes a namespace.")
+@JsonTypeName("dropNamespace")
+public record DropNamespaceCommand(
+    @NotNull
+        @Pattern(regexp = "[a-zA-Z][a-zA-Z0-9_]*")
+        @Pattern(regexp = "^(?!system$).*")
+        @Size(min = 1, max = 48)
+        @Schema(description = "Name of the namespace")
+        String name)
+    implements GeneralCommand {
+  // TODO add NoOptions extension
+}
--- a/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/DropNamespaceCommand.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/DropNamespaceCommand.java
@@ -12,7 +12,6 @@ import org.eclipse.microprofile.openapi.annotations.media.Schema;
 public record DropNamespaceCommand(
     @NotNull
         @Pattern(regexp = "[a-zA-Z][a-zA-Z0-9_]*")
-        @Pattern(regexp = "^(?!system$).*")
         @Size(min = 1, max = 48)
         @Schema(description = "Name of the namespace")
         String name)
--- a/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/DropNamespaceCommand.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/DropNamespaceCommand.java
@@ -2,6 +2,7 @@ package io.stargate.sgv2.jsonapi.api.model.command.impl;
 
 import com.fasterxml.jackson.annotation.JsonTypeName;
 import io.stargate.sgv2.jsonapi.api.model.command.GeneralCommand;
+import io.stargate.sgv2.jsonapi.api.model.command.NoOptionsCommand;
 import javax.validation.constraints.NotNull;
 import javax.validation.constraints.Pattern;
 import javax.validation.constraints.Size;
@@ -15,6 +16,4 @@ public record DropNamespaceCommand(
         @Size(min = 1, max = 48)
         @Schema(description = "Name of the namespace")
         String name)
-    implements GeneralCommand {
-  // TODO add NoOptions extension
-}
+    implements GeneralCommand, NoOptionsCommand {}
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DropNamespaceOperation.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DropNamespaceOperation.java
@@ -1,0 +1,33 @@
+package io.stargate.sgv2.jsonapi.service.operation.model.impl;
+
+import io.smallrye.mutiny.Uni;
+import io.stargate.bridge.proto.QueryOuterClass;
+import io.stargate.sgv2.jsonapi.api.model.command.CommandResult;
+import io.stargate.sgv2.jsonapi.service.bridge.executor.QueryExecutor;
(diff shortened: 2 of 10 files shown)


src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/DropNamespaceCommand.java
```diff
@@ -0,0 +1,21 @@
...
+import javax.validation.constraints.Pattern;
+import javax.validation.constraints.Size;
+import org.eclipse.microprofile.openapi.annotations.media.Schema;
+
+@Schema(description = "Command that deletes a namespace.")
+@JsonTypeName("dropNamespace")
+public record DropNamespaceCommand(
+    @NotNull
+        @Pattern(regexp = "[a-zA-Z][a-zA-Z0-9_]*")
+        @Pattern(regexp = "^(?!system$).*")
```
ivansenic, line 15: Seems wrong considering these are all system keyspaces:

```
 keyspace_name      | durable_writes | graph_engine | replication
--------------------+----------------+--------------+-------------------------------------------------------------------------------------
        system_auth |           True |         null | {'class': 'org.apache.cassandra.locator.SimpleStrategy', 'replication_factor': '1'}
      system_schema |           True |         null |                             {'class': 'org.apache.cassandra.locator.LocalStrategy'}
   dse_system_local |           True |         null |                             {'class': 'org.apache.cassandra.locator.LocalStrategy'}
         dse_system |           True |         null |                        {'class': 'org.apache.cassandra.locator.EverywhereStrategy'}
         dse_leases |           True |         null | {'class': 'org.apache.cassandra.locator.SimpleStrategy', 'replication_factor': '1'}
         solr_admin |           True |         null |                        {'class': 'org.apache.cassandra.locator.EverywhereStrategy'}
       dse_insights |           True |         null |                        {'class': 'org.apache.cassandra.locator.EverywhereStrategy'}
 dse_insights_local |           True |         null |                             {'class': 'org.apache.cassandra.locator.LocalStrategy'}
 system_distributed |           True |         null | {'class': 'org.apache.cassandra.locator.SimpleStrategy', 'replication_factor': '3'}
     system_backups |           True |         null |                        {'class': 'org.apache.cassandra.locator.EverywhereStrategy'}
             system |           True |         null |                             {'class': 'org.apache.cassandra.locator.LocalStrategy'}
           dse_perf |           True |         null | {'class': 'org.apache.cassandra.locator.SimpleStrategy', 'replication_factor': '1'}
 data_endpoint_auth |           True |         null | {'class': 'org.apache.cassandra.locator.SimpleStrategy', 'replication_factor': '1'}
      system_traces |           True |         null | {'class': 'org.apache.cassandra.locator.SimpleStrategy', 'replication_factor': '2'}
       dse_security |           True |         null | {'class': 'org.apache.cassandra.locator.SimpleStrategy', 'replication_factor': '1'}
```

Anybody with idea how to solve this?

jeffreyscarpenter: Do we enforce similar restriction on other APIs? 

One way to protect against this would be through permissions.

ivansenic: We don't enforce in the other apis..

maheshrajamani: I think we need to enforce on data_endpoint_auth keyspace, since it's created by stargate and storage may not restrict it.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DropNamespaceOperation.java
```diff
@@ -0,0 +1,33 @@
...
+public record DropNamespaceOperation(String name) implements Operation {
+
+  // simple pattern for the cql
+  private static final String DROP_KEYSPACE_CQL = "DROP KEYSPACE IF EXISTS \"%s\";";
+
+  /** {@inheritDoc} */
+  @Override
+  public Uni<Supplier<CommandResult>> execute(QueryExecutor queryExecutor) {
+    QueryOuterClass.Query query =
+        QueryOuterClass.Query.newBuilder().setCql(DROP_KEYSPACE_CQL.formatted(name)).build();
```
tatu-at-datastax, line 24: I guess the assumption is that since we validate namespace name for Command, no escaping is needed here, right?

tatu-at-datastax (commented):
LGTM (but letting Mahesh approve)

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DropNamespaceOperation.java
ivansenic: Yup, not needed..

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/DropNamespaceCommand.java
ivansenic: I think I will remove any limitations on the keyspace name here. As @tatu-at-datastax pointed, the jsonapi should not be responsible in protecting which keyspaces can be deleted or not.

Even in case of the `data_endpoint_auth` keyspace, it should be the responsibility up in the chain (coordinator).

Furthermore, as @jeffreyscarpenter mentioned yesterday, there is no metadata for the keyspaces in cassandra, so there is no way for us to "label" a keyspace during a creation process and know exatcly which ones are created with the jsonapi.

The only additional idea which comes to my mind (and is out of the scope for this PR), would be to allow deletion of a kespace only if it empty (no table) or all tables are schema-wise the jsonapi tables. Similar technique, we can use in the `findNamespaces` and list namesapces only if they are empty or at least one table is schema-wise a jsonapi table. Would like to hear opinions about this?

ivansenic:
@johnsmartco What's the process for the spec updates? We need to add spec for this new command `dropNamespace`, but in the spec I also see `createNamespace` was not mentioned at all. Furthermore both commands belong to the `Namespace` section, thus maybe it would be better if you take it over? How does it work, should I create a ticket for it?

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/DropNamespaceCommand.java
tatu-at-datastax: I like the idea of only allowing dropping if no tables exist in the namespace.

Question on listing is interesting... I can see benefits of only listing only ones that have at least one Collection, although it might be problem if user creates namespace and wants to verify it exists before creating collection.
Maybe there would be an option for "show empty namespaces".

tatu-at-datastax (approved):

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/DropNamespaceCommand.java
ivansenic: @tatu-at-datastax Yea I included the empty keyspace as well:

* for listing: no tables or at least one table is jsonapi schema
* for dropping: no tables or all are jsonapi schema

johnsmartco:
@ivansenic , thanks. In a separate PR, I'll add createNamespace and dropNamespace to the spec. I think I can piece together the needed info from your PR 321 here, and will ask questions as needed.

vkarpov15 (commented):
Thanks for this. Can't wait to try this out, will make cleaning up db after tests much easier.

PR #342, opened 2023-04-07
Fixes #329: Support `$slice` modifier for Projection

### LINKED ISSUE #329: Add support for `$slice` Array Projection modifier
Now that #300 (general Projection) has been implemented we should also implement special operator `$slice` that may be used for Array-valued fields. General format is:

```
 { <arrayField>: { $slice: <number> } }

  { <arrayField>: { $slice: [ <number-to-skip>, <number-to-return> ] } }
```

that is, value is a JSON Object (instead of Number 0 or 1, or Boolean) and for Numbers:

* Specify a positive number n to return the first n elements.
* Specify a negative number n to return the last n elements.

and "number-to-skip" also allows positive numbers (simple skip) and negative (skip "array.length + negative-skip"; that is, `-1` would mean starting at the last element).

"number-to-return" needs to be positive number, starting after skipped elements. In all cases values are truncated to valid index ranges as appropriate (so, for example, for array of 10 elements, skip of 5 and "number-to-return" of 8 would just return last 5 values without error).

tatu-at-datastax:
**What this PR does**:

Implements support for `$slice` projection modifier

Fixes #329

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/src/test/java/io/stargate/sgv2/jsonapi/api/v1/FindOneIntegrationTest.java
+++ b/src/test/java/io/stargate/sgv2/jsonapi/api/v1/FindOneIntegrationTest.java
@@ -22,11 +22,9 @@ import org.junit.jupiter.api.TestMethodOrder;
 @QuarkusIntegrationTest
 @QuarkusTestResource(DseTestResource.class)
 public class FindOneIntegrationTest extends CollectionResourceBaseIntegrationTest {
-
   @Nested
   @TestMethodOrder(MethodOrderer.OrderAnnotation.class)
   class FindOne {
-
     private static final String DOC1_JSON =
         """
         {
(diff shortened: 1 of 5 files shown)


src/test/java/io/stargate/sgv2/jsonapi/api/v1/FindOneIntegrationTest.java
```diff
@@ -260,42 +258,6 @@ public void byIdNotFound() {
           .body("errors", is(nullValue()));
     }
 
-    @Test
```
tatu-at-datastax, line 263: Moved to new IT class below (not removed)

vkarpov15 (commented):
LGTM. One potential suggestion: maybe worth testing `$slice` with `nestedArray`?

tatu-at-datastax:
> LGTM. One potential suggestion: maybe worth testing `$slice` with `nestedArray`?

Sounds reasonable, will do. Thanks!

PR #343, opened 2023-04-11
Handle collection name with case sensitivity.

### LINKED ISSUE #337: Keyspace and table names not handling case-sensitivity (missing escaping in CQL statements)
The `createNamespace` command correctly escapes the keyspace names when creating them, to achieve case-sensitivity using: `CREATE KEYSPACE IF NOT EXISTS "%s" WITH REPLICATION = %s;`

However, other commands are not taking care of this, and are not having keyspace and table names in the double quotes. Thus means that if I create a namespace with upper-case chars, I will not be able to execute any command on it.

This is a major bug imo. Easily reproducible. Assigning to @maheshrajamani for further check.

maheshrajamani:
**What this PR does**:
Handle collection name with case sensitive. 
Keyspace name for casing already handled.

Fixes #337 

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/CreateCollectionOperation.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/CreateCollectionOperation.java
@@ -34,7 +34,7 @@ public record CreateCollectionOperation(CommandContext commandContext, String na
 
   protected QueryOuterClass.Query getCreateTable(String keyspace, String table) {
     String createTable =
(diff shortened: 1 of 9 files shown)


tatu-at-datastax (approved):
LGTM

ivansenic:
Do we need some int tests here? Maybe simple ensuring that collection name in the `CollectionResourceBaseIntegrationTest` is including upper case chars as well.. Maybe I can do this and maybe also check if we can go away form the cql base tests, we don't need the cql anymore in the int tests..

PR #347, opened 2023-04-12
Fix broken postman workflow

### LINKED ISSUE #227: Indicate minimum required versions of docker and docker-compose in README
https://github.com/stargate/jsonapi#development-guide does a good job of specifying that the project requires Java 17. But would also be great to indicate which versions of docker and docker-compose you need.

For example, I had docker-compose v1.25 installed, which is what you get if you `sudo apt install docker-compose` on Ubuntu. Got the following error.

```
~/Workspace/meanIT/jsonapi/docker-compose$ sudo ./start_dse_68_dev_mode.sh 
[sudo] password for val: 
Running with DSE 6.8.32, Stargate v2.0.9, JSON API v1.0.0-ALPHA-2-SNAPSHOT
ERROR: The Compose file './docker-compose-dev-mode.yml' is invalid because:
Unsupported config option for services.coordinator: 'healthcheck'
Unsupported config option for services.jsonapi: 'healthcheck'
services.jsonapi.depends_on contains an invalid type, it should be an array
```

I upgraded to docker-compose v2.17.0-rc1 by downloading from docker-compose's release page, and that worked fine.

Perhaps we recommend docker-compose 2.x and docker engine 20.x?

jeffreyscarpenter:
This fixes the GH Actions workflow that runs the JSON postman collection against a docker-compose configuration.

Also Fixes #227 

**Checklist**
- [x] Changes manually tested


.github/workflows/postman-docker.yaml
```diff
@@ -65,6 +58,13 @@ jobs:
           cd docker-compose
           ./start_dse_68_dev_mode.sh -j $JSONTAG -t $SGTAG
 
+      - name: Install Postman CLI
```
jeffreyscarpenter, line 61: even with the changes made in this PR to run against docker compose v2 (and thus ensure the `--wait` flag is used), it is still not guaranteed that postman will succeed on its first few attempts to call the JSON API. For some reason even though the JSON API health check passes and docker compose command exits, it appears the JSON API port 8080 is not immediately available on the Docker host. This change moves the postman CLI installation to after docker compose is started to make sure the port is available. Admittedly a bit of a hack.

docker-compose/README.md
```diff
@@ -2,59 +2,66 @@
 
 This directory provides two ways to start the JSON API and Stargate coordinator with DSE 6.8 using `docker compose`.
 
-## Stargate JSON API with 3-node DSE 6.8 cluster
+## Prerequisites
+
+### Docker / Docker Compose Versions
+
+Make sure that you have Docker engine 20.x installed, which should include Docker compose 2.x. Our compose files rely on features only available in the Docker compose v2 file format.
```
jeffreyscarpenter, line 9: Addresses #227

docker-compose/README.md
```diff
@@ -2,59 +2,66 @@
...
+
+Follow instructions under the [Script options](#script-options) section to use the locally built image.
+
+## Stargate JSON API with 3-node DSE 6.8 cluster
 
 You can start a simple Stargate configuration with the following command:
 
 ```
-docker-compose up -d
+./start_dse_68.sh
```
jeffreyscarpenter, line 27: rewrote instructions to prefer the wrapper script so users get the Docker version check.

tatu-at-datastax (approved):
LGTM

PR #354, opened 2023-04-14
Add $size

johnsmartco:
**What this PR does**:

Fixes #<issue number>

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/docs/jsonapi-spec.md
+++ b/docs/jsonapi-spec.md
@@ -1441,7 +1441,7 @@ TODO:
 The `$size` operation allows users to match any array with the number of elements specified by the argument. Example:
 
 ```json5
-db.collection.find( { city: { $size: 2 } } );
+db.collection.find( { field: { $size: 2 } } );
 ```
 
 That expression returns, e.g.: `{ field: [ red, green ] }` and `{ field: [ apple, lime ] }`, but not `{ field: fruit }` or `{ field: [ orange, lemon, grapefruit ] }`.
--- a/docs/jsonapi-spec.md
+++ b/docs/jsonapi-spec.md
@@ -1435,8 +1435,6 @@ That expression returns, e.g.: `{ field: [ red, green ] }` and `{ field: [ apple
 
 If the given field is not an array, there's no match. `$size` should ignore non-arrays.
 
-Queries cannot use indexes for the $size portion of a query, although the other portions of a query can use indexes if applicable.
-
 
 ### Projection Clause
 
--- a/docs/jsonapi-spec.md
+++ b/docs/jsonapi-spec.md
@@ -403,19 +403,27 @@ The maximum length of an array is 100 elements.
 
 ### Equality handling with arrays and subdocs
 
-Given the query `{'foo' : "bar"}`, this matches: 
+Given the query `{"foo" : "bar"}`, this matches: 
 
 * `foo` as a string with value `bar` 
 
 * `foo` is an array that contains at least one item that is the string `"bar"`. The length of the array is not important.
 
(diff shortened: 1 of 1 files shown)


johnsmartco:
Ready for follow-up review @maheshrajamani.  Thanks!

docs/jsonapi-spec.md
```diff
@@ -1419,26 +1419,26 @@ The `$all` operation allows users to check for documents that have an array fiel
...
 ```
 
-#### $elemMatch operation
+That expression returns, e.g.: `{ field: [ red, green ] }` and `{ field: [ apple, lime ] }`, but not `{ field: fruit }` or `{ field: [ orange, lemon, grapefruit ] }`.
 
-TODO:
+If the given field is not an array, there's no match. `$size` should ignore non-arrays.
 
-#### $size operation
+`$size` does not accept ranges of values. To select documents based on fields with different numbers of elements, a user could create a counter field to increment when adding elements to a field.
```
maheshrajamani, line 1438: These 3 lines are not required. The $size is based on pre build index. Also we don't support counter columns.

maheshrajamani (changes requested):

docs/jsonapi-spec.md
johnsmartco: thanks will fix...

docs/jsonapi-spec.md
```diff
@@ -1419,26 +1419,24 @@ The `$all` operation allows users to check for documents that have an array fiel
...
 ```
 
-#### $elemMatch operation
+That expression returns, e.g.: `{ field: [ red, green ] }` and `{ field: [ apple, lime ] }`, but not `{ field: fruit }` or `{ field: [ orange, lemon, grapefruit ] }`.
 
-TODO:
+If the given field is not an array, there's no match. `$size` should ignore non-arrays.
 
-#### $size operation
+Queries cannot use indexes for the $size portion of a query, although the other portions of a query can use indexes if applicable.
```
maheshrajamani, line 1438: Should have seen this before, $size index can be used along other filters and it uses the index. Can we remove this too?

maheshrajamani (approved):

docs/jsonapi-spec.md
johnsmartco: yes will do

docs/jsonapi-spec.md
```diff
@@ -1419,26 +1419,22 @@ The `$all` operation allows users to check for documents that have an array fiel
...
 
-... is equivalent to:
+#### $size operation
 
-```json5
-{ $and: [ { tags: "ssl" }, { tags: "security" } ] }
+The `$size` operation allows users to match any array with the number of elements specified by the argument. Example:
 
+```json5
+db.collection.find( { field: { $size: 2 } } );
```
vkarpov15, line 1431: This isn't valid JSON. I think this should be `{ "field": { "$size": 2 } }`

docs/jsonapi-spec.md
```diff
@@ -1419,26 +1419,22 @@ The `$all` operation allows users to check for documents that have an array fiel
...
 ```
 
-#### $elemMatch operation
-
-TODO:
+That expression returns, e.g.: `{ field: [ red, green ] }` and `{ field: [ apple, lime ] }`, but not `{ field: fruit }` or `{ field: [ orange, lemon, grapefruit ] }`.
```
vkarpov15, line 1434: Quotes around "field" and string values, so `{ "field": ["red", "green"] }`. We should use valid JSON where possible for consistency and so code samples are easier to copy/paste.

PR #358, opened 2023-04-15
NoSQLBench workloads - auto token generation

jeffreyscarpenter:
Update JSON API NoSQLBench workloads to take advantage of automatic token generation. 

Also includes minor changes to emulate improvements made to Docs API NB workloads under https://github.com/nosqlbench/nosqlbench/pull/1201

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/nosqlbench/http-jsonapi-crud-basic.yaml
+++ b/nosqlbench/http-jsonapi-crud-basic.yaml
@@ -1,8 +1,8 @@
-min_version: "4.17.15"
+min_version: "5.17.1"
 
 # nb5 -v run driver=http yaml=http-jsonapi-crud-basic jsonapi_host=my_jsonapi_host auth_token=$AUTH_TOKEN
 
-description: |
+description: >2
   This workload emulates CRUD operations for the Stargate JSON API.
   It generates a simple JSON document to be used for writes and updates.
   Note that jsonapi_port should reflect the port where the JSON API is exposed (defaults to 8080).
@@ -21,13 +21,14 @@ bindings:
   #   single host: jsonapi_host=host1
   #   multiple hosts: jsonapi_host=host1,host2,host3
   #   multiple weighted hosts: jsonapi_host=host1:3,host2:7
-  weighted_hosts: WeightedStrings('<<jsonapi_host:jsonapi>>')
+  weighted_hosts: WeightedStrings('<<jsonapi_host:stargate>>')
 
   # spread into different spaces to use multiple connections
   space: HashRange(1,<<connections:20>>); ToString();
 
   # http request id
   request_id: ToHashedUUID(); ToString();
+  token: Discard(); Token('<<auth_token:>>','<<uri:http://localhost:8081/v1/auth>>', '<<uid:cassandra>>', '<<pswd:cassandra>>');
 
   seq_key: Mod(<<docscount:10000000>>); ToString() -> String
   random_key: Uniform(0,<<docscount:10000000>>); ToString() -> String
@@ -54,7 +55,7 @@ blocks:
         X-Cassandra-Token: "<<auth_token:my_auth_token>>"
         Content-Type: "application/json"
         ok-body: ".*\"ok\":1.*"
-        body: |
+        body: >2
           {
             "createNamespace": {
               "name": "<<namespace:jsonapi_crud_basic>>"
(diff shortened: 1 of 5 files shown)


jeffreyscarpenter:
This has been tested with locally compiled version of nb5, but should not be merged until the release of NB 5.17.3.

jeffreyscarpenter:
> This has been tested with locally compiled version of nb5, but should not be merged until the release of NB 5.17.3.

Actually this is not correct. Tests pass with NB 5.17.2, and this can be approved and merged. https://github.com/stargate/jsonapi/actions/runs/4708744043

nosqlbench/http-jsonapi-crud-basic.yaml
```diff
@@ -1,8 +1,8 @@
-min_version: "4.17.15"
+min_version: "5.17.1"
 
 # nb5 -v run driver=http yaml=http-jsonapi-crud-basic jsonapi_host=my_jsonapi_host auth_token=$AUTH_TOKEN
 
-description: |
+description: >2
```
ivansenic, line 5: is this change wanted? what s `>2` doing?

nosqlbench/http-jsonapi-crud-basic.yaml
```diff
@@ -21,13 +21,14 @@ bindings:
...
   #   multiple weighted hosts: jsonapi_host=host1:3,host2:7
-  weighted_hosts: WeightedStrings('<<jsonapi_host:jsonapi>>')
+  weighted_hosts: WeightedStrings('<<jsonapi_host:stargate>>')
 
   # spread into different spaces to use multiple connections
   space: HashRange(1,<<connections:20>>); ToString();
 
   # http request id
   request_id: ToHashedUUID(); ToString();
+  token: Discard(); Token('<<auth_token:>>','<<uri:http://localhost:8081/v1/auth>>', '<<uid:cassandra>>', '<<pswd:cassandra>>');
```
ivansenic, line 31: will this do an auth request on each cycle?

ivansenic (approved):
The `token` is added only in the `nosqlbench/http-jsonapi-crud-basic.yaml`, other files don't have this change.. Will this be updated?

nosqlbench/http-jsonapi-crud-basic.yaml
jeffreyscarpenter: no, NB caches the token

nosqlbench/http-jsonapi-crud-basic.yaml
tatu-at-datastax: I think that's a YAML feature, where there's a myriad of ways to specify how text blocks are included (String values, not YAML markup -- often embedded JSON, XML, CSV, shell scripts)
Which can be convenient but also confusing until you learn all of them when someone uses them (at least I find them confusing).

jeffreyscarpenter:
> The `token` is added only in the `nosqlbench/http-jsonapi-crud-basic.yaml`, other files don't have this change.. Will this be updated?

Good catch, thanks. I have pushed a fix so you can re-review.

nosqlbench/http-jsonapi-crud-basic.yaml
```diff
@@ -21,13 +21,14 @@ bindings:
   #   single host: jsonapi_host=host1
   #   multiple hosts: jsonapi_host=host1,host2,host3
   #   multiple weighted hosts: jsonapi_host=host1:3,host2:7
-  weighted_hosts: WeightedStrings('<<jsonapi_host:jsonapi>>')
+  weighted_hosts: WeightedStrings('<<jsonapi_host:stargate>>')
```
ivansenic, line 24: I am still like why don't we have `localhost` as default? In addition, this change is also only available in this file, not the others.

nosqlbench/http-jsonapi-crud-basic.yaml
ivansenic: Good. but can you add a small comment here, explaining what the parameters are.. I guess first parameter is token given from the command line and if given, don't call the auth api? In addition it's good to know that the result of the token is saved in the `{{token}}` var.

ivansenic (approved):
Looks good, have a look on the remaining comments before merging..

jeffreyscarpenter:
FYI @maheshrajamani - I made a couple additional small improvements to address feedback from @ivansenic re-review.

PR #359, opened 2023-04-17
use upper-case collection and namespace names in the int tests

ivansenic:
**What this PR does**:
Adds integration tests for the #343, by simply changing how collection and namespace names are generated.. 

I also refactored the tests and moved away from the `CqlEnabledBaseIntegrationTest`, as now we can create and drop keyspaces using the jsonapi directly.. 

I was also able to implement a workaround and do creation/dropping of namespaces/collections in `@AfterAll/@BeforeAll` methods.. This provides one big advantage -> ability to run a single test from the class.. This was not possible before as it was depending on the order of the executed methods..

Note that majority of changes in this PR is simply changing `keyspaceId.asInternal() -> namespaceName` and was executed using the IDE. The important changes are in the two base int test classes.

Followup of #343

### DIFF
--- a/src/test/java/io/stargate/sgv2/jsonapi/api/v1/FindIntegrationTest.java
+++ b/src/test/java/io/stargate/sgv2/jsonapi/api/v1/FindIntegrationTest.java
@@ -296,7 +296,7 @@ public class FindIntegrationTest extends AbstractCollectionIntegrationTestBase {
           .contentType(ContentType.JSON)
           .body(json)
           .when()
-          .post(CollectionResource.BASE_PATH, keyspaceId.asInternal(), collectionName)
+          .post(CollectionResource.BASE_PATH, namespaceName, collectionName)
           .then()
           .statusCode(200)
           .body("data.count", is(0))


src/test/java/io/stargate/sgv2/jsonapi/api/v1/CollectionResourceBaseIntegrationTest.java
```diff
@@ -20,7 +20,9 @@
 
 @TestMethodOrder(MethodOrderer.OrderAnnotation.class)
 public abstract class CollectionResourceBaseIntegrationTest extends CqlEnabledIntegrationTestBase {
```
ivansenic, line 22: The namespace name is not easy to change as it depends on the `CqlEnabledIntegrationTestBase`.. I think we should go away from the CQL based tests, as now we have all the commands to run the tests purely using jsonapi commands.. Should I adapt that in this PR?

tatu-at-datastax: If it's simple enough to do, yes, I think that makes sense.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/AbstractNamespaceIntegrationTestBase.java
```diff
@@ -0,0 +1,85 @@
...
+import io.restassured.RestAssured;
+import io.restassured.http.ContentType;
+import io.stargate.sgv2.api.common.config.constants.HttpConstants;
+import org.apache.commons.lang3.RandomStringUtils;
+import org.eclipse.microprofile.config.ConfigProvider;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.TestInstance;
+
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
```
ivansenic, line 17: This is the only thing that is not standard anymore and still annoys me.. I can remove this, but then I need to make the `createNamespace()` method `static` as well, meaning I need a way to reference a `namespace` as a static var.. But then all test would have the same namespace and that is not what we want.. 

Static methods are also not overriable, so I can not allow sub-classes to control the name as well..

I guess we must live with this..

tatu-at-datastax (approved):
LGTM

PR #366, opened 2023-04-17
Fixes #365: indicate max docs, actual docs inserted in error (fail) message

### LINKED ISSUE #365: Improve failure message on "inserting too many documents" to include limit, number sent
Looks like currently we have `MaxInsertManyDocuments` and `MaxInsertManyDocumentsValidation` classes that give static fail message if attempting insert more documents than specified maximum:

    Request invalid, the field postCommand.command.documents not valid: amount of documents to insert is over the max limit.

since the limit may be reconfigured, this can be difficult for users to reason about: instead we should include:

* Maximum number of documents allowed
* Number of documents sent

which would make it easier for caller to adjust number being sent.

tatu-at-datastax:
**What this PR does**:

Adds `" (<actual docs to insert> vs <max docs to insert>)"` in error message when validating number of documents to insert, so caller gets indication of what limit was vioated.

Fixes #365

**Checklist**
- [ ] Changes manually tested


kathirsvn (approved):
LGTM

PR #367, opened 2023-04-18
closes #313: remove data.count from the findOne command responses

### LINKED ISSUE #313: Spec: should `findOne()` have `count` entry in response?
Working on #300, I noticed what seems like a minor discrepancy in our spec: inclusion (or not) of `count` property in response for `find()`, `findOne()` and `findOneAndModify()`.
It seems to me that:

* Spec suggests that ONLY `find()` should have it (see https://github.com/stargate/jsonapi/blob/main/docs/jsonapi-spec.textile#commandFind -- "'data' -- Present with fields : docs, count and nextPageState." -- but NOT `findOne()` ("'data' -- Present with fields : docs only"
* Our implementation always adds `count` in `data`, as long as `docs` is a non-null array.

So, it looks like either Spec should be updated to state that `count` is included for `findOne()`/`findOneAndUpdate()` OR our implementation should drop it from them.

ivansenic:
**What this PR does**:
Removes `data.count` in the `findOneXXX` operations.

The current implementation is based on the `limit > 1` in order to decide when to include the count. However, this also results in the problem that if user executes `find` command with `options.limit: 1`, the count would not be present (see failing test).. Now to avoid this, we need to extend the `FindOperation` introduce yet another parameter and then change all the resolvers, tests, etc. I am just wandering if this is worth it, was this such a big issue? @maheshrajamani @tatu-at-datastax @vkarpov15 please advise and let me know how to continue!

Fixes #313


tatu-at-datastax:
My understanding is that it would be "nice" to remove it, but that it is not a real Problem on its own.

So if removal is difficult/intrusive at this point there is no big upside. But I'll let @vkarpov15 comment on his view.

ivansenic:
> > ntation is based on the `limit > 1` in order to decide when to include the count. However, this also results in the problem that if user executes `find` command with `options.limit: 1`, the count would not be present (see failing test).. Now to avoid this, we need to extend the `FindOperation` introduce yet another parameter and then change all the resolvers, tests, etc. I am just wandering if this is worth it, was this such a big issue? @maheshrajamani @tatu-at-datastax @vkarpov15 please advise and let me know how to continue!
> 
> IMO., not sure if count is ever used from find operation response. If @vkarpov15 is ok we should remove it for all commands.

@vkarpov15 Valeri can you advise here please?

vkarpov15 (approved):
LGTM. @tatu-at-datastax @maheshrajamani are both right - removing `count` is a "nice to have", not critical. But would be ideal to remove it from all commands, with the exception of `countDocuments()`.

tatu-at-datastax:
Let's do this!

ivansenic:
Ack, will remove it completely from code and spec today.

src/main/java/io/stargate/sgv2/jsonapi/StargateJsonApi.java
```diff
@@ -431,12 +429,11 @@
...
-                            "matchedCount": 1,
-                            "modifiedCount": 1
-                          }
+                          ]
+                        },
+                        "status": {
+                          "matchedCount": 1,
+                          "modifiedCount": 1
                         }
                       }
```
ivansenic, line 438: all examples had `status` in the `data` field, that's wrong so I fixed it

ivansenic:
@amorton FYI the team agreed to remove count from the `data` field in response.. Please ack and approve this change..

maheshrajamani (approved):
LGTM

PR #348, opened 2023-04-13
`$in` operator support for `_id` field

maheshrajamani:
**What this PR does**:
`$in` support only for `_id` field
**Which issue(s) this PR fixes**:
Fixes #346 

**Checklist**
- [x] Changes manually tested
- [x] Automated Tests added/updated
- [ ] Documentation added/updated
- [x] CLA Signed: [DataStax CLA](https://cla.datastax.com/)

PR #368, opened 2023-04-18
`$in` operator support for empty array

maheshrajamani:
**What this PR does**:
Changes based on review comments in #348 
1. Support empty array in filter and return no result. 
2. Create new configuration to check maximum $in operator array size.

**Checklist**
- [x] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DBFilterBase.java
```diff
@@ -237,6 +237,7 @@ public List<BuiltCondition> getAll() {
                   getDocumentIdValue(values.get(0))));
 
         case IN:
+          if (values.isEmpty()) return List.of();
```
ivansenic, line 240: Why is this supported? and if it's empty array is the whole `$in` condition effectively ignored?

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/FindOperation.java
```diff
@@ -263,14 +268,19 @@ private List<List<BuiltCondition>> buildConditions(DBFilterBase.IDFilter additio
...
+        return null;
+      } else {
+        return inSplit.stream()
+            .map(
+                idCondition -> {
+                  List<BuiltCondition> conditionsWithId = new ArrayList<>(conditions);
+                  conditionsWithId.add(idCondition);
+                  return conditionsWithId;
+                })
+            .collect(Collectors.toList());
```
ivansenic, line 282: Wait so is the primary key supporting the Cassandra `IN` or not? Or we anyway want to run the `IN` as multiple queries?

ivansenic (approved):
Looks good..

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DBFilterBase.java
ivansenic: In addition, how would this actually reflect the query:

```
{
  key: { $in: [] },
  smth-else: { $eq: 'geroge' }
}
```

I guess this would not work atm? If it would, can you add a test please..

maheshrajamani: No condition is not ignored, basically no document is sent back in response.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/FindOperation.java
maheshrajamani: There are 2 reasons in conditions on primary key is run as multiple queries. 
1. Reduce load on Coordinator which eventually does the same thing.
2. Cassandra does not allow additional condition if IN is used on primary key.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DBFilterBase.java
maheshrajamani: > I guess this would not work atm? If it would, can you add a test please..

If there is $in with empty array we don't build and run query irrespective of other conditions available as part of filter(everything is `and` operations), this will return empty response.

ivansenic: Perfect thanks :+1:

PR #369, opened 2023-04-18
closes #350: errorCode and exceptionClass to be reported always with …

### LINKED ISSUE #350: Errors in response
If there is an error thrown during deserialization of commands, response has array of errors as below. The error message in index 0 adds no value. In my opinion it needs to be removed. 
Also the error code is missing on the raised error, in this case expecting `"errorCode" : INVALID_FILTER_EXPRESSION`

`{
    "errors": [
        {
            "message": "$in operator must have at least one value (through reference chain: io.stargate.sgv2.jsonapi.api.model.command.impl.FindOneCommand[\"filter\"])",
            "exceptionClass": "JsonMappingException"
        },
        {
            "message": "$in operator must have at least one value",
            "exceptionClass": "JsonApiException"
        }
    ]
}`

@ivansenic @amorton @tatu-at-datastax

ivansenic:
**What this PR does**:
As discussed in the https://github.com/stargate/jsonapi/issues/350#issuecomment-1513166081 I pushed the implementation that fixes `errorCode` missing. We still need to agree on the handling of the cause and if we do want the error array to include the cause..

Fixes #350

**Checklist**
- [x] Agree on cause handling

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/exception/JsonApiException.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/exception/JsonApiException.java
@@ -43,11 +43,8 @@ public class JsonApiException extends RuntimeException implements Supplier<Comma
       message = errorCode.getMessage();
     }
 
-    // add error code as error field
-    Map<String, Object> fields = Map.of("errorCode", errorCode.name());
-
     // construct and return
-    CommandResult.Error error = new CommandResult.Error(message, fields);
+    CommandResult.Error error = getCommandResultError(message);
 
     // handle cause as well
     Throwable cause = getCause();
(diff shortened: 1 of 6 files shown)


src/main/java/io/stargate/sgv2/jsonapi/exception/JsonApiException.java
```diff
@@ -59,6 +56,12 @@ public CommandResult get() {
     }
   }
 
+  public CommandResult.Error getCommandResultError(String message) {
+    Map<String, Object> fields =
+        Map.of("errorCode", errorCode.name(), "exceptionClass", this.getClass().getSimpleName());
```
ivansenic, line 61: the `JsonApiException` was not including the `exceptionClass` before so I added it, I guess this is also helpful information..

maheshrajamani (approved):

vkarpov15 (commented):
LGTM :+1:

Issue #362, opened 2023-04-17
Add first-class support for Date-Time values for JSON-API


tatu-at-datastax:
Currently we support a subset of datatypes Mongoose supports, basically matching native JSON types (Objects/sub-docs, Arrays, Number, Strings, Booleans, null), and one pseudo-type (Doc Id).

It would make sense to also support Date/Time, so users need not handle conversion to/from JSON String or Number types, but can use native Javascript Date type (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date) on client-side and get it efficiently handled on JSON API server side.

Current thinking is to convention defined by EJSON (https://docs.meteor.com/api/ejson.html), in which numeric (milliseconds since 1.1.1970 timestamp, standard used by Java and Javascript as well as many other platforms) is encapsulates as a single-entry JSON Object like:

```
{ "$date": 1358205756553 }
```

This allows for reliable auto-detection because JSON API Object/sub-doc properties can not start with `$` character.

Details of this approach will be added here and after/during prototyping Spec needs to be updated. We will also file separate issues for different parts of handling; so tasks include:

* Update Spec to describe storage and logic
* Create new index field for Date/Time values, update hashing, shredder (#370)
* Change Insert and Update operation to update timestamp index column (#378)
* Change doc-id (`_id`) handling to recognize and support new id type (#371)
* Test that (and change if need be) update operators work with Date/Time values (#372)
* Test that (and change if necessary) Projection works with Date/Time values (#373)
* Change in-memory sort to auto-detect and convert Date/Time values appropriately for sorting (#374)
* Change query filters to auto-detect and convert Date/Time values appropriately for equality queries (#375)
* Add support for `$currentDate` update operation (#397)

vkarpov15:
Just a couple of details to confirm:

1. Mongoose only has one Date type that corresponds to a Date in JavaScript and a [Date in MongoDB](https://www.mongodb.com/docs/manual/reference/bson-types/#date). Mongoose doesn't have first class support for [MongoDB's internal timestamp type](https://www.mongodb.com/docs/manual/reference/bson-types/#timestamps).
2. Mongoose does not expect timezone information to be stored with the date. Dates are 64 bit ints in MongoDB.
[...]

PR #376, opened 2023-04-18
Add "query_timestamp_values" column in Collections, update shredder

### LINKED ISSUE #370: (Date/Time support) Add DB column `query_timestamp_values`, Shredding support
The first sub-task for Date/Time support (#362) is to add database query column -- `query_timestamp_values` -- and change Shredder to populate this column. Also modify storage in other columns appropriately: this requires changes to value hashing.

tatu-at-datastax:
**What this PR does**:

Adds `query_timestamp_values` columns and will populate it via shredder. Also includes hash value addition for Date/Time values.

Note that this is just the first small part of full Date/Time handling (see #362 for full scope)

Fixes #370

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/shredding/model/AtomicValue.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/shredding/model/AtomicValue.java
@@ -1,6 +1,7 @@
 package io.stargate.sgv2.jsonapi.service.shredding.model;
 
 import java.math.BigDecimal;
+import java.util.Date;
 
 /**
  * Simple immutable value container class used when shredding documents, to contain "full" value as
@@ -23,6 +24,11 @@ public record AtomicValue(DocValueType type, String typedFullValue, DocValueHash
   public static AtomicValue forNumber(BigDecimal num) {
     // For Numbers just make sure not to use Engineering notation:
     return create(DocValueType.NUMBER, num.toPlainString());
+  }
+
+  public static AtomicValue forTimestamp(Date dt) {
+    // Long-value of timestamp still fits without truncation
+    return create(DocValueType.TIMESTAMP, String.valueOf(dt.getTime()));
   }
 
   static AtomicValue create(DocValueType type, String fullValue) {
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/shredding/model/WritableShreddedDocument.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/shredding/model/WritableShreddedDocument.java
@@ -5,8 +5,10 @@ import com.fasterxml.jackson.databind.node.ArrayNode;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import io.stargate.sgv2.jsonapi.service.shredding.JsonPath;
 import io.stargate.sgv2.jsonapi.service.shredding.ShredListener;
+import io.stargate.sgv2.jsonapi.util.JsonUtil;
 import java.math.BigDecimal;
 import java.util.Collections;
+import java.util.Date;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedHashSet;
@@ -33,6 +35,7 @@ public record WritableShreddedDocument(
     Map<JsonPath, Boolean> queryBoolValues,
     Map<JsonPath, BigDecimal> queryNumberValues,
     Map<JsonPath, String> queryTextValues,
+    Map<JsonPath, Date> queryTimestampValues,
     Set<JsonPath> queryNullValues) {
   public static Builder builder(DocValueHasher hasher, DocumentId id, UUID txID, String docJson) {
     return new Builder(hasher, id, txID, docJson);
@@ -65,6 +68,7 @@ public record WritableShreddedDocument(
     private Map<JsonPath, Boolean> queryBoolValues;
     private Map<JsonPath, BigDecimal> queryNumberValues;
     private Map<JsonPath, String> queryTextValues;
+    private Map<JsonPath, Date> queryTimestampValues;
     private Set<JsonPath> queryNullValues;
 
     public Builder(DocValueHasher hasher, DocumentId id, UUID txID, String docJson) {
@@ -94,6 +98,7 @@ public record WritableShreddedDocument(
           _nonNull(queryBoolValues),
           _nonNull(queryNumberValues),
           _nonNull(queryTextValues),
+          _nonNull(queryTimestampValues),
           _nonNull(queryNullValues));
     }
 
@@ -112,13 +117,29 @@ public record WritableShreddedDocument(
      */
 
     @Override
-    public void shredObject(JsonPath path, ObjectNode obj) {
-      addKey(path);
+    public boolean shredObject(JsonPath path, ObjectNode obj) {
+      // Either Sub-doc or EJSON-encoded Date/time value:
+      Date dtValue = JsonUtil.extractEJsonDate(obj);
+      if (dtValue != null) {
+        shredDateTime(path, dtValue);
+        return false; // We are done
+      }
 
+      addKey(path);
       if (subDocEquals == null) {
         subDocEquals = new HashMap<>();
       }
       subDocEquals.put(path, hasher.hash(obj).hash());
+      return true; // proceed to shred individual entries too
+    }
+
+    private void shredDateTime(JsonPath path, Date dtValue) {
+      addKey(path);
+      if (queryTimestampValues == null) {
+        queryTimestampValues = new HashMap<>();
+      }
+      queryTimestampValues.put(path, dtValue);
+      addArrayContains(path, hasher.timestampValue(dtValue).hash());
     }
 
     @Override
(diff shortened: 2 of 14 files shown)


src/main/java/io/stargate/sgv2/jsonapi/service/shredding/model/WritableShreddedDocument.java
```diff
@@ -65,6 +68,7 @@ public static class Builder implements ShredListener {
     private Map<JsonPath, Boolean> queryBoolValues;
     private Map<JsonPath, BigDecimal> queryNumberValues;
     private Map<JsonPath, String> queryTextValues;
+    private Map<JsonPath, Date> queryTimestampValues;
```
maheshrajamani, line 71: Are we not going to keep it java.time.Instant?

src/main/java/io/stargate/sgv2/jsonapi/util/JsonUtil.java
```diff
@@ -58,4 +61,22 @@ public static boolean equalsOrdered(JsonNode node1, JsonNode node2) {
...
+   * @return
+   */
+  public static Date extractEJsonDate(JsonNode json) {
+    if (json.isObject() && json.size() == 1) {
+      JsonNode value = json.path(EJSON_VALUE_KEY_DATE);
+      if (value.isIntegralNumber() && value.canConvertToLong()) {
+        return new Date(value.longValue());
+      }
+    }
+    return null;
```
maheshrajamani, line 80: Do we need to throw error if the incoming value is not a number?

maheshrajamani (approved):
LGTM, May need to handle if the value is not a proper epoch.

src/main/java/io/stargate/sgv2/jsonapi/service/shredding/model/WritableShreddedDocument.java
tatu-at-datastax: We can choose that if it makes more sense -- I don't have strong preference. All we need is a wrapper for 64-bit `long`. Let me read a bit about `java.time.Instant`.

src/main/java/io/stargate/sgv2/jsonapi/util/JsonUtil.java
tatu-at-datastax: Hmmmh. Good question. Yes, probably should, will add.

tatu-at-datastax: @maheshrajamani Added better validation, now catches every JSON Object that looks like EJSON encoded (1 entry, key starts with "$").

src/main/java/io/stargate/sgv2/jsonapi/service/shredding/model/WritableShreddedDocument.java
tatu-at-datastax: Ok, as per JDK Javadocs:

"The range of an instant requires the storage of a number larger than a long. To achieve this, the class stores a long representing epoch-seconds and an int representing nanosecond-of-second, which will always be between 0 and 999,999,999. "

so I think that for our purposes it is more efficient to use otherwise outdated `java.util.Date`; that uses less memory.

I am ok going with alternatives (including `java.time.Instant`) if we find reasons to do so, but I think for now will go with `java.util.Date` just as a light-weight wrapper.

ivansenic: I am for `java.time.Instant` as well.. Date is outdated :smile:

src/main/java/io/stargate/sgv2/jsonapi/service/shredding/model/AtomicValue.java
```diff
@@ -25,6 +26,11 @@ public static AtomicValue forNumber(BigDecimal num) {
     return create(DocValueType.NUMBER, num.toPlainString());
   }
 
+  public static AtomicValue forTimestamp(Date dt) {
+    // Long-value of timestamp still fits without truncation
+    return create(DocValueType.TIMESTAMP, String.valueOf(dt.getTime()));
```
ivansenic, line 31: Not sure if matters but why are we storing the long as a String here?

src/main/java/io/stargate/sgv2/jsonapi/util/JsonUtil.java
```diff
@@ -58,4 +63,47 @@ public static boolean equalsOrdered(JsonNode node1, JsonNode node2) {
...
+   * value -- that is, JSON Object with 1 entry, name of which starts with "$" character -- without
+   * decoding value. No validation is done beyond feasibility: caller is likely to want to call
+   * other methods like {@link #extractEJsonDate} for further processing and validation/
+   *
+   * @param json JSON value to check
+   * @return {@code true} if value looks like it has to be EJSON value (valid or invalid); {@code
+   *     false} if not.
+   */
+  public static boolean looksLikeEJsonValue(JsonNode json) {
+    return json.isObject() && json.size() == 1 && json.fieldNames().next().startsWith("$");
```
ivansenic, line 78: should we just have a getter here? `json.fieldNames().next()` -> `json.fieldNames().get(0)`, avoids creating iterator for a single value

src/main/java/io/stargate/sgv2/jsonapi/util/JsonUtil.java
```diff
@@ -58,4 +63,47 @@ public static boolean equalsOrdered(JsonNode node1, JsonNode node2) {
...
+   *
+   * @param json JSON value to check
+   * @return
+   */
+  public static Date extractEJsonDate(JsonNode json, Object path) {
+    if (json.isObject() && json.size() == 1) {
+      JsonNode value = json.get(EJSON_VALUE_KEY_DATE);
+      if (value != null) {
+        if (value.isIntegralNumber() && value.canConvertToLong()) {
+          return new Date(value.longValue());
```
ivansenic, line 94: Do we need to to wrap this into a date or instant at all?

src/test/java/io/stargate/sgv2/jsonapi/service/shredding/ShredderTest.java
```diff
@@ -207,6 +209,73 @@ public void shredSimpleWithNumberIdWithTrailingZeroes() {
     }
   }
 
+  @Nested
+  class EJSONDateTime {
+    @Test
+    public void shredDocWithDateTimeColumn() {
+      final long testTimestamp = defaultTestDate().getTime();
```
ivansenic, line 216: would remove the `defaultTestDate()` and simply use `Instant.now()`.. Make tests more robust as date changes all the time..

ivansenic (commented):
Ahh this was merged already, still few comments.

src/main/java/io/stargate/sgv2/jsonapi/service/shredding/model/WritableShreddedDocument.java
ivansenic: That s what I pointed out in one of the comments, maybe we should simply operate on long values.

maheshrajamani: Agree. @tatu-at-datastax Can we have this as just `long`?

src/main/java/io/stargate/sgv2/jsonapi/service/shredding/model/AtomicValue.java
tatu-at-datastax: This is for Digest value ("hash"); we cannot have polymorphism, so everything ends up as String, when stored in `array_contains`: which is used for equality filtering (except for special case of document id). So same done for every value type during shredding (Arrays and Objects are composed of digests of leaf values along with path).

src/main/java/io/stargate/sgv2/jsonapi/util/JsonUtil.java
tatu-at-datastax: To use `get()` would need to know the name to look for, and here we want to see if that one key that exists starts with "$". I don't think there is an accessor that would give us that key without knowing it, or creating Iterator or equivalent.
Would be happy to avoid iterator if there's an alternate mechanism.

src/main/java/io/stargate/sgv2/jsonapi/util/JsonUtil.java
tatu-at-datastax: Not on server side; we just want something to hold on to that wraps a `long` and is type-safe for documentation.
`Instant` has 2 `longs` (millisecond like `Date`, and then nanosecond counter).

Mongo has only one date-related operator (`$currentDate`), for everything else we just convert to/from JSON, or get 64-bit value (for sorting).

src/test/java/io/stargate/sgv2/jsonapi/service/shredding/ShredderTest.java
tatu-at-datastax: To me it's exactly the opposite when value changes for any test. I guess matter of preference. :)
(that is, test is not necessarily stable, could theoretically pass or fail depending on luck).

But I can change this along with other changes.

src/main/java/io/stargate/sgv2/jsonapi/service/shredding/model/WritableShreddedDocument.java
tatu-at-datastax: I think `Date` is better for type-safety (as sort of documentation), over wrapper type `Long`.
I guess it is possible to document things with naming conventions if we really want to do it, but I am not sure there is much benefit from doing that. To me `java.util.Date` seems like a convenient abstraction to use here.

If we really want to change this, sure, it can be done, but I don't think it greatly matters.

PR #380, opened 2023-04-19
Changes for date filter

### LINKED ISSUE #375: (Date/Time support) Support full filtering on Date/Time valued fields
(part of #362 implementation -- full support for Date/Time values)

Beyond id-based queries (#371) it should be possible to filter based on regular columns with Date/Time values.
This requires changes to hash-calculation for Date/Time values; this is likely already done earlier (perhaps even for #370) but will definitely be needed here.

maheshrajamani:
**What this PR does**:
Changes for date filter support
Fixes #375 

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DBFilterBase.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DBFilterBase.java
@@ -194,7 +194,7 @@ public abstract class DBFilterBase implements Supplier<BuiltCondition> {
     private final Date dateValue;
 
     public DateFilter(String path, Operator operator, Date value) {
-      super("query_dbl_values", path, operator, value);
+      super("query_timestamp_values", path, operator, value);
       this.dateValue = value;
     }
 
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DBFilterBase.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DBFilterBase.java
@@ -15,6 +15,7 @@ import io.stargate.sgv2.jsonapi.service.bridge.serializer.CustomValueSerializers
 import io.stargate.sgv2.jsonapi.service.shredding.model.DocValueHasher;
 import io.stargate.sgv2.jsonapi.service.shredding.model.DocumentId;
 import java.math.BigDecimal;
+import java.util.Date;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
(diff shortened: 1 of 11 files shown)


tatu-at-datastax (commented):
Looks good except for possibly wrong field name in one place.

tatu-at-datastax (approved):
Approving in anticipation of that one possible misnaming getting resolved.

tatu-at-datastax (approved):

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DBFilterBase.java
```diff
@@ -188,6 +189,26 @@ boolean canAddField() {
     }
   }
 
+  /** Filters db documents based on a date field value */
+  public static class DateFilter extends MapFilterBase<Date> {
+    private final Date dateValue;
+
+    public DateFilter(String path, Operator operator, Date value) {
+      super("query_dbl_values", path, operator, value);
```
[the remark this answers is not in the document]
maheshrajamani: It's not used currently because equals operator uses array_contains column. Will fix it.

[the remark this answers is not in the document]
tatu-at-datastax: Ok makes sense then (that no tests fail). Good to match in case our logic changes wrt using index column.

vkarpov15 (commented):
LGTM :+1:

src/test/java/io/stargate/sgv2/jsonapi/api/model/command/deserializers/FilterClauseDeserializerTest.java
```diff
@@ -135,6 +136,37 @@ public void mustHandleBoolean() throws Exception {
       assertThat(filterClause.comparisonExpressions()).hasSize(1).contains(expectedResult);
     }
 
+    @Test
+    public void mustHandleDate() throws Exception {
+      String json = """
+            {"dateType": {"$date": 1672531200000}}
```
vkarpov15, line 142: I took a look and MongoDB has an [Extended JSON specification](https://www.mongodb.com/docs/manual/reference/mongodb-extended-json/#mongodb-bsontype-Date). It looks like v2 of the spec requires dates be represented as either `{ $date: '2023-04-22T01:13:08.562Z' }` or `{ $date: { $numberLong: '1682126010903' } }` .

`{ $date: 1682126010903 }` is consistent with "shell mode" in v1 of MongoDB's EJSON spec: https://www.mongodb.com/docs/manual/reference/mongodb-extended-json-v1/#date, so this syntax is not quite consistent with modern extended JSON. What do you think about changing this to require `$date: '2023-04-22T01:13:08.562Z'` ISO 8601 date string instead of millis since epoch? That is valid in both EJSON v1 and EJSON v2.

tatu-at-datastax: Converting between textual format and number is very inefficient so ideally we would use timestamp. So while ISO-8601 String is more readable, it is rather wasteful with regards to CPU costs. :-(

One question is the value for `$numberLong` really String as well, and not JSON number?

vkarpov15: Number timestamp is fine if you have a strong preference, which it sounds like you do :) No need to abide exactly by the EJSON spec, it's just a convenience.

I suspect why they made the `$numberLong` string change for EJSON v2 is that the JSON spec doesn't explicitly specify the maximum size of JSON numbers, so there may be some JSON parser for some language that doesn't

tatu-at-datastax: @vkarpov15 Ok that makes sense. And yes, I agree -- it'd be nice to be compliant with something that exists, but we are not bound. One thing we could support if it makes sense to also allow textual values. But let's use timestamp for now.

PR #393, opened 2023-04-24
Fix #373: add basic tests to verify "$date" wrt Projection

### LINKED ISSUE #373: (Date/Time support) Ensure Projection of Date/Time valued fields works
(part of #362 issue, adding full support for Date/Time values)

Projection operations should work with Date/Time valued fields: this may or may not need changes to handling, but at least we need to add tests to ensure Projections work.

tatu-at-datastax:
**What this PR does**:

Adds simple testing to ensure Projection works with Date valued fields

Fixes #373

**Checklist**
- [ ] Changes manually tested


maheshrajamani (approved):
There is a test case that excludes sub document which does the same, but can you also add a date field exclusion as separate test case.

tatu-at-datastax:
> There is a test case that excludes sub document which does the same, but can you also add a date field exclusion as separate test case.

Will do.

vkarpov15 (commented):
LGTM :+1:

Issue #400, opened 2023-04-26
Spec - Support deleteOne() and updateOne() with sort


maheshrajamani:
Update the spec for deleteOne() and updateOne() with sort support

PR #401, opened 2023-04-26
Spec update for sort clause, updateOne and deleteOne

### LINKED ISSUE #156: Design + Spec service side implementation of sort + offset
Ticket to collect design ideas for how we can implement this in the API tier

- sort on a single field 
- what restrictions do we have on the field, e.g. can you sort on an element in an array ? 
- how will we do the sort in memory, do we need to fully deserialise the whole document or can we (sometimes or all the time) use the values in the table columns ? 
- What is a reasonable limit for sort ? Based on the number of docs, or the total size of docs ? Do we need to store the byte size of a doc on the row so we know it before deserialising (probably good to know anyway) 
- How do we support cursor based pagination of sorted results? This is different to limit + offset, it is where the requested number of rows (size) is greater than our cursor page size and we need to send the result in multiple responses. Maybe we do not support cursors over sorted results. 

This would be *nice to have* for the demo release, but not at expense of other things working.

maheshrajamani:
**What this PR does**:
Fixed the spec for sort clause
Updated spec for deleteOne and updateOne

Fixes #156 #400 

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/docs/jsonapi-spec.md
+++ b/docs/jsonapi-spec.md
@@ -598,6 +598,15 @@ If an error occurs the command will still return `status` as some documents may 
 
 ### deleteOne Command
 
+*Syntax:*
+
+```bnf
+<delete-one-command> ::= delete-one <delete-one-command-payload> 
+
+<delete-one-command-payload> ::= <filter-clause>?
+                             <sort-clause>? 
+```
+
 #### deleteOne Command Options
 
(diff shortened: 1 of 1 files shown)


johnsmartco (approved):
LGTM

vkarpov15 (commented):
LGTM

PR #402, opened 2023-04-27
closes #141: introduce single and multi document responses

### LINKED ISSUE #141: consistent "document" / "documents" fields for insert commands
Ensure we have a consistent and document use of "document" / "documents" for the insertOne and insertMany commands - and this matches what we use in the response document. 

ensure same in:
- server code
- spec
- driver code
- postman
-

ivansenic:
**What this PR does**:

Introduces the multi and single document responses, based on the commands used. For find one, we will always use the `single` response.

Fixes #141

**Checklist**
- [x] Apply changes to single commands
- [x] ~Fix limit issue when creating docs~
- [x] Integration tests
- [x] Ensure `@JsonIgnore` is kicking in..
- [x] Fix wrong IDE replacements
- [x] Swagger updates
- [x] Spec updates

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/resolver/model/impl/DeleteManyCommandResolver.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/resolver/model/impl/DeleteManyCommandResolver.java
@@ -52,7 +52,7 @@ public class DeleteManyCommandResolver extends FilterableResolver<DeleteManyComm
   private FindOperation getFindOperation(CommandContext commandContext, DeleteManyCommand command) {
     List<DBFilterBase> filters = resolve(commandContext, command);
     // Read One extra document than delete limit so return moreData flag
-    return new FindOperation(
+    return FindOperation.unsorted(
         commandContext,
         filters,
         DocumentProjector.identityProjector(),
@@ -60,10 +60,6 @@ public class DeleteManyCommandResolver extends FilterableResolver<DeleteManyComm
         operationsConfig.maxDocumentDeleteCount() + 1,
         operationsConfig.defaultPageSize(),
         ReadType.KEY,
-        objectMapper,
-        null,
-        0,
-        0,
-        false);
+        objectMapper);
   }
 }
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/resolver/model/impl/DeleteManyCommandResolver.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/resolver/model/impl/DeleteManyCommandResolver.java
@@ -63,6 +63,7 @@ public class DeleteManyCommandResolver extends FilterableResolver<DeleteManyComm
         objectMapper,
         null,
         0,
-        0);
+        0,
+        false);
   }
 }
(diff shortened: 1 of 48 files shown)


src/main/java/io/stargate/sgv2/jsonapi/service/resolver/model/impl/DeleteManyCommandResolver.java
```diff
@@ -63,6 +63,7 @@ private FindOperation getFindOperation(CommandContext commandContext, DeleteMany
         objectMapper,
         null,
         0,
-        0);
+        0,
+        false);
```
maheshrajamani, line 67: Can we change here to use  FindOperation.unsorted() instead of direct constructor.

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/model/impl/UpdateManyCommandResolver.java
```diff
@@ -77,6 +77,7 @@ private FindOperation getFindOperation(CommandContext commandContext, UpdateMany
         objectMapper,
         null,
         0,
-        0);
+        0,
```
maheshrajamani, line 80: Same as DeleteManyCommandResolver, can we use FindOperation.unsorted() instead of direct constructor.

maheshrajamani (approved):
This is carried forward from the existing code. Now that we have 4 different factory method for FindOperation constructor, can that be used everywhere. Apart from DeleteManyCommandResolver and UpdateManyCommandResolver, direct FindOperation constructor is used in all the *CommandResolverTest and *OperationTest.

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/model/impl/DeleteManyCommandResolver.java
ivansenic: Yes of course..

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/model/impl/UpdateManyCommandResolver.java
ivansenic: Yes of course..

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/model/impl/DeleteManyCommandResolver.java
ivansenic: Done.

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/model/impl/UpdateManyCommandResolver.java
ivansenic: Done.

ivansenic:
@maheshrajamani Another quick look, added spec updates?

ivansenic:
> > @maheshrajamani Another quick look, added spec updates?
> 
> Spec and swagger updates looks good. Planning to change the Test classes to use the FindOperation factory methods? We can take it up as different PR if you think it will make this one very big.

Can do in separate PR.. Test classes should anyway not use constructors, but assert properties of the operation.. many of tests does that as well, so maybe it's a minor change..

PR #415, opened 2023-05-04
Http status code change

### LINKED ISSUE #407: JSON API HTTP error codes
# Intention 

Our intention is to use HTTP status codes to transmit information when there is a reasonable expectation that intermediate server / services between the client and the API may want to create a return with that status. Any time the API service code is solely responsible for generating the HTTP response, and we cannot foresee a situation where that would change, the response should be a HTTP 200 with embedded error objects. 

This should be thought of as the responses and status code our API services return, those that may be generated inside the Astra infrastructure but not by our code, and those that may be generated by intermediate HTTP services between the client and the front door to Astra. We need an approach that covers all these situations, and tells the driver(s) how to handle things.

# Changes 

Situations where the response [status code](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes) would not be a 200 with an embedded error object: 

* 401 Unauthorized - when the request has not been successfully authenticated, that is the client is not able to connect to the API. While this may currently be generated from the JSON API code, we can imagine a situation where Proxy / Gateway / Frontdoor of the whole astra platform may check that incoming requests are from customers we want to allow in. e.g. we want to blacklist a customer / tenant at the proxy level and not let their requests be processed because they have not paid their bill, or their services have been compromised and they want to stop all access to data. 
 * 404 Not Found - when the prefix path to the JSON API endpoint is incorrect, errors to do with missing namespaces or collections are application errors and returned using the 200 code. e.g. the user tries to access the API at /json/v11 rather than /json/v1
* 405 Method Not Allowed - Attempt to use the wrong verb on an otherwise valid URL, e.g. trying to run a GET 

* 5XX - Two situations: 1) Any response returned by intermediate servers between the client and our front door, such as 503 Service Unavailable. 2) An unhandled or unexpected server side error in the Astra platform, either in the infrastructure or the API code, could result in a 500 Internal Server Error.

maheshrajamani:
**What this PR does**:
Http status code change

Fixes #407 

**Checklist**
- [x] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/api/model/command/CommandResult.java
```diff
@@ -168,7 +168,10 @@ public List<JsonNode> getResponseDocuments() {
             implementation = String.class)
       })
   public record Error(
-      String message, @JsonAnyGetter @Schema(hidden = true) Map<String, Object> fields) {
+      String message,
+      @JsonAnyGetter @Schema(hidden = true) Map<String, Object> fields,
+      // Http status code to be used in the response, defaulted to 200
+      @JsonIgnore int errorCode) {
```
ivansenic, line 174: I would rather accept something that is not int, but HttpStatus code object from JAX-RS.. Now I can set status to `-522` and would fail of course at some point later on.. Since we are anyway ignoring the field, I guess that's better option imo..

src/main/java/io/stargate/sgv2/jsonapi/api/v1/response/MapToRestResult.java
```diff
@@ -0,0 +1,28 @@
...
+public class MapToRestResult {
+  /**
+   * Maps CommandResult to RestResponse. Except for few selective errors, all errors are mapped to
+   * http status 200. In case of 401, 500, 502 and 504 response is sent with appropriate status
+   * code.
+   *
+   * @param commandResult
+   * @return
+   */
+  public static RestResponse map(CommandResult commandResult) {
```
ivansenic, line 16: this can also be  a method  `CommandResult`

src/main/java/io/stargate/sgv2/jsonapi/service/processor/CommandProcessor.java
```diff
@@ -67,7 +67,6 @@ public <T extends Command> Uni<CommandResult> processCommand(
               if (t instanceof JsonApiException jsonApiException) {
                 return jsonApiException;
               }
-
```
ivansenic, line 70: nit: no changes, remove from commit

src/main/java/io/stargate/sgv2/jsonapi/api/exception/WebApplicationExceptionMapper.java
```diff
@@ -11,6 +13,13 @@ public class WebApplicationExceptionMapper {
 
   @ServerExceptionMapper
   public RestResponse<CommandResult> genericExceptionMapper(WebApplicationException e) {
+    // Return 405 for method not allowed and 404 for not found
+    if (e instanceof NotAllowedException) {
+      return RestResponse.status(RestResponse.Status.METHOD_NOT_ALLOWED);
+    }
+    if (e instanceof NotFoundException) {
+      return RestResponse.status(RestResponse.Status.NOT_FOUND);
+    }
```
ivansenic, line 22: but here we are not adhering to the spec in case of the error? can we still create our body and report given statuses? seems strange now the mixup:

* some errors return `200` and a body from spec
* some errors return `4xx` and completely different body

src/main/java/io/stargate/sgv2/jsonapi/exception/JsonApiException.java
```diff
@@ -59,7 +59,7 @@ public CommandResult get() {
   public CommandResult.Error getCommandResultError(String message) {
     Map<String, Object> fields =
         Map.of("errorCode", errorCode.name(), "exceptionClass", this.getClass().getSimpleName());
-    return new CommandResult.Error(message, fields);
+    return new CommandResult.Error(message, fields, 200);
```
ivansenic, line 62: fixed? so all `JsonApiExceptions` would return 200 always? would add a method into `ErrorCode.getHttpStatusCode()` which can return `200` by default.. then we can easily change for some error code..

src/test/java/io/stargate/sgv2/jsonapi/api/v1/response/MapToRestResponseTest.java
```diff
@@ -0,0 +1,79 @@
...
+import io.stargate.sgv2.jsonapi.api.model.command.CommandResult;
+import io.stargate.sgv2.jsonapi.api.model.command.CommandStatus;
+import java.util.List;
+import java.util.Map;
+import org.jboss.resteasy.reactive.RestResponse;
+import org.junit.jupiter.api.Nested;
+import org.junit.jupiter.api.Test;
+
+@QuarkusTest
+@TestProfile(NoGlobalResourcesTestProfile.Impl.class)
```
ivansenic, line 17: no need for a quarkus test, this is a simple unit test

src/test/java/io/stargate/sgv2/jsonapi/api/v1/HttpStatusCodeIntegrationTest.java
```diff
@@ -0,0 +1,118 @@
...
+import io.restassured.http.ContentType;
+import io.stargate.sgv2.api.common.config.constants.HttpConstants;
+import io.stargate.sgv2.jsonapi.testresource.DseTestResource;
+import org.junit.jupiter.api.Test;
+
+@QuarkusIntegrationTest
+@QuarkusTestResource(DseTestResource.class)
+public class HttpStatusCodeIntegrationTest extends AbstractCollectionIntegrationTestBase {
+  @Test
+  public void unauthenticated() {
```
ivansenic, line 22: Don't we have a set of those in integration tests for each resource? I think the idea was, general errors, validations, etc, are validated in the:

* `GeneralResourceIntegrationTest`
* `NamespaceResourceIntegrationTest`
* `CollectionResourceIntegrationTest`

for each endpoint we have..

ivansenic (changes requested):

src/main/java/io/stargate/sgv2/jsonapi/exception/JsonApiException.java
maheshrajamani: Yes any error other than specific one in the ticket will return error message with http status code 200 as per #407

src/main/java/io/stargate/sgv2/jsonapi/api/exception/WebApplicationExceptionMapper.java
maheshrajamani: This is what is decided, refer #407

src/main/java/io/stargate/sgv2/jsonapi/service/processor/CommandProcessor.java
maheshrajamani: Will do

src/main/java/io/stargate/sgv2/jsonapi/api/v1/response/MapToRestResult.java
maheshrajamani: Will do.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/CommandResult.java
maheshrajamani: @ivansenic Make sense. Not sure if we should be using JAX-RS enum everywhere in the code. Better to create separate enum for the error code and later map it when result is generated?

ivansenic: I would use the JAX-RS directlym this is anyway API level code.. Having our own is really an overhead imo..

src/main/java/io/stargate/sgv2/jsonapi/exception/JsonApiException.java
ivansenic: OK.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/CommandResult.java
maheshrajamani: Thought of not using JAX-RS enum in classes like ThrowableToErrorMapper.

src/main/java/io/stargate/sgv2/jsonapi/api/exception/WebApplicationExceptionMapper.java
ivansenic: OK, seems now there is no body at all.. Just status and empty response, I though this is using generic web application error body response..

ivansenic (approved):

PR #420, opened 2023-05-09
closes #414: adapts java args for docker-comspose

### LINKED ISSUE #414: Docker-comspse memory settings cause coordinator OOM
Memory ops for the coordinator in the docker compose scrips are allowing the max heap size to be same as the max memory given to the container, causing the OOM kill. Off heap memory is not taken into account.. 

https://github.com/stargate/jsonapi/blob/6473f8723d1e80da34a0a95b68cb8da991e8e2b1/docker-compose/docker-compose.yml#L75-L77

We should go with Quarkus memory settings here, and set these env vars:

```
JAVA_MAX_MEM_RATIO: '75'
JAVA_INITIAL_MEM_RATIO: '50'
```

ivansenic:
**What this PR does**:
Fixes memory java args in the docker-compose files:

* remove low new heap size for the dse
* set reasonable heap maximum sizes for all JVMs on a 2G container mem limit

After updates:

Fixes #414

### DIFF
--- a/docker-compose/docker-compose-dev-mode.yml
+++ b/docker-compose/docker-compose-dev-mode.yml
@@ -10,7 +10,7 @@ services:
       - "8081:8081"
     mem_limit: 2G
     environment:
-      - JAVA_OPTS="-Xmx2G"
+      - JAVA_OPTS="-Xmx1536M"
       - CLUSTER_NAME=dse-${DSETAG}-cluster
       - CLUSTER_VERSION=6.8
       - DSE=1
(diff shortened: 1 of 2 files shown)


maheshrajamani:
@ivansenic Why the max memory needs to be reduced to 1.5G?

ivansenic:
> @ivansenic Why the max memory needs to be reduced to 1.5G?

@maheshrajamani container limits are at 2G, the heap needs to be smaller so there is enough space for non-heap memory regions..

maheshrajamani (approved):

PR #423, opened 2023-05-11
add fallout smoke test

ivansenic:
**What this PR does**:
Adds smoke test for Fallout, including the target nosqlbench file

**Checklist**
- [ ] add README
- [x] resolve open questions

### DIFF
--- a/fallout/smoke-test-fallout.yaml
+++ b/fallout/smoke-test-fallout.yaml
@@ -12,10 +12,15 @@ client_nb_version: 5.17.3-release
 # keyspace setup
 keyspace: jsonapi_smoke_test
 
-# documents 10K to enable sorting
-docs_count: 10000
+# 100K documents and 25 groups, so that sort and count target ~4K docs
+# adapt if needed, but keep in mind that sort supports max 10K docs per operation
+docs_count: 100000
+sort_groups: 25
 
-# load setup
+# load setup:
+# - total of 1M ops split between 10 operations equally
+# - each operation limited to 25ops/s, totals to 250ops/s
+# - each operation will be executed by 3 clients, totals to 30 clients
 op_cycles: 100000
 op_rate: 25
 op_connections: 3
@@ -26,11 +31,11 @@ ops:
   - op: find-one-by-city
   - op: find-multi-by-married-name-exists
   - op: find-multi-by-married-name-exists-with-projection
-  - op: find-multi-find-by-gender-sort-by-name
-  - op: find-by-name-newest-and-update-city
+  - op: find-multi-find-by-group-sort-by-name
+  - op: find-one-by-name-newest-and-update-city
   - op: delete-one-id
   - op: find-one-id-and-replace-upsert
-  - op: count-all
+  - op: update-many-by-country
   - op: count-by-gender
 
 ---
@@ -121,6 +126,7 @@ workload:
             - path_prefix=/api/json
             - connections={{op_connections}}
             - docscount={{docs_count}}
+            - sort_groups={{sort_groups}}
             - rate={{op_rate}}
       {{/ops}}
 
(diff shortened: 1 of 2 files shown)


fallout/smoke-test-fallout.yaml
```diff
@@ -0,0 +1,133 @@
...
+    # set schema phase, ensures collection table is created from scratch
+    - nb_set_schema:
+        module: nosqlbench
+        properties:
+          client.size: 1
+          alias: schema
+          service_type: stargate
+          args:
+            - driver=http
+            - workload=https://raw.githubusercontent.com/stargate/jsonapi/ise-fallout/fallout/smoke-test-nosqlbench.yaml
```
ivansenic, line 74: Should the file or branch be parametrized? In any way, it should point to `main` branch by default..

fallout/smoke-test-fallout.yaml
```diff
@@ -0,0 +1,133 @@
+# region setup
```
ivansenic, line 1: should we separate tests info folders? `fallout/smoke-test/..`

fallout/smoke-test-fallout.yaml
```diff
@@ -0,0 +1,133 @@
...
+            - threads=30
+            - errors=stop
+            - namespace={{keyspace}}
+            - protocol=https
+            - jsonapi_port=443
+            - path_prefix=/api/json
+            - connections=30
+            - docscount={{docs_count}}
+
+    -
```
ivansenic, line 104: should I add support for sequential execution?

fallout/smoke-test-fallout.yaml
```diff
@@ -0,0 +1,133 @@
...
+  - op: find-one-id
+  - op: find-one-by-city
+  - op: find-multi-by-married-name-exists
+  - op: find-multi-by-married-name-exists-with-projection
+  - op: find-multi-find-by-gender-sort-by-name
+  - op: find-by-name-newest-and-update-city
+  - op: delete-one-id
+  - op: find-one-id-and-replace-upsert
+  - op: count-all
+  - op: count-by-gender
```
ivansenic, line 34: @vkarpov15 We agreed to ask you about a more realistic scenario structure.. Currently we run these in parallel, and each weights the same, so each op contributes with 10%.. Also it would be great if you have any additions to this, some operation we should include, etc?

vkarpov15: @ivansenic it might be helpful to add insert operations, and maybe some more update operations. In particular, an updateMany. However, Mongoose apps currently tend to be read heavy, so this isn't too far off.

I think we should take a look at the results of this scenario, because this scenario covers a good range of features. If there are performance issues, it's worth investigating.

In the future, we can add scenarios that are more in line with basic apps, like simulating what happens with our sample apps under load.

tatu-at-datastax: I think @maheshrajamani found one thing that may help improve performance (count(1) instead of count(column)), we'll see how much that helps (fix has been merged).

ivansenic: * Added update many instead of the count all
* Adapted count and sort ops so that we can control how many docs should be targeted in those ops

fallout/smoke-test-fallout.yaml
ivansenic: Will not do, following same as in the `/nosqlbench/` directory..

fallout/smoke-test-fallout.yaml
ivansenic: Skipped.

ivansenic:
Test run [493fc5b9](http://fallout.aws.dsinternal.org/tests/ui/ivan.senic@datastax.com/jsonapi_smoke_test/493fc5b9-bd89-4e4d-a60b-354b76c368c6/artifacts) with these two files.

jeffreyscarpenter (approved):
Looks great. We can always make improvements later if needed.

PR #427, opened 2023-05-12
Count operation - fix to use count(1) instead of count("key") 

### LINKED ISSUE #424: Use count(1) in count operation query
Change the query to use count(1) instead of count(key) for countDocuments command. This will avoid checking for not null check of the key field.

Before query
`select count(key) from table`
After query
`select count(1) from table`

maheshrajamani:
**What this PR does**:
Count operation - fix to use count(1) instead of count("key") 
Fixes #424 

**Checklist**
- [x] Changes manually tested


tatu-at-datastax (approved):
LGTM

PR #432, opened 2023-05-17
Add Number size limits (max 50 characters), enforcement

### LINKED ISSUE #172: Decide on the size of numbers we accept
Split from #135  

For the [Document Value Limits](https://github.com/stargate/jsonapi/blob/main/docs/jsonapi-spec.textile#document-value-limits) for a number we ended with: 

> TODO: define max num that is well handled by BigDecimal

Need to work out what limits we may have, because JSON and BigDecimal both support unlimited size  but we need a limit to protect the server from bad workloads and know what we need to test against.

With reference to the [BigDecimal](https://docs.oracle.com/javase/8/docs/api/java/math/BigDecimal.html) we could do something like say the integer component can be no bigger than 128 bites . By reading [unscaledValue](https://docs.oracle.com/javase/8/docs/api/java/math/BigDecimal.html#unscaledValue--) and then getting the number of bits from the [BigInteger](https://docs.oracle.com/javase/8/docs/api/java/math/BigInteger.html)

@tatu-at-datastax  did you have some opinions ?

tatu-at-datastax:
**What this PR does**:

Will impose limit to maximum length of JSON Number values we accept in JSON Documents (based on textual length): initial value configured to 50 characters.
In theory we could limit to lower (some documentation suggest maximum number of significant digits would be 21) but since the goal here is to reduce likelihood of DoS style resource consumption, we only need to prevent thousands of digits -- performance testing done by Jackson project suggested change not below 1000 digits.

Fixes #172

**Checklist**
- [ ] Changes manually tested


tatu-at-datastax:
Hmmmh. Looks like I found a Record-deserialization regression b/w Jackson 2.14 and 2.15. Yikes. Need to troubleshoot....

ivansenic:
@tatu-at-datastax Guess not needed anymore with #439 ?

tatu-at-datastax:
@ivansenic We still need 2.15.x (likely 2.15.2) to get Number length limits. But I need to get 2.15.2 released first, to resolve issue uncovered by my "interesting" usage of setter method on `record`s (fix is in 2.15 branch).

tatu-at-datastax:
Now that Jackson 2.15.2 is out, with relevant fix, can unblock this. Will first get ITs to run to make sure everything still works, then add tighter number length limits and matching tests.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/Command.java
```diff
@@ -38,10 +38,7 @@
  *
  * <p>Each command should validate itself using the <i>javax.validation</i> framework.
  */
-@JsonTypeInfo(
-    use = JsonTypeInfo.Id.NAME,
-    include = JsonTypeInfo.As.WRAPPER_OBJECT,
-    property = "commandName")
```
tatu-at-datastax, line 44: "property" only used with `As.PROPERTY`, not with WRAPPER_OBJECT.

src/main/java/io/stargate/sgv2/jsonapi/config/DocumentLimitsConfig.java
```diff
@@ -1,10 +1,15 @@
...
 import io.smallrye.config.ConfigMapping;
 import io.smallrye.config.WithDefault;
 import jakarta.validation.constraints.Positive;
 
-/** Configuration Object that defines limits on Documents managed by JSON API. */
+/**
+ * Configuration Object that defines limits on Documents managed by JSON API. Needed early for
+ * providers so has to be declared as {@link StaticInitSafe}.
+ */
+@StaticInitSafe
```
tatu-at-datastax, line 12: Injection into ObjectMapper provider failed without adding this annotation.

pom.xml
```diff
@@ -10,6 +10,8 @@
   <version>1.0.0-ALPHA-10-SNAPSHOT</version>
   <properties>
     <stargate.version>${project.parent.version}</stargate.version>
+    <!-- 17-May-2023, tatu: [json-api#172] Need Jackson 2.15.x -->
+    <jackson.version>2.15.2</jackson.version>
```
tatu-at-datastax, line 14: Quarkus 3.1.x will likely upgrade, but until we use that, override since we need this version for length limit validation.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/InsertIntegrationTest.java
```diff
@@ -100,16 +100,16 @@ public void insertDocument() {
     public void insertDocumentWithDateValue() {
       String json =
           """
-        {
```
tatu-at-datastax, line 103: Not done on purpose but `./mvnw fmt:format` sometimes/somehow seems to decide it wants to reformat. :-(

src/main/java/io/stargate/sgv2/jsonapi/config/DocumentLimitsConfig.java
ivansenic: Are you sure this is not preventing the values to be changed during the runtime?

pom.xml
ivansenic: Just checked with 3.1.0, it's still on `2.15.0`..

ivansenic (approved):
Looks fine, just one thing to check imo..

tatu-at-datastax:
@ivansenic 

> Are you sure this is not preventing the values to be changed during the runtime?

I am not sure it would be possible to change these during runtime (to have effect dynamically), as they are typically injected into other handlers. I guess I haven't tried that, and in kubernetes environment I was assuming change to pod would need to restart it anyway (to pass system properties or env variables).

But I'll read bit more about this annotation to understand it better.

tatu-at-datastax:
Since injection of `DocumentLimitsConfig` otherwise fails, preventing injection of `ObjectMapper` which in turns in needed by unit tests, right now annotation `@StaticInitSafe` is required, so will merge.

PR #436, opened 2023-05-23
Refactor id-handling part of Shredding

### LINKED ISSUE #425: Rewrite handling of `_id` to occur before document processed by `Shredder`
(follow-up to #421)

As suggested in #421 code review, handling of document id by `Shredder` is bit convoluted wrt initial addition and/or reodering before doing actual shredding. Shredding should not have to modify the passed document: this could be achieved by extracting pre-processing logic to specialized class.
This would simplify `Shredder` a bit.

tatu-at-datastax:
**What this PR does**:

Refactors handling of Document Id (_id) in `Shredder` to try to isolate part that needs to modify Doc (or create new one) from shredding that is read-only operation.
Ideally caller could call modify part before shredding but this change does not quite get there yet; it just prepares for further changes.

Fixes #425

**Checklist**
- [ ] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/service/shredding/Shredder.java
```diff
@@ -93,24 +78,53 @@ public WritableShreddedDocument shred(JsonNode doc, UUID txId) {
     validateDocument(documentLimits, docWithId, docJson);
 
     final WritableShreddedDocument.Builder b =
-        WritableShreddedDocument.builder(new DocValueHasher(), docId, txId, docJson, docWithId);
```
tatu-at-datastax, line 96: No need to construct it here, pass, constructed by builder itself.

src/main/java/io/stargate/sgv2/jsonapi/service/shredding/Shredder.java
```diff
@@ -93,24 +78,52 @@ public WritableShreddedDocument shred(JsonNode doc, UUID txId) {
...
+    // First: see if we have Object Id present or not
+    JsonNode idNode = doc.get(DocumentConstants.Fields.DOC_ID);
+    ObjectNode docWithoutId;
+
+    if (idNode == null) { // No id: generate, doc fine to use as source
+      docWithoutId = (ObjectNode) doc;
+      idNode = generateDocumentId();
+    } else { // Has id, use as-is, but create a copy of Doc without id
+      docWithoutId = objectMapper.createObjectNode();
+      docWithoutId.setAll((ObjectNode) doc);
```
maheshrajamani, line 106: Code is doing setAll twice in most case, can the docWithoutId outside if condition and add everything back to doc object itself? By that way the reference to document in the Operation class still holds good. I made a fix before to set the document to WritableShreddedDocument that's not needed.

maheshrajamani (approved):
Made a suggestion, to maintain the document reference.

src/main/java/io/stargate/sgv2/jsonapi/service/shredding/Shredder.java
tatu-at-datastax: Ok, so, I think I can remove one of `setAll()` cases but will not be able to avoid construction of a new `ObjectNode` in all cases unfortunately.

The problem is that due to the way `LinkedHashMap` ordering works, it is not possible (as far as I know) to reorder existing entries (when instance uses insertion-order which is the case with Jackson): new entries will be appended at the end. So it is not possible to move or insert document id as the first field.

But I think I should avoid one of copies since it is fine to just add all fields, including object id, as long as object id already exists in target `ObjectNode`. So there is no need for intermediate `docWithoutId` after all, I think.
I will make changes and ask you to re-review.

tatu-at-datastax:
@maheshrajamani Ok change passes tests (at least non-native image) so I think this is ready for re-review.

Unfortunately, as per my comments, I don't think we can avoid copying `ObjectNode` given without changing API.
I guess if we changed `Shredder` to require 2 calls -- or use another object/class -- we could split things into 2 steps:

1. Get "normalized" `ObjectNode`, basically what new `normalizeDocumentId()` does. Does not need `txId`.
2. Call "shred()" method with normalized document ObjectNode which will not be changed in any way.

The reason I didn't yet do this is that APIs that require sequence of calls seem like bad design.
But maybe adding new class would help.

Either way I was thinking of merging this in and then trying to think of next steps if we want to eliminate calling access method you added.

PR #437, opened 2023-05-23
Add command name to count metrics

### LINKED ISSUE #411: Add command name and errors to the server-side HTTP metrics
Since everything goes through `POST` and targets a single endpoint our server-side metrics are not showing any information about the commands begin executed. For example:

```
# HELP http_server_requests_seconds  
# TYPE http_server_requests_seconds summary
http_server_requests_seconds_count{method="POST",module="sgv2-jsonapi",outcome="SUCCESS",status="200",uri="/v1/{namespace}/{collection}",} 40000.0
http_server_requests_seconds_sum{method="POST",module="sgv2-jsonapi",outcome="SUCCESS",status="200",uri="/v1/{namespace}/{collection}",} 620.394043765
http_server_requests_seconds_count{method="POST",module="sgv2-jsonapi",outcome="SUCCESS",status="200",uri="/v1",} 2.0
http_server_requests_seconds_sum{method="POST",module="sgv2-jsonapi",outcome="SUCCESS",status="200",uri="/v1",} 1.928188204
http_server_requests_seconds_count{method="POST",module="sgv2-jsonapi",outcome="SUCCESS",status="200",uri="/v1/{namespace}",} 4.0
http_server_requests_seconds_sum{method="POST",module="sgv2-jsonapi",outcome="SUCCESS",status="200",uri="/v1/{namespace}",} 14.483613413
http_server_requests_seconds_count{method="POST",module="sgv2-jsonapi",outcome="CLIENT_ERROR",status="RESET",uri="/v1/{namespace}/{collection}",} 40000.0
http_server_requests_seconds_sum{method="POST",module="sgv2-jsonapi",outcome="CLIENT_ERROR",status="RESET",uri="/v1/{namespace}/{collection}",} 617.666413847
http_server_requests_seconds_count{method="POST",module="sgv2-jsonapi",outcome="CLIENT_ERROR",status="RESET",uri="/v1/{namespace}",} 4.0
http_server_requests_seconds_sum{method="POST",module="sgv2-jsonapi",outcome="CLIENT_ERROR",status="RESET",uri="/v1/{namespace}",} 12.24249949
http_server_requests_seconds_count{method="GET",module="sgv2-jsonapi",outcome="SUCCESS",status="200",uri="/favicon.ico",} 1.0
http_server_requests_seconds_sum{method="GET",module="sgv2-jsonapi",outcome="SUCCESS",status="200",uri="/favicon.ico",} 0.012527204
http_server_requests_seconds_count{method="POST",module="sgv2-jsonapi",outcome="CLIENT_ERROR",status="RESET",uri="/v1",} 2.0
http_server_requests_seconds_sum{method="POST",module="sgv2-jsonapi",outcome="CLIENT_ERROR",status="RESET",uri="/v1",} 1.926456839
```

Meaning I have no idea what those client errors are and where are they coming from.. We should add a tag that clearly specifies the command being executed.

Also we should think about error responses and how can we count those. As everything goes into `200`, there no chance to know if error occurred or not, and even worse if this was client side or server side error.

maheshrajamani:
**What this PR does**:
Add command name to count metrics

Fixes #411 

**Checklist**
- [x] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/api/security/ErrorChallengeSender.java
```diff
@@ -28,18 +34,38 @@ public class ErrorChallengeSender implements ChallengeSender {
   /** Result is always constant */
   private final CommandResult commandResult;
 
+  /** The {@link MeterRegistry} to report to. */
+  private final MeterRegistry meterRegistry;
+
+  /** The tag for error being true, created only once. */
+  private final Tag errorTag = Tag.of("error", "true");
```
tatu-at-datastax, line 41: These can probably be `static`? (unlike injected things like meterRegistry)

src/main/java/io/stargate/sgv2/jsonapi/api/security/ErrorChallengeSender.java
```diff
@@ -59,8 +85,13 @@ public Uni<Boolean> apply(RoutingContext context, ChallengeData challengeData) {
...
 
-      // always set status to 200
-      context.response().setStatusCode(200);
+      // Return the status code from the challenge data
+      context.response().setStatusCode(challengeData.status);
+
+      // Add metrics
+      Tags tags = Tags.of(apiTag, commandTag, statusCodeTag, errorTag);
+      // record
+      meterRegistry.counter("http_server_requests_custom_seconds_count", tags).increment();
```
tatu-at-datastax, line 94: Not sure if there's a good place to define these constants? If not, inlining is fine.

src/main/java/io/stargate/sgv2/jsonapi/api/v1/metrics/RequestMetricsFilter.java
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  /**
+   * Filter that this bean produces.
+   *
+   * @param requestContext {@link ContainerRequestContext}
+   * @param responseContext {@link ContainerResponseContext}
+   */
+  @ServerResponseFilter
+  public void record(
```
tatu-at-datastax, line 54: Would it make sense to use different name even if `record` is not technically a reserved word in Java?

src/main/java/io/stargate/sgv2/jsonapi/api/v1/metrics/RequestMetricsFilter.java
```diff
@@ -0,0 +1,99 @@
...
+   *
+   * @param inputStream {@link InputStream}
+   * @return the command name
+   */
+  private String getCommandName(InputStream inputStream) {
+    try {
+      // reset the stream to fetch from beginning of stream
+      inputStream.reset();
+      // get body from requestContext
+      final Iterator<String> fieldIterator = objectMapper.readTree(inputStream).fieldNames();
```
tatu-at-datastax, line 79: This is rather expensive, parsing the whole JSON contents. Is there no other way? It also requires deeper knowledge of payload than should be needed in a filter.

src/main/java/io/stargate/sgv2/jsonapi/api/v1/metrics/RequestMetricsFilter.java
```diff
@@ -0,0 +1,99 @@
...
+  /**
+   * Filter that this bean produces.
+   *
+   * @param requestContext {@link ContainerRequestContext}
+   * @param responseContext {@link ContainerResponseContext}
+   */
+  @ServerResponseFilter
+  public void record(
+      ContainerRequestContext requestContext, ContainerResponseContext responseContext) {
+    String commandName = getCommandName(requestContext.getEntityStream());
```
tatu-at-datastax, line 56: Is it guaranteed this stream can be `reset()` in all cases?

tatu-at-datastax (commented):
I have some concerns about having to decode JSON for every command additional time, but that is probably necessary due to design of API. So I guess I am ok with that. There are some additional suggestions (wrt constant made `static`), but more generally I think I'd like someone else (Ivan) to have a look.

src/main/java/io/stargate/sgv2/jsonapi/api/v1/metrics/RequestMetricsFilter.java
tatu-at-datastax: Ok, we can actually avoid most of decoding, although do need to stream through content more than once.

Since we only need to see:

```
{ "command-name"
```

it is possible to create `JsonParser` using `ObjectMapper` and use `nextToken()` to look only at first 2 tokens: they must be:

1. `JsonToken.START_OBJECT`
2. `JsonToken.FIELD_NAME`

and after (2), can call `jsonParser.currentName()` to get command name we need.
After that `JsonParser` should be `close()`d (frees/recycles buffers).

This would remove most of processing overhead, as long as request payload is buffered.

ivansenic (changes requested):
I would like to propose different solution, which has pros and cons, but definitely avoids doing that gymnastics in the filter, which imo can have negative performance for the requests..

### Proposal

How about we don't count record on the HTTP level, but rather record the performance of the `CommandProcessor`. There we could easily use `MeterRegistry` directly and measure time for the command execution. We could include tags that we want and that we can take from the exceptions, context, tenant info provider, etc.

The implementation would be way easier and we would read information directly from the model objects, thus we would have way better performance.

The other libraries often have such metrics as well, and do report on in-flight messages, error rates, etc, so it wouldn't be anything unusual. For example, I know Axon has something similar for their [command gateway](https://grafana.com/grafana/dashboards/12963-axon-spring-boot-applications-statistics/).

The implementation can be very simple:

```
  public <T extends Command> Uni<CommandResult> processCommand(
      CommandContext commandContext, T command) {
     Timer timer = meterRegistry.timer(...);

      return commandResolverService
        .resolverForCommand(command)

        // resolver can be null, not handled in CommandResolverService for now
        .flatMap(
            resolver -> {
              // if we have resolver, resolve operation and execute
              Operation operation = resolver.resolveCommand(commandContext, command);
              return operation.execute(queryExecutor);
            })

       .onItemOrFailure((item, ex) -> {
           // capture metrics & handle result or exception 
       });
  }
```

#### Pros over current solution
* can measure time and not only count
* does not manipulation of the response input stream
* can include tenant if available (can the current solution as well)
* can include detailed information on the error, for example error code or error class
* better performance

#### Cons over current solution
* no information on `4xx` response as they never reach the command gateway

I would like to sync about this and thus putting *Request changes* until we agree on the approach..

src/main/java/io/stargate/sgv2/jsonapi/api/security/ErrorChallengeSender.java
maheshrajamani: Will add it. The class is marked application scope so only one instance is maintained.

src/main/java/io/stargate/sgv2/jsonapi/api/security/ErrorChallengeSender.java
maheshrajamani: Will add it to the Constants

src/main/java/io/stargate/sgv2/jsonapi/api/v1/metrics/RequestMetricsFilter.java
maheshrajamani: Will change it to recordMetric

src/main/java/io/stargate/sgv2/jsonapi/api/v1/metrics/RequestMetricsFilter.java
maheshrajamani: Will make the change

src/main/java/io/stargate/sgv2/jsonapi/api/v1/metrics/RequestMetricsFilter.java
maheshrajamani: This is done post command is run, so after reset is not needed. I will add a null check, if Stream object present reset won't fail.

tatu-at-datastax:
@ivansenic I like this idea: access from model does give more information.

Wrt 4xx responses; perhaps there could be different mechanism to deal with those? And if we only processed stream content (... if we must) for 4xx, it'd be less of performance concern as well. Or, alternatively, just have non-command-specific 4xx counts: not optimal but at least could alert.

ivansenic:
The 4xx would still be captured in the http metrics out of the box, so we would have info about those as well, just without the command name. @maheshrajamani and I agreed to go and record in the command processor and we agreed to pick up error codes, exception classes and even the info if error is caused by user or not.

tatu-at-datastax:
> The 4xx would still be captured in the http metrics out of the box, so we would have info about those as well, just without the command name. @maheshrajamani and I agreed to go and record in the command processor and we agreed to pick up error codes, exception classes and even the info if error is caused by user or not.

Excellent! Sounds like a plan.

src/main/java/io/stargate/sgv2/jsonapi/service/processor/CommandProcessor.java
```diff
@@ -46,36 +60,92 @@ public CommandProcessor(
...
-            resolver -> {
-              // if we have resolver, resolve operation and execute
-              Operation operation = resolver.resolveCommand(commandContext, command);
-              return operation.execute(queryExecutor);
-            })
+    final Tag commandTag = Tag.of(MetricsConstants.COMMAND, command.getClass().getSimpleName());
+    final String tenant =
+        stargateRequestInfo != null && stargateRequestInfo.getTenantId().isPresent()
+            ? stargateRequestInfo.getTenantId().get()
+            : MetricsConstants.UNKNOWN_VALUE;
```
ivansenic, line 67: `stargateRequestInfo` is never null here.. just `stargateRequestInfo.getTenantId().orElse(MetricsConstants.UNKNOWN_VALUE)`

src/main/java/io/stargate/sgv2/jsonapi/service/processor/CommandProcessor.java
```diff
@@ -46,36 +60,92 @@ public CommandProcessor(
...
+    final Tag tenantTag = Tag.of(MetricsConstants.TENANT, tenant);
+    final long start = System.currentTimeMillis();
```
ivansenic, line 69: You shoud not measure time yourself, but rather use `Timer.Sample sample = Timer.start(registry);`

See: https://github.com/micrometer-metrics/micrometer-docs/blob/main/src/docs/concepts/timers.adoc#storing-start-state-in-timersample

src/main/java/io/stargate/sgv2/jsonapi/service/processor/CommandProcessor.java
```diff
@@ -46,36 +60,92 @@ public CommandProcessor(
...
+    return Uni.createFrom()
+        .item(start)
+        .onItem()
+        .transformToUni(
+            startTime -> {
```
ivansenic, line 74: why this?

src/main/java/io/stargate/sgv2/jsonapi/service/processor/CommandProcessor.java
```diff
@@ -46,36 +60,92 @@ public CommandProcessor(
...
-        .ifNotNull()
-        .transform(Supplier::get);
+                        } else {
+                          // return the tags and command result
+                          return item;
+                        }
+                      })
+                  .onItem()
+                  .ifNotNull()
+                  .transform(Supplier::get)
```
ivansenic, line 104: I am not sure why you changed all the code above, when you anyway inspect the results from the final result?

src/main/java/io/stargate/sgv2/jsonapi/service/processor/CommandProcessor.java
```diff
@@ -46,36 +60,92 @@ public CommandProcessor(
...
+                                          MetricsConstants.ERROR_CODE,
+                                          MetricsConstants.UNKNOWN_VALUE);
+                          errorCodeTag = Tag.of(MetricsConstants.ERROR_CODE, errorCode);
+                        }
+                        Tags tags =
+                            Tags.of(commandTag, tenantTag, errorTag, errorClassTag, errorCodeTag);
+                        // add metrics
+                        meterRegistry
+                            .counter(MetricsConstants.COUNT_METRICS_NAME, tags)
+                            .increment();
```
ivansenic, line 140: not needed, remove

src/main/java/io/stargate/sgv2/jsonapi/service/processor/CommandProcessor.java
```diff
@@ -46,36 +60,92 @@ public CommandProcessor(
...
+                        meterRegistry
+                            .timer(MetricsConstants.TIMER_METRICS_NAME, tags)
+                            .record(
+                                System.currentTimeMillis() - start,
+                                java.util.concurrent.TimeUnit.MILLISECONDS);
+                        // return the command result
+                        return result;
+                      });
+            });
```
ivansenic, line 149: First of all I still think metrics could be added in the `MeteredCommandProcessor`, which wraps the current one and does the metrics only.. But even if you don't want to do this, the code should be simply something like:

```
public <T extends Command> Uni<CommandResult> processCommand(
      CommandContext commandContext, T command) {

Timer.Sample sample = Timer.start(registry);

// same code as now
return ...

         // if we have a non-null item
        // call supplier get to map to the command result
        .onItem()
        .ifNotNull()
        .transform(Supplier::get)

        .onItem()
        .call(commandResult -> {
            Tags tags = resolveTags(command, commandResult);
            sample.stop(registry.timer("my.timer", tags));
        });
```

Resolving tags should be extracted to a method for clarity and easier reading.

ivansenic: Again having `MeteredCommandProcessor` seems like a better solution, as you can wrap the current one and thus extract only metric catching to this class.. This also enables you to easily tests everything cause you can always mock the delegating CommandProcessor and thus flush different results..

```
public <T extends Command> Uni<CommandResult> processCommand(
      CommandContext commandContext, T command) {

Timer.Sample sample = Timer.start(registry);

return delegate.processCommand(commandContext, command)

        .onItem()
        .call(commandResult -> {
            Tags tags = resolveTags(command, commandResult);
            sample.stop(registry.timer("my.timer", tags));
        });
```

src/test/java/io/stargate/sgv2/jsonapi/api/v1/CountIntegrationTest.java
```diff
@@ -708,4 +712,13 @@ public void byBooleanColumn() {
           .body("errors", is(nullValue()));
     }
   }
+
+  @Nested
+  @Order(2)
+  class Metrics {
+    @Test
+    public void checkMetrics() {
+      CountIntegrationTest.super.checkMetrics("CountDocumentsCommands");
```
ivansenic, line 721: This is fine, but we need unit tests that test the error cases, tenant exist yes/no, etc..

ivansenic (changes requested):
I think I provided a lot of good suggestions, please have a look..

src/test/java/io/stargate/sgv2/jsonapi/api/v1/CountIntegrationTest.java
maheshrajamani: Added MeteredCommandProcessorTest.java for unit testing

src/main/java/io/stargate/sgv2/jsonapi/api/v1/CollectionResource.java
```diff
@@ -51,11 +51,11 @@ public class CollectionResource {
 
   public static final String BASE_PATH = "/v1/{namespace}/{collection}";
 
-  private final CommandProcessor commandProcessor;
+  private final MeteredCommandProcessor meteredCommandProcessor;
 
   @Inject
-  public CollectionResource(CommandProcessor commandProcessor) {
-    this.commandProcessor = commandProcessor;
+  public CollectionResource(MeteredCommandProcessor meteredCommandProcessor) {
```
ivansenic, line 57: There was no need to change this to metered processor, you could have just removed the `@ApplicationScoped` from the `CommandProcessor`.. Then `MeteredCommandProcessor` would be the only `CommandProcessor` in the app that could be injected..

This could in future allow us to disable processor metering using the config property and have a configuration method that deciedes what instance is used.

src/main/java/io/stargate/sgv2/jsonapi/api/v1/metrics/JsonApiMetricsConfig.java
```diff
@@ -0,0 +1,32 @@
+package io.stargate.sgv2.jsonapi.api.v1.metrics;
+
+import io.smallrye.config.ConfigMapping;
+import io.smallrye.config.WithDefault;
+import jakarta.validation.Valid;
+import jakarta.validation.constraints.NotBlank;
+import jakarta.validation.constraints.NotNull;
+
+@ConfigMapping(prefix = "stargate.metrics")
```
ivansenic, line 9: maybe just `"stargate.jsonapi.metric` and all props directly in the top-level class

src/main/java/io/stargate/sgv2/jsonapi/service/processor/MeteredCommandProcessor.java
```diff
@@ -0,0 +1,116 @@
...
+    if (null != result.errors() && !result.errors().isEmpty()) {
+      errorTag = errorTrue;
+      String errorClass =
+          (String) result.errors().get(0).fields().getOrDefault("exceptionClass", UNKNOWN_VALUE);
+      errorClassTag = Tag.of(customMetricsConfig.errorClass(), errorClass);
+      String errorCode =
+          (String) result.errors().get(0).fields().getOrDefault("errorCode", UNKNOWN_VALUE);
+      errorCodeTag = Tag.of(customMetricsConfig.errorCode(), errorCode);
+    }
+    Tags tags = Tags.of(commandTag, tenantTag, errorTag, errorClassTag, errorCodeTag);
```
ivansenic, line 113: I would only have `errorTag` set to true if it's only real server side error, this way everything is set to errors, so we can not use alerting here.. Is it possible to do it here?

src/test/java/io/stargate/sgv2/jsonapi/service/processor/MeteredCommandProcessorTest.java
```diff
@@ -0,0 +1,131 @@
...
+          .satisfies(
+              lines -> {
+                assertThat(lines.size()).isEqualTo(3);
+                lines.forEach(
+                    line -> {
+                      assertThat(line).contains("command=\"CountDocumentsCommands\"");
+                      assertThat(line).contains("tenant=\"test-tenant\"");
+                      assertThat(line).contains("error=\"false\"");
+                      assertThat(line).contains("error_code=\"NA\"");
+                      assertThat(line).contains("error_class=\"NA\"");
```
ivansenic, line 76: should this be `none` and not `NA`? because on happy path we should not set to NA, as we set when there is an error, but there's no error code?.. Not sure though..

src/test/java/io/stargate/sgv2/jsonapi/service/processor/MeteredCommandProcessorTest.java
```diff
@@ -0,0 +1,131 @@
...
+                      assertThat(line).contains("module=\"sgv2-jsonapi\"");
+                    });
+              });
+    }
+
+    @Test
+    public void errorMetrics() throws Exception {
```
ivansenic, line 83: can you add one more test with the error case, but when we don't have error code in the response?

src/test/java/io/stargate/sgv2/jsonapi/service/processor/MeteredCommandProcessorTest.java
```diff
@@ -0,0 +1,131 @@
...
+      CommandContext commandContext = new CommandContext("namespace", "collection");
+      Map<String, Object> fields = new HashMap<>();
+      fields.put("exceptionClass", "TestExceptionClass");
+      fields.put("errorCode", "TestErrorCode");
+      CommandResult.Error error = new CommandResult.Error("message", fields, Response.Status.OK);
+      CommandResult commandResult = new CommandResult(Collections.singletonList(error));
+      Mockito.when(commandProcessor.processCommand(commandContext, countCommand))
+          .thenReturn(Uni.createFrom().item(commandResult));
+      Mockito.when(stargateRequestInfo.getTenantId()).thenReturn(Optional.of("test-tenant"));
+      meteredCommandProcessor.processCommand(commandContext, countCommand).await().indefinitely();
```
ivansenic, line 104: nit: formatting json, clear separation between test setup, execution and verification/assertion would improve readability here..

src/test/java/io/stargate/sgv2/jsonapi/service/processor/MeteredCommandProcessorTest.java
```diff
@@ -0,0 +1,131 @@
...
+          """;
+
+      CountDocumentsCommands countCommand =
+          objectMapper.readValue(json, CountDocumentsCommands.class);
+      CommandContext commandContext = new CommandContext("namespace", "collection");
+      CommandResult commandResult = new CommandResult(Collections.emptyList());
+      Mockito.when(commandProcessor.processCommand(commandContext, countCommand))
+          .thenReturn(Uni.createFrom().item(commandResult));
+      Mockito.when(stargateRequestInfo.getTenantId()).thenReturn(Optional.of("test-tenant"));
+      meteredCommandProcessor.processCommand(commandContext, countCommand).await().indefinitely();
```
ivansenic, line 55: usually bad to wait forever, if and when this fails CI gets stuck forever..

ivansenic (approved):
Please check what comments do make and adapt.. Approving to remove the need for wait..

src/main/java/io/stargate/sgv2/jsonapi/api/v1/metrics/JsonApiMetricsConfig.java
ivansenic: Also `CONFIGURATION.md` should be updated..

src/main/java/io/stargate/sgv2/jsonapi/api/v1/CollectionResource.java
maheshrajamani: Discussed with @ivansenic, leaving it as it is.

src/test/java/io/stargate/sgv2/jsonapi/service/processor/MeteredCommandProcessorTest.java
maheshrajamani: As discussed keeping this as NA

src/main/java/io/stargate/sgv2/jsonapi/service/processor/MeteredCommandProcessor.java
maheshrajamani: Will create a separate PR, we are not currently tracking gRPC error in the ErrorCode enum.

PR #444, opened 2023-06-05
(WIP) Fix handling of "findOneAndReplace" wrt #435

### LINKED ISSUE #435: findOneAndReplace with update operators throws unreadable error
The following JSON API command:

```
{
  "findOneAndReplace": {
    "filter": {},
    "replacement": {
      "$set": {
        "username": "aaronm"
      },
      "_id": "646cc1ef474ed1af6f8cf461"
    },
    "options": {
      "upsert": true
    }
  }
}

```

throws the following error:

```
Error: Command "findOneAndReplace" failed with the following errors: [{"message":"Cannot invoke \"com.fasterxml.jackson.databind.JsonNode.getNodeType()\" because \"node2\" is null","exceptionClass":"NullPointerException"}]
```

This command should throw an error, update operators aren't allowed with `findOneAndReplace()`. But the error message should be more helpful.

tatu-at-datastax:
**What this PR does**:

Fixes handling of problem case outlined in #435 wrt findOneAndReplace

Fixes #435

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/src/test/java/io/stargate/sgv2/jsonapi/api/v1/FindOneAndReplaceIntegrationTest.java
+++ b/src/test/java/io/stargate/sgv2/jsonapi/api/v1/FindOneAndReplaceIntegrationTest.java
@@ -337,6 +337,60 @@ public class FindOneAndReplaceIntegrationTest extends AbstractCollectionIntegrat
     }
 
     @Test
+    public void withUpsertNewId() {
+      final String newId = "new-id-1234";
+      String json =
+          """
+                {
+                  "findOneAndReplace": {
+                    "filter" : {},
+                    "replacement" : {
+                      "username": "aaronm",
+                      "_id": "%s"
+                    },
+                    "options" : {
+                      "returnDocument": "after",
+                      "upsert": true
+                    }
+                  }
+                }
+                """
+              .formatted(newId);
+      given()
+          .header(HttpConstants.AUTHENTICATION_TOKEN_HEADER_NAME, getAuthToken())
+          .contentType(ContentType.JSON)
+          .body(json)
+          .when()
+          .post(CollectionResource.BASE_PATH, namespaceName, collectionName)
+          .then()
+          .statusCode(200)
+          .body("data.document._id", is(newId))
+          .body("status.matchedCount", is(0))
+          .body("status.modifiedCount", is(0))
+          .body("status.upsertedId", is(nullValue()))
+          .body("errors", is(nullValue()));
+
+      // assert state after update
+      json =
+          """
+                {
+                  "find": {
+                    "filter" : {"username" : "aaronm"}
+                  }
+                }
+                """;
+      given()
+          .header(HttpConstants.AUTHENTICATION_TOKEN_HEADER_NAME, getAuthToken())
+          .contentType(ContentType.JSON)
+          .body(json)
+          .when()
+          .post(CollectionResource.BASE_PATH, namespaceName, collectionName)
+          .then()
+          .statusCode(200)
+          .body("data.documents[0]._id", is(newId));
+    }
+
+    @Test
     public void withUpsertNoId() {
       String json =
           """


tatu-at-datastax:
Fixed NPE but looks like there is a problem with logic: although document id provided, it is not being used (new one generated instead), which I assume is different from expected.

@vkarpov15 is my assumption correct in that a new document should be created, inserted, using given `_id`?

src/main/java/io/stargate/sgv2/jsonapi/service/updater/DocumentUpdater.java
```diff
@@ -82,7 +82,7 @@ private boolean replace(ObjectNode docToUpdate, boolean docInserted) {
     ObjectNode compareDoc = docToUpdate.deepCopy();
     JsonNode idNode = compareDoc.remove(DocumentConstants.Fields.DOC_ID);
     // The replace document cannot specify an _id value that differs from the replaced document.
-    if (replaceDocumentId != null) {
+    if (replaceDocumentId != null && idNode != null) {
```
tatu-at-datastax, line 85: This avoids NPE for case where no document exists (hence no `idNode`). Or rather, we have "empty document".

src/main/java/io/stargate/sgv2/jsonapi/service/updater/DocumentUpdater.java
```diff
@@ -93,9 +93,13 @@ private boolean replace(ObjectNode docToUpdate, boolean docInserted) {
     if (JsonUtil.equalsOrdered(compareDoc, replaceDocument())) {
       return false;
     }
-    // remove all data and add _id as first field
+    // remove all data and add _id as first field; either from original document or from replacement
     docToUpdate.removeAll();
-    if (idNode != null) docToUpdate.set(DocumentConstants.Fields.DOC_ID, idNode);
+    if (idNode != null) {
+      docToUpdate.set(DocumentConstants.Fields.DOC_ID, idNode);
+    } else if (replaceDocumentId != null) {
```
tatu-at-datastax, line 100: Without this, `_id` from replacement document not added since it was removed from `replaceDocument()` earlier.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/FindOneAndReplaceIntegrationTest.java
```diff
@@ -336,6 +336,61 @@ public void withUpsert() {
...
+          .when()
+          .post(CollectionResource.BASE_PATH, namespaceName, collectionName)
+          .then()
+          .statusCode(200)
+          .body("errors", is(nullValue()))
+          .body("status.matchedCount", is(0))
+          .body("status.modifiedCount", is(0))
+          .body("data.document._id", is(newId))
+          // Should we return id of new document as upsertedId?
+          .body("status.upsertedId", is(newId));
```
tatu-at-datastax, line 372: This is something I am not 100% sure about; should there instead be `insertedId`, or is it fine for this to cover both replacement of existing and insert of now document?

tatu-at-datastax:
@maheshrajamani I think this is ready: the only unclear thing for me is whether `upsertedId` should also be used when inserting a new document. I assume this is correct but wanted to know what you think.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/FindOneAndReplaceIntegrationTest.java
vkarpov15: I think `upsertedId` is fine

src/main/java/io/stargate/sgv2/jsonapi/service/updater/DocumentUpdater.java
```diff
@@ -93,9 +93,13 @@ private boolean replace(ObjectNode docToUpdate, boolean docInserted) {
...
+      docToUpdate.set(DocumentConstants.Fields.DOC_ID, replaceDocumentId);
```
vkarpov15, line 101: One potential issue here: `findOneAndReplace()` shouldn't allow changing the document's id. So if `_id` is specified in the replacement document and it is exactly equal to the existing document's `_id` (if there is an existing document), then `findOneAndReplace()` should succeed. However, if `_id` is specified in the replacement document and it is different than the existing document's `_id`, `findOneAndReplace()` should fail.

tatu-at-datastax: Yes: I think existing test `byIdWithDifferentId()` (added by @maheshrajamani ) catches this case, verifies it is caught and handled as expected.

maheshrajamani (approved):

PR #451, opened 2023-06-09
Fix #442: avoid EPIPE by using higher low-level limit; stricter limit…

### LINKED ISSUE #442: Write EPIPE error while inserting doc over 1 MB
While trying to insert a document with size greater than 1MB, the server is expected to return below message in the `errors` field in the response.

```json
[{"message":"Request failed with status code 413"}]
```

But instead of the above, intermittently, we get back error message as `write EPIPE`, which the stargate-mongoose CI tests fail intermittently (attached the logs from one of the failed runs)
[stargate_mongoose_ci_fail_logs.txt](https://github.com/stargate/jsonapi/files/11619106/stargate_mongoose_ci_fail_logs.txt)
.

```json
[{"message":"write EPIPE"}]
```

Request (document with size 1 MB):
```json
insertOne: {
    document: {
      username: 'aaaaaaaa'... 1048566 more characters
    }
  }
}
```

tatu-at-datastax:
**What this PR does**:

Configures low-level Quarkus max-request limit back to 10 megs, to let 1 meg limit enforced by App-level logic (in `Shredder`); this to reduce likelihood of legit requests hitting "EPIPE" response due to low-level http connection closure.

See #442 for more discussion on chosen approach.

Fixes #442

**Checklist**
- [ ] Changes manually tested


src/main/resources/application.yaml
```diff
@@ -49,8 +49,10 @@ quarkus:
           policy: authenticated
 
     limits:
-      # Let's lower maximum HTTP request size from 10 megs to 1
-      max-body-size: 1M
+      # Let's limit low-level maximum HTTP request size to 10 megs: stricter limits
+      # are applied at the JSON API level. Low-level limits may result in EPIPE/413 errors
+      # whereas at higher level we can use regular JSON API error responses
+      max-body-size: 10M
```
ivansenic, line 55: what's the default here, do we even need to explicitly define it?

ivansenic (approved):

kathirsvn:
@tatu-at-datastax Looking at the error response below, it looks like an error response 'constructed' on the server side with 200 status code, and so it doesn't seem to be an error thrown from the quarkus http low level APIs. Might it be an error thrown by the coordinator to the JSON API or by Cassandra/DSE to the coordinator?

```json
[{"message":"write EPIPE"}]
```

src/test/java/io/stargate/sgv2/jsonapi/api/v1/HTTPLimitsIntegrationTest.java
```diff
@@ -23,8 +23,8 @@ public class HTTPLimitsIntegrationTest extends AbstractNamespaceIntegrationTestB
   @Test
   @SuppressWarnings("deprecation") // CoreProtocolPNames deprecated
   public void tryToSendTooBigInsert() {
-    // try sending 1.01 MB
-    RandomBytesInputStream inputStream = new RandomBytesInputStream(1024 * 1024 + 1024);
+    // try sending 10.01 MB (current limit in "application.yaml" is 10 megs
+    RandomBytesInputStream inputStream = new RandomBytesInputStream(10 * 1024 * 1024 + 1024);
```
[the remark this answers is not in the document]
tatu-at-datastax: It's a fair question. I am not 100% sure -- I think we mostly want to make sure that below limit we document things work, and above there's failure.

So I can go ahead and actually remove this test.

src/main/resources/application.yaml
tatu-at-datastax: I think it is 10 megs. We do not absolutely have to define it, although it may make sense for documentation purposes.

tatu-at-datastax:
> @tatu-at-datastax Looking at the error response below, it looks like an error response 'constructed' on the server side with 200 status code, and so it doesn't seem to be an error thrown from the quarkus http low level APIs. Might it be an error thrown by the coordinator to the JSON API or by Cassandra/DSE to the coordinator?
> 
> ```json
> [{"message":"write EPIPE"}]
> ```

Good point; it cannot be just due to closing of underlying connection.

Does the JSON Object returned contain anything else beyond `message`?

tatu-at-datastax:
@kathirsvn Is there an easy way to try to reproduce this issue? I could try to at least see the difference between existing `main` and this PR.

kathirsvn:
> @kathirsvn Is there an easy way to try to reproduce this issue? I could try to at least see the difference between existing `main` and this PR.

@tatu-at-datastax I'm not able to reproduce on my local machine. Let me add a logger (to print the full response) next time when this happens on running the IT in Github Actions and share the output when this happens the next time.

tatu-at-datastax:
Sounds good @kathirsvn

PR #452, opened 2023-06-13
closes #174: finalizes find command tests and options

### LINKED ISSUE #174: Implement options for `find` command
**EDIT**: as per @ivansenic 's comment, this is about remaining verification of options just for `find` command: other work has been completed.

ivansenic:
**What this PR does**:
Standard stuff:

* more assertions in int tests
* refactored unit tests
* simplified validation

One important thing to figure out is limit, currently:

* it's unbounded, meaning I can ask for 10000 documents if I want
* second, it's used as a limit to the CQL op, which does no make too much sense for me (but could be valid)
* I am pritty sure limit does not work in sort, or if it does then it's a hack

Fixes #174

Tasks:

* [x] figure our where sort int tests are are refactor, assert
* [x] max for the limit value
* [x] limit should not influence sort page size fetch

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/FindCommand.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/FindCommand.java
@@ -10,6 +10,7 @@ import io.stargate.sgv2.jsonapi.api.model.command.clause.filter.FilterClause;
 import io.stargate.sgv2.jsonapi.api.model.command.clause.sort.SortClause;
 import io.stargate.sgv2.jsonapi.api.model.command.validation.CheckFindOption;
 import jakarta.validation.Valid;
+import jakarta.validation.constraints.Positive;
 import jakarta.validation.constraints.PositiveOrZero;
 import javax.annotation.Nullable;
 import org.eclipse.microprofile.openapi.annotations.enums.SchemaType;
@@ -28,7 +29,7 @@ public record FindCommand(
   public record Options(
 
       // limit of returned documents
-      @PositiveOrZero(message = "limit should be greater than or equal to `0`")
+      @Positive(message = "limit should be greater than `0`")
           @Schema(
               description =
                   "Maximum number of document that can be fetched for the command. If value is higher than the default page size, amount of returned documents will be limited to the default page size and paging state will be returned in the response, so a caller can to continue paging through documents.",
--- a/src/test/java/io/stargate/sgv2/jsonapi/api/model/command/impl/FindCommandTest.java
+++ b/src/test/java/io/stargate/sgv2/jsonapi/api/model/command/impl/FindCommandTest.java
@@ -64,7 +64,30 @@ public class FindCommandTest {
       assertThat(result)
           .isNotEmpty()
           .extracting(ConstraintViolation::getMessage)
-          .contains("limit should be greater than or equal to `0`");
+          .contains("limit should be greater than `0`");
+    }
+
+    @Test
+    public void invalidOptionsZeroLimit() throws Exception {
+      String json =
+          """
+            {
+            "find": {
+                "filter" : {"username" : "user1"},
+                "options" : {
+                  "limit" : 0
+                }
+              }
+            }
+            """;
+
+      FindCommand command = objectMapper.readValue(json, FindCommand.class);
+      Set<ConstraintViolation<FindCommand>> result = validator.validate(command);
+
+      assertThat(result)
+          .isNotEmpty()
+          .extracting(ConstraintViolation::getMessage)
+          .contains("limit should be greater than `0`");
     }
 
     @Test


src/main/java/io/stargate/sgv2/jsonapi/service/resolver/model/impl/FindCommandResolver.java
```diff
@@ -40,24 +40,40 @@ public Class<FindCommand> getCommandClass() {
...
           commandContext,
           filters,
           command.buildProjector(),
           pagingState,
           // For in memory sorting if no limit provided in the request will use
           // documentConfig.defaultPageSize() as limit
+          // TODO this is an issue, the limit in the command should be used amount of returned
+          // documents
+          //  and not for the amount of rows fetched
           Math.min(limit, operationsConfig.defaultPageSize()),
```
ivansenic, line 77: @maheshrajamani this makes no sense, if I ask for sort and set a limit, we will decrease the fetch size? we should not do this imo..

maheshrajamani: This is not reducing the fetch size, it is reducing number of records returned in the response. By default api always return back  defaultPageSize (20).

ivansenic: But then why are we not having the same in the unsorted option?

src/test/java/io/stargate/sgv2/jsonapi/api/v1/FindOperationWithSortIntegrationTest.java
```diff
@@ -93,18 +93,14 @@ public void sortWithSkipLimit() {
...
-      try {
-        for (int i = 0; i < 20; i++) {
-          if (i >= 10) {
-            arrayNode.add(
-                objectMapper.readTree(
-                    objectMapper
-                        .writerWithDefaultPrettyPrinter()
-                        .writeValueAsString(testDatas.get(i))));
-          }
+      for (int i = 0; i < 20; i++) {
```
vkarpov15, line 96: It would be great to have 21 documents here, or reduce `skip` to 9, just to make sure `limit` applies. Right now we're only testing that `skip` works, this test would pass even if `limit` was ignored.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/FindOperationWithSortIntegrationTest.java
```diff
@@ -115,12 +111,14 @@ public void sortWithSkipLimit() {
           .post(CollectionResource.BASE_PATH, namespaceName, collectionName)
           .then()
           .statusCode(200)
+          .body("status", is(nullValue()))
+          .body("errors", is(nullValue()))
+          .body("data.documents", hasSize(10))
           .body("data.documents", jsonEquals(arrayNode.toString()));
```
vkarpov15, line 117: Can we also test with sorting by `{ username: -1 }` just to make sure sort is being applied? AFAIK this test would pass even if sort option was ignored.

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/model/impl/FindCommandResolver.java
```diff
@@ -40,24 +40,40 @@ public Class<FindCommand> getCommandClass() {
...
-              ? options.skip()
-              : 0;
       return FindOperation.sorted(
           commandContext,
           filters,
           command.buildProjector(),
           pagingState,
           // For in memory sorting if no limit provided in the request will use
           // documentConfig.defaultPageSize() as limit
+          // TODO this is an issue, the limit in the command should be used amount of returned
```
vkarpov15, line 74: Is this comment accurate? If so, this could be problematic for `skip`. For example, if skip is 11 and limit is 10, would I only get 9 docs back?

vkarpov15 (commented):
A couple of questions/comments

src/test/java/io/stargate/sgv2/jsonapi/api/v1/FindOperationWithSortIntegrationTest.java
ivansenic: There are 25 documents in total inserted in the test..

src/test/java/io/stargate/sgv2/jsonapi/api/v1/FindOperationWithSortIntegrationTest.java
ivansenic: Hmm, there should be a test for desc as well.. Will re-check..

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/model/impl/FindCommandResolver.java
ivansenic: Could be wrong, that's why TODO is there :smile:

src/test/java/io/stargate/sgv2/jsonapi/api/v1/FindOperationWithSortIntegrationTest.java
ivansenic: Yes there is `sortDescendingTextValue` with `-1`

maheshrajamani (approved):
LGTM other than the limit validation. Think I got it wrong before.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/FindCommand.java
```diff
@@ -25,20 +26,26 @@ public record FindCommand(
     implements ReadCommand, Filterable, Projectable {
 
   public record Options(
-      @Valid
+
+      // limit of returned documents
+      @PositiveOrZero(message = "limit should be greater than or equal to `0`")
```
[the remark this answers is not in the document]
ivansenic: Cool will adapt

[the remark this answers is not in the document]
ivansenic: Done.

PR #462, opened 2023-07-10
Add logs to json api

### LINKED ISSUE #404: JSON API Logging
Currently the JSON API only logs a bit of info on startup (driven by Quarkus), and nothing really after that. 

We need to determine what logging JSON API will do, at a minimum it seems like there should be some error conditions that are logged. 

Would be useful to investigate what logging we have for other v2 APIs for comparison. For example, I seem to remember that other APIs are logging health check requests.

maheshrajamani:
**What this PR does**:
Added few error and warn logs in Json API.
If request-response need to be printed for debugging it can be done with adding env arguments provided by Quarkus.

Fixes #404 

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/grpc/retries/impl/JsonApiGrpcRetryPolicy.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/grpc/retries/impl/JsonApiGrpcRetryPolicy.java
@@ -8,6 +8,8 @@ import io.stargate.bridge.proto.QueryOuterClass;
 import io.stargate.sgv2.api.common.grpc.retries.GrpcRetryPredicate;
 import jakarta.enterprise.context.ApplicationScoped;
 import java.util.Objects;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * Default gRPC retry policy used in the project. The policy defines retries when:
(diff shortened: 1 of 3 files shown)


kathirsvn (approved):

src/main/java/io/stargate/sgv2/jsonapi/grpc/retries/impl/JsonApiGrpcRetryPolicy.java
```diff
@@ -56,7 +58,10 @@ private boolean isValidServerSideTimeout(Metadata trailers) {
     // if we have trailers
     if (null != trailers) {
       // read, write and CAS write timeouts will include one of two trailers
-      return trailers.containsKey(READ_TIMEOUT_KEY) || trailers.containsKey(WRITE_TIMEOUT_KEY);
+      if (trailers.containsKey(READ_TIMEOUT_KEY) || trailers.containsKey(WRITE_TIMEOUT_KEY)) {
+        logger.warn("DEADLINE_EXCEEDED with read/write timeout trailers, retrying");
```
tatu-at-datastax, line 62: This could be bit noisy as it has very little information on which request, by which client, for what operation etc it is for. Or does tagging add such metadata?
It should perhaps also included whether this is for reads or writes.

I guess we can add more information as needed based on how this works in production.

PR #474, opened 2023-07-26
Find and findOne command changes to support vector search

### LINKED ISSUE #459: Find* command changes for vector search
Change sort clause to handle $vector field. This special field will accept an array of floats which will be passed to CQL “ORDER BY query_vector_value ANN OF []”. Default response document limit will be 20 (jsonapi page-size). If the limit option comes in, it will be used as a limit.
Filter clause will only support the $exists operator for the $vector field.
	
```
{
  "find" : {
    "filter" : {},
    "sort" : {"$vector" : [0.11,0.12,...]},
    "options" : {"limit" : 5}
  }
}
```

Translated CQL query
`SELECT * FROM <namespace>.<collection> WHERE <> ORDER BY query_vector_value ANN OF [0.11,0.12,...] limit 5`

maheshrajamani:
**What this PR does**:
Find and findOne command changes to support vector search

Fixes #459 

**Checklist**
- [x] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/api/model/command/deserializers/SortClauseDeserializer.java
```diff
@@ -50,25 +54,48 @@ public SortClause deserialize(JsonParser parser, DeserializationContext ctxt)
...
+            JsonNode element = arrayNode.get(i);
+            if (!element.isNumber()) {
+              throw new JsonApiException(
+                  ErrorCode.SHRED_BAD_VECTOR_VALUE, ErrorCode.SHRED_BAD_VECTOR_VALUE.getMessage());
+            }
+            arrayVals[i] = element.floatValue();
+          }
+        }
+        SortExpression exp = SortExpression.vsearch(arrayVals);
+        sortExpressions.clear();
```
tatu-at-datastax, line 79: So I think this means that we have vector-search order by, anything else passed will be ignored?
I think this is Cassandra limitation so has to be done, but would it rather fail in such cases instead of hiding the problem?

Also: in case we get sort-by-$vector first, then other types, it seems that we would not ignore other cases?
(so we would end up with vector-order as first, then one or more "regular" order by clauses.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/deserializers/SortClauseDeserializer.java
```diff
@@ -50,25 +54,48 @@ public SortClause deserialize(JsonParser parser, DeserializationContext ctxt)
...
+        sortExpressions.add(exp);
+        break;
+      } else {
+        if (!inner.getValue().isInt()
+            || !(inner.getValue().intValue() == 1 || inner.getValue().intValue() == -1)) {
+          throw new JsonMappingException(
```
tatu-at-datastax, line 85: Would it make more sense to throw JsonApiException (as this is not JSON parsing/decoding problem but higher level)?

src/main/java/io/stargate/sgv2/jsonapi/util/SortClauseUtil.java
```diff
@@ -8,10 +8,20 @@
 public class SortClauseUtil {
   public static List<FindOperation.OrderBy> resolveOrderBy(SortClause sortClause) {
     if (sortClause == null || sortClause.sortExpressions().isEmpty()) return null;
+    if (sortClause.sortExpressions().stream().findFirst().get().path().equals("$vector"))
```
tatu-at-datastax, line 11: This might be more readable by adding convenience method in `SortClause` (something like `hasVectorSort()` or whatever), called from here.

src/main/java/io/stargate/sgv2/jsonapi/util/SortClauseUtil.java
```diff
@@ -8,10 +8,20 @@
...
     return sortClause.sortExpressions().stream()
         .map(
             sortExpression ->
                 new FindOperation.OrderBy(sortExpression.path(), sortExpression.ascending()))
         .collect(Collectors.toList());
   }
+
+  public static float[] resolveVsearch(SortClause sortClause) {
+    if (sortClause == null || sortClause.sortExpressions().isEmpty()) return null;
+    if (sortClause.sortExpressions().stream().findFirst().get().path().equals("$vector")) {
```
tatu-at-datastax, line 22: Calculates `sortClause.sortExpressions().stream().findFirst().get()` twice, could assign result to a var, reuse.

tatu-at-datastax (approved):

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/deserializers/SortClauseDeserializer.java
maheshrajamani: Added validation

vkarpov15 (commented):
LGTM

PR #479, opened 2023-07-27
Allow bigger arrays for vector embeddings

### LINKED ISSUE #475: Relax array size validation for $vector field
Currently there is a validation for array field to have not more than 100 elements, This needs to be relaxed for $vector field. Other than vector index we only do exists index for this field so size of array should not be an issue.

tatu-at-datastax:
**What this PR does**:

Changes array length validation logic for our vector embedding field and verifies that we use higher configured limit (instead of `100` or so for "regular" arrays use something like `8000` (backend will have lower limit)).
Also need to ensure caller can pass matching vectors for search.

Fixes #475

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/config/DocumentLimitsConfig.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/config/DocumentLimitsConfig.java
@@ -59,8 +59,9 @@ public interface DocumentLimitsConfig {
 
   /**
    * @return Maximum length of Vector ($vector) array JSON API allows -- NOTE: backend data store
-   *     likely limits length to a lower value; but we want to prevent handling of huge arrays
-   *     before trying to pass them to DB.
+   *     may limit length to a lower value; but we want to prevent handling of huge arrays before
+   *     trying to pass them to DB. Or, conversely, if data store does not limit length, to impose
+   *     something reasonable from JSON API perspective (for service-protection reasons).
    */
   @Positive
   @WithDefault("16000")
--- a/src/main/java/io/stargate/sgv2/jsonapi/config/DocumentLimitsConfig.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/config/DocumentLimitsConfig.java
@@ -63,6 +63,6 @@ public interface DocumentLimitsConfig {
    *     before trying to pass them to DB.
    */
   @Positive
-  @WithDefault("8000")
+  @WithDefault("16000")
   int maxVectorEmbeddingLength();
 }
--- a/src/main/java/io/stargate/sgv2/jsonapi/config/DocumentLimitsConfig.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/config/DocumentLimitsConfig.java
@@ -56,4 +56,13 @@ public interface DocumentLimitsConfig {
   @Positive
   @WithDefault("100")
   int maxArrayLength();
+
+  /**
+   * @return Maximum length of Vector ($vector) array JSON API allows -- NOTE: backend data store
+   *     likely limits length to a lower value; but we want to prevent handling of huge arrays
+   *     before trying to pass them to DB.
+   */
+  @Positive
+  @WithDefault("8000")
+  int maxVectorEmbeddingLength();
 }
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/shredding/Shredder.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/shredding/Shredder.java
@@ -231,26 +231,28 @@ public class Shredder {
     ++depth;
     validateDocDepth(limits, depth);
 
-    // One special case: vector embeddings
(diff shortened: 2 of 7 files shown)


src/main/java/io/stargate/sgv2/jsonapi/config/DocumentLimitsConfig.java
```diff
@@ -56,4 +56,13 @@ public interface DocumentLimitsConfig {
...
   @WithDefault("100")
   int maxArrayLength();
+
+  /**
+   * @return Maximum length of Vector ($vector) array JSON API allows -- NOTE: backend data store
+   *     likely limits length to a lower value; but we want to prevent handling of huge arrays
+   *     before trying to pass them to DB.
+   */
+  @Positive
+  @WithDefault("8000")
```
vkarpov15, line 66: 8000 is probably ok given that OpenAI's recent models output vectors of length 1536, and other libs I've worked with generally output vectors of length around 1000-2000.

However, [older OpenAI models would output embeddings of around 12k](https://openai.com/blog/new-and-improved-embedding-model). So would there be anything blocking us from increasing this limit to 20k or even 50k to be safe?

tatu-at-datastax: @vkarpov15 Just memory usage and decoding overhead, but that should be ok at this range.
How about I just set it as 16000 then for now.

src/main/java/io/stargate/sgv2/jsonapi/service/shredding/Shredder.java
```diff
@@ -212,24 +212,38 @@ private void validateDocument(DocumentLimitsConfig limits, ObjectNode doc, Strin
...
 
-  private void validateArrayValue(DocumentLimitsConfig limits, JsonNode arrayValue, int depth) {
+  private void validateArrayValue(
+      DocumentLimitsConfig limits, String referringPropertyName, JsonNode arrayValue, int depth) {
     ++depth;
     validateDocDepth(limits, depth);
 
-    if (arrayValue.size() > limits.maxArrayLength()) {
+    // One special case: vector embeddings
+    if (DocumentConstants.Fields.VECTOR_EMBEDDING_FIELD.equals(referringPropertyName)) {
```
maheshrajamani, line 235: Should we first do if arrayValue.size() > limits.maxArrayLength() first, if true check the field name. This way, don't need to compare the field name for every array node.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/VectorSearchIntegrationTest.java
```diff
@@ -125,6 +156,93 @@ public void insertVectorSearch() {
           .body("errors", is(nullValue()));
     }
 
+    // Test to verify vector embedding size can exceed general Array length limit
+    @Test
+    public void insertBigVectorThenSearch() {
```
maheshrajamani, line 161: Can you add a test case for vector size error? And a test case that verifies $set and findOneAndReplace command?

maheshrajamani (approved):

src/main/java/io/stargate/sgv2/jsonapi/service/shredding/Shredder.java
tatu-at-datastax: Was wondering which way to but yes that would probably slightly more efficient. Will change.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/VectorSearchIntegrationTest.java
tatu-at-datastax: Not sure about vector size (I assume you mean passing wrong sized vector?) -- I don't think validation via `Shredder` will handle that part as it doesn't have access, and I think it needs to be done in some part.

Will look into adding ITs for `$set` and `findOneAndReplace`

tatu-at-datastax: Test for $set added, looking into findOneAndReplace next.

maheshrajamani: > Not sure about vector size (I assume you mean passing wrong sized vector?) -- I don't think validation via `Shredder` will handle that part as it doesn't have access, and I think it needs to be done in some part.

What I meant was to create a test case with $vector array size > 16K which will error?

tatu-at-datastax: @maheshrajamani Ahh! Yes, that makes sense. Thank you for clarifying this!

tatu-at-datastax: Added test for max vector column length -- and since it turns out this was not enforced for `createCollection()`, added code for that as well.

PR #496, opened 2023-08-08
Revisit docker mem allocation

Yuqi-Du:
**What this PR does**:
- fixed JSON_API native image (build time is longer, since need to build two time for ARM and AMD platforms. Currently didn’t see other options)
- JSONAPI use native image by default and give option to use java-based image
- reallocate memory for JSONAPI java-based image, native image, dse node.

Fixes #<494>, part of [issue 2607](https://github.com/stargate/stargate/issues/2607)

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/.quarkus/cli/plugins/quarkus-cli-catalog.json
+++ b/.quarkus/cli/plugins/quarkus-cli-catalog.json
@@ -1,5 +1,0 @@
-{
-  "version" : "v1",
-  "lastUpdate" : "07/08/2023 14:34:27",
-  "plugins" : { }
-}


jeffreyscarpenter:
I suggest that we have just the docker compose fixes in this PR and handle the image building changes in a separate PR.

Yuqi-Du:
> I suggest that we have just the docker compose fixes in this PR and handle the image building changes in a separate PR.

I understand. 
But the thing is that our native image for JSONAPI on dockerhub is unusable.
The changes in docker compose will use native image by default, then users will have an issue running JSONAPI by pulling this unusable native image.

jeffreyscarpenter:
> The changes in docker compose will use native image by default, then users will have an issue running JSONAPI by pulling this unusable native image.

We should probably back that part out for now then. The original focus of the issue was to update the memory settings. Making native image the default is a good idea but since we are having issues let's address separately.

PR #500, opened 2023-08-08
Projection clause support for similarity function score

### LINKED ISSUE #463: Include vector similarity in vector search result
Ideally there should be a way to return the computed similarity between the search vector and the result vectors.

maheshrajamani:
**What this PR does**:
Change to support $similarity for find and findOne.
$similarity will return the similarity score for every document returned with respect to vector search.

Fixes #463 

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/bridge/executor/NamespaceCache.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/bridge/executor/NamespaceCache.java
@@ -63,9 +63,9 @@ public class NamespaceCache {
   private Uni<CollectionProperty> getVectorProperties(String collectionName) {
     return queryExecutor
         .getSchema(namespace, collectionName)
-        .onItemOrFailure()
+        .onItem()
         .transform(
-            (table, error) -> {
+            table -> {
               if (table.isPresent()) {
                 Boolean vectorEnabled =
                     table.get().getColumnsList().stream()
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/bridge/executor/NamespaceCache.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/bridge/executor/NamespaceCache.java
@@ -125,7 +125,7 @@ public class NamespaceCache {
 
       public static SimilarityFunction fromString(String similarityFunction) {
         if (similarityFunction == null) return UNDEFINED;
-        return switch (similarityFunction) {
+        return switch (similarityFunction.toLowerCase()) {
           case "cosine" -> COSINE;
           case "euclidean" -> EUCLIDEAN;
           case "dot_product" -> DOT_PRODUCT;
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/bridge/executor/NamespaceCache.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/bridge/executor/NamespaceCache.java
@@ -4,9 +4,12 @@ import com.github.benmanes.caffeine.cache.Cache;
 import com.github.benmanes.caffeine.cache.Caffeine;
 import io.grpc.StatusRuntimeException;
 import io.smallrye.mutiny.Uni;
+import io.stargate.bridge.proto.Schema;
 import io.stargate.sgv2.jsonapi.config.constants.DocumentConstants;
 import io.stargate.sgv2.jsonapi.exception.ErrorCode;
+import io.stargate.sgv2.jsonapi.exception.JsonApiException;
 import java.time.Duration;
+import java.util.Optional;
 
 /** Caches the vector enabled status for the namespace */
 public class NamespaceCache {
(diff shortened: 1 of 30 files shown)


src/main/java/io/stargate/sgv2/jsonapi/service/bridge/executor/NamespaceCache.java
```diff
@@ -57,22 +60,80 @@ protected Uni<Boolean> isVectorEnabled(String collectionName) {
...
+
+      public static SimilarityFunction fromString(String similarityFunction) {
+        if (similarityFunction == null) return UNDEFINED;
+        return switch (similarityFunction) {
+          case "cosine" -> COSINE;
+          case "euclidean" -> EUCLIDEAN;
+          case "dot_product" -> DOT_PRODUCT;
+          default -> throw new JsonApiException(
+              ErrorCode.VECTOR_SEARCH_INVALID_FUCTION_NAME,
+              ErrorCode.VECTOR_SEARCH_INVALID_FUCTION_NAME.getMessage() + similarityFunction);
```
vkarpov15, line 134: Does this need a space between `getMessage()` and `similarityFunction` so we don't end up with `Invalid vector search function namefoo`?

src/test/java/io/stargate/sgv2/jsonapi/service/resolver/model/impl/FindOneAndDeleteCommandResolverTest.java
```diff
@@ -205,5 +208,27 @@ public void idFilterWithProjection() throws Exception {
...
     }
+
+    @Test
+    public void idFilterWithProjectionSimilarity() throws Exception {
+      String json =
+          """
+                  {
+                    "findOneAndDelete": {
+                      "filter" : {"_id" : "id"},
+                      "projection" : { "$similarity" : 1 }
```
vkarpov15, line 219: Something worth considering: it may be worthwhile to require using `$similarity: { $meta: 1 }` or something similar, instead of `$similarity: 1`. Because, otherwise, `$similarity: 1` looks like an inclusive projection, which makes it difficult to tell JSON API "get me the document with all fields, plus `$similarity`".

Mongoose [has logic to avoid treating objects in projections as a clause that defines whether the projection as inclusive or exclusive](https://github.com/Automattic/mongoose/blob/7287420d94e7917049ad78b6ce0e03d9459473ad/lib/helpers/projection/isDefiningProjection.js#L12-L16). Even if JSON API treats `$similarity` as a special key that doesn't define the projection as inclusive or exclusive, we would also have to add support for that to Mongoose. Which is why I think the `$meta` approach is worth considering.

vkarpov15 (commented):
I left a suggestion to consider supporting `$similarity: { $meta: 1 }` as a projection syntax which tells JSON API to get the similarity score, but not treat the projection as an inclusive projection.

Also, a smaller question: what does $similarity contain if the document doesn't have a `$vector` field? Ideally `null` or `undefined`, but looks like it might be `0`?

src/main/java/io/stargate/sgv2/jsonapi/service/bridge/executor/NamespaceCache.java
maheshrajamani: Added space as suggested. This error will never be hit because its validated while collection creation,

src/test/java/io/stargate/sgv2/jsonapi/service/resolver/model/impl/FindOneAndDeleteCommandResolverTest.java
maheshrajamani: @vkarpov15 The design was to ask for required column when similarity score is needed. Should we add it as a command options also, so it can return along with full document?

maheshrajamani:
> Also, a smaller question: what does $similarity contain if the document doesn't have a `$vector` field? Ideally `null` or `undefined`, but looks like it might be `0`?

Vector search doesn't return document that don't have $vector field in it. `0` is returned if the query is not vector search but the projection had asked for $similarity. May be we can ignore this.

jeffreyscarpenter (approved):
looks good.

maheshrajamani:
> > Also, a smaller question: what does $similarity contain if the document doesn't have a `$vector` field? Ideally `null` or `undefined`, but looks like it might be `0`?
> 
> Vector search doesn't return document that don't have $vector field in it. `0` is returned if the query is not vector search but the projection had asked for $similarity. May be we can ignore this.

@vkarpov15 Changed it to not add `0` for the default $similarity projection for a non vector search query.

vkarpov15:
As per discussion with @maheshrajamani today, we should add a `includeSimilarity` option to `find()`, `findOne()`, etc. - any place where we support `$similarity: 1` in projection. This will be an alternative syntax to better support the use case of "return the whole document with $similarity".

src/main/java/io/stargate/sgv2/jsonapi/service/projection/DocumentProjector.java
```diff
@@ -161,6 +207,20 @@ PathCollector collectFromObject(JsonNode ob, String parentPath) {
           addSlice(parentPath, entry.getValue());
           continue;
         }
+        // This `or` is needed because the method is called in loop
+        includeSimilarityScore = includeSimilarityScore || includeSimilarity;
+        if (parentPath == null
+            && DocumentConstants.Fields.VECTOR_FUNCTION_PROJECTION_FIELD.equals(path)) {
+          JsonNode value = entry.getValue();
+          if (BigDecimal.ZERO.equals(value.decimalValue()) && includeSimilarity) {
```
vkarpov15, line 215: Do we also need to check for `$projection: 1` with `includeSimilarity: false`?

vkarpov15 (commented):
LGTM :+1:

PR #506, opened 2023-08-16
Merge array_equals and sub_doc_equals field into query_text_value

### LINKED ISSUE #505: Reduce index columns
Reduce the number of index columns in the schema by merging content of array_equals and sub_doc_equals into query_text_values field.

maheshrajamani:
Merge array_equals and sub_doc_equals field into query_text_value field.
Fixes #505 

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/schema/model/JsonapiTableMatcher.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/schema/model/JsonapiTableMatcher.java
@@ -13,21 +13,34 @@ public class JsonapiTableMatcher implements Predicate<Schema.CqlTable> {
 
   private final Predicate<QueryOuterClass.ColumnSpec> columnsPredicate;
 
+  private final Predicate<QueryOuterClass.ColumnSpec> columnsPredicateVector;
+
   public JsonapiTableMatcher() {
     primaryKeyPredicate = new CqlColumnMatcher.Tuple("key", Basic.TINYINT, Basic.VARCHAR);
     columnsPredicate =
         new CqlColumnMatcher.BasicType("tx_id", Basic.TIMEUUID)
             .or(new CqlColumnMatcher.BasicType("doc_json", Basic.VARCHAR))
             .or(new CqlColumnMatcher.Set("exist_keys", Basic.VARCHAR))
-            .or(new CqlColumnMatcher.Map("sub_doc_equals", Basic.VARCHAR, Basic.VARCHAR))
             .or(new CqlColumnMatcher.Map("array_size", Basic.VARCHAR, Basic.INT))
-            .or(new CqlColumnMatcher.Map("array_equals", Basic.VARCHAR, Basic.VARCHAR))
             .or(new CqlColumnMatcher.Set("array_contains", Basic.VARCHAR))
             .or(new CqlColumnMatcher.Map("query_bool_values", Basic.VARCHAR, Basic.TINYINT))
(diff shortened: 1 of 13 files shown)


jeffreyscarpenter (approved):
Assume we don't need any new tests here. Approving

maheshrajamani:
> Assume we don't need any new tests here. Approving

Yes, existing unit tests are modified to check usage of merged field.

src/main/java/io/stargate/sgv2/jsonapi/service/schema/model/JsonapiTableMatcher.java
```diff
@@ -57,7 +70,8 @@ public boolean test(Schema.CqlTable cqlTable) {
     }
 
     List<QueryOuterClass.ColumnSpec> columns = cqlTable.getColumnsList();
-    if (columns.size() != 12 || !columns.stream().allMatch(columnsPredicate)) {
+    if (!(columns.stream().allMatch(columnsPredicate)
+        || columns.stream().allMatch(columnsPredicateVector))) {
```
tatu-at-datastax, line 74: Would it not be sufficient to only check the second, as it is a superset of former, assuming I understand how `allMatch` works -- that all columns table does have are recognized?
(I am not sure if the check overall is correct since a table with just one "correct" column would seem to be considered a JSON API collection, but that may be fine).

tatu-at-datastax:
Makes sense and should work as intended as hashed-values have type prefix so array/object/string values should not collide.

I have some questions/notes but nothing that requires immediate action:

1. Should we update other documentation (spec(s)?) to indicate removal of 2 columns, extended usage of `query_text_values`?
2. Name of `query_text_values` is now bit misleading... but it's probably not worth the hassle to rename it.

src/main/java/io/stargate/sgv2/jsonapi/service/schema/model/JsonapiTableMatcher.java
maheshrajamani: Yup make sense, I will create separate PR to use only columnsPredicateVector predicate. Misunderstood the `allMatch` doc.

maheshrajamani:
> 1. Should we update other documentation (spec(s)?) to indicate removal of 2 columns, extended usage of `query_text_values`

I looked to places where `array_equals` and `sub_doc_equals` and changed it to `query_text_values`.

> 2\. Name of `query_text_values` is now bit misleading... but it's probably not worth the hassle to rename it.

Yes, not its more than just text values. Thought about it, but didn't get a better field name. Any suggestions?

tatu-at-datastax:
>> 2. Name of query_text_values is now bit misleading... but it's probably not worth the hassle to rename it.
>
> Yes, now its more than just text values. Thought about it, but didn't get a better field name. Any suggestions?

Can't think of anything obvious, given that we have `query_dbl_values` and so on.
Could do `query_misc_values` or `query_other_values` but those are not very descriptive.

I think we do need to finalize names now, before (more) public launch fwtw. I am ok leaving it as `query_text_values` if we cannot think of a better name.

jeffreyscarpenter:
Here is my argument for leaving the name `query_text_value` as-is: this PR doesn't change the column definition. The type is still: `map<text, text>`. The difference is in how the application code is using the column.

tatu-at-datastax:
I think the name is slightly misleading as its semantics changed. But. Given hassle for changing it in all places, docs, I am fine in leaving it as-is.
Let's do that.

PR #509, opened 2023-08-17
Use DataStoreProperties instead of DataStoreConfig

### LINKED ISSUE #508: DataStoreConfig is not produced, use DataStoreProperties instead
In JSON API, we need to examine things like SAI, Vector Search are supported by cassandra backend or not.
When create collection ([CreateCollectionCommandResolver](https://github.com/stargate/jsonapi/blob/main/src/main/java/io/stargate/sgv2/jsonapi/service/resolver/model/impl/CreateCollectionCommandResolver.java)), we use DataStoreConfig. However, the injected dataStoreConfig object is the default one and can not respond to backend change.

Yuqi-Du:
**What this PR does**: fix createCollectionCommandResolver, use DataStoreProperties instead of using DataStoreConfig

Fixes #508

**Checklist**
- [x] Changes manually tested


jeffreyscarpenter:
what's happening with this PR @Yuqi-Du @tatu-at-datastax

tatu-at-datastax:
I think I'd like @maheshrajamani to indicate whether this should be done -- I do not have good enough understanding of pros and cons.

PR #518, opened 2023-08-30
fix NPE when concurrent delete

Yuqi-Du:
**What this PR does**: 
findOneAndDelete with projection under concurrent situation.
For example, some requests read `a` successfully, but delete will fail because `a` has been deleted. In the retry process, JSON API will try to read `a` again, this time it read a `null`. Then when apply projection, there will be a NPE.

This issue will occur not with vectors, but also when doing findOneAndDelete with normal filter +projection.

The fix has passed the IT and also benchmark test.

Fixes #<517>

**Checklist**
- [x] Changes manually tested


tatu-at-datastax (approved):
LGTM

src/test/java/io/stargate/sgv2/jsonapi/api/v1/VectorSearchIntegrationTest.java
```diff
@@ -1683,6 +1679,114 @@ public void findOneAndReplace() {
     }
   }
 
+  @Nested
+  @Order(8)
+  @TestMethodOrder(MethodOrderer.OrderAnnotation.class)
```
tatu-at-datastax, line 1684: These seem to be duplicate for `FindOneAndDeleteIntegrationTest`, but just with vector column.
I think I would not add these since it's more code to maintain and does not add much more verification.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/FindOneAndDeleteIntegrationTest.java
```diff
@@ -321,4 +317,151 @@ public void checkMetrics() {
       FindOneAndDeleteIntegrationTest.super.checkMetrics("FindOneAndDeleteCommand");
     }
   }
+
+  @Nested
+  class ConcurrentDelete {
```
tatu-at-datastax, line 322: Good! If these failed (at least for some runs) without the fix, and pass (reliably) with the fix (couple of sample runs should be enough to prove), that's good.

PR #519, opened 2023-08-30
Fixes #516: add validation to fail on attempts to filter on $vector (except with $exists)

### LINKED ISSUE #516: findOne filtering by `$vector` returns no results
The following query returns no results:

```
{
  "findOne": {
    "filter": {
      "$vector": [
        100,
        1
      ]
    },
    "options": {}
  }
}
```

But executing the following:

```
{
  "find": {
    "filter": {}
  }
}
```

Returns:

```
[
  {
    _id: new ObjectId("64ea8e193e3aec5cd0b356b2"),
    '$vector': [ 1, 100 ],
    name: 'Test vector 1',
    __v: 0
  },
  {
    _id: new ObjectId("64ea8e193e3aec5cd0b356b3"),
    '$vector': [ 100, 1 ],
    name: 'Test vector 2',
    __v: 0
  }
]
```

Are you able to query by exact match for `$vector`?

tatu-at-datastax:
**What this PR does**:

Adds explicit fail on attempts to filter by $vector, except for one supported operation ($exists).

Fixes #516

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/deserializers/FilterClauseDeserializer.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/deserializers/FilterClauseDeserializer.java
@@ -1,6 +1,5 @@
 package io.stargate.sgv2.jsonapi.api.model.command.deserializers;
 
-import com.fasterxml.jackson.core.JacksonException;
 import com.fasterxml.jackson.core.JsonParser;
 import com.fasterxml.jackson.databind.DeserializationContext;
 import com.fasterxml.jackson.databind.JsonNode;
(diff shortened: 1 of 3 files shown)


vkarpov15 (approved):
LGTM. It would be nice to support exact match queries on $vector, but it isn't critical. And throwing an error is a big improvement over silently failing, so thanks for fixing this.

tatu-at-datastax:
> LGTM. It would be nice to support exact match queries on $vector, but it isn't critical. And throwing an error is a big improvement over silently failing, so thanks for fixing this.

I may have forgotten to mention it, but this is limitation of Cassandra itself (wrt filtering on explicit column we have): vector columns can only be filtered on if they are part of the key (partition or clustering column).
This is usually not something one wants to do.

In theory we could support it same way as Array / Object values (by calculating hash on contents) but I suspect this is not worth the work or, especially, overhead of storage and processing.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/deserializers/FilterClauseDeserializer.java
```diff
@@ -79,6 +77,16 @@ private void validate(List<ComparisonExpression> expressionList) {
   }
 
   private void validate(String path, FilterOperation<?> filterOperation) {
+    // First: $vector can only be used with $exists operator
+    if (path.equals(DocumentConstants.Fields.VECTOR_EMBEDDING_FIELD)
+        && ElementComparisonOperator.EXISTS != filterOperation.operator()) {
+      throw new JsonApiException(
+          ErrorCode.INVALID_FILTER_EXPRESSION,
+          String.format(
+              "Cannot filter on '%s' field using operator '%s': only '$exists' is supported",
```
kathirsvn, line 86: Do we have a test case for `findOne` with `$exists` for `$vector` field? Can we add one if it is not?

kathirsvn (approved):

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/deserializers/FilterClauseDeserializer.java
tatu-at-datastax: Excellent question! I believe we do since the unit tests were failing for my first version. That's how I found out that `$exists` works on Vectors -- first version simply failed on every type of expression and lead to test failure(s).

So yes, this case is fortunately covered; I think @maheshrajamani originally added that test.

PR #522, opened 2023-09-05
Fix #521: allow hyphen in names, prevent empty String

### LINKED ISSUE #521: JSON API should allow hyphens (`-`) in document property names (but should not allow empty String)
Currently only following characters are as follows:

```
[a-zA-Z0-9_]*
```

meaning that ascii letters (both cases), ten-based numbers, and underscore are allowed.
But nothing else.

It seems that there is wide-spread usage of hyphens for Mongoose apps, so we should probably allow that as well: we handle validation and do not rely on Cassandra's column name limitations.

In addition, we should NOT allow empty String as property name as that is not allowed by other databases.

tatu-at-datastax:
**What this PR does**:

Improves property name validation wrt #521 so that:

- Allow hyphen (`-`) in names and name segments
- Empty names no longer allowed (shouldn't have been, oversight)

Fixes #521

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/src/test/java/io/stargate/sgv2/jsonapi/api/v1/InsertIntegrationTest.java
+++ b/src/test/java/io/stargate/sgv2/jsonapi/api/v1/InsertIntegrationTest.java
@@ -96,7 +96,7 @@ public class InsertIntegrationTest extends AbstractCollectionIntegrationTestBase
 
     // [https://github.com/stargate/jsonapi/issues/521]: allow hyphens in property names
     @Test
-    public void insertDocumentWithHyphen() {
+    public void insertDocumentWithHyphenatedColumn() {
       String json =
           """
               {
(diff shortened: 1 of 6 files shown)


vkarpov15 (approved):
LGTM :+1:

tatu-at-datastax:
@maheshrajamani Could you also review this when you get a chance?

maheshrajamani (approved):
@tatu-at-datastax Can you add insert and find (filter) IT for this case?

tatu-at-datastax:
> @tatu-at-datastax Can you add insert and find (filter) IT for this case?

Yes that sounds reasonable.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/InsertIntegrationTest.java
```diff
@@ -67,22 +67,47 @@ public void insertDocument() {
           .body("status.insertedIds[0]", is("doc3"))
           .body("data", is(nullValue()))
           .body("errors", is(nullValue()));
+      given()
```
tatu-at-datastax, line 70: Existing not change, just inlined JSON payloads for bit more compact code.

tatu-at-datastax:
@maheshrajamani Added insert and find test cases, will merge once CI passes.

PR #528, opened 2023-09-15
Add embedding support api

### LINKED ISSUE #524: [Vectorize] Implement embedding service call
Implement client for different embedding services which can pass text data to get the embedding vector.
Following api to be supported:

- Huggingface
- OpenAI
- Vertex AI

Following command needs to be changed to support $vectorize
- InsertOne
- InsertMany
- All command that support sort
- All update clause supported commands.

maheshrajamani:
**What this PR does**:

Support for vectorize method
Fixed the update operation for vector data

Fixes #524

**Checklist**
- [x] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/update/UpdateOperation.java
```diff
@@ -59,6 +59,15 @@ protected static String validateUpdatePath(UpdateOperator oper, String path) {
           ErrorCode.UNSUPPORTED_UPDATE_FOR_VECTOR.getMessage() + ": " + oper.operator());
     }
 
+    if (DocumentConstants.Fields.VECTOR_EMBEDDING_TEXT_FIELD.equals(path)
+        && !(oper.operator().equals("$set")
+            || oper.operator().equals("$unset")
```
vkarpov15, line 64: Does `$unset` make sense with `$vectorize`? Looks like it should be equivalent to `$unset` on `$vector`. What happens if you `$unset` `$vector` but `$set` on `$vectorize`?

src/test/java/io/stargate/sgv2/jsonapi/service/embedding/operation/VectorizeDataTest.java
```diff
@@ -0,0 +1,169 @@
...
+            "update" : {"$set" : {"$vectorize" : "New York"}}
+          }
+        }
+        """;
+      FindOneAndUpdateCommand command = objectMapper.readValue(json, FindOneAndUpdateCommand.class);
+      UpdateClause updateClause = command.updateClause();
+      VectorizeData vectorizeData = new VectorizeData(testService, objectMapper.getNodeFactory());
+      vectorizeData.vectorizeUpdateClause(updateClause);
+      final ObjectNode setNode = updateClause.updateOperationDefs().get(UpdateOperator.SET);
+      assertThat(setNode.has("$vectorize")).isFalse();
```
vkarpov15, line 119: Are we not storing `$vectorize`? I thought we were planning on storing `$vectorize`?

vkarpov15 (commented):
Overall LGTM, just a couple of comments

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/update/UpdateOperation.java
maheshrajamani: Good point. I will add validation that vectorize and vector can't be used together.


> Does `$unset` make sense with `$vectorize`?

Not Sure if the user uses vectorize to load, thought will use the same to unset. I can remove it if you think it's not required.

src/test/java/io/stargate/sgv2/jsonapi/service/embedding/operation/VectorizeDataTest.java
maheshrajamani: Not a big change will add it to document json.

pom.xml
```diff
@@ -172,6 +176,21 @@
           </execution>
         </executions>
       </plugin>
+      <plugin>
+        <!-- https://github.com/wildfly/jandex-maven-plugin -->
+        <groupId>org.jboss.jandex</groupId>
+        <artifactId>jandex-maven-plugin</artifactId>
```
tatu-at-datastax, line 182: What does this do? Documentation by project is quite sparse.
Is it still actively developed? On https://github.com/wildfly/jandex-maven-plugin it's marked as archived, but maybe it just moved to somewhere else.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/sort/SortClause.java
```diff
@@ -31,4 +31,13 @@ public boolean hasVsearchClause() {
             .path()
             .equals(DocumentConstants.Fields.VECTOR_EMBEDDING_FIELD);
   }
+
+  public boolean hasVectorizeSearchClause() {
+    return sortExpressions != null
+        && sortExpressions.stream()
+            .findFirst()
```
tatu-at-datastax, line 38: Is it guaranteed that we can only have one sort expression with vectors? (I assume so based on code).
Maybe add a comment here stating that invariant.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/update/UpdateOperation.java
tatu-at-datastax: Should `$unset` on `$vector` automatically clear `$vectorize` field if one exists (now that we store it)?

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/deserializers/SortClauseDeserializer.java
```diff
@@ -85,6 +85,21 @@ public SortClause deserialize(JsonParser parser, DeserializationContext ctxt)
           sortExpressions.add(exp);
           break;
         }
+      } else if (DocumentConstants.Fields.VECTOR_EMBEDDING_TEXT_FIELD.equals(path)) {
+        if (totalFields > 1) {
```
tatu-at-datastax, line 89: Maybe add a comment indicating logic here (vector embedding update must be the only clause).

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/deserializers/SortClauseDeserializer.java
```diff
@@ -85,6 +85,21 @@ public SortClause deserialize(JsonParser parser, DeserializationContext ctxt)
...
+              ErrorCode.VECTOR_SEARCH_USAGE_ERROR.getMessage());
+        }
+        if (!inner.getValue().isTextual()) {
+          throw new JsonApiException(
+              ErrorCode.SHRED_BAD_VECTORIZE_VALUE,
+              ErrorCode.SHRED_BAD_VECTORIZE_VALUE.getMessage());
+        }
+        SortExpression exp = SortExpression.vectorizeSearch(inner.getValue().textValue());
+        sortExpressions.clear();
+        sortExpressions.add(exp);
```
tatu-at-datastax, line 101: Is there a problem (both here and for vector field) that one could reverse order and avoid failure?
That is, there first sort clause if for `$vector`/`$vectorize`, but there's second one for "regular" property? It seems like that would not result in failure here.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/update/UpdateOperation.java
maheshrajamani: In case set or unset come with $vector field, the logic should remove existing $vectorize field in the main document?

pom.xml
maheshrajamani: Added it when I had OpenAI client library used and native image creation was failing. Have removed this now.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/sort/SortClause.java
maheshrajamani: Yes, check for this is in SortClauseDeserializer.java where validation will throw VECTOR_SEARCH_USAGE_ERROR if it has more than one expression used.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/deserializers/SortClauseDeserializer.java
maheshrajamani: Added comment `Vector search can't be used with other sort clause`

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/deserializers/SortClauseDeserializer.java
maheshrajamani: it checks as `totalFields > 1` where totalFields is size of incoming expressions.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/CommandContext.java
```diff
@@ -14,13 +16,18 @@ public record CommandContext(
...
+    NamespaceCache.CollectionProperty.SimilarityFunction similarityFunction,
+    EmbeddingService embeddingService) {
 
   public CommandContext(String namespace, String collection) {
-    this(namespace, collection, false, null);
+    this(namespace, collection, false, null, null);
   }
 
-  private static final CommandContext EMPTY = new CommandContext(null, null, false, null);
+  public CommandContext(Optional<String> tenant, String namespace, String collection) {
```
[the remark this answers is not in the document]
maheshrajamani: Removed this constructor

src/test/java/io/stargate/sgv2/jsonapi/service/embedding/operation/VectorizeDataTest.java
maheshrajamani: After discussion added $vectorize to the stored document.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/CreateEmbeddingServiceCommand.java
```diff
@@ -0,0 +1,33 @@
...
+        @Size(min = 1, max = 48)
+        @Schema(description = "Service name to be created")
+        String name,
+    @Nullable
+        @Pattern(
+            regexp = "(openai|vertexai|huggingface)",
+            message = "supported providers are 'openai', 'vertexai' or 'huggingface'")
+        @Schema(
+            description = "Embedding service provider service name",
+            defaultValue = "cosine",
```
tatu-at-datastax, line 27: Probably not "cosine"... :)

src/main/java/io/stargate/sgv2/jsonapi/api/v1/CollectionResource.java
```diff
@@ -164,12 +168,23 @@ else if (error instanceof JsonApiException jsonApiException) {
                 // otherwise use generic for now
                 return Uni.createFrom().item(new ThrowableCommandResultSupplier(error));
               } else {
+                EmbeddingService embeddingService = null;
+                if (collectionProperty.vectorizeServiceName() != null
+                    && collectionProperty.modelName() != null) {
+                  embeddingService =
```
tatu-at-datastax, line 174: So we always fetch embeddingService if the Collection has one configured? Although we don't yet know if one is needed?

src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCode.java
```diff
@@ -68,6 +71,9 @@ public enum ErrorCode {
 
   UNSUPPORTED_UPDATE_OPERATION("Unsupported update operation"),
 
+  UNSUPPORTED_COMMAND_EMBEDDING_SERVICE(
+      "Unsupported command createEmbeddingService since application is configured for property based embedding"),
```
tatu-at-datastax, line 75: Include (single) quotes around createEmbeddingService -> 'createEmbeddingService' to be consistent with other messages.

src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCode.java
```diff
@@ -91,7 +98,14 @@ public enum ErrorCode {
   VECTOR_SEARCH_SIMILARITY_PROJECTION_NOT_SUPPORTED(
       "$similarity projection is not supported for this command"),
 
-  VECTOR_SEARCH_FIELD_TOO_BIG("Vector embedding field '$vector' length too big");
+  VECTOR_SEARCH_FIELD_TOO_BIG("Vector embedding field '$vector' length too big"),
+  VECTORIZE_SERVICE_NOT_REGISTERED("Vectorize service name provided is not registered : "),
+
+  VECTORIZE_SERVICE_TYPE_NOT_ENABLED("Vectorize service type not enabled : "),
+  VECTORIZE_SERVICE_TYPE_UNSUPPORTED("Vectorize service type unsupporterd : "),
```
tatu-at-datastax, line 105: typo in "unsupporterd" -> "unsupported"

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/VectorizeData.java
```diff
@@ -0,0 +1,114 @@
...
+import io.stargate.sgv2.jsonapi.config.constants.DocumentConstants;
+import io.stargate.sgv2.jsonapi.exception.ErrorCode;
+import io.stargate.sgv2.jsonapi.exception.JsonApiException;
+import io.stargate.sgv2.jsonapi.service.embedding.operation.EmbeddingService;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class VectorizeData {
```
tatu-at-datastax, line 20: It'd be good to have at least some comments explaining functioning of this class...

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/VectorizeData.java
```diff
@@ -0,0 +1,114 @@
...
+  private final EmbeddingService embeddingService;
+  private final JsonNodeFactory nodeFactory;
+
+  public VectorizeData(EmbeddingService embeddingService, JsonNodeFactory nodeFactory) {
+    this.embeddingService = embeddingService;
+    this.nodeFactory = nodeFactory;
+  }
+
+  public void vectorize(List<JsonNode> documents) {
```
tatu-at-datastax, line 29: Javadoc? So this modifies given documents if they have `$vectorize` field?

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/VectorizeData.java
```diff
@@ -0,0 +1,114 @@
...
+        }
+
+        vectorizeTexts.add(jsonNode.asText());
+        vectorizeMap.put(vectorDataPosition, position);
+        vectorDataPosition++;
+      }
+    }
+
+    if (!vectorizeTexts.isEmpty()) {
+      List<float[]> vectors = embeddingService.vectorize(vectorizeTexts);
```
tatu-at-datastax, line 56: Ah. So we collect vectorizable fields from all entries (within same collection), and make what is hopefully a batch call.

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/configuration/PropertyBasedEmbeddingServiceConfig.java
```diff
@@ -0,0 +1,62 @@
...
+import javax.annotation.Nullable;
+
+@ConfigMapping(prefix = "stargate.jsonapi.embedding.service")
+public interface PropertyBasedEmbeddingServiceConfig {
+
+  @Nullable
+  OpenaiConfig openai();
+
+  @Nullable
+  HuggingFaceConfig hf();
```
tatu-at-datastax, line 14: Maybe just use full name instead "hf"?

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/configuration/PropertyBasedEmbeddingServiceConfig.java
```diff
@@ -0,0 +1,62 @@
...
+  CustomConfig custom();
+
+  public interface OpenaiConfig {
+    @WithDefault("false")
+    boolean enabled();
+
+    @WithDefault("https://api.openai.com/v1")
+    String url();
+
+    @WithDefault("Bearer")
```
tatu-at-datastax, line 29: What is the meaning of default, if any? (as in, why not just empty String)

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/configuration/PropertyBasedEmbeddingServiceConfig.java
```diff
@@ -0,0 +1,62 @@
...
+
+  @Nullable
+  CustomConfig custom();
+
+  public interface OpenaiConfig {
+    @WithDefault("false")
+    boolean enabled();
+
+    @WithDefault("https://api.openai.com/v1")
+    String url();
```
tatu-at-datastax, line 27: Should this be `java.net.URL` (or `URI`) to force validation?

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/configuration/PropertyBasedEmbeddingServiceConfig.java
```diff
@@ -0,0 +1,62 @@
...
+
+    @WithDefault("Bearer")
+    String apiKey();
+  }
+
+  public interface CustomConfig {
+    @WithDefault("false")
+    boolean enabled();
+
+    @WithDefault("Class not defined")
```
tatu-at-datastax, line 59: Could possibly be `Class<?>` valued -- default could be `java.lang.Void` f.ex.

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/EmbeddingServiceCache.java
```diff
@@ -0,0 +1,79 @@
...
+          final EmbeddingService customEmbeddingClient =
+              (EmbeddingService) clazz.getConstructor().newInstance();
+          return customEmbeddingClient;
+        } catch (NoSuchMethodException
+            | ClassNotFoundException
+            | InstantiationException
+            | IllegalAccessException
+            | InvocationTargetException
+            | ClassCastException e) {
+          e.printStackTrace();
```
tatu-at-datastax, line 67: Remove

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/EmbeddingServiceCache.java
```diff
@@ -0,0 +1,79 @@
...
+          Class<?> clazz = Class.forName(configuration.className());
+          final EmbeddingService customEmbeddingClient =
+              (EmbeddingService) clazz.getConstructor().newInstance();
+          return customEmbeddingClient;
+        } catch (NoSuchMethodException
+            | ClassNotFoundException
+            | InstantiationException
+            | IllegalAccessException
+            | InvocationTargetException
+            | ClassCastException e) {
```
tatu-at-datastax, line 66: Probably easier to just catch `Exception` here?

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/EmbeddingServiceCache.java
```diff
@@ -0,0 +1,79 @@
...
+          e.printStackTrace();
+          throw new JsonApiException(
+              ErrorCode.VECTORIZE_SERVICE_TYPE_UNAVAILABLE,
+              ErrorCode.VECTORIZE_SERVICE_TYPE_UNAVAILABLE.getMessage()
+                  + configuration.className());
```
tatu-at-datastax, line 71: Should probably include exception message here, or if not, add original exception as cause (if `JsonApiException` allows that)

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/HuggingFaceEmbeddingClient.java
```diff
@@ -0,0 +1,54 @@
...
+    this.modelName = modelName;
+    this.baseUrl = baseUrl;
+    embeddingService =
+        QuarkusRestClientBuilder.newBuilder()
+            .baseUri(URI.create(baseUrl))
+            .build(HuggingFaceEmbeddingService.class);
+  }
+
+  public HuggingFaceEmbeddingClient(String apiKey, String modelName) {
+    this("https://api-inference.huggingface.co", apiKey, modelName);
```
tatu-at-datastax, line 31: Should maybe define in one of constants classes? (if there's natural place for it) -- I think config class duplicates this as default URL? (so maybe define as constant there actually).

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/OpenAiEmbeddingClient.java
```diff
@@ -0,0 +1,57 @@
...
+    this.modelName = modelName;
+    this.baseUrl = baseUrl;
+    embeddingService =
+        QuarkusRestClientBuilder.newBuilder()
+            .baseUri(URI.create(baseUrl))
+            .build(OpenAiEmbeddingService.class);
+  }
+
+  public OpenAiEmbeddingClient(String apiKey, String modelName) {
+    this("https://api.openai.com/v1", apiKey, modelName);
```
tatu-at-datastax, line 30: Similarly would be good to refer to a constant.

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/test/TestEmbeddingService.java
```diff
@@ -0,0 +1,45 @@
...
+  ;
+
+  @Override
+  public List<float[]> vectorize(List<String> texts) {
+    List<float[]> response = new ArrayList<>(texts.size());
+    for (String text : texts) {
+      if (TEST_DATA.containsKey(text)) {
+        response.add(TEST_DATA.get(text));
+      } else {
+        response.add(new float[] {0.25f, 0.25f, 0.25f});
```
tatu-at-datastax, line 40: Is it intentional that real matches have 5 values, fallback only 3?

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/CreateCollectionCommand.java
```diff
@@ -54,5 +60,21 @@ public VectorSearchConfig(Integer size, String function) {
         this.function = function == null ? "cosine" : function;
       }
     }
+
+    public record VectorizeConfig(
```
tatu-at-datastax, line 64: I think this class as well as `VectorSearchConfig` should be main level class, not nested in`CreateCollectionCommand` -- since it's not really part of CCC and is used elsewhere.

Also: if this is what is stored in CQL table comment property, I think it should be done in a way that can extended: I suspect we will be adding more configuration properties to persist. So `VectorizeConfig` would be a root-level property of general "StoredCollectionConfig" (or whatever config container is named).

src/main/java/io/stargate/sgv2/jsonapi/service/bridge/executor/NamespaceCache.java
```diff
@@ -99,10 +105,26 @@ private Uni<CollectionProperty> getVectorProperties(String collectionName) {
...
                     }
                   }
-                  return new CollectionProperty(vectorEnabled, function);
+                  final String comment = table.get().getOptionsOrDefault("comment", null);
+                  CreateCollectionCommand.Options.VectorizeConfig vectorizeConfig = null;
+                  if (comment != null && !comment.isBlank()) {
+                    try {
+                      vectorizeConfig =
+                          objectMapper.readValue(
+                              comment, CreateCollectionCommand.Options.VectorizeConfig.class);
```
tatu-at-datastax, line 114: As per my note next to class, I think we should use a general CollectionConfiguration object and have `VectorizeConfig` be main-level property. This allows easier extension of configuration that we will probably need -- f.ex for Text tokenization, likely.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/CreateCollectionOperation.java
```diff
@@ -15,17 +15,23 @@ public record CreateCollectionOperation(
     String name,
     boolean vectorSearch,
     int vectorSize,
-    String vectorFunction)
+    String vectorFunction,
+    String vectorize)
```
tatu-at-datastax, line 19: Is this a JSON serialization of Vectorize service configuration?

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/CreateCollectionOperation.java
```diff
@@ -66,6 +72,9 @@ protected QueryOuterClass.Query getCreateTable(String keyspace, String table) {
               + vectorSize
               + ">, "
               + "    PRIMARY KEY (key))";
+      if (vectorize != null) {
+        createTableWithVector = createTableWithVector + " WITH comment = '" + vectorize + "'";
```
tatu-at-datastax, line 76: Not sure if it matters, but if `vectorize` is serialized JSON it could still have characters that require escaping (esp. apostrophes themselves). Although I guess those would lead to CQL query failure, not corrupted data.

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/CommandResolverService.java
```diff
@@ -40,7 +41,7 @@ public <T extends Command> Uni<CommandResolver<T>> resolverForCommand(T command)
     // try to get from the map of resolvers
     return Uni.createFrom()
         .item((CommandResolver<T>) resolvers.get(command.getClass()))
-
+        .runSubscriptionOn(Infrastructure.getDefaultWorkerPool())
```
tatu-at-datastax, line 44: Could use a comment on purpose of this? (sounds like it's async operation for some reason?)

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/model/impl/CreateCollectionCommandResolver.java
```diff
@@ -52,8 +59,17 @@ public Operation resolveCommand(CommandContext ctx, CreateCollectionCommand comm
                 vectorSize,
                 documentLimitsConfig.maxVectorEmbeddingLength()));
       }
+      String vectorize = null;
+      if (command.options().vectorize() != null) {
+        try {
+          vectorize = objectMapper.writeValueAsString(command.options().vectorize());
+        } catch (JsonProcessingException e) {
+          // This should never happen because the object is extracted from json request
```
tatu-at-datastax, line 67: which is why should probably re-throw as RuntimeException just in case it somehow happened. In theory value object serialization could fail.

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/model/impl/CreateEmbeddingServiceCommandResolver.java
```diff
@@ -0,0 +1,55 @@
...
+@ApplicationScoped
+public class CreateEmbeddingServiceCommandResolver
+    implements CommandResolver<CreateEmbeddingServiceCommand> {
+
+  private EmbeddingServiceConfigStore embeddingServiceConfigStore;
+  private StargateRequestInfo stargateRequestInfo;
+
+  @Inject
+  public CreateEmbeddingServiceCommandResolver(
+      Instance<EmbeddingServiceConfigStore> embeddingServiceConfigStore,
```
tatu-at-datastax, line 27: Why is `Instance` used here? Are embedding service config stores created asynchronously? (or perhaps there's cyclic dependencies somewhere)

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/model/impl/DeleteOneCommandResolver.java
```diff
@@ -50,7 +51,10 @@ public Class<DeleteOneCommand> getCommandClass() {
   private FindOperation getFindOperation(CommandContext commandContext, DeleteOneCommand command) {
     List<DBFilterBase> filters = resolve(commandContext, command);
     final SortClause sortClause = command.sortClause();
-
+    if (sortClause != null && commandContext.embeddingService() != null) {
+      new VectorizeData(commandContext.embeddingService(), objectMapper.getNodeFactory())
```
tatu-at-datastax, line 55: as per earlier comments, I think `VectorizeData` name could be improved, as it's more like DataVectorizer or something.

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/model/impl/InsertManyCommandResolver.java
```diff
@@ -29,6 +33,11 @@ public Class<InsertManyCommand> getCommandClass() {
 
   @Override
   public Operation resolveCommand(CommandContext ctx, InsertManyCommand command) {
+    if (ctx.embeddingService() != null) {
+      new VectorizeData(ctx.embeddingService(), objectMapper.getNodeFactory())
```
tatu-at-datastax, line 37: One thought, seeing this pattern being repeated multiple times: maybe `CommandContext` could have a method like `tryVectorize(List<Document>)` since it has "embeddingService" (if any), and likely also `ObjectMapper`.
So call would be:

    ctxt.tryVectory(command.documents());

and could check existence (or lack thereof) of embedding service within method.

src/test/java/io/stargate/sgv2/jsonapi/api/model/command/deserializers/SortClauseDeserializerTest.java
```diff
@@ -131,6 +131,50 @@ public void vectorSearchInvalidSortClause() {
...
+        {
+         "$vectorize" : {}
+        }
+        """;
+
+      Throwable throwable = catchThrowable(() -> objectMapper.readValue(json, SortClause.class));
+
+      assertThat(throwable).isInstanceOf(JsonApiException.class);
+      assertThat(throwable.getMessage()).contains("$vectorize search needs to be text value");
+    }
```
tatu-at-datastax, line 176: Should also have failing test for case where there is non-vectorize sort clause in addition to vectorize one (assuming verification is done) -- unless it'd be part of one of follow-up issues.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/VectorizeSearchIntegrationTest.java
```diff
@@ -0,0 +1,820 @@
...
+      given()
+          .header(HttpConstants.AUTHENTICATION_TOKEN_HEADER_NAME, getAuthToken())
+          .contentType(ContentType.JSON)
+          .body(json)
+          .when()
+          .post(CollectionResource.BASE_PATH, namespaceName, collectionName)
+          .then()
+          .statusCode(200)
+          .body("data.documents[0]._id", is("1"))
+          .body("data.documents[0].$vector", is(notNullValue()))
```
tatu-at-datastax, line 114: Should probably also check that we get an array of floats (with known size but no need to match values)

src/test/java/io/stargate/sgv2/jsonapi/api/v1/VectorizeSearchIntegrationTest.java
```diff
@@ -0,0 +1,820 @@
...
+        .header(HttpConstants.AUTHENTICATION_TOKEN_HEADER_NAME, getAuthToken())
+        .contentType(ContentType.JSON)
+        .body(json)
+        .when()
+        .post(CollectionResource.BASE_PATH, namespaceName, collectionName)
+        .then()
+        .statusCode(200)
+        .body("errors", is(nullValue()))
+        .extract()
+        .path("status.moreData");
```
tatu-at-datastax, line 266: What does this test? (that is, is extract + path needed here?)

src/test/java/io/stargate/sgv2/jsonapi/service/embedding/operation/TestEmbeddingService.java
```diff
@@ -0,0 +1,48 @@
...
+  ;
+
+  @Override
+  public List<float[]> vectorize(List<String> texts) {
+    List<float[]> response = new ArrayList<>(texts.size());
+    for (String text : texts) {
+      if (TEST_DATA.containsKey(text)) {
+        response.add(TEST_DATA.get(text));
+      } else {
+        response.add(new float[] {0.25f, 0.25f, 0.25f});
```
tatu-at-datastax, line 43: Similar to earlier, not sure if it matters but size of default vector different from TEST_DATA entries.

src/test/java/io/stargate/sgv2/jsonapi/service/embedding/operation/EmbeddingServiceCacheTest.java
```diff
@@ -0,0 +1,95 @@
...
+                  embeddingServiceCache.getConfiguration(
+                      Optional.empty(), "vertexai", "vertexai-model"));
+      assertThat(failure)
+          .isInstanceOf(JsonApiException.class)
+          .hasFieldOrPropertyWithValue("errorCode", ErrorCode.VECTORIZE_SERVICE_TYPE_NOT_ENABLED)
+          .hasFieldOrPropertyWithValue("message", "Vectorize service type not enabled : vertexai");
+    }
+  }
+
+  public static class PropertyBasedOverrideProfile implements QuarkusTestProfile {
```
tatu-at-datastax, line 75: Since this is shared by multiple test classes, maybe extract as main level class?

tatu-at-datastax (approved):
Ok, looks good overall -- I added (quite) a few suggestions; feel free to use or ignore whatever makes (or doesn't make) sense.

src/main/java/io/stargate/sgv2/jsonapi/api/v1/CollectionResource.java
maheshrajamani: Yes, Embedding service is light weight object which is cached.

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/VectorizeData.java
maheshrajamani: Yes, in case of insert all, we collect all the $vectorize field from different document and call the embedding service as batch.

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/configuration/InMemoryEmbeddingServiceConfigStore.java
```diff
@@ -0,0 +1,27 @@
...
+
+import com.github.benmanes.caffeine.cache.Cache;
+import com.github.benmanes.caffeine.cache.Caffeine;
+import io.quarkus.arc.lookup.LookupIfProperty;
+import jakarta.enterprise.context.ApplicationScoped;
+import java.util.Optional;
+
+@LookupIfProperty(name = "stargate.jsonapi.embedding.config.store", stringValue = "in-memory")
+@ApplicationScoped
+public class InMemoryEmbeddingServiceConfigStore implements EmbeddingServiceConfigStore {
```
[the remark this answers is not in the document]
maheshrajamani: @tatu-at-datastax : I needed something like this to finish the `createEmbeddingService`. This will eventually this will change to some security manager store. That will be separate PR once we have details on how it's going to be hasted.

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/configuration/PropertyBasedEmbeddingServiceConfig.java
maheshrajamani: Changed this to  resolve to Optional<Class<?>>

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/HuggingFaceEmbeddingClient.java
maheshrajamani: Removed these constructor as they are defined as part of property.

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/test/TestEmbeddingService.java
maheshrajamani: This value is never used, for consistency will change this to return array of size 5.

src/main/java/io/stargate/sgv2/jsonapi/service/bridge/executor/NamespaceCache.java
maheshrajamani: Made the code to resolve the values using JsonNode, so the class need not be moved.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/CreateCollectionOperation.java
maheshrajamani: Yes, the value is stored part of table's comment.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/CreateCollectionOperation.java
maheshrajamani: Not sure what kind of characters to handle since this is just a text value, If needed I will add permitted character for the vectorize options.

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/model/impl/CreateEmbeddingServiceCommandResolver.java
maheshrajamani: Embedding Service store is created using @Produces based on the config property stargate.jsonap.embedding.config.store

src/test/java/io/stargate/sgv2/jsonapi/service/embedding/operation/VectorizeDataTest.java
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    @Test
+    public void updateClauseSetOnInsertValues() throws Exception {
+      String json =
+          """
+        {
+          "findOneAndUpdate": {
+            "filter" : {"_id" : "id"},
+            "update" : {"$setOnInsert" : {"$vectorize" : "New York"}}
```
Yuqi-Du, line 132: what is the difference between this $setOnInsert and upper $set

src/test/java/io/stargate/sgv2/jsonapi/api/model/command/deserializers/SortClauseDeserializerTest.java
maheshrajamani: Added additional test case

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/VectorizeData.java
Yuqi-Du: That is cool. OpenAI/Vertex/Hugging Face, they all provide batch calls?

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/EmbeddingService.java
```diff
@@ -0,0 +1,15 @@
...
+import java.util.List;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+
+public interface EmbeddingService {
+  int EXECUTOR_SERICE_THREADS = 10;
+  int MAX_RETRIES = 3;
+  ExecutorService executor =
+      Executors.newFixedThreadPool(
+          EXECUTOR_SERICE_THREADS, r -> new Thread(r, "EMBEDDING_SERVICE_REQUEST"));
```
Yuqi-Du, line 12: EXECUTOR_SERICE_THREADS and RETRIES are not used in implementation.
That is because we are still doing API call mocking? and wait for the api keys from the company.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/VectorizeSearchIntegrationTest.java
maheshrajamani: This is not test, this is for initial test data setup. It is used to delete data if there are more than paging size. Here we load only few data I have not paginated the delete.

src/test/java/io/stargate/sgv2/jsonapi/service/embedding/operation/VectorizeDataTest.java
maheshrajamani: When upsert option is set to true during update operation, we will create new document if no document found based on condition. setOnInsert will work when this a new document.

tatu-at-datastax: `$setOnInsert` is only applied in case a new document is inserted, but not if one exists.
Used for things like adding id field(s), or creation-time.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/VectorizeSearchIntegrationTest.java
tatu-at-datastax: Ok, my question was related to extraction of response and getting `status.moreData`, which is not used. So it seems unnecessary (last 2 lines). Not a big deal, was just wondering if there was some specific purpose.

PR #531, opened 2023-09-20
Fix update vector

maheshrajamani:
**What this PR does**:
Fix the issue where query_vector_value is not updated during updates.

**Checklist**
- [x] Changes manually tested


tatu-at-datastax (approved):
Looks good (although I am starting worry about duplication of code for vector/non-vector cases; and esp. if we needed to do something similar for tokenization/text-search).

But should there be an IT that verifies change actually takes effect?

PR #535, opened 2023-09-21
In support

Yuqi-Du:
**What this PR does**: add in operation for jsonapi

[Issue 291](https://github.com/stargate/jsonapi/issues/291)
[PR 348](https://github.com/stargate/jsonapi/pull/348)

**Checklist**
- [x] Changes manually tested


Yuqi-Du:
The current test fail is because of that querybuilder quarkus-common change has not been merged.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DBFilterBase.java
```diff
@@ -288,13 +285,85 @@ boolean canAddField() {
       }
     }
   }
+
+  /**
+   * Filters db documents based on non document id field values if id use "$in" operator, use
+   * IDFilter, since partition key 'or' is not supported besides id, "$in" operator, use INFilter
+   */
+  public static class INFilter extends DBFilterBase {
```
tatu-at-datastax, line 293: Capitalization should be `InFilter` (otherwise it'd be read as... "infinity filter"? :) )

src/test/java/io/stargate/sgv2/jsonapi/api/v1/FindIntegrationTest.java
```diff
@@ -387,11 +387,73 @@ public void inConditionNonIdField() {
...
+          "{\"_id\":\"doc1\", \"username\":\"user1\", \"active_user\":true, \"date\" : {\"$date\": 1672531200000}}";
+      given()
+          .header(HttpConstants.AUTHENTICATION_TOKEN_HEADER_NAME, getAuthToken())
+          .contentType(ContentType.JSON)
+          .body(json)
+          .when()
+          .post(CollectionResource.BASE_PATH, namespaceName, collectionName)
+          .then()
+          .statusCode(200)
+          .body("data.documents", hasSize(1))
```
vkarpov15, line 406: Would be great to have a test that tests we get multiple results when using `$in`. All of these tests assert that there's exactly 1 result, so they don't cover the full functionality of `$in`.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DBFilterBase.java
```diff
@@ -16,10 +16,7 @@
 import io.stargate.sgv2.jsonapi.service.shredding.model.DocumentId;
 import io.stargate.sgv2.jsonapi.util.JsonUtil;
 import java.math.BigDecimal;
-import java.util.Date;
-import java.util.List;
-import java.util.Map;
-import java.util.Objects;
+import java.util.*;
```
maheshrajamani, line 19: I was asked before by Ivan to keep the import statement expanded. Can we have explicit imports?

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DBFilterBase.java
```diff
@@ -288,13 +285,85 @@ boolean canAddField() {
...
+
   /**
    * DB filter / condition for testing a set value Note: we can only do CONTAINS until SAI indexes
-   * are updated
+   * are updated Now, cassandra supports or operation, we can add $in
    */
   public abstract static class SetFilterBase<T> extends DBFilterBase {
     public enum Operator {
-      CONTAINS;
+      CONTAINS,
```
maheshrajamani, line 366: Can we revert this?

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DBFilterBase.java
tatu-at-datastax: FWTW, we as a team are free to decide whether to expand or not these imports. There's no universal consensus on "Right" approach.

I know some devs want fully expanded; others not. I think IDEs default to expanded for up to 4 from same packages, then wildcard if there are 5 or more imports.
This is a compromise between verbosity and visibility.

I can work with either approach.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DBFilterBase.java
Yuqi-Du: yes

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DBFilterBase.java
Yuqi-Du: solved

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DBFilterBase.java
Yuqi-Du: solved

src/test/java/io/stargate/sgv2/jsonapi/api/v1/FindIntegrationTest.java
Yuqi-Du: thanks, added.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/FindOperation.java
```diff
@@ -381,34 +386,75 @@ public ReadDocument getNewDocument() {
...
-                    .build());
-          } else {
-            QueryOuterClass.Query builtQuery = getVectorSearchQuery(condition);
-            final List<QueryOuterClass.Value> valuesList =
-                builtQuery.getValuesOrBuilder().getValuesList();
-            final QueryOuterClass.Values.Builder builder = QueryOuterClass.Values.newBuilder();
-            valuesList.forEach(builder::addValues);
-            builder.addValues(CustomValueSerializers.getVectorValue(vector()));
-            queries.add(QueryOuterClass.Query.newBuilder(builtQuery).setValues(builder).build());
+    AtomicBoolean hasInFilterBesidesIdField = new AtomicBoolean(false);
```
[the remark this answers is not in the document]
Yuqi-Du: sure

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/FindOperation.java
```diff
@@ -381,34 +386,75 @@ public ReadDocument getNewDocument() {
...
+    filters.forEach(
+        filter -> {
+          if (filter instanceof DBFilterBase.INFilter) {
+            hasInFilterBesidesIdField.set(true);
+            return; // need to break
           }
         });
 
-    return queries;
+    if (hasInFilterBesidesIdField.get()) {
```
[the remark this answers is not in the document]
Yuqi-Du: sorry, Mahesh, after looking into it for a long time.
I do not think this is doable right now, since resolve() method returns List<DBFilterBase>, I need to iterate through the list.  I can move some codes forward into FindOneAndDeleteCommandResolver, but does not make a big difference.
Do you have some suggestions.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/FindIntegrationTest.java
```diff
@@ -409,6 +409,31 @@ public void inConditionNonIdField() {
...
+              """;
+      given()
+          .header(HttpConstants.AUTHENTICATION_TOKEN_HEADER_NAME, getAuthToken())
+          .contentType(ContentType.JSON)
+          .body(json)
+          .when()
+          .post(CollectionResource.BASE_PATH, namespaceName, collectionName)
+          .then()
+          .statusCode(200)
+          .body("data.documents", hasSize(2))
```
vkarpov15, line 432: Can we also test that the document ids are what we expect? Or at least test that usernames are "user1" and "user2"?

Yuqi-Du: sure. Thanks.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DBFilterBase.java
```diff
@@ -288,9 +288,81 @@ boolean canAddField() {
       }
     }
   }
+
+  /**
+   * Filters db documents based on non document id field values if id use "$in" operator, use
```
tatu-at-datastax, line 293: maybe "based on values of fields other than document id: for filtering on id field use IDFilter..."?

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DBFilterBase.java
```diff
@@ -288,9 +288,81 @@ boolean canAddField() {
...
+   * IDFilter, since partition key 'or' is not supported besides id, "$in" operator, use INFilter
+   */
+  public static class InFilter extends DBFilterBase {
+    private final Object arrayValue;
```
tatu-at-datastax, line 297: Can we not use anything more specific than `Object`

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DBFilterBase.java
```diff
@@ -288,9 +288,81 @@ boolean canAddField() {
...
+    JsonNode asJson(JsonNodeFactory nodeFactory) {
+      return DBFilterBase.getJsonNode(nodeFactory, arrayValue);
+    }
+
+    @Override
+    boolean canAddField() {
+      return false;
+    }
+
+    public enum Operator {
```
tatu-at-datastax, line 310: Will we have other Operator values at some point? Seems strange to define an Enum with just 1 value.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DBFilterBase.java
```diff
@@ -288,9 +288,81 @@ boolean canAddField() {
...
+      InFilter inFilter = (InFilter) o;
+      return operator == inFilter.operator && Objects.equals(arrayValue, inFilter.arrayValue);
+    }
+
+    @Override
+    public int hashCode() {
+      return Objects.hash(arrayValue, operator);
+    }
+
+    private static final String DATA_CONTAINS = "array_contains";
```
tatu-at-datastax, line 333: If this refers to name of DB column to use, I think we have a container class, `DocumentConstants` where we have added some column/field names. So perhaps add it there, instead of class private constant.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DBFilterBase.java
```diff
@@ -288,9 +288,81 @@ boolean canAddField() {
...
+
+    @Override
+    public BuiltCondition get() {
+      // For IN filter we always use getALL() method
```
tatu-at-datastax, line 337: If so, maybe throw an exception so that if this was accidentally called, we'd know (caller probably gets an NPE now)? Could use `UnsupportedOperationException` f.ex.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DBFilterBase.java
```diff
@@ -288,9 +288,81 @@ boolean canAddField() {
...
+        default:
+          throw new JsonApiException(
+              ErrorCode.UNSUPPORTED_FILTER_OPERATION,
+              String.format("Unsupported %s column operation %s", getPath(), operator));
+      }
+    }
+  }
+
   /**
    * DB filter / condition for testing a set value Note: we can only do CONTAINS until SAI indexes
```
tatu-at-datastax, line 364: Should first part be now removed?

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DBFilterBase.java
Yuqi-Du: sure!

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/FindOperation.java
```diff
@@ -472,7 +519,71 @@ private QueryOuterClass.Query getVectorSearchQuery(List<BuiltCondition> conditio
...
+              .from(commandContext.namespace(), commandContext.collection())
+              .where(expression)
+              .limit(limit)
+              .vsearch(DocumentConstants.Fields.VECTOR_SEARCH_INDEX_COLUMN_NAME)
+              .build();
+        }
+        default -> {
+          throw new JsonApiException(
+              ErrorCode.VECTOR_SEARCH_INVALID_FUCTION_NAME,
+              ErrorCode.VECTOR_SEARCH_INVALID_FUCTION_NAME.getMessage());
```
tatu-at-datastax, line 573: Append name of function in question; otherwise caller doesn't know offending value?

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DBFilterBase.java
Yuqi-Du: sure, I am following other filter's definition. Let me add some doc.

tatu-at-datastax: But those other filters probably have multiple values (or could conceivably get more)?
If there won't be others, I don't think separate type should be added.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DBFilterBase.java
Yuqi-Du: solved

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DBFilterBase.java
tatu-at-datastax: This class is weird, too (not fault of this PR :) ) -- `DBFilterBase` is a base class, yet there are nested sub-classes that implement it for some reason. That seems a bit wrong.
So maybe eventually these would be good to refactor out of containing parent class.
(not in this PR of course)

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DBFilterBase.java
Yuqi-Du: solved

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DBFilterBase.java
Yuqi-Du: solved

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/FindOperation.java
Yuqi-Du: solved

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/DBFilterBase.java
Yuqi-Du: solved, use List<Object>

maheshrajamani (approved):

PR #559, opened 2023-10-09
Return create collection options as part of findCollections response

### LINKED ISSUE #550: Provide a way to display collection options
Originally JSON API collections had no attributes, so the `findCollections` command was sufficient to list the names of the collections. With the addition of vector search, collections can now have attributes, so we need to provide a way to return them.

The easiest way to do this might be to extend `findCollections`:

- add an optional field to provide the name of a specific collection
- add an option to indicate if the attributes should be returned, and default to true.

The specification and docs will need to be updated accordingly

maheshrajamani:
**What this PR does**:
1) Refactor CollectionProperty as separate class so it can be used in find collections operation
2) Find Collections with explain changes.

Fixes #550 

**Checklist**
- [x] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/service/bridge/executor/CollectionProperty.java
```diff
@@ -0,0 +1,123 @@
...
+ * Refactored as seperate class that represent a collection property.
+ *
+ * @param collectionName
+ * @param vectorEnabled
+ * @param vectorSize
+ * @param similarityFunction
+ * @param vectorizeServiceName
+ * @param modelName
+ */
+public record CollectionProperty(
```
maheshrajamani, line 23: Refactored this as separate class so that it can be used from the create collection command.

tatu-at-datastax: Nice!

Minor nit: name itself could probably be improved a bit. Not sure if `CollectionSettings` would be a better name? (CollectionConfig possibly, but we use `-Config` for things configured with configuration files and this is more dynamic so `-Settings` seems better).

src/test/java/io/stargate/sgv2/jsonapi/api/v1/FindCollectionsIntegrationTest.java
```diff
@@ -68,11 +74,88 @@ public void happyPath() {
...
+                }
+                """
+              .formatted("collection2");
+
+      json =
+          """
+              {
+                "findCollections": {
+                  "options": {
+                    "explain" : true
```
vkarpov15, line 137: What happens if you don't pass `explain` here? Do you not get the `options` back?

vkarpov15 (approved):
LGTM, just one minor question

src/main/java/io/stargate/sgv2/jsonapi/service/bridge/executor/CollectionProperty.java
maheshrajamani: Done.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/FindCollectionsIntegrationTest.java
maheshrajamani: Yes., it will only return the collection names.

src/main/java/io/stargate/sgv2/jsonapi/service/bridge/executor/CollectionSettings.java
```diff
@@ -0,0 +1,123 @@
...
+ * @param collectionName
+ * @param vectorEnabled
+ * @param vectorSize
+ * @param similarityFunction
+ * @param vectorizeServiceName
+ * @param modelName
+ */
+public record CollectionSettings(
+    String collectionName,
+    Boolean vectorEnabled,
```
tatu-at-datastax, line 25: Can this be "unknown"? If not (which I think is the case), should perhaps be primitive `boolean` instead? Unless the reason is to allow exclusion of `null` Boolean value on serialization.

src/main/java/io/stargate/sgv2/jsonapi/service/bridge/executor/CollectionSettings.java
```diff
@@ -0,0 +1,123 @@
...
+                      i.getColumnName()
+                          .equals(DocumentConstants.Fields.VECTOR_SEARCH_INDEX_COLUMN_NAME))
+              .findFirst();
+      CollectionSettings.SimilarityFunction function = CollectionSettings.SimilarityFunction.COSINE;
+      if (vectorIndex.isPresent()) {
+
+        if (vectorIndex
+            .get()
+            .getOptionsMap()
+            .containsKey(DocumentConstants.Fields.VECTOR_INDEX_FUNCTION_NAME)) {
```
tatu-at-datastax, line 78: Instead of `containsKey()` first, `get()` then, could just do `get()` with null check?

src/main/java/io/stargate/sgv2/jsonapi/service/bridge/executor/CollectionSettings.java
```diff
@@ -0,0 +1,123 @@
...
+                      .get(DocumentConstants.Fields.VECTOR_INDEX_FUNCTION_NAME));
+        }
+      }
+      final String comment = table.getOptionsOrDefault("comment", null);
+      if (comment != null && !comment.isBlank()) {
+        try {
+          JsonNode vectorizeConfig = objectMapper.readTree(comment);
+          String vectorizeServiceName =
+              vectorizeConfig != null && vectorizeConfig.has("service")
+                  ? vectorizeConfig.get("service").textValue()
```
tatu-at-datastax, line 93: I think you can use `vectorizeConfig.path("service").textValue()`.

"has()" will return "MissingNode" if no value exist, not `null`, so can call `textValue()` -- which in turn returns `null` for non-Text `JsonNode` values.

src/main/java/io/stargate/sgv2/jsonapi/service/bridge/executor/CollectionSettings.java
```diff
@@ -0,0 +1,123 @@
...
+                  : null;
+          String modelName = null;
+          final JsonNode optionsNode =
+              vectorizeConfig != null && vectorizeConfig.has("options")
+                  ? vectorizeConfig.get("options")
```
tatu-at-datastax, line 98: And in here, just use `vectorizeConfig.get("options")`, will return `null` if property has no value.

src/main/java/io/stargate/sgv2/jsonapi/service/bridge/executor/CollectionSettings.java
```diff
@@ -0,0 +1,123 @@
...
+                      .getOptions()
+                      .get(DocumentConstants.Fields.VECTOR_INDEX_FUNCTION_NAME));
+        }
+      }
+      final String comment = table.getOptionsOrDefault("comment", null);
+      if (comment != null && !comment.isBlank()) {
+        try {
+          JsonNode vectorizeConfig = objectMapper.readTree(comment);
+          String vectorizeServiceName =
+              vectorizeConfig != null && vectorizeConfig.has("service")
```
tatu-at-datastax, line 92: I think `readTree()` should never return `null`; if content missing, should throw some parsing exception.
Because of this, none of null checks for `vectorizeConfig` should be needed.

If `null` could be returned, could use:

    if (vectorizeConfig == null) {
       vectorizeConfig = objectMapper.missingNode();
    }

src/main/java/io/stargate/sgv2/jsonapi/service/bridge/executor/CollectionSettings.java
```diff
@@ -0,0 +1,123 @@
...
+                  ? vectorizeConfig.get("service").textValue()
+                  : null;
+          String modelName = null;
+          final JsonNode optionsNode =
+              vectorizeConfig != null && vectorizeConfig.has("options")
+                  ? vectorizeConfig.get("options")
+                  : null;
+          if (optionsNode != null && optionsNode.has("modelName")) {
+            modelName = optionsNode.get("modelName").textValue();
```
tatu-at-datastax, line 101: Actually... even simpler,

    // Can pass String for JsonPointer
    modelName = vectorizeConfig.at("/options/modelName").textValue();

src/main/java/io/stargate/sgv2/jsonapi/service/bridge/executor/SchemaCache.java
```diff
@@ -19,7 +19,7 @@ public class SchemaCache {
   private final Cache<CacheKey, NamespaceCache> schemaCache =
       Caffeine.newBuilder().maximumSize(1000).build();
 
-  public Uni<NamespaceCache.CollectionProperty> getCollectionProperties(
+  public Uni<CollectionSettings> getCollectionProperties(
```
tatu-at-datastax, line 22: getCollectionSettings() ?

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/FindCollectionsOperation.java
```diff
@@ -58,23 +69,60 @@ public Uni<Supplier<CommandResult>> execute(QueryExecutor queryExecutor) {
...
-  private record Result(List<String> collections) implements Supplier<CommandResult> {
+  private record Result(boolean explain, List<CollectionSettings> collections)
+      implements Supplier<CommandResult> {
 
     @Override
     public CommandResult get() {
-      Map<CommandStatus, Object> statuses = Map.of(CommandStatus.EXISTING_COLLECTIONS, collections);
-      return new CommandResult(statuses);
+      if (explain) {
+        final List<CreateCollectionCommand> createCollectionCommands =
```
tatu-at-datastax, line 89: Add a comment explaining that "CreateCollectionCommand" created for convenience; we are not creating collections? (but want to expose same set of properties that we take when creating).

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/FindCollectionsOperation.java
```diff
@@ -58,23 +69,60 @@ public Uni<Supplier<CommandResult>> execute(QueryExecutor queryExecutor) {
...
+            collections.stream()
+                .map(
+                    collectionProperty -> {
+                      CreateCollectionCommand.Options options = null;
```
tatu-at-datastax, line 93: This code is a good example of where `record` type construction gets pretty hairy -- would be cleaner with Builder pattern. But that's not yet available with JDK so not much we can do.
Just mentioning :)

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/FindCollectionsOperation.java
```diff
@@ -25,7 +28,11 @@
  * @param commandContext {@link CommandContext}
  */
 public record FindCollectionsOperation(
-    SchemaManager schemaManager, JsonapiTableMatcher tableMatcher, CommandContext commandContext)
+    boolean explain,
```
tatu-at-datastax, line 31: Should add `@param` for `explain` in Javadoc above

src/test/java/io/stargate/sgv2/jsonapi/api/v1/FindCollectionsIntegrationTest.java
```diff
@@ -68,11 +74,88 @@ public void happyPath() {
...
+          .body("status.ok", is(1));
+
+      String expected1 =
+          """
+                  {
+                    "name": "%s"
+                    }
+                  }
+                    """
+              .formatted("collection1");
```
tatu-at-datastax, line 117: nit: looks like there's unbalanced curly braces? (1 opening, 2 closing)

tatu-at-datastax (approved):
LGTM -- added some suggestions for improvements but nothing super important.

src/main/java/io/stargate/sgv2/jsonapi/service/bridge/executor/CollectionSettings.java
maheshrajamani: "has()" is returning boolean.

tatu-at-datastax: Ah sorry, my bad! `path()`, NOT `has()`. That will return `MissingNode`.

PR #561, opened 2023-10-10
Vertex AI embedding client fix

maheshrajamani:
**What this PR does**:
Fixed the response data parsing for vertex AI.
Not added any test case because it needs API keys.
Tested manually.

**Checklist**
- [x] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/VertexAIEmbeddingClient.java
```diff
@@ -29,24 +31,86 @@ public interface VertexAIEmbeddingService {
...
+  private record EmbeddingRequest(List<Content> instances) {
     public record Content(String content) {}
+  }
+
+  private static class EmbeddingResponse {
+    public EmbeddingResponse() {}
+
+    private List<Prediction> predictions;
+
+    @JsonIgnore private Object metadata;
```
tatu-at-datastax, line 49: Is this to ignore `metadata` from serialization? If so makes sense; although can alternatively put next to `getMetadata()` for same effect (shouldn't matter, devs have different preferences so this is fine :) ).

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/VertexAIEmbeddingClient.java
```diff
@@ -29,24 +31,86 @@ public interface VertexAIEmbeddingService {
...
+
+    public Object getMetadata() {
+      return metadata;
+    }
+
+    public void setMetadata(Object metadata) {
+      this.metadata = metadata;
+    }
+
+    private static class Prediction {
```
tatu-at-datastax, line 67: Probably better to keep package protected (no modifier) if and when this needs to be serialized; Jackson can access private classes too, but has to force access and there's (I think) minor performance penalty (not enough to be measurable probably tho).

tatu-at-datastax (approved):

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/VertexAIEmbeddingClient.java
maheshrajamani: The response from vertexAI has data other than embedding value which are not used. So added JsonIgnore so these data are not deserialized.

PR #565, opened 2023-10-11
Support and or

Yuqi-Du:
**What this PR does**:
support $and and $or operator 
fully rely on expression to support logical operations in filter

Fixes [547](https://github.com/stargate/jsonapi/issues/547) [543](https://github.com/stargate/jsonapi/issues/543)

**Checklist**
- [x] Changes manually tested


Yuqi-Du:
All integration tests and unit tests have passed in my local, will need stargate v2.1 release to passed the CI of github action.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/FindIntegrationTest.java
```diff
@@ -1269,6 +1270,134 @@ public void byDateColumn() {
...
+    }
+
+    @Test
+    public void nestedAndOr() {
+      String json =
+          """
+{
+    "find": {
+        "filter": {
+            "$or": [
```
vkarpov15, line 1319: I'd recommend making this test use a `$and` of `$or`, rather than a `$or` of `$and`. That's the most common use case for `$and`, because there's no difference between `$and: [{ username: 'user3' }, { 'subdoc.id': 'abc' }]` and `{ username: 'user3', 'subdoc.id': 'abc' }`. You would typically use `$and` like this:

```javascript
{
  $and: [
    { $or: [{ username: 'user1' }, { 'subdoc.id': 'test1' }] },
    { $or: [{ username: 'user2' }, { 'subdoc.id': 'test2' }] }
  ]
}
```

Yuqi-Du: thank, Val. Fixed

src/test/java/io/stargate/sgv2/jsonapi/api/v1/FindIntegrationTest.java
```diff
@@ -1269,6 +1270,143 @@ public void byDateColumn() {
...
+          .header(HttpConstants.AUTHENTICATION_TOKEN_HEADER_NAME, getAuthToken())
+          .contentType(ContentType.JSON)
+          .body(json)
+          .when()
+          .post(CollectionResource.BASE_PATH, namespaceName, collectionName)
+          .then()
+          .statusCode(200)
+          .body("status", is(nullValue()))
+          .body("errors", is(nullValue()))
+          .body("data.documents", hasSize(1));
```
vkarpov15, line 1369: Can we test that the document is equal to the expected result?

Yuqi-Du: solved, thanks

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/filter/ComparisonExpression.java
```diff
@@ -19,9 +17,29 @@
  * {"username" : {"$eq" : "aaron"}} In here we expand the shortcut into a canonical long form, so it
  * is all the same.
  */
-public record ComparisonExpression(
-    @NotBlank(message = "json node path can not be null in filter") String path,
-    @Valid @NotEmpty List<FilterOperation<?>> filterOperations) {
+public class ComparisonExpression {
+
+  @NotBlank(message = "json node path can not be null in filter")
```
tatu-at-datastax, line 22: If changing record into class, you probably need to move validation annotations to constructor parameters for them to work, due to differences in how records work compared to regular POJOs.
Not sure if we have tests verifying that validation occurs or not, that should tell whether using annotations on fields work.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/filter/LogicalExpression.java
```diff
@@ -0,0 +1,95 @@
...
+    return new LogicalExpression(LogicalOperator.AND, 0, new ArrayList<>(), new ArrayList<>());
+  }
+
+  public static LogicalExpression or() {
+    return new LogicalExpression(LogicalOperator.OR, 0, new ArrayList<>(), new ArrayList<>());
+  }
+
+  public void addLogicalExpression(LogicalExpression logicalExpression) {
+    // skip empty logic expression
+    if (logicalExpression.logicalExpressions.isEmpty()
```
tatu-at-datastax, line 56: Maybe add `isEmpty()` method in this class so there's no need to reference internal fields from here?

So something like

```
public boolean isEmpty() {
   return logicalExpressions.isEmpty();
}
```

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/deserializers/FilterClauseDeserializer.java
```diff
@@ -50,42 +41,114 @@ public FilterClause deserialize(
...
+      LogicalExpression.LogicalOperator fromLogicalRelation) {
+    if (fromLogicalRelation.equals(LogicalExpression.LogicalOperator.OR)
+        && path.equals(DocumentConstants.Fields.DOC_ID)) {
       throw new JsonApiException(
           ErrorCode.INVALID_FILTER_EXPRESSION,
           String.format(
-              "Cannot filter on '%s' field using operator '%s': only '$exists' is supported",
-              DocumentConstants.Fields.VECTOR_EMBEDDING_FIELD,
-              filterOperation.operator().getOperator()));
+              "Cannot filter on '%s' field within '%s'",
```
tatu-at-datastax, line 148: Should "within" be "with" instead?

src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCode.java
```diff
@@ -21,6 +21,9 @@ public enum ErrorCode {
 
   FILTER_UNRESOLVABLE("Unable to resolve the filter"),
 
+  FILTER_MULTIPLE_ID_FILTER(
+      "Should be just have one _id filter, key cannot be restricted by more than one relation if it includes an Equal"),
```
tatu-at-datastax, line 25: Need to reword "be just have one"; maybe "Should only have one _id filter, document id cannot" (I think we refer to 'document id' everywhere, not 'key').

maheshrajamani (approved):
LGTM., Can you make changes wrt @tatu-at-datastax comments?

tatu-at-datastax (approved):
Great changes, LGTM!

Yuqi-Du:
@maheshrajamani  Can I merge it now?

src/test/java/io/stargate/sgv2/jsonapi/service/resolver/model/impl/DeleteOneCommandResolverTest.java
```diff
@@ -1,294 +1,324 @@
-package io.stargate.sgv2.jsonapi.service.resolver.model.impl;
```
tatu-at-datastax, line 1: @Yuqi-Du This change was hidden ("large diffs not displayed by default")

PR #566, opened 2023-10-12
Add error message in createCollection()

### LINKED ISSUE #546: Handling of createCollection command
Recreating table different vector option which already exist. This should error out.

Hazel-Datastax:
**What this PR does**:
Introduces error messaging for scenarios where a user attempts to create a collection with a name that already exists, but with differing settings. The error is triggered in two specific cases:
1. When an existing collection is non-vector, but the user attempts to create a vector-based collection with the same name.
2. When the existing vector collection's settings do not align with the vector options specified by the user.

Fixes #546 

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCode.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCode.java
@@ -109,7 +109,11 @@ public enum ErrorCode {
   VECTORIZE_SERVICE_TYPE_UNSUPPORTED("Vectorize service type unsupported : "),
 
   VECTORIZE_SERVICE_TYPE_UNAVAILABLE("Vectorize service unavailable : "),
-  VECTORIZE_USAGE_ERROR("Vectorize search can't be used with other sort clause");
+  VECTORIZE_USAGE_ERROR("Vectorize search can't be used with other sort clause"),
+
+  VECTORIZECONFIG_CHECK_FAIL("VectorizeConfig check null fail");
+
+
 
   private final String message;
 
--- a/src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCode.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCode.java
@@ -111,7 +111,7 @@ public enum ErrorCode {
   VECTORIZE_SERVICE_TYPE_UNAVAILABLE("Vectorize service unavailable : "),
   VECTORIZE_USAGE_ERROR("Vectorize search can't be used with other sort clause"),
 
-  VECTORIZECONFIG_CHECK_FAIL("VectorizeConfig check null fail");
+  VECTORIZECONFIG_CHECK_FAIL("Internal server error: VectorizeConfig check fail");
 
   private final String message;
 
--- a/src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCode.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCode.java
@@ -113,8 +113,6 @@ public enum ErrorCode {
 
   VECTORIZECONFIG_CHECK_FAIL("VectorizeConfig check null fail");
 
-
-
   private final String message;
 
   ErrorCode(String message) {
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/bridge/executor/CollectionSettings.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/bridge/executor/CollectionSettings.java
@@ -81,17 +81,8 @@ public record CollectionSettings(
       }
       final String comment = table.getOptionsOrDefault("comment", null);
       if (comment != null && !comment.isBlank()) {
-        try {
-          JsonNode vectorizeConfig = objectMapper.readTree(comment);
-          String vectorizeServiceName = vectorizeConfig.get("service").textValue();
-          final JsonNode optionsNode = vectorizeConfig.get("options");
-          String modelName = optionsNode.get("modelName").textValue();
-          return new CollectionSettings(
-              collectionName, vectorEnabled, vectorSize, function, vectorizeServiceName, modelName);
-        } catch (JsonProcessingException e) {
-          // This should never happen
-          throw new RuntimeException(e);
-        }
+        return createCollectionSettingsFromJson(
+            collectionName, vectorEnabled, vectorSize, function, comment, objectMapper);
       } else {
         return new CollectionSettings(
             collectionName, vectorEnabled, vectorSize, function, null, null);
@@ -116,25 +107,31 @@ public record CollectionSettings(
       ObjectMapper objectMapper) {
     // parse vectorize to get vectorizeServiceName and modelName
     if (vectorize != null && !vectorize.isBlank()) {
-      try {
-        JsonNode vectorizeConfig = objectMapper.readTree(vectorize);
-        String vectorizeServiceName_ = vectorizeConfig.get("service").textValue();
-        final JsonNode optionsNode = vectorizeConfig.get("options");
-        String modelName_ = optionsNode.get("modelName").textValue();
-        return new CollectionSettings(
-            collectionName,
-            vectorEnabled,
-            vectorSize,
-            similarityFunction,
-            vectorizeServiceName_,
-            modelName_);
-      } catch (JsonProcessingException e) {
-        // This should never happen
-        throw new RuntimeException(e);
-      }
+      return createCollectionSettingsFromJson(
+          collectionName, vectorEnabled, vectorSize, similarityFunction, vectorize, objectMapper);
     } else {
       return new CollectionSettings(
           collectionName, vectorEnabled, vectorSize, similarityFunction, null, null);
+    }
+  }
+
+  private static CollectionSettings createCollectionSettingsFromJson(
+      String collectionName,
+      boolean vectorEnabled,
+      int vectorSize,
+      SimilarityFunction function,
+      String vectorize,
+      ObjectMapper objectMapper) {
+    try {
+      JsonNode vectorizeConfig = objectMapper.readTree(vectorize);
+      String vectorizeServiceName = vectorizeConfig.get("service").textValue();
+      JsonNode optionsNode = vectorizeConfig.get("options");
+      String modelName = optionsNode.get("modelName").textValue();
+      return new CollectionSettings(
+          collectionName, vectorEnabled, vectorSize, function, vectorizeServiceName, modelName);
+    } catch (JsonProcessingException e) {
+      // This should never happen
+      throw new RuntimeException(e);
     }
   }
 }
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/bridge/executor/CollectionSettings.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/bridge/executor/CollectionSettings.java
@@ -1,5 +1,7 @@
 package io.stargate.sgv2.jsonapi.service.bridge.executor;
 
+import static io.stargate.sgv2.jsonapi.exception.ErrorCode.VECTORIZECONFIG_CHECK_FAIL;
+
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
@@ -9,8 +11,6 @@ import io.stargate.sgv2.jsonapi.config.constants.DocumentConstants;
 import io.stargate.sgv2.jsonapi.exception.ErrorCode;
 import io.stargate.sgv2.jsonapi.exception.JsonApiException;
 import java.util.Optional;
-
-import static io.stargate.sgv2.jsonapi.exception.ErrorCode.VECTORIZECONFIG_CHECK_FAIL;
 
 /**
  * Refactored as seperate class that represent a collection property.
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/bridge/executor/CollectionSettings.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/bridge/executor/CollectionSettings.java
@@ -124,14 +124,21 @@ public record CollectionSettings(
       ObjectMapper objectMapper) {
     try {
       JsonNode vectorizeConfig = objectMapper.readTree(vectorize);
-      String vectorizeServiceName = vectorizeConfig.get("service").textValue();
-      JsonNode optionsNode = vectorizeConfig.get("options");
-      String modelName = optionsNode.get("modelName").textValue();
-      return new CollectionSettings(
-          collectionName, vectorEnabled, vectorSize, function, vectorizeServiceName, modelName);
+      String vectorizeServiceName = vectorizeConfig.path("service").textValue();
+      JsonNode optionsNode = vectorizeConfig.path("options");
+      String modelName = optionsNode.path("modelName").textValue();
+      if (vectorizeServiceName != null
+          && !vectorizeServiceName.isEmpty()
+          && modelName != null
+          && !modelName.isEmpty()) {
+        return new CollectionSettings(
+            collectionName, vectorEnabled, vectorSize, function, vectorizeServiceName, modelName);
+      } else {
+        throw new RuntimeException("Invalid json string");
+      }
     } catch (JsonProcessingException e) {
       // This should never happen
-      throw new RuntimeException(e);
+      throw new RuntimeException("Invalid json string");
     }
   }
 }
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/bridge/executor/CollectionSettings.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/bridge/executor/CollectionSettings.java
@@ -115,21 +115,26 @@ public record CollectionSettings(
       String vectorize,
       ObjectMapper objectMapper) {
     // parse vectorize to get vectorizeServiceName and modelName
-    try {
-      JsonNode vectorizeConfig = objectMapper.readTree(vectorize);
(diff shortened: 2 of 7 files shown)


src/main/java/io/stargate/sgv2/jsonapi/service/bridge/executor/CollectionSettings.java
```diff
@@ -106,4 +97,41 @@ public static CollectionSettings getCollectionSettings(
...
+      ObjectMapper objectMapper) {
+    try {
+      JsonNode vectorizeConfig = objectMapper.readTree(vectorize);
+      String vectorizeServiceName = vectorizeConfig.get("service").textValue();
+      JsonNode optionsNode = vectorizeConfig.get("options");
+      String modelName = optionsNode.get("modelName").textValue();
+      return new CollectionSettings(
+          collectionName, vectorEnabled, vectorSize, function, vectorizeServiceName, modelName);
+    } catch (JsonProcessingException e) {
+      // This should never happen
```
tatu-at-datastax, line 133: It would happen if String that is supposed to contain JSON was corrupted and isn't valid JSON, so should probably throw some more specific exception that indicates invalid JSON?

src/main/java/io/stargate/sgv2/jsonapi/service/bridge/executor/CollectionSettings.java
```diff
@@ -106,4 +97,41 @@ public static CollectionSettings getCollectionSettings(
...
+  private static CollectionSettings createCollectionSettingsFromJson(
+      String collectionName,
+      boolean vectorEnabled,
+      int vectorSize,
+      SimilarityFunction function,
+      String vectorize,
+      ObjectMapper objectMapper) {
+    try {
+      JsonNode vectorizeConfig = objectMapper.readTree(vectorize);
+      String vectorizeServiceName = vectorizeConfig.get("service").textValue();
```
tatu-at-datastax, line 127: If there is no `service`, this would throw `NullPointerException` as `get()` returns `null`.

But if you use `vectorizeConfig.path("service")` it will return `MissingNode`, not `null`. So I recommend changing that.

src/main/java/io/stargate/sgv2/jsonapi/service/bridge/executor/CollectionSettings.java
```diff
@@ -106,4 +97,41 @@ public static CollectionSettings getCollectionSettings(
...
+      JsonNode optionsNode = vectorizeConfig.get("options");
+      String modelName = optionsNode.get("modelName").textValue();
```
tatu-at-datastax, line 129: Same here, `get()` -> `path()`.

May also want to validate that values are non-empty Strings; unless it is validated somewhere later.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/CreateCollectionOperation.java
```diff
@@ -1,41 +1,126 @@
...
     boolean vectorSearch,
     int vectorSize,
     String vectorFunction,
     String vectorize)
     implements Operation {
 
+  private static final Function<String, Uni<? extends Schema.CqlKeyspaceDescribe>>
+      MISSING_KEYSPACE_FUNCTION =
+          keyspace -> {
+            String message = "Unknown namespace %s, you must create it first.".formatted(keyspace);
```
tatu-at-datastax, line 34: Add single quotes around `%s`,following convention for error messages (quoting collection and namespace names)

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/CreateCollectionOperation.java
```diff
@@ -1,41 +1,126 @@
...
   }
 
   @Override
   public Uni<Supplier<CommandResult>> execute(QueryExecutor queryExecutor) {
+    return schemaManager
+        .getTable(commandContext.namespace(), name, MISSING_KEYSPACE_FUNCTION)
+        .onItemOrFailure()
+        .transformToUni(
+            (table, error) -> {
+              // if getTable return null, continue
```
tatu-at-datastax, line 74: I am not sure comment explain logic here... looks like comment and check for `error` not being null are unrelated?

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/CreateCollectionOperation.java
```diff
@@ -1,41 +1,126 @@
...
+              if (error != null) {
+                return executeQuery(queryExecutor);
```
tatu-at-datastax, line 76: So.... if we have an error, why/what are we executing then? Should we check what kind of `error` we have, and take action based on that, or fail?

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/CreateCollectionOperation.java
```diff
@@ -1,41 +1,126 @@
...
+              }
+              // table doesn't exist
+              if (table == null) {
+                return executeQuery(queryExecutor);
```
tatu-at-datastax, line 80: Ok so we will proceed with table creation if `table` does not exist.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/CreateCollectionOperation.java
```diff
@@ -1,41 +1,126 @@
...
+                              ErrorCode.INVALID_COLLECTION_NAME,
+                              "The provided collection already exists with a non-vector setting"));
+                }
+              }
+              // if table exists and user want to create a non-vector collection, continue
+              return executeQuery(queryExecutor);
+            });
+  }
+
+  private Uni<Supplier<CommandResult>> executeQuery(QueryExecutor queryExecutor) {
```
tatu-at-datastax, line 123: If this is not just generic query execution, maybe name differently: this tries to do the actual table creation, right? If so, something like `executeTableCreation()` or somehting.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/CreateCollectionIntegrationTest.java
```diff
@@ -47,6 +48,152 @@ public void happyPath() {
           .statusCode(200)
           .body("status.ok", is(1));
     }
+
+    @Test
+    public void duplicateCollectionName() {
```
tatu-at-datastax, line 53: I think this test method is getting long and should be split into multiple separate test methods, to keep it more readable.

src/main/java/io/stargate/sgv2/jsonapi/service/bridge/executor/CollectionSettings.java
```diff
@@ -106,4 +97,48 @@ public static CollectionSettings getCollectionSettings(
...
+          && modelName != null
+          && !modelName.isEmpty()) {
+        return new CollectionSettings(
+            collectionName, vectorEnabled, vectorSize, function, vectorizeServiceName, modelName);
+      } else {
+        throw new RuntimeException("Invalid json string");
+      }
+    } catch (JsonProcessingException e) {
+      // This should never happen
+      throw new RuntimeException("Invalid json string");
```
tatu-at-datastax, line 141: This needs to include actual reason: add message from `e` for example, or chain it (add `e` as second argument to `RuntimeException` constructor). Otherwise there is no clue as to what is causing the exception.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/CreateCollectionOperation.java
```diff
@@ -1,41 +1,127 @@
...
+      String name) {
+    return new CreateCollectionOperation(
+        commandContext, objectMapper, schemaManager, name, false, 0, null, null);
   }
 
   @Override
   public Uni<Supplier<CommandResult>> execute(QueryExecutor queryExecutor) {
+    return schemaManager
+        .getTable(commandContext.namespace(), name, MISSING_KEYSPACE_FUNCTION)
+        .onItemOrFailure()
```
tatu-at-datastax, line 72: Maybe it would be better to separate `onItem` and `onFailure` cases? Failure can never proceed with actual operation, only needs to surface failure in question.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/CreateCollectionOperation.java
```diff
@@ -1,41 +1,127 @@
...
+        .transformToUni(
+            (table, error) -> {
+              // if getTable return null, continue
+              if (error != null) {
+                return executeCollectionCreation(queryExecutor);
+              }
+              // table doesn't exist
+              if (table == null) {
```
tatu-at-datastax, line 80: Do we have existing Integration Tests for missing (misspelled etc) case? If not, would need to add... but I think ones probably exist.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/CreateCollectionOperation.java
tatu-at-datastax: Based on our discussion, this is necessary to work around problem with a unit test, mocking.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/CreateCollectionOperation.java
tatu-at-datastax: Same as above, problem case is one wrt unit tests which fails (due to missing mock for `getTable()`), and as a work-around, failing case (where `table` would be `null`) is taken to "succeed", just to test passes.
Resolving, to allow for figuring out better handling.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/CreateCollectionIntegrationTest.java
```diff
@@ -47,6 +48,152 @@ public void happyPath() {
...
+          .header(HttpConstants.AUTHENTICATION_TOKEN_HEADER_NAME, getAuthToken())
+          .contentType(ContentType.JSON)
+          .body(createVectorCollection)
+          .when()
+          .post(NamespaceResource.BASE_PATH, namespaceName)
+          .then()
+          .statusCode(200)
+          .body("status.ok", is(1));
+
+      // create a non vector collection with the same name
```
vkarpov15, line 169: This seems a bit strange: I would expect `createCollection()` would fail if calling `createCollection()` with no options, but there already is a collection with different options.

FWIW, MongoDB always fails if you call `createCollection()` on a collection that already exists, even if with same option. Below is a Mongoose example:

```
'use strict';

const mongoose = require('mongoose');

run().catch(err => console.log(err));

async function run() {
  await mongoose.connect('mongodb://127.0.0.1:27017/mongoose_test');

  await mongoose.connection.dropCollection('foobar').catch(() => {});
  await mongoose.connection.createCollection('foobar', { capped: true, size: 1024 });
  // Throws "Collection already exists. NS: mongoose_test.foobar"
  await mongoose.connection.createCollection('foobar', {});
  console.log('Done', await mongoose.connection.db.listCollections().toArray());
}
```

tatu-at-datastax: I think this is not what user would expect -- if they want non-vector Collection, service should not quietly just leave existing vector Collection in place and proceed.
It should indicate an error.

maheshrajamani: I think we should error if the configuration are different. It doesn't matter if existing collection is vector enabled or not. Also this will be helpful for future enhancement like upsert support for vector collection. 
If a collection is created for vector and user want to use it as non vector collection, they can still use it but configuration can't be changed.

Hazel-Datastax: Got it, thank you for pointing it out! Solved now!

src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCode.java
```diff
@@ -109,7 +109,9 @@ public enum ErrorCode {
   VECTORIZE_SERVICE_TYPE_UNSUPPORTED("Vectorize service type unsupported : "),
 
   VECTORIZE_SERVICE_TYPE_UNAVAILABLE("Vectorize service unavailable : "),
-  VECTORIZE_USAGE_ERROR("Vectorize search can't be used with other sort clause");
+  VECTORIZE_USAGE_ERROR("Vectorize search can't be used with other sort clause"),
+
+  VECTORIZECONFIG_CHECK_FAIL("VectorizeConfig check null fail");
```
tatu-at-datastax, line 114: This is a pretty bad failure message to give to user -- what can they do with it? It refers to some internal details that caller has no access to or knowledge of.

We may need to instead indicate some sort of "internal processing failure", related to bad configuration for vectorization?

src/main/java/io/stargate/sgv2/jsonapi/service/bridge/executor/CollectionSettings.java
```diff
@@ -106,4 +99,49 @@ public static CollectionSettings getCollectionSettings(
...
+      String modelName = optionsNode.path("modelName").textValue();
+      if (vectorizeServiceName != null
+          && !vectorizeServiceName.isEmpty()
+          && modelName != null
+          && !modelName.isEmpty()) {
+        return new CollectionSettings(
+            collectionName, vectorEnabled, vectorSize, function, vectorizeServiceName, modelName);
+      } else {
+        // This should never happen, VectorizeConfig check null, unless it fails
+        throw new JsonApiException(VECTORIZECONFIG_CHECK_FAIL);
```
tatu-at-datastax, line 140: Instead of that hard-coded message, should probably say something like "Vectorization configuration corrupt: missing 'service' entry" (or such).

src/main/java/io/stargate/sgv2/jsonapi/service/bridge/executor/CollectionSettings.java
```diff
@@ -106,4 +99,49 @@ public static CollectionSettings getCollectionSettings(
...
+      }
+    } catch (JsonProcessingException e) {
+      // This should never happen, already check if vectorize is a valid JSON
+      throw new RuntimeException("Invalid json string", e);
```
tatu-at-datastax, line 144: Should indicate what JSON content was problematic: vector(ization?) config.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/CreateCollectionOperation.java
```diff
@@ -1,41 +1,134 @@
...
+                      CollectionSettings.SimilarityFunction.fromString(vectorFunction),
+                      vectorize,
+                      objectMapper);
+              // if table exists and user want to create a vector collection with the same name
+              if (vectorSearch) {
+                // if existing collection is a vector collection
+                if (collectionSettings.vectorEnabled()) {
+                  if (collectionSettings.equals(collectionSettings_cur)) {
+                    // if settings are equal, no error
+                    return executeCollectionCreation(queryExecutor);
```
tatu-at-datastax, line 98: Fine not to error out, but can't we just return ok/good at this point since no creation is needed? (table exists with same config).
Is there a query to be executed? Or is `executeCollectionCreation()` just misleading name and it will do nothing?

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/CreateCollectionOperation.java
```diff
@@ -1,41 +1,134 @@
...
+                  } else {
+                    // if settings are not equal, error out
+                    return Uni.createFrom()
+                        .failure(
+                            new JsonApiException(
+                                ErrorCode.INVALID_COLLECTION_NAME,
+                                "The provided collection already exists with a different vector setting"));
```
tatu-at-datastax, line 105: Should probably include name of collection, recreation of which is attempted.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/CreateCollectionOperation.java
Hazel-Datastax: The query executed in `executeCollectionCreation()` is `CREATE TABLE IF NOT EXISTS`. Since the table already exists, the database will do nothing and return 'ok', and we will return 'ok' status to the user as well. Would you prefer that we directly return 'ok' here instead of interacting with the database? 

I am not sure about this situation: If the database is experiencing issues, bypassing the database interaction would prevent us from receiving any error messages. This could create a misleading situation where the user believes they have successfully created a collection, but may encounter problems when trying to interact with it later.

tatu-at-datastax: Good question. But when attempting to use "CREATE TABLE IF NOT EXISTS", I am not sure interaction occurs, beyond checking DB exists (which we just did).
So in that sense I think it would be best to just return ok directly.

@maheshrajamani @amorton WDYT?

This is something we can discuss on team daily meeting, to resolve it quickly.

maheshrajamani: @tatu-at-datastax Wanted to run it again so in case any index is missed it will be created. This can happen if index failed because of guardrail and user decides to drop a different collection and create this again.

tatu-at-datastax (approved):
LGTM

Hazel-Datastax:
Solved #546

PR #582, opened 2023-10-24
Cache changes

kathirsvn:
**What this PR does**:
CQL session caching changes

**Checklist**
- [X] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/config/OperationsConfig.java
```diff
@@ -109,4 +113,41 @@ interface LwtConfig {
     @WithDefault("3")
     int retries();
   }
+
+  /** Cassandra/AstraDB related configurations. */
+  @NotNull
+  @Valid
+  DatabaseConfig databaseConfig();
+
+  interface DatabaseConfig {
```
tatu-at-datastax, line 122: Does this need to be a nested inner class as opposed to main level class?

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/CQLSessionCache.java
```diff
@@ -1,3 +1,104 @@
...
+@ApplicationScoped
+public class CQLSessionCache {
+  /** Configuration for the JSON API operations. */
+  @Inject OperationsConfig operationsConfig;
+
+  /** Stargate request info. */
+  @Inject StargateRequestInfo stargateRequestInfo;
+
+  /** Time to live for CQLSession in cache in seconds. */
+  private static final long CACHE_TTL_SECONDS = 300;
```
tatu-at-datastax, line 31: Since this is time after last use (access) -- I think? -- would this be typical value to use? Or should it be lower, like 60 seconds? (to reduce number of idle sessions).

PR #591, opened 2023-10-25
provide additional detail on swagger docs for insertMany

jeffreyscarpenter:
During testing some users have become confused by the behavior of the `insertMany` command with respect to how partial failure is handled. Users don't realize that setting the `ordered` option to false will allow them to bypass the default fail fast behavior, which returns on the first insert failure, even a failure as simple as inserting a duplicate `_id`. 

Instead of referencing "specifications" that most users will not know how to find, this PR updates the swagger docs to pull in the detail from the spec. 

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/InsertManyCommand.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/InsertManyCommand.java
@@ -36,7 +36,7 @@ public record InsertManyCommand(
   public record Options(
       @Schema(
               description =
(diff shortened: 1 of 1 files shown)


jeffreyscarpenter:
As far as I could tell, this was the only place in swagger docs that references the "specification" so hopefully we will not have other issues like this.

tatu-at-datastax (approved):

tatu-at-datastax:
Really good improvement, let's do it!

vkarpov15 (commented):
LGTM :+1:

PR #595, opened 2023-10-26
Verify max 5 collections creation (per namespace)

### LINKED ISSUE #577: Enforce limit of maximum 5 Collections per database
Due to limits for maximum SAIs, we can only create up to 5 Collections per tenant/Database (note: not per namespace): this is to prevent failures on index creation during collection creation.

But currently we do not enforce this limit. We should.

tatu-at-datastax:
**What this PR does**:

Adds validation that no more than 5 (configurable) Collections are created per Namespace.

Fixes #577

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCode.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCode.java
@@ -59,6 +59,8 @@ public enum ErrorCode {
 
   INVALID_COLLECTION_NAME("Invalid collection name "),
 
+  TOO_MANY_COLLECTIONS("Too many collections"),
+
   UNSUPPORTED_FILTER_DATA_TYPE("Unsupported filter data type"),
 
   UNSUPPORTED_FILTER_OPERATION("Unsupported filter operator"),
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/CreateCollectionOperation.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/CreateCollectionOperation.java
@@ -7,6 +7,7 @@ import io.stargate.bridge.proto.Schema;
 import io.stargate.sgv2.api.common.schema.SchemaManager;
 import io.stargate.sgv2.jsonapi.api.model.command.CommandContext;
 import io.stargate.sgv2.jsonapi.api.model.command.CommandResult;
+import io.stargate.sgv2.jsonapi.config.DatabaseLimitsConfig;
 import io.stargate.sgv2.jsonapi.exception.ErrorCode;
 import io.stargate.sgv2.jsonapi.exception.JsonApiException;
 import io.stargate.sgv2.jsonapi.service.bridge.executor.CollectionSettings;
@@ -19,6 +20,7 @@ import java.util.function.Supplier;
 
 public record CreateCollectionOperation(
     CommandContext commandContext,
+    DatabaseLimitsConfig dbLimitsConfig,
     ObjectMapper objectMapper,
     SchemaManager schemaManager,
     String name,
@@ -40,6 +42,7 @@ public record CreateCollectionOperation(
 
   public static CreateCollectionOperation withVectorSearch(
       CommandContext commandContext,
+      DatabaseLimitsConfig dbLimitsConfig,
       ObjectMapper objectMapper,
       SchemaManager schemaManager,
       String name,
@@ -48,6 +51,7 @@ public record CreateCollectionOperation(
       String vectorize) {
     return new CreateCollectionOperation(
         commandContext,
+        dbLimitsConfig,
         objectMapper,
         schemaManager,
         name,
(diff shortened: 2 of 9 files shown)


src/main/resources/application.yaml
```diff
@@ -2,6 +2,9 @@
 # please keep properties in the alphabetical order
 
 stargate:
+  database:
+    limits:
+      maxCollections: 5
```
tatu-at-datastax, line 7: Not necessarily strictly speaking (Config class has defaults) but added as documentation.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/AbstractNamespaceIntegrationTestBase.java
```diff
@@ -38,6 +38,10 @@ public static void enableLog() {
 
   @BeforeAll
   public void createNamespace() {
+    createNamespace(namespaceName);
+  }
+
+  protected void createNamespace(String nsToCreate) {
```
tatu-at-datastax, line 44: Refactored to be usable by IT sub-classes.

src/main/resources/application.yaml
```diff
@@ -3,6 +3,10 @@
 
 stargate:
 
+  database:
+    limits:
+      max-collections: 5
```
tatu-at-datastax, line 8: Not 100% required since `DatabaseLimitsConfig` has defaults, but shown as documentation

src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCode.java
```diff
@@ -59,6 +59,8 @@ public enum ErrorCode {
 
   INVALID_COLLECTION_NAME("Invalid collection name "),
 
+  TOO_MANY_COLLECTIONS("Too many collections"),
```
Yuqi-Du, line 62: do we want to say maximum 5 here? not just too many collections?

Yuqi-Du (approved):

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/CreateCollectionOperation.java
```diff
@@ -149,6 +158,21 @@ private Uni<Supplier<CommandResult>> executeCollectionCreation(QueryExecutor que
     return indexResult.onItem().transform(res -> new SchemaChangeResult(res));
   }
 
+  void verifyLimits(List<Schema.CqlTable> tables) {
+    final int MAX_COLLECTIONS = dbLimitsConfig.maxCollections();
+    if (tables.size() >= MAX_COLLECTIONS) {
+      throw new JsonApiException(
+          ErrorCode.TOO_MANY_COLLECTIONS,
+          String.format(
+              "%s:  number of Collections cannot exceed %d, already have %d",
```
vkarpov15, line 167: Why uppercase "C" in "Collections"?

Also, would be nice to specify "number of collections in database cannot exceed 5", because this collection limit is database-specific, not global.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/CreateCollectionOperation.java
```diff
@@ -59,17 +63,22 @@ public static CreateCollectionOperation withVectorSearch(
...
   }
 
   @Override
   public Uni<Supplier<CommandResult>> execute(QueryExecutor queryExecutor) {
     return schemaManager
-        .getTable(commandContext.namespace(), name, MISSING_KEYSPACE_FUNCTION)
+        .getTables(commandContext.namespace(), MISSING_KEYSPACE_FUNCTION)
+        .collect()
+        .asList()
+        .invoke(tables -> verifyLimits(tables))
```
maheshrajamani, line 80: If a user rerun the same createCollection command which is already created, this logic will throw error  ErrorCode.TOO_MANY_COLLECTIONS instead of duplicate error. Is this the expectation?

maheshrajamani (approved):
Left a clarification., otherwise LGTM.

src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCode.java
Yuqi-Du: oh, my mistake

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/CreateCollectionOperation.java
tatu-at-datastax: Good point; I was wondering about that. Probably should only throw exception if new would be created -- will change, test that case.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/CreateCollectionOperation.java
tatu-at-datastax: Thank, I will see how other error messages do capitalization of "collections" and follow that (i.e. capitalize if already doing that; otherwise not). And will add reference to database.

tatu-at-datastax: Ok yes, no capitalization used for "collection", removed: added ref to "in databse".

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/CreateCollectionOperation.java
tatu-at-datastax: @maheshrajamani Fixed logic so that max-collections only checked if table does not yet exist; added verification of this case in test too.

jeffreyscarpenter:
I'm sorry I didn't catch this @tatu-at-datastax. But we need to enforce 5 collections per database, not per namespace.

tatu-at-datastax:
Right. That needs more work -- I'll create a follow-up ticket. For now it should work for most usage since there will be one-namespace per-database limit?

Issue #569, opened 2023-10-17
Replace `X-Cassandra-Token` with `Token`


jeffreyscarpenter:
We need to rename the header that we are using for accessing the JSON API. There is no need to reference Cassandra in this context. Will add a note with the new name when we have decided on it.

jeffreyscarpenter:
The new name will be "Token". We will also support "X-Cassandra-Token" for backward compatibility.

maheshrajamani:
@jeffreyscarpenter For this we need to just start the json api with this VM arg set., https://github.com/stargate/stargate/blob/main/apis/sgv2-quarkus-common/CONFIGURATION.md#:~:text=stargate.auth.header%2Dbased.header%2Dname

PR #607, opened 2023-10-31
token header name "X-Casssandra-Token" -> "Token"

Yuqi-Du:
**What this PR does**:
change token header name, "X-Casssandra-Token" -> "Token" 
also supports X-Cassandra-Token for backward compatibility

Fixes [#569 ](https://github.com/stargate/jsonapi/issues/569)

**Checklist**
- [x] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/api/security/ErrorChallengeSender.java
```diff
@@ -39,7 +40,7 @@ public ErrorChallengeSender(
     // create the response
     String message =
         "Role unauthorized for operation: Missing token, expecting one in the %s header."
-            .formatted(headerName);
```
tatu-at-datastax, line 42: Is `headerName` argument still needed/wanted? It is injected from `stargate.auth.header-based.header-name` which would usually override default of some kind. If not, can probably remove argument.

Yuqi-Du: Want to show "Token" to users, I am not using the injected one, which is "X-Cassandra-Token"

johnsmartco: Mentioned in slack thread but just to cover all the bases:

Thanks Yuqi  for the reminder.  If the header is still needed (see Tatu's question above) ... then yes I think we would need to also update the Headers tab of commands in (e.g.) ...: 

https://www.postman.com/datastax/workspace/stargate-cassandra/request/25879866-2a91dfda-fc9b-4536-8b20-7dfb18d13384

johnsmartco: Screenshot 2023-10-31 at 1 38 33 PM

tatu-at-datastax: > Want to show "Token" to users, I am not using the injected one, which is "X-Cassandra-Token"

Right, so you can then just remove `String headerName` from constructor I think.
Since it is no longer being used for anything.

Yuqi-Du: Perhaps no.
It is binded with configuration
If remove, can not even start quarkus
Pasted Graphic 4

tatu-at-datastax: Sorry, you can drop `headerName` but `ObjectMapper` still needs to be injected. So injection is needed but not headerName.

Yuqi-Du: understand, fixed it

src/main/java/io/stargate/sgv2/jsonapi/api/v1/TokenFilter.java
```diff
@@ -0,0 +1,27 @@
+package io.stargate.sgv2.jsonapi.api.v1;
+
+import static io.stargate.sgv2.jsonapi.config.constants.HttpConstants.DEPRECATED_AUTHENTICATION_TOKEN_HEADER_NAME;
+
+import io.quarkus.vertx.http.runtime.filters.Filters;
+import io.vertx.core.MultiMap;
+import jakarta.enterprise.context.RequestScoped;
+import jakarta.enterprise.event.Observes;
+
+@RequestScoped
```
tatu-at-datastax, line 10: Should add a JAvadoc to explain logic of why we want to (temporarily) add secondary Header from primary.

tatu-at-datastax (approved):

src/main/java/io/stargate/sgv2/jsonapi/config/constants/HttpConstants.java
```diff
@@ -0,0 +1,10 @@
+package io.stargate.sgv2.jsonapi.config.constants;
+
+public interface HttpConstants {
+
+  /** JSON API Authentication token header name. */
+  String AUTHENTICATION_TOKEN_HEADER_NAME = "Token";
+
+  /** JSON API also supports X-Cassandra-Token for backward compatibility. */
+  String DEPRECATED_AUTHENTICATION_TOKEN_HEADER_NAME = "X-Cassandra-Token";
```
vkarpov15, line 9: Do we have a test that tests the deprecated header still works? It looks like all the tests just use the `AUTHENTICATION_TOKEN_HEADER_NAME`

PR #612, opened 2023-11-02
fix collection not exist

Yuqi-Du:
when collection does not exist, operation involves vsearch will not error out as "collection not exist", but "vector search not enabled on this collection" which is misleading.

Fixes https://github.com/stargate/jsonapi/issues/609

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/bridge/executor/NamespaceCache.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/bridge/executor/NamespaceCache.java
@@ -6,6 +6,7 @@ import com.github.benmanes.caffeine.cache.Caffeine;
 import io.grpc.StatusRuntimeException;
 import io.smallrye.mutiny.Uni;
 import io.stargate.sgv2.jsonapi.exception.ErrorCode;
+import io.stargate.sgv2.jsonapi.exception.JsonApiException;
 import java.time.Duration;
 
 /** Caches the vector enabled status for the namespace */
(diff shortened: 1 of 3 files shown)


src/main/java/io/stargate/sgv2/jsonapi/service/bridge/executor/NamespaceCache.java
```diff
@@ -41,14 +42,24 @@ protected Uni<CollectionSettings> getCollectionProperties(String collectionName)
           .transformToUni(
               (result, error) -> {
                 if (null != error) {
+                  // collection does not exist
+                  if (error instanceof RuntimeException rte
```
tatu-at-datastax, line 46: Is there no `cause` to check for type or something, instead of relying exception message (which may change)?

tatu-at-datastax (approved):

src/main/java/io/stargate/sgv2/jsonapi/service/bridge/executor/NamespaceCache.java
```diff
@@ -41,14 +42,24 @@ protected Uni<CollectionSettings> getCollectionProperties(String collectionName)
...
+                      && rte.getMessage()
+                          .startsWith(ErrorCode.INVALID_COLLECTION_NAME.getMessage())) {
```
vkarpov15, line 48: Not wrong, but in my experience it is usually best practice to check against the error code or some other constant other than the error message, because (1) error messages may change, (2) it is possible that there may be multiple types of errors that start with the same message. Not a blocker, just figured I'd point this out.

PR #624, opened 2023-11-08
Fix #622: add Integration tests verifying both biggest allowed, and bigger-than-allowed JSON docs, values

### LINKED ISSUE #608: Improve max Collection limit verification to be based on available SAIs, not existing Tables
(follow-up to #577)

Currently JSON API enforces maximum of 5 (configurable) Collections per-namespace (implemented via #595).

But number of Collections we create is actually limited by number of indexes (SAIs) that can be created.
So we need to check if we could still create enough indexes for a new Collection, not by number of Table that exists, for 2 reasons:

1. Different Collections need different number of Indexes so number of Collections is not necessarily static
2. It may be possible to also create "raw" (non-Collection) tables, and number of indexes those create can be basically anything.

NOTE: to relax SAI count limitation from 50 to 100, we have https://datastax.jira.com/jira/software/c/projects/VECTOR/issues/VECTOR-184

### LINKED ISSUE #622: Add Integration Tests for max String property length; max Document size to ensure ability to store max valid values, docs
Currently we have unit tests for checking that maximum document and document component limits are enforced by shredder; mostly focusing on catching "too big" cases. There are some tests for large-but-valid cases too.

But we do not have tests to validate that "large but valid" docs, entities can actually be stored in the database.
Verifying this aspect is necessary to ensure that SAI does not have problems indexing "big" String values (for example).

tatu-at-datastax:
**What this PR does**:

Adds Integration tests to verify allowed maximum docs wrt max doc size, max string value length etc.

Fixes #622

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/config/DocumentLimitsConfig.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/config/DocumentLimitsConfig.java
@@ -13,15 +13,26 @@ import jakarta.validation.constraints.Positive;
 @ConfigMapping(prefix = "stargate.jsonapi.document.limits")
 public interface DocumentLimitsConfig {
 
-  /** Defines the max size of filter fields, default is 64 fields. */
+  /** Defines the default max size of filter fields. */
   int DEFAULT_MAX_FILTER_SIZE = 64;
 
+  /** Defines the default maximum document size. */
+  int DEFAULT_MAX_DOCUMENT_SIZE = 1_000_000;
+
+  /** Defines the default maximum length (in elements) of a single Array value */
+  int DEFAULT_MAX_ARRAY_LENGTH = 100;
+
+  /** Defines the maximum length of property names in JSON documents */
+  int DEFAULT_MAX_PROPERTY_NAME_LENGTH = 48;
+
+  /** Defines the default maximum length of a single String value */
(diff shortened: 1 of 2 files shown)


src/main/java/io/stargate/sgv2/jsonapi/config/DocumentLimitsConfig.java
```diff
@@ -13,15 +13,26 @@
 @ConfigMapping(prefix = "stargate.jsonapi.document.limits")
 public interface DocumentLimitsConfig {
 
-  /** Defines the max size of filter fields, default is 64 fields. */
+  /** Defines the default max size of filter fields. */
```
tatu-at-datastax, line 16: Adding constants (refactoring) because we cannot inject `DocumentLimitsConfig` into integration tests.

Yuqi-Du (approved):
LGTM

Yuqi-Du:
> LGTM

oh, it is still a draft PR. Sorry

tatu-at-datastax:
> > LGTM
> 
> oh, it is still a draft PR. Sorry

Np, thank you for pre-preview!

Yuqi-Du (approved):

vkarpov15 (commented):
LGTM :+1:

PR #626, opened 2023-11-09
$in with empty array should find nothing

Yuqi-Du:
**What this PR does**:
$in with empty array should find nothing

https://github.com/stargate/jsonapi/issues/625

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/ExpressionBuilder.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/ExpressionBuilder.java
@@ -16,6 +16,12 @@ public class ExpressionBuilder {
 
   public static List<Expression<BuiltCondition>> buildExpressions(
       LogicalExpression logicalExpression, DBFilterBase.IDFilter additionalIdFilter) {
+    // an empty filter should find everything
+    if (logicalExpression.isEmpty()) {
+      ArrayList<Expression<BuiltCondition>> list = new ArrayList<>();
+      list.add(null);
+      return list;
+    }
     // after validate in FilterClauseDeserializer,
     // partition key column key will not be nested under OR operator
     // so we can collect all id_conditions, then do a combination to generate separate queries
@@ -43,8 +49,7 @@ public class ExpressionBuilder {
     if (idFilters.isEmpty()
         && additionalIdFilter == null) { // no idFilters in filter clause and no additionalIdFilter
       if (expressionWithoutId == null) {
-        expressionsWithId.add(null); // should find everything
-        return expressionsWithId;
+        return null; // should find nothing
       } else {
         return List.of(expressionWithoutId);
       }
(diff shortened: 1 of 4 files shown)


src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/ExpressionBuilder.java
```diff
@@ -16,6 +16,12 @@ public class ExpressionBuilder {
 
   public static List<Expression<BuiltCondition>> buildExpressions(
       LogicalExpression logicalExpression, DBFilterBase.IDFilter additionalIdFilter) {
+    // an empty filter should find everything
+    if (logicalExpression.isEmpty() && additionalIdFilter == null) {
+      ArrayList<Expression<BuiltCondition>> list = new ArrayList<>();
```
tatu-at-datastax, line 21: Could just use

    return Collections.singletonList(null);

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/ExpressionBuilder.java
```diff
@@ -115,6 +125,14 @@ private static Expression<BuiltCondition> buildExpressionRecursive(
     if (conditionExpressions.isEmpty()) {
       return null;
     }
+    // when having an empty array $in, if $in occurs within an $and logic, entire $and should match
+    // nothing
+    if (hasInFilterThisLevel
```
tatu-at-datastax, line 130: Instead of these state flags, is it not possible to translate empty `$in` clause into something in CQL that evaluates to `false`? Instead of trying to use special states, checks, at this level. I am mostly worried that due to possible nesting, these checks might get complicated and not handle all cases.

Yuqi-Du: same concerns, but it seems there is no keywords like 'false' you can add into cql where clause

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/ExpressionBuilder.java
```diff
@@ -91,12 +96,17 @@ private static Expression<BuiltCondition> buildExpressionRecursive(
       }
       conditionExpressions.add(subExpressionCondition);
     }
+
+    boolean hasInFilterThisLevel = false;
+    boolean InFilterThisLevelWithEmptyArray = true;
```
tatu-at-datastax, line 101: naming convention: should be `inFilterThisLevelWithEmptyArray` (start with lower-case)

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/ExpressionBuilder.java
tatu-at-datastax: One thing that I think does work is actually using dummy `ID` clause for our own `key` column:

```
   WHERE key in ( )
```

which works as expected, matching nothing. And can then be composed with other clauses, like

```
   WHERE key in ( ) OR array_contains contains "column value"
```

(or however it would be generated)

Yuqi-Du: unfortunately, this is also a no.
You can not use OR for partition key in cassandra.
image

tatu-at-datastax: Interesting. Ok, never mind then... Cassandra not smart enough to optimize that away (or validates usage before optimizations).

Let's go with code as is, then.

tatu-at-datastax (approved):

src/test/java/io/stargate/sgv2/jsonapi/api/v1/FindIntegrationTest.java
```diff
@@ -393,6 +393,102 @@ public void inConditionEmptyArray() {
...
+          .contentType(ContentType.JSON)
+          .body(json)
+          .when()
+          .post(CollectionResource.BASE_PATH, namespaceName, collectionName)
+          .then()
+          .statusCode(200)
+          .body("data.documents", hasSize(1))
+          .body("status", is(nullValue()))
+          .body("errors", is(nullValue()))
+          .body("data.documents[0]", jsonEquals(expected1));
```
vkarpov15, line 489: It might be better to have a test case that returns no results here, because there's no way of telling whether the document `expected1` matched the first condition in `$or`, the 2nd, or both.

vkarpov15 (commented):
One minor comment, but LGTM overall

PR #629, opened 2023-11-09
fix pagination

Yuqi-Du:
fix pagination in cql branch

**Checklist**
- [x]  Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/service/bridge/executor/QueryExecutor.java
```diff
@@ -77,12 +77,13 @@ public Uni<QueryOuterClass.ResultSet> executeRead(
    */
   public Uni<AsyncResultSet> executeRead(
       SimpleStatement simpleStatement, Optional<String> pagingState, int pageSize) {
+    simpleStatement =
+        simpleStatement
+            .setPageSize(pageSize)
```
tatu-at-datastax, line 82: Good catch! So basically was missing both page size and consistency level unless we had paging state -- and that would only happen for the non-first page of result set. Which would not occur when page size (limit) was not even set.

tatu-at-datastax (approved):

PR #636, opened 2023-11-10
Fix #630: limit max properties per doc to 1000

### LINKED ISSUE #630: Add `DocumentLimitsConfig` limit for total Properties per Document (1000?)
Currently `DocumentLimitsConfig` only has limit for per-Object properties (64). But since we allow documents of up to 1MB in length, one could theoretically create documents with tens of thousands of properties.

So let's add per-document limit: start with, say, 1000.

There are couple of ways we could try to do limit:

1. Leaf values -- but would this count Array elements too? Should it? (we have per-Array max element setting of 100)
2. All property names at every level: longer paths somewhat penalized

For initial limit let's go with (2); this is easier to relax than if we tried another approach.

tatu-at-datastax:
**What this PR does**:

Adds new document constraint, validation, to ensure total number of properties in doc does not exceed 1000 (configurable)

Fixes #630

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/shredding/Shredder.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/shredding/Shredder.java
@@ -17,6 +17,7 @@ import java.io.IOException;
 import java.util.Iterator;
 import java.util.Map;
 import java.util.UUID;
+import java.util.concurrent.atomic.AtomicInteger;
 
 /**
  * Shred an incoming JSON document into the data we need to store in the DB, and then de-shred.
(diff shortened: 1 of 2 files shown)


src/main/java/io/stargate/sgv2/jsonapi/service/shredding/Shredder.java
```diff
@@ -214,22 +215,40 @@ private void validateDocument(DocumentLimitsConfig limits, ObjectNode doc, Strin
     }
 
     // Second: traverse to check for other constraints
-    validateObjectValue(limits, null, doc, 0);
+    AtomicInteger totalProperties = new AtomicInteger(0);
```
Yuqi-Du, line 218: Reason of usinh AtomicInteger here is **just** for wrapping the counter, and enabling to change it in method calls?
Is that correct?

Yuqi-Du (approved):

src/main/java/io/stargate/sgv2/jsonapi/service/shredding/Shredder.java
tatu-at-datastax: Yes, exactly. Could use `int[1]` as an alternative but `AtomicInteger` seems better (even if giving wrong idea of needing atomicity).

vkarpov15 (commented):
LGTM :+1:

PR #638, opened 2023-11-13
Fix native-cql branch: $in with multiple same values

Yuqi-Du:
Bug in native-cql branch : if have multiple same values in the $in value array, json api will error out.
This pr is for the fix.
Fixes https://github.com/stargate/jsonapi/issues/623

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/ExpressionBuilder.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/ExpressionBuilder.java
@@ -138,7 +138,7 @@ public class ExpressionBuilder {
 
   public static List<Object> getExpressionValuesInOrder(Expression<BuiltCondition> expression) {
     List<Object> values = new ArrayList<>();
-    if(expression!=null) {
+    if (expression != null) {
       populateValuesRecursive(values, expression);
     }
     return values;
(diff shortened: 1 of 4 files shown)


src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/ExpressionBuilder.java
```diff
@@ -135,4 +135,31 @@ private static Expression<BuiltCondition> buildExpressionRecursive(
     return ExpressionUtils.buildExpression(
         conditionExpressions, logicalExpression.getLogicalRelation().getOperator());
   }
+
+  public static List<Object> getExpressionValuesInOrder(Expression<BuiltCondition> expression) {
```
tatu-at-datastax, line 139: Could you add a Javadoc explaining what this does?

tatu-at-datastax (approved):
LGTM -- ideally would add Javadoc to explain intended behavior (although name is good and gives some idea).

tatu-at-datastax:
@Yuqi-Du Is this not needed for `main`?

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/ExpressionBuilder.java
Yuqi-Du: sure!

Yuqi-Du:
> @Yuqi-Du Is this not needed for `main`?

Yes, only for CQL branch.

tatu-at-datastax:
@Yuqi-Du may want to merge/rebase from `native_cql_changes` to get Mahesh' IT fixes, to let CI verify there are only those 4 Authorization fails and nothing else

PR #640, opened 2023-11-13
Add to #630: update `jsonapi-spec.md` too

### LINKED ISSUE #630: Add `DocumentLimitsConfig` limit for total Properties per Document (1000?)

tatu-at-datastax:
**What this PR does**:

Updates documentation wrt limit introduced in #630

Fixes #630

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/docs/jsonapi-spec.md
+++ b/docs/jsonapi-spec.md
@@ -389,6 +389,21 @@ The maximum length of a [field name](#json-document-field-names) is 48 character
 
 The maximum number of fields allowed in a single JSON object is 64.
 
+The maximum number of fields allowed in a single JSON document is 1000.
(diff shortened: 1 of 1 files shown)


tatu-at-datastax:
Need to force push since this does not trigger CI jobs.

PR #654, opened 2023-11-16
Fix #639: add testing of CreateCollection with mixed-case name, fix quoting for FindCollections

### LINKED ISSUE #639: Verify handling of Collection with capital letter(s)
(user reported)

Use of mixed-case letters is allowed for Collections (which map to Cassandra Tables). But it seems that when reporting the name, it is sometimes surrounded by double-quotes, likely due to internal handling for CQL.
This should not be exposed to JSON API users: if and when quoting is needed for CQL that should be handled transparently.

Two places where this may be exposed would be:

1. Response when creating Collection (`CreateCollection` command)
2. `FindCollections` command

and so need to add ITs and fix handling if necessary.

tatu-at-datastax:
**What this PR does**:

Adds tests for #639 to check that mixed-case Collection names are supported, and also fix one issue found: that of unwanted quoting of Collection name in mixed-case case.

Fixes #639

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/CollectionSettings.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/CollectionSettings.java
@@ -59,7 +59,8 @@ public record CollectionSettings(
 
   public static CollectionSettings getCollectionSettings(
       TableMetadata table, ObjectMapper objectMapper) {
-    String collectionName = table.getName().asCql(true);
+    // [jsonapi#639]: get internal name to avoid quoting of case-sensitive names
+    String collectionName = table.getName().asInternal();
     // get vector column
     final Optional<ColumnMetadata> vectorColumn =
         table.getColumn(DocumentConstants.Fields.VECTOR_SEARCH_INDEX_COLUMN_NAME);
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/CollectionSettings.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/CollectionSettings.java
@@ -85,7 +85,7 @@ public record CollectionSettings(
         if (functionName != null)
           function = CollectionSettings.SimilarityFunction.fromString(functionName);
       }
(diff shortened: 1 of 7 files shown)


src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/CollectionSettings.java
```diff
@@ -59,7 +59,8 @@ public static SimilarityFunction fromString(String similarityFunction) {
 
   public static CollectionSettings getCollectionSettings(
       TableMetadata table, ObjectMapper objectMapper) {
-    String collectionName = table.getName().asCql(true);
+    // [jsonapi#639]: get internal name to avoid quoting of case-sensitive names
+    String collectionName = table.getName().asInternal();
```
tatu-at-datastax, line 63: This is the actual fix: prevents name from being surrounded by double-quotes (needed for CQL usage, but not in JSON)

src/test/java/io/stargate/sgv2/jsonapi/api/v1/AbstractCollectionIntegrationTestBase.java
```diff
@@ -26,21 +26,23 @@ public abstract class AbstractCollectionIntegrationTestBase
...
-          "createCollection": {
-            "name": "%s"
-          }
-        }
-        """
-            .formatted(collectionName);
+    createCollection(this.collectionName);
+  }
 
+  protected void createCollection(String collectionToCreate) {
```
tatu-at-datastax, line 32: Minor refactoring to make it easy to create custom collections from ITs.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/FindCollectionsIntegrationTest.java
```diff
@@ -79,7 +74,7 @@ public void happyPath() {
     }
 
     @Test
-    @Order(1)
+    @Order(2)
```
tatu-at-datastax, line 77: Had same order, assuming accidentally, fixed.

Hazel-Datastax (approved):
LGTM

vkarpov15 (commented):
LGTM :+1:

PR #658, opened 2023-11-17
Add truncate/empty collection support

### LINKED ISSUE #631: Feature request: Add truncate/empty collection support
Please add a truncate operation equivalent to the CQL truncate table operation.

Hazel-Datastax:
**What this PR does**:
Add truncate/empty collection support

Fixes #631 

**Checklist**
- [x] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/TruncateCollectionOperation.java
```diff
@@ -0,0 +1,30 @@
...
+import io.stargate.sgv2.jsonapi.api.model.command.CommandContext;
+import io.stargate.sgv2.jsonapi.api.model.command.CommandResult;
+import io.stargate.sgv2.jsonapi.service.cqldriver.executor.QueryExecutor;
+import io.stargate.sgv2.jsonapi.service.operation.model.Operation;
+import java.util.function.Supplier;
+
+/**
+ * Implementation of the truncate collection.
+ *
+ * @param context Command context, carries namespace and collectoin of the collection.
```
tatu-at-datastax, line 14: nit: "collectoin" -> "Collection". But also... not sure what "collection of the collection" means? Maybe just "name of the collection"?

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/model/impl/DeleteManyCommandResolver.java
```diff
@@ -35,6 +36,10 @@ public DeleteManyCommandResolver(OperationsConfig operationsConfig, ObjectMapper
 
   @Override
   public Operation resolveCommand(CommandContext commandContext, DeleteManyCommand command) {
+    // if there is no filter or is an empty filter, delete all the rows or truncate the table.
```
tatu-at-datastax, line 39: "If there is no filter or filter is empty, use Truncate operation instead of Delete"?

src/test/java/io/stargate/sgv2/jsonapi/api/v1/DeleteManyIntegrationTest.java
```diff
@@ -208,17 +208,20 @@ public void byColumn() {
           .body("errors", is(nullValue()));
     }
 
+    /**
+     * when there is no filter, it will delete all data in collection, like truncate operation. it
```
tatu-at-datastax, line 212: Should be reworded to explain what is being verified/tested. Something like

"Verify that DeleteMany with no (or empty) filter will use truncate and delete all documents in a collection, instead of using Delete with maximum number to delete".

src/test/java/io/stargate/sgv2/jsonapi/api/v1/DeleteManyIntegrationTest.java
```diff
@@ -253,17 +257,69 @@ public void noFilter() {
           .body("errors", is(nullValue()));
     }
 
+    /**
```
tatu-at-datastax, line 260: Reword: can basically say that this tests same behavior as `noFilter()` but with "empty" filter.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/DeleteManyIntegrationTest.java
```diff
@@ -253,17 +257,69 @@ public void noFilter() {
...
+      given()
+          .header(HttpConstants.AUTHENTICATION_TOKEN_HEADER_NAME, getAuthToken())
+          .contentType(ContentType.JSON)
+          .body(json)
+          .when()
+          .post(CollectionResource.BASE_PATH, namespaceName, collectionName)
+          .then()
+          .statusCode(200)
+          .body("status.ok", is(1))
+          .body("status.deletedCount", is(nullValue()))
```
vkarpov15, line 284: Is there any way to also get the deleted count without having to execute a separate count operation? Our dccs say that `deleteMany()` returns a `deletedCount` that contains the number of docs deleted, it would be a bit clunky if `deletedCount` isn't set in some cases. I wouldn't say that missing `deletedCount` is a critical issue, more a nice-to-have to ensure the API is consistent. If we're unable to return `deletedCount` in the case of a truncate, we should document that.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/TruncateCollectionOperation.java
vkarpov15: :+1: this comment is unclear

vkarpov15 (commented):
A couple of small comments. I'm mostly concerned about the missing `deletedCount`.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/DeleteManyIntegrationTest.java
Hazel-Datastax: To accurately return the total number of rows, a separate count operation is necessary, which may reduce efficiency. Aaron suggested [here](https://github.com/stargate/jsonapi/issues/631) using -1 as the value for deletedCount. What are your thoughts on this approach?

jeffreyscarpenter: @amorton can you comment on this? I would like to get this merged if possible.

vkarpov15 (approved):

Yuqi-Du:
LGTM

Yuqi-Du (approved):

PR #676, opened 2023-11-22
test with DSE 7 image

jeffreyscarpenter:
This is a draft PR to see if we can use the Stargate coordinator with `persistence-dse-next` module in front of DSE 7 as a backend, or if we'll end up having to create a new `persistence-dse-7.0` module in the main Stargate repo.


jeffreyscarpenter:
The new docker compose file in this PR appears to work correctly and allows everything to come up, but when I try to run ITs the coordinator and cassandra nodes don't seem to gossip correctly, and the tests fail. 

To reproduce the IT failure, clone this branch and then do a simple command like `./mvnw verify -DskipUnitTests`.
Then you have to look in the failsafe logs to really see what is going on, i.e. after a test run look in a file such as `target/failsafe-reports/TEST-io.stargate.sgv2.jsonapi.api.v1.CollectionResourceIntegrationTest.xml`

toward the bottom of the file you will see that that the coordinator and cassandra containers start up and are trying to connect, and then you see a line like this from the cassandra container every 5 seconds until the test times out:

```
INFO  [docker-java-stream-533490291] 2023-11-22 12:24:01,113 null:-1 - [CASSANDRA] STDOUT: INFO  [ScheduledTasks:1] 2023-11-22 19:24:01,109 TokenMetadata.java:532 - Updating topology for all endpoints that have changed
```

Because the gossip doesn't complete successfully, the coordinator node does not report ready status, and TestContainers times out.

Maybe @maheshrajamani or @kathirsvn has some idea of what is going on? I suspected it was datacenter/rack name but I'm now overriding these and it still doesn't work.

jeffreyscarpenter:
Closing this PR, as it duplicates work under #733. We can use that PR for continued work

PR #681, opened 2023-11-28
JsonTerm Refactoring for unit test

maheshrajamani:
**What this PR does**:
Converted JsonTerm to extend Literal instead of marker to solve the unit test issue.
QueryBuilder asserts for Literal value for the rhs in case where clause use map entry.

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/JsonTerm.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/JsonTerm.java
@@ -1,9 +1,10 @@
 package io.stargate.sgv2.jsonapi.service.operation.model.impl;
 
-import io.stargate.sgv2.api.common.cql.builder.Marker;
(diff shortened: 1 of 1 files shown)


src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/JsonTerm.java
```diff
@@ -1,9 +1,10 @@
 package io.stargate.sgv2.jsonapi.service.operation.model.impl;
 
-import io.stargate.sgv2.api.common.cql.builder.Marker;
+import io.stargate.bridge.grpc.Values;
+import io.stargate.sgv2.api.common.cql.builder.Literal;
 import java.util.Objects;
 
-public class JsonTerm extends Marker {
+public class JsonTerm extends Literal {
```
tatu-at-datastax, line 7: Would it make sense to add simple Javadoc?

tatu-at-datastax (approved):

PR #687, opened 2023-11-29
Fix #686: add ITs to verify Max-String-Value-Length constaints

### LINKED ISSUE #686: Add IT for maximum String value length
Most Document limits (maximum doc size, max number of properties, maximum number value length etc) are already covered by both unit and integration tests.

But it looks like maximum String length (16,000 characters) is currently only tested by unit tests. There have been reports that long (over 5kB) String values may be problematic with Storage-Attached Index updates, so let's add ITs (max allowed, too long) to cover this gap.

tatu-at-datastax:
**What this PR does**:

Adds Integration tests checking handling of both too-long-String-values and as-long-as-possible-String values (already have unit tests)

Fixes #686

**Checklist**
- [ ] Changes manually tested


tatu-at-datastax:
Not having much success: updated `dse-next` version; next will get a new Stargate 2.1 build to try to get limits-enforcement by backend working as it should.

PR #693, opened 2023-12-01
Throw error on invalid json api table schema

### LINKED ISSUE #659: Provide better error message when accessing non-JSON API table
A user reported the following:

> We’re working on a chatbot using a db loaded w/ wikipedia data. The table has already been loaded w/ the embeddings but when running a find using my TS client I always get a [{"message":"Vector search is not enabled for the collection pages","errorCode":"VECTOR_SEARCH_NOT_SUPPORTED"}] .

After investigation it was determined that the underlying table for this collection was created via CQL. While the table did have a vector column, it was not created via JSON API, which is the real root issue. We need a more correct error message in this case that addresses the larger problem.

maheshrajamani:
**What this PR does**:
Fix to throw error on invalid json api table schema

Fixes #659 

**Checklist**
- [x] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCode.java
```diff
@@ -59,6 +59,8 @@ public enum ErrorCode {
 
   INVALID_COLLECTION_NAME("Invalid collection name "),
 
+  INVALID_JSONAPI_COLLECTION_SCHEMA("Not a valid json api collection schema : "),
```
tatu-at-datastax, line 62: nit: no space before colon (similar to other messages)

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/NamespaceCache.java
```diff
@@ -72,13 +86,20 @@ protected Uni<CollectionSettings> getCollectionProperties(String collectionName)
     }
   }
 
+  @VisibleForTesting
```
tatu-at-datastax, line 89: Is this needed? I thought it was for case where method has "too much" visibility for purposes of testing?
(as per https://stackoverflow.com/questions/33267820/how-to-use-visiblefortesting-for-pure-junit-tests)

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/NamespaceCache.java
```diff
@@ -42,6 +44,18 @@ protected Uni<CollectionSettings> getCollectionProperties(String collectionName)
...
               (result, error) -> {
                 if (null != error) {
+                  // not a valid collection schema
+                  if (error instanceof JsonApiException
+                      && ((JsonApiException) error).getErrorCode()
+                          == ErrorCode.VECTORIZECONFIG_CHECK_FAIL) {
+                    return Uni.createFrom()
+                        .failure(
+                            new JsonApiException(
+                                ErrorCode.COLLECTION_NOT_EXIST,
```
tatu-at-datastax, line 54: But wouldn't this give "Collection does not exist, collection name: " failure? Would it better to indicate that there is a table but one that is not valid Collection? Or maybe that happens somewhere else?

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/NamespaceCache.java
maheshrajamani: Good catch, was trying something, will remove it.

tatu-at-datastax (approved):

PR #698, opened 2023-12-04
for local cassandra, check if sai is enabled

Yuqi-Du:
**What this PR does**:
For local cassandra backend, check if sai is enabled.
Previous we did this startup check by bridge, now use CqlSession

Fixes https://github.com/stargate/jsonapi/issues/674

**Checklist**
- [x] Changes manually tested


tatu-at-datastax:
Let's wait as it's not clear we actually want to do these checks. To be discussed on today's meeting.

tatu-at-datastax (changes requested):
Let's wait for discussion on whether this is really needed; or if we just document "JSON API requires SAI support" as a pre-req.

PR #706, opened 2023-12-07
Change the consistency level for vector search

### LINKED ISSUE #691: Use lower consistency in CQL vector search queries
On the CQL-native based version of JSON API introduced in v1.0.0-BETA-4, we now see log messages like the following at the warning level when performing vector search queries:

```
Query '[1 values] SELECT key, tx_id, doc_json FROM my_namespace.my_collection ORDER BY query_vector_value ANN OF ? LIMIT 1 [v0=[0.1, 0.9]]' generated server side warning(s): Top-K queries can only be run with consistency level ONE / LOCAL_ONE / NODE_LOCAL. Consistency level LOCAL_QUORUM was requested. Downgrading the consistency level to LOCAL_ONE.
```

We should change the consistency level used for these queries.

Hazel-Datastax:
**What this PR does**:
Use lower consistency in CQL vector search queries

Fixes #691 #716

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/config/OperationsConfig.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/config/OperationsConfig.java
@@ -202,7 +202,7 @@ public interface OperationsConfig {
       @WithConverter(ConsistencyLevelConverter.class)
       ConsistencyLevel reads();
 
-      /** @return Consistency for queries reading the data. */
+      /** @return Consistency for vector search queries. */
       @WithDefault("LOCAL_ONE")
       @NotNull
       @WithConverter(ConsistencyLevelConverter.class)
--- a/src/main/java/io/stargate/sgv2/jsonapi/config/OperationsConfig.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/config/OperationsConfig.java
@@ -201,6 +201,12 @@ public interface OperationsConfig {
       @NotNull
       @WithConverter(ConsistencyLevelConverter.class)
       ConsistencyLevel reads();
+
+      /** @return Consistency for queries reading the data. */
+      @WithDefault("LOCAL_ONE")
+      @NotNull
+      @WithConverter(ConsistencyLevelConverter.class)
+      ConsistencyLevel vectorSearch();
     }
   }
 }
--- a/src/main/java/io/stargate/sgv2/jsonapi/config/OperationsConfig.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/config/OperationsConfig.java
@@ -177,7 +177,7 @@ public interface OperationsConfig {
     ConsistencyConfig consistency();
 
     /** @return Serial Consistency for queries. */
-    @WithDefault("SERIAL")
+    @WithDefault("LOCAL_SERIAL")
     @WithConverter(ConsistencyLevelConverter.class)
     ConsistencyLevel serialConsistency();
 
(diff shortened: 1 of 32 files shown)


src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/FindOperation.java
```diff
@@ -315,20 +315,48 @@ public Uni<FindResponse> getDocuments(
...
+              queries,
+              pageSize,
+              objectMapper(),
+              new ChainedComparator(orderBy(), objectMapper()),
+              orderBy().size(),
+              skip(),
+              limit(),
+              maxSortReadLimit(),
+              projection(),
+              false);
```
maheshrajamani, line 330: Instead of if condition can we just pass the argument as  `vector() != null`

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/FindOperation.java
```diff
@@ -315,20 +315,48 @@ public Uni<FindResponse> getDocuments(
...
-            queries,
-            pageSize,
-            objectMapper(),
-            new ChainedComparator(orderBy(), objectMapper()),
-            orderBy().size(),
-            skip(),
-            limit(),
-            maxSortReadLimit(),
-            projection());
+        if (vector() == null) {
```
maheshrajamani, line 318: Instead of if-else block can you send the last argument as `vector() == null`

maheshrajamani (approved):
LGTM., a minor code change suggested

src/main/java/io/stargate/sgv2/jsonapi/config/OperationsConfig.java
```diff
@@ -201,6 +201,12 @@ interface ConsistencyConfig {
       @NotNull
       @WithConverter(ConsistencyLevelConverter.class)
       ConsistencyLevel reads();
+
+      /** @return Consistency for queries reading the data. */
```
tatu-at-datastax, line 205: nit: queries reading data -> vector search queries

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/QueryExecutor.java
```diff
@@ -42,14 +42,25 @@ public QueryExecutor(OperationsConfig operationsConfig, StargateRequestInfo star
...
+   * @param vectorSearch - whether the query uses vector search -- use ANN OF
    * @return AsyncResultSet
    */
   public Uni<AsyncResultSet> executeRead(
-      SimpleStatement simpleStatement, Optional<String> pagingState, int pageSize) {
-    simpleStatement =
-        simpleStatement
-            .setPageSize(pageSize)
-            .setConsistencyLevel(operationsConfig.queriesConfig().consistency().reads());
+      SimpleStatement simpleStatement,
```
tatu-at-datastax, line 49: I wonder if it'd be cleaner to have 2 methods -- executeRead() and executeVectorSearch() instead of one more parameter? Mostly it would be more obvious from call since I think it is easy to get two mixed.
(note: I am not sure this is better way, just suggesting as consideration)

@maheshrajamani WDYT?

Hazel-Datastax: Changed the boolean flag to the additional method. @tatu-at-datastax Could you help me do the last review? Thanks!

tatu-at-datastax: Ok @maheshrajamani did the review -- looks good to me too, thank you for doing this @Hazel-Datastax !

PR #719, opened 2023-12-08
Fix #710: use byte-length for max String value, not characters

### LINKED ISSUE #710: Improve max-String-value limit checking to be based on UTF-8 (byte) length, not char count
Currently maximum String value length JSON API allows is 16,000 characters. But the backend limitations for indexing (SAI) define limit in terms of bytes: so assuming limit for DB was 16,000 bytes, we might hit the limit with multi-byte UTF-8 characters.

Let's change validation for Strings to be in units of bytes of serialized UTF-8: this allows as to maintain precise limits.

Note: this does add some overhead, but without doing this we cannot reliably align limits across JSON API and DB, leading to bad user experience.

Note: implementation should be able to use one of Commons libraries (Guava or Apache commons) to avoid allocation of new byte arrays.

tatu-at-datastax:
**What this PR does**:

Changes validation of max String value length to use UTF-8 encoded byte length, instead of Java characters: this because that's what SAI uses.

Fixes #710

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -71,12 +71,16 @@
       <artifactId>quarkus-smallrye-health</artifactId>
     </dependency>
     <dependency>
+      <groupId>io.quarkus</groupId>
+      <artifactId>quarkus-rest-client-reactive-jackson</artifactId>
+    </dependency>
+    <dependency>
       <groupId>org.apache.commons</groupId>
       <artifactId>commons-lang3</artifactId>
     </dependency>
     <dependency>
-      <groupId>io.quarkus</groupId>
-      <artifactId>quarkus-rest-client-reactive-jackson</artifactId>
+      <groupId>com.google.guava</groupId>
+      <artifactId>guava</artifactId>
     </dependency>
     <dependency>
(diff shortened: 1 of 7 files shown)


pom.xml
```diff
@@ -70,13 +70,17 @@
...
+      <artifactId>quarkus-rest-client-reactive-jackson</artifactId>
+    </dependency>
     <dependency>
       <groupId>org.apache.commons</groupId>
       <artifactId>commons-lang3</artifactId>
     </dependency>
     <dependency>
-      <groupId>io.quarkus</groupId>
-      <artifactId>quarkus-rest-client-reactive-jackson</artifactId>
+      <groupId>com.google.guava</groupId>
```
tatu-at-datastax, line 82: Add Guava dependency for UTF-8 string length calculation method. Looks like some code already depended on Guava but we did not have explicit dependency but transitive one. Safer to have explicit.

Hazel-Datastax (approved):
LGTM

maheshrajamani (approved):
LGTM

PR #730, opened 2023-12-13
CQLSessionCache - evictionListener changed to removalListener

### LINKED ISSUE #645: Closing session with possible in-flight queries
In the native cql driver based implementation, the driver cqlsession object is getting closed on getting evicted from the cache. But there is no check if there are any queries in-flight before closing the session. In other words, is there a graceful close for session?

kathirsvn:
**What this PR does**:

In `CQLSessionCache`, Caffeine cache's `evictionListener` is changed to `removalListener`.

1. The configured removal listener is invoked after the entry has been removed from the cache, so any session that gets evicted will first be removed from the cache so it will no longer be found for any key.
2. The configured removal listener closes the `CqlSession` that is passed which is a graceful shut down (i.e. any in-flight queries will be completed before shutting down the channels)

Fixes #645 

**Checklist**
- [X] Changes manually tested

### DIFF
--- a/.github/workflows/continuous-integration.yaml
+++ b/.github/workflows/continuous-integration.yaml
@@ -79,7 +79,7 @@ jobs:
             profile: ''
         exclude:
           - type: native
-            profile: '-Pnative'
+            #profile: '-Pnative'
 
     steps:
       - uses: actions/checkout@v3
--- a/.github/workflows/continuous-integration.yaml
+++ b/.github/workflows/continuous-integration.yaml
@@ -77,7 +77,7 @@ jobs:
         include:
           - type: docker
             profile: ''
-
+        exclude:
           - type: native
             profile: '-Pnative'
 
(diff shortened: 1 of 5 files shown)


kathirsvn:
@stargate/core The IT for the native image fails in this PR.
https://github.com/stargate/jsonapi/actions/runs/7195421867/job/19598287312?pr=730.

I reported the same [here](https://github.com/quarkusio/quarkus/issues/30744#issuecomment-1855093527). Can we disable native image test for now until this is fixed on the Quarkus side?

jeffreyscarpenter:
> Can we disable native image test for now until this is fixed on the Quarkus side?

Yes, I agree this makes sense @kathirsvn

kathirsvn:
Excluded the Native Image CI tests and the tests look good now.

Yuqi-Du (approved):
LGTM

PR #733, opened 2023-12-14
Run JSON API with DSE 7 in IT / docker-compose

### LINKED ISSUE #600: Run JSON API with DSE 7
There are now Docker images being published for DataStax Enterprise 7 on [Docker Hub](https://hub.docker.com/layers/datastax/dse-server/7.0.0-alpha.1/images/sha256-d97e7ae7110e7ff367be8976fc3b6237bddf155e8f99c9ad7ec2c9a61954bca6?context=explore).

Since DSE 7 is based on https://github.com/datastax/cassandra, most SAI features required by JSON API are available or will be available soon. We can start exploring DSE 7 support as an experimental feature.

Hazel-Datastax:
**What this PR does**:
Create an additional docker-compose configuration that runs JSON API in front of a single DSE 7 node, ~with no coordinator~

Fixes #600 

**Checklist**
- [ ] Changes manually tested


docker-compose/docker-compose-dse7.yml
```diff
@@ -0,0 +1,53 @@
+version: '2'
+services:
+  dse-7:
+    image: datastax/dse-server:7.0.0-alpha.3
```
tatu-at-datastax, line 4: Should the tag come from `.env` file (like `DSE7_TAG`)?

Hazel-Datastax: Sorry, forgot to remove the image hardcode for testing purposes. I've updated to use the reference `DSE7TAG` which is the latest DSE7 image tag in Docker Hub. The command is `DSE7TAG="$(curl -s https://registry.hub.docker.com/v2/repositories/datastax/dse-server/tags/ | jq -r '[.results[] | select(.name | startswith("7.0.0"))] | sort_by(.last_updated) | last | .name')"` in `start_dse7.sh`.

jeffreyscarpenter:
> Looks good. One last question: do we need default value for `DSE7Tag` in `docker-compose/.env`? I am not 100% sure this is needed since script defaults to another value.
> 
> @jeffreyscarpenter Do you think `docker-compose/.env` file is still used/needed?

Yes, I think that is a good fix to include.

I'm on the fence about whether we should even continue to support the "dse-next" image although technically it is an open source alternative to using the DSE-7 image, it is something we built purely for testing so it's not officially supported. I mention that because I think maybe we might decide to remove usage of DSE-next. If we do that, we'd keep using the DSETAG name and have that refer to the DSE 7 version only.

jeffreyscarpenter:
I would like to see this PR updated so that the integration tests run against DSE 7 instead of dse-next.

The piece that might be tricky is how to get an authentication token. Our ITs are set up so that we get a token from the Auth API in the coordinator. I guess we could run a coordinator node in dev mode (so that it doesn't attempt to connect to the DSE node), but this seems like a hack.

@kathirsvn can you suggest what alternatives we might consider to get auth tokens? At one time there was a docker-compose setup for Stargate v1 on the main stargate repo that ran keycloak to generate JWTs, don't know if something like this would make sense?

jeffreyscarpenter:
This PR is no longer current and work has been continued in other PRs, closing.

PR #648, opened 2023-11-15
Timeout retry

maheshrajamani:
**What this PR does**:
Retry on read or write timeout

**Checklist**
- [ ] Changes manually tested
- [ ] Automated Tests added/updated
- [ ] Documentation added/updated
- [ ] CLA Signed: [DataStax CLA](https://cla.datastax.com/)

PR #734, opened 2023-12-14
Same Node Retry Policy added

kathirsvn:
**What this PR does**:

- Added a retry policy that retries on the same node in all scenarios. This is useful when we have a proxy like cqlrouter to communicate with the backend.
- The `SameNodeRetryPolicy` inherits `DefaultRetryPolicy` provided by the driver and overrides few error scenarios to retry on the same node where the `DefaultRetryPolicy` retries on the next available node.
- This can be enabled by setting `-Ddatastax-java-driver.advanced.retry-policy.class=io.stargate.sgv2.jsonapi.service.cqldriver.SameNodeRetryPolicy`

Fixes #648 

**Checklist**
- [X] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/SameNodeRetryPolicy.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/SameNodeRetryPolicy.java
@@ -17,7 +17,8 @@ import org.slf4j.LoggerFactory;
 /**
  * A retry policy that retries to the same node in the query plan. DefaultRetryPolicy retries to the
  * next node in the query plan in some of the error cases and this implementation overrides that
- * behavior.
+ * behavior. This is useful when we have a proxy like cqlrouter to communicate with the backend
+ * cassandra.
  */
 public class SameNodeRetryPolicy extends DefaultRetryPolicy {
 
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/SameNodeRetryPolicy.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/SameNodeRetryPolicy.java
@@ -1,0 +1,100 @@
+package io.stargate.sgv2.jsonapi.service.cqldriver;
+
+import com.datastax.oss.driver.api.core.ConsistencyLevel;
+import com.datastax.oss.driver.api.core.connection.ClosedConnectionException;
+import com.datastax.oss.driver.api.core.connection.HeartbeatException;
+import com.datastax.oss.driver.api.core.context.DriverContext;
+import com.datastax.oss.driver.api.core.retry.RetryDecision;
+import com.datastax.oss.driver.api.core.servererrors.CoordinatorException;
+import com.datastax.oss.driver.api.core.servererrors.ReadFailureException;
+import com.datastax.oss.driver.api.core.servererrors.WriteFailureException;
+import com.datastax.oss.driver.api.core.session.Request;
+import com.datastax.oss.driver.internal.core.retry.DefaultRetryPolicy;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * A retry policy that retries to the same node in the query plan. DefaultRetryPolicy retries to the
+ * next node in the query plan in some of the error cases and this implementation overrides that
+ * behavior.
+ */
+public class SameNodeRetryPolicy extends DefaultRetryPolicy {
+
+  private static final Logger LOG = LoggerFactory.getLogger(DefaultRetryPolicy.class);
+  private final String logPrefix;
+
+  public SameNodeRetryPolicy(DriverContext context, String profileName) {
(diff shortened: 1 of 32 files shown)


kathirsvn:
The implementation of these methods are marked as deprecated on the RetryPolicy.java
https://github.com/apache/cassandra-java-driver/blob/8d5849cb38995b312f29314d18256c0c3e94cf07/core/src/main/java/com/datastax/oss/driver/api/core/retry/RetryPolicy.java#L184

But, the alternate method provided still uses the deprecated method internally
https://github.com/apache/cassandra-java-driver/blob/8d5849cb38995b312f29314d18256c0c3e94cf07/core/src/main/java/com/datastax/oss/driver/api/core/retry/RetryPolicy.java#L209

So, following the implementation in the `DefaultRetryPolicy` marking the overridden methods also as deprecated for now.

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/SameNodeRetryPolicy.java
```diff
@@ -0,0 +1,100 @@
...
+import com.datastax.oss.driver.api.core.servererrors.ReadFailureException;
+import com.datastax.oss.driver.api.core.servererrors.WriteFailureException;
+import com.datastax.oss.driver.api.core.session.Request;
+import com.datastax.oss.driver.internal.core.retry.DefaultRetryPolicy;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * A retry policy that retries to the same node in the query plan. DefaultRetryPolicy retries to the
```
tatu-at-datastax, line 18: Could you add a brief mention why this is useful for our case and more generally due to the way proxies like CQLRouter work?

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/SameNodeRetryPolicy.java
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    return decision;
+  }
+
+  /**
+   * {@inheritDoc}
+   *
+   * <p>This implementation rethrows read and write failures, and retries other errors on the same
+   * node.
```
tatu-at-datastax, line 82: What other types of exceptions are known to exist?

tatu-at-datastax:
> The implementation of these methods are marked as deprecated on the RetryPolicy.java https://github.com/apache/cassandra-java-driver/blob/8d5849cb38995b312f29314d18256c0c3e94cf07/core/src/main/java/com/datastax/oss/driver/api/core/retry/RetryPolicy.java#L184
> 
> But, the alternate method provided still uses the deprecated method internally https://github.com/apache/cassandra-java-driver/blob/8d5849cb38995b312f29314d18256c0c3e94cf07/core/src/main/java/com/datastax/oss/driver/api/core/retry/RetryPolicy.java#L209
> 
> So, following the implementation in the `DefaultRetryPolicy` marking the overridden methods also as deprecated for now.

That works, but is there a reason for not just overriding new, un-deprecated methods?
Presumably the idea of deprecation, new methods, is to let new code use new methods, without breaking backwards compatibility of old methods?

tatu-at-datastax (approved):
LGTM -- added some questions but nothing major.

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/SameNodeRetryPolicy.java
Yuqi-Du: I guess it is hard to know a thorough list of exceptions this sameNodeRetry can solve.
Probably we need answer from driver team.

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/SameNodeRetryPolicy.java
kathirsvn: Sure, added a comment in [this commit](https://github.com/stargate/jsonapi/pull/734/commits/3e05ad4110be60b6018ccc23c23f0cbbd8c8542e).

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/SameNodeRetryPolicy.java
kathirsvn: DefaultRetryPolicy which is provided by the driver covers all the retry-able errors. We just extend that and replace wherever it goes to the next node with the same node, so we can use this against cqlrouter like component which is a single node from the driver's perspective.

tatu-at-datastax: What I am suggesting is to add some note on comment itself as to not just what is re-thrown but what kinds are not -- not a complete list, but some idea. PR comments are not as easy to see as code comments.

So comment by @kathirsvn could be added (or some parts) in that comment.

kathirsvn:
Closing this PR since this is already taken care in #750

PR #736, opened 2023-12-14
Limit number of Collections created by both max-collections and SAIs available

### LINKED ISSUE #608: Improve max Collection limit verification to be based on available SAIs, not existing Tables

tatu-at-datastax:
**What this PR does**:

It will improve calculation of limits for new Collection creation to use 2 limits:

1. Total number of Collections (but not general Tables) allowed to be created
2. Number of SAIs that can be created (each Collection needs certain number)

instead of current subset of (1) (basically, number of Tables within Keyspace, which is inaccurate)

Fixes #608

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/.github/workflows/continuous-integration.yaml
+++ b/.github/workflows/continuous-integration.yaml
@@ -63,8 +63,8 @@ jobs:
     needs: [ build ]
     runs-on: ubuntu-latest
 
-    # max run time 30 minutes
-    timeout-minutes: 30
+    # max run time 45 minutes
+    timeout-minutes: 45
 
     strategy:
 
--- a/.github/workflows/continuous-integration.yaml
+++ b/.github/workflows/continuous-integration.yaml
@@ -40,8 +40,8 @@ jobs:
     name: Unit tests
     runs-on: ubuntu-latest
 
-    # max run time 10 minutes
-    timeout-minutes: 10
+    # max run time 15 minutes
+    timeout-minutes: 15
 
     steps:
       - uses: actions/checkout@v3
--- a/.github/workflows/continuous-integration.yaml
+++ b/.github/workflows/continuous-integration.yaml
@@ -63,8 +63,8 @@ jobs:
     needs: [ build ]
     runs-on: ubuntu-latest
 
-    # max run time 60 minutes
-    timeout-minutes: 60
+    # max run time 30 minutes
+    timeout-minutes: 30
 
     strategy:
 
--- a/.github/workflows/continuous-integration.yaml
+++ b/.github/workflows/continuous-integration.yaml
@@ -40,8 +40,8 @@ jobs:
     name: Unit tests
     runs-on: ubuntu-latest
 
-    # max run time 15 minutes
-    timeout-minutes: 15
+    # max run time 12 minutes
+    timeout-minutes: 12
 
     steps:
       - uses: actions/checkout@v3
@@ -63,8 +63,8 @@ jobs:
     needs: [ build ]
     runs-on: ubuntu-latest
 
-    # max run time 30 minutes
-    timeout-minutes: 30
+    # max run time 20 minutes
+    timeout-minutes: 20
 
     strategy:
 
--- a/.github/workflows/continuous-integration.yaml
+++ b/.github/workflows/continuous-integration.yaml
@@ -63,8 +63,8 @@ jobs:
     needs: [ build ]
     runs-on: ubuntu-latest
 
-    # max run time 20 minutes
-    timeout-minutes: 20
+    # max run time 30 minutes
+    timeout-minutes: 30
 
     strategy:
 
--- a/.github/workflows/continuous-integration.yaml
+++ b/.github/workflows/continuous-integration.yaml
@@ -63,8 +63,8 @@ jobs:
     needs: [ build ]
     runs-on: ubuntu-latest
 
-    # max run time 45 minutes
-    timeout-minutes: 45
+    # max run time 60 minutes
+    timeout-minutes: 60
 
     strategy:
 
--- a/.github/workflows/continuous-integration.yaml
+++ b/.github/workflows/continuous-integration.yaml
@@ -77,9 +77,9 @@ jobs:
         include:
           - type: docker
             profile: ''
-
+        exclude:
           - type: native
-            profile: '-Pnative'
+            #profile: '-Pnative'
 
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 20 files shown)


.github/workflows/continuous-integration.yaml
```diff
@@ -40,8 +40,8 @@ jobs:
     name: Unit tests
     runs-on: ubuntu-latest
 
-    # max run time 10 minutes
-    timeout-minutes: 10
+    # max run time 15 minutes
```
tatu-at-datastax, line 43: Increasing due to observed new timeouts. Didn't think anything here changed wrt unit tests, so maybe there's just incremental increase over time.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/CreateCollectionIntegrationTest.java
```diff
@@ -269,74 +268,6 @@ public void duplicateVectorCollectionNameWithDiffSetting() {
     }
   }
 
-  @Nested
```
tatu-at-datastax, line 272: Moved to separate class to be able to use different system property overrides for passing/failing tests.

src/test/java/io/stargate/sgv2/jsonapi/testresource/StargateTestResource.java
```diff
@@ -73,6 +73,14 @@ public Map<String, String> start() {
       propsBuilder.put("stargate.int-test.coordinator.cql-port", cqlPort);
       propsBuilder.put("stargate.int-test.cluster.persistence", getPersistenceModule());
 
+      // Many ITs create more Collections than default Max 5, use more than 50 indexes so:
```
tatu-at-datastax, line 76: Added this after observing majority of ITs failing with 5 Collection max: but with 10 Collections all pass.
May need to be tuned over time if ITs create more Collections.

src/test/java/io/stargate/sgv2/jsonapi/testresource/StargateTestResource.java
```diff
@@ -250,6 +258,14 @@ private String getAuthToken(String host, int authPort) {
     }
   }
 
+  public int getMaxCollectionsPerDBOverride() {
```
tatu-at-datastax, line 261: Defined as overridable methods for sub-classes (per-test custom test resources) to override.

.github/workflows/continuous-integration.yaml
tatu-at-datastax: Looks like this worked: unit test suite passed in bit over 13 minutes. But there's definitely some increase for time spend.

.github/workflows/continuous-integration.yaml
```diff
@@ -63,8 +63,8 @@ jobs:
     needs: [ build ]
     runs-on: ubuntu-latest
 
-    # max run time 20 minutes
-    timeout-minutes: 20
+    # max run time 30 minutes
```
tatu-at-datastax, line 66: Similar to unit tests, integration tests take longer -- should also not have increased a lot but here we are.

.github/workflows/continuous-integration.yaml
tatu-at-datastax: Run time is back to 8-10 minutes, but I think it is good to allow 15 minutes just so that slightly increased latency will not cause transient failures.

.github/workflows/continuous-integration.yaml
tatu-at-datastax: Not needed, will revert

.github/workflows/continuous-integration.yaml
```diff
@@ -40,8 +40,8 @@ jobs:
     name: Unit tests
     runs-on: ubuntu-latest
 
-    # max run time 10 minutes
-    timeout-minutes: 10
+    # max run time 12 minutes
```
tatu-at-datastax, line 43: Leaving in this increase as highest runtime I saw was 10 minutes (it varies, but don't want bogus failure just because CI system may be slow).

src/test/java/io/stargate/sgv2/jsonapi/api/v1/CreateCollectionTooManyIndexesIntegrationTest.java
```diff
@@ -0,0 +1,110 @@
...
+import org.junit.jupiter.api.TestClassOrder;
+
+/**
+ * Separate integration test from {@code CreateCollectionIntegrationTest} to test case of too many
+ * Indexes being created: that is, cannot create enough indexes for a new Collection.
+ */
+@QuarkusIntegrationTest
+@QuarkusTestResource(
+    value = CreateCollectionTooManyIndexesIntegrationTest.TooManyIndexesTestResource.class,
+    restrictToAnnotatedClass = true)
```
tatu-at-datastax, line 27: This property is VERY important -- without it this test resources is applied to EVERY TEST CLASS!

Yuqi-Du (approved):
LGTM

src/test/java/io/stargate/sgv2/jsonapi/api/v1/CreateCollectionTooManyIndexesIntegrationTest.java
```diff
@@ -0,0 +1,100 @@
...
+          .when()
+          .post(NamespaceResource.BASE_PATH, NS)
+          .then()
+          .statusCode(200)
+          .body("status", is(nullValue()))
+          .body("data", is(nullValue()))
+          .body(
+              "errors[0].message",
+              matchesPattern(
+                  "Too many indexes: cannot create a new collection; \\d+ indexes already created in database, maximum 20"))
```
vkarpov15: One minor suggestion: maybe indicate that deleting an existing collection would help to resolve this issue? When I see "Too many indexes", my first thought is to figure out how to drop an index, which isn't what we want users to do here.

vkarpov15 (commented):
One minor text comment, but LGTM overall

PR #738, opened 2023-12-18
Setting ignore bridge flag to true by default

kathirsvn:
**What this PR does**:
Since the JSON API backend is changed to use cql driver instead of gRPC bridge service on the coordinator, setting the `stargate.data-store.ignore-bridge` flag to true by default will make it not to try to connect to the gRPC bridge service during the startup to get the database features and instead of that it will be returned from the configuration properties.

**Checklist**
- [X] Changes manually tested


tatu-at-datastax (approved):

tatu-at-datastax:
Approved, triggered CI re-run

Yuqi-Du (approved):

PR #750, opened 2023-12-22
Retry policy changes

kathirsvn:
**What this PR does**:

- Added a custom retry policy that retries only on unavailable exception and rethrows in all other scenarios. This is useful while executing queries using a proxy layer that takes care of retries in all other scenarios.
- Marked INSERT IF NOT EXISTS,  DELETE IF, CAS and DDL statements as idempotent.

**Checklist**
- [X] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/CqlProxyRetryPolicy.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/CqlProxyRetryPolicy.java
@@ -1,0 +1,136 @@
+package io.stargate.sgv2.jsonapi.service.cqldriver;
+
+import static com.datastax.oss.driver.internal.core.retry.DefaultRetryPolicy.RETRYING_ON_UNAVAILABLE;
+
+import com.datastax.oss.driver.api.core.ConsistencyLevel;
+import com.datastax.oss.driver.api.core.context.DriverContext;
+import com.datastax.oss.driver.api.core.retry.RetryDecision;
+import com.datastax.oss.driver.api.core.retry.RetryPolicy;
+import com.datastax.oss.driver.api.core.retry.RetryVerdict;
+import com.datastax.oss.driver.api.core.servererrors.CoordinatorException;
+import com.datastax.oss.driver.api.core.servererrors.WriteType;
+import com.datastax.oss.driver.api.core.session.Request;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * This retry policy while executing cql statements is useful when the requests are delegated to a
+ * proxy layer which is responsible for retrying the requests. This policy will only retry once if
+ * the intermediate layer is unavailable for some reason. In all other cases, this will rethrow the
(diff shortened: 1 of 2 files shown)


src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/CqlProxyRetryPolicy.java
```diff
@@ -0,0 +1,136 @@
...
+import com.datastax.oss.driver.api.core.servererrors.CoordinatorException;
+import com.datastax.oss.driver.api.core.servererrors.WriteType;
+import com.datastax.oss.driver.api.core.session.Request;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * This retry policy while executing cql statements is useful when the requests are delegated to a
+ * proxy layer which is responsible for retrying the requests. This policy will only retry once if
```
tatu-at-datastax, line 19: But weren't we thinking that we do NOT want proxy layer doing retries, since we anyway require non-proxy retries & try to avoid "double retry"?
So is there need for this policy?

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/QueryExecutor.java
```diff
@@ -115,8 +116,10 @@ public Uni<AsyncResultSet> executeSchemaChange(SimpleStatement boundStatement) {
             cqlSessionCache
                 .getSession()
                 .executeAsync(
-                    boundStatement.setSerialConsistencyLevel(
-                        operationsConfig.queriesConfig().consistency().schemaChanges())));
+                    boundStatement
+                        .setIdempotent(true)
```
tatu-at-datastax, line 120: Are Schema change (DDL) really idempotent? I guess they can be if (and only if) we use "CREATE TABLE ... IF NOT EXISTS" and other LWT-based queries. Otherwise successful but timed out queries will be retried and fail (like trying to add same column again)

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/CqlProxyRetryPolicy.java
kathirsvn: No, we would to have the proxy layer do the retry and have the JSON API driver retry only when the proxy layer is unavailable.

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/QueryExecutor.java
kathirsvn: Agreed. All the DDL statements in JSON API have IF NOT EXISTS kind of check which makes them idempotent.

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/CqlProxyRetryPolicy.java
tatu-at-datastax: Ah ok. So whether this policy is used is separate from it being available for use. Makes sense.

PR #754, opened 2024-01-02
CQL Query logging changes.

kathirsvn:
**What this PR does**:
Enables CQL Query logging for slow and errored queries.

**Checklist**
- [X] Changes manually tested

### DIFF
--- a/src/main/resources/application.conf
+++ b/src/main/resources/application.conf
@@ -27,7 +27,6 @@ datastax-java-driver {
       enabled = [pool.open-connections, pool.available-streams, pool.in-flight, cql-messages, errors.request.unsent, errors.request.write-timeouts, errors.request.read-timeouts, errors.request.unavailables, retries.total, speculative-executions]
     }
   }
-  basic.request.timeout = 10 seconds
   advanced.request-tracker {
     classes = [RequestLogger]
     logs {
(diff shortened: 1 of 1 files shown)


src/main/resources/application.conf
```diff
@@ -28,6 +28,22 @@ datastax-java-driver {
...
+    classes = [RequestLogger]
+    logs {
+      success.enabled = false
+      slow {
+        threshold = 1 second
+        enabled = true
+      }
+      error.enabled = true
+      max-query-length = 500
+      show-values = true
```
jeffreyscarpenter, line 41: I think we should set this to false so that we don't get any customer data in the logs

jeffreyscarpenter (changes requested):

src/main/resources/application.conf
```diff
@@ -28,6 +28,22 @@ datastax-java-driver {
     }
   }
   basic.request.timeout = 10 seconds
+  advanced.request-tracker {
```
tatu-at-datastax, line 31: nit: should come before `basic.request.timeout` to keep things in alphabetic order.

src/main/resources/application.conf
tatu-at-datastax: That does seem safer default choice, +1.

kathirsvn: Agreed. Changed to `false` in [this commit](https://github.com/stargate/jsonapi/pull/754/commits/b70a3266bab05de5f9c68f73bd0963e3a3e399f2).

src/main/resources/application.conf
kathirsvn: Changed in [this commit](https://github.com/stargate/jsonapi/pull/754/commits/b70a3266bab05de5f9c68f73bd0963e3a3e399f2).

jeffreyscarpenter (approved):

PR #755, opened 2024-01-02
Add metrics for serialization to doc_json

### LINKED ISSUE #669: Add timing metrics for serialization (shredding output) to `doc_json`
Similar to #668 we may be able to produce metrics on serialization of updated Documents into `doc_json` (for storage): this is done after Shredding when inserting/upserting, or updating Documents.

(actual serialization of response is handled by Quarkus and would need to be reported by Q, possibly there's config setting but we cannot directly output such metrics).

Hazel-Datastax:
**What this PR does**:
Add metrics for serialization to `doc_json`

Fixes #669 

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/api/v1/metrics/JsonApiMetricsConfig.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/api/v1/metrics/JsonApiMetricsConfig.java
@@ -27,6 +27,14 @@ public interface JsonApiMetricsConfig {
   String sortType();
 
   @NotBlank
+  @WithDefault("serialization.json")
+  String serializationJson();
+
+  @NotBlank
+  @WithDefault("command.processor.serialization")
+  String serializationMetricsName();
+
+  @NotBlank
   @WithDefault("command.processor.process")
   String metricsName();
 
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/resolver/model/impl/InsertManyCommandResolver.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/resolver/model/impl/InsertManyCommandResolver.java
@@ -36,7 +36,9 @@ public class InsertManyCommandResolver implements CommandResolver<InsertManyComm
     ctx.tryVectorize(objectMapper.getNodeFactory(), command.documents());
 
     final List<WritableShreddedDocument> shreddedDocuments =
-        command.documents().stream().map(shredder::shred).toList();
+        command.documents().stream()
+            .map(doc -> shredder.shred(doc, command.getClass().getSimpleName()))
+            .toList();
 
     // resolve ordered
     InsertManyCommand.Options options = command.options();
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/shredding/Shredder.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/shredding/Shredder.java
@@ -4,6 +4,11 @@ import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.node.ArrayNode;
 import com.fasterxml.jackson.databind.node.ObjectNode;
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.Tag;
+import io.micrometer.core.instrument.Tags;
+import io.micrometer.core.instrument.Timer;
+import io.stargate.sgv2.jsonapi.api.v1.metrics.JsonApiMetricsConfig;
 import io.stargate.sgv2.jsonapi.config.DocumentLimitsConfig;
 import io.stargate.sgv2.jsonapi.config.constants.DocumentConstants;
 import io.stargate.sgv2.jsonapi.exception.ErrorCode;
@@ -36,10 +41,20 @@ public class Shredder {
 
   private final DocumentLimitsConfig documentLimits;
 
+  private final MeterRegistry meterRegistry;
+
+  private final JsonApiMetricsConfig jsonApiMetricsConfig;
+
   @Inject
-  public Shredder(ObjectMapper objectMapper, DocumentLimitsConfig documentLimits) {
+  public Shredder(
+      ObjectMapper objectMapper,
+      DocumentLimitsConfig documentLimits,
+      MeterRegistry meterRegistry,
+      JsonApiMetricsConfig jsonApiMetricsConfig) {
     this.objectMapper = objectMapper;
     this.documentLimits = documentLimits;
+    this.meterRegistry = meterRegistry;
+    this.jsonApiMetricsConfig = jsonApiMetricsConfig;
   }
 
   /**
@@ -48,11 +63,11 @@ public class Shredder {
    * @param document {@link JsonNode} to shred.
    * @return WritableShreddedDocument
    */
-  public WritableShreddedDocument shred(JsonNode document) {
-    return shred(document, null);
+  public WritableShreddedDocument shred(JsonNode document, String commandName) {
+    return shred(document, commandName, null);
   }
 
-  public WritableShreddedDocument shred(JsonNode doc, UUID txId) {
+  public WritableShreddedDocument shred(JsonNode doc, String commandName, UUID txId) {
     // Although we could otherwise allow non-Object documents, requirement
     // to have the _id (or at least place for it) means we cannot allow that.
     if (!doc.isObject()) {
@@ -70,8 +85,13 @@ public class Shredder {
     // Need to re-serialize document now that _id is normalized.
     // Also gets rid of pretty-printing (if any) and unifies escaping.
     try {
+      Timer.Sample sample = Timer.start(meterRegistry);
       // Important! Must use configured ObjectMapper for serialization, NOT JsonNode.toString()
       docJson = objectMapper.writeValueAsString(docWithId);
+      Tag commandTag = Tag.of(jsonApiMetricsConfig.command(), commandName);
+      Tag serializationTag = Tag.of(jsonApiMetricsConfig.serializationJson(), "true");
+      Tags tags = Tags.of(commandTag, serializationTag);
+      sample.stop(meterRegistry.timer(jsonApiMetricsConfig.serializationMetricsName(), tags));
     } catch (IOException e) { // never happens but signature exposes it
       throw new RuntimeException(e);
     }
--- a/src/test/java/io/stargate/sgv2/jsonapi/service/shredding/ShredderTest.java
+++ b/src/test/java/io/stargate/sgv2/jsonapi/service/shredding/ShredderTest.java
@@ -50,7 +50,7 @@ public class ShredderTest {
                       }
                       """;
       final JsonNode inputDoc = objectMapper.readTree(inputJson);
-      WritableShreddedDocument doc = shredder.shred(inputDoc);
+      WritableShreddedDocument doc = shredder.shred(inputDoc, "");
       assertThat(doc.id()).isEqualTo(DocumentId.fromString("abc"));
       List<JsonPath> expPaths =
           Arrays.asList(
@@ -120,7 +120,7 @@ public class ShredderTest {
                       }
                       """;
       final JsonNode inputDoc = objectMapper.readTree(inputJson);
-      WritableShreddedDocument doc = shredder.shred(inputDoc);
+      WritableShreddedDocument doc = shredder.shred(inputDoc, "");
 
       assertThat(doc.id()).isInstanceOf(DocumentId.StringId.class);
       // should be auto-generated UUID:
@@ -160,7 +160,7 @@ public class ShredderTest {
                       }
                       """;
       final JsonNode inputDoc = objectMapper.readTree(inputJson);
-      WritableShreddedDocument doc = shredder.shred(inputDoc);
+      WritableShreddedDocument doc = shredder.shred(inputDoc, "");
       assertThat(doc.id()).isEqualTo(DocumentId.fromBoolean(true));
 
       JsonNode jsonFromShredded = objectMapper.readTree(doc.docJson());
@@ -185,7 +185,7 @@ public class ShredderTest {
                       }
                       """;
       final JsonNode inputDoc = fromJson(inputJson);
-      WritableShreddedDocument doc = shredder.shred(inputDoc);
+      WritableShreddedDocument doc = shredder.shred(inputDoc, "");
       assertThat(doc.id()).isEqualTo(DocumentId.fromNumber(new BigDecimal(123L)));
 
       JsonNode jsonFromShredded = fromJson(doc.docJson());
@@ -206,7 +206,7 @@ public class ShredderTest {
     @Test
     public void shredSimpleWithNumberIdWithTrailingZeroes() {
       final String inputJson = "{\"_id\":30}";
-      WritableShreddedDocument doc = shredder.shred(fromJson(inputJson));
+      WritableShreddedDocument doc = shredder.shred(fromJson(inputJson), "");
       assertThat(doc.id()).isEqualTo(DocumentId.fromNumber(new BigDecimal(30L)));
       // Verify that we do NOT have '{"_id":3E+1}':
       assertThat(doc.docJson()).isEqualTo(inputJson);
@@ -230,7 +230,7 @@ public class ShredderTest {
               """
               .formatted(testTimestamp);
       final JsonNode inputDoc = fromJson(inputJson);
-      WritableShreddedDocument doc = shredder.shred(inputDoc);
+      WritableShreddedDocument doc = shredder.shred(inputDoc, "");
       assertThat(doc.id()).isEqualTo(DocumentId.fromNumber(new BigDecimal(123L)));
 
       JsonNode jsonFromShredded = fromJson(doc.docJson());
@@ -254,7 +254,8 @@ public class ShredderTest {
     public void badEJSONDate() {
       Throwable t =
           catchThrowable(
-              () -> shredder.shred(objectMapper.readTree("{ \"date\": { \"$date\": false } }")));
+              () ->
+                  shredder.shred(objectMapper.readTree("{ \"date\": { \"$date\": false } }"), ""));
 
       assertThat(t)
           .isNotNull()
@@ -266,7 +267,7 @@ public class ShredderTest {
     @Test
     public void badEmptyVectorData() {
       Throwable t =
-          catchThrowable(() -> shredder.shred(objectMapper.readTree("{ \"$vector\": [] }")));
+          catchThrowable(() -> shredder.shred(objectMapper.readTree("{ \"$vector\": [] }"), ""));
 
       assertThat(t)
           .isNotNull()
@@ -278,7 +279,7 @@ public class ShredderTest {
     public void badInvalidVectorData() {
       Throwable t =
           catchThrowable(
-              () -> shredder.shred(objectMapper.readTree("{ \"$vector\": [0.11, \"abc\"] }")));
+              () -> shredder.shred(objectMapper.readTree("{ \"$vector\": [0.11, \"abc\"] }"), ""));
 
       assertThat(t)
           .isNotNull()
@@ -292,7 +293,7 @@ public class ShredderTest {
           catchThrowable(
               () ->
                   shredder.shred(
-                      objectMapper.readTree("{ \"value\": { \"$unknownType\": 123 } }")));
+                      objectMapper.readTree("{ \"value\": { \"$unknownType\": 123 } }"), ""));
 
       assertThat(t)
           .isNotNull()
@@ -306,7 +307,7 @@ public class ShredderTest {
 
     @Test
     public void docBadJSONType() {
-      Throwable t = catchThrowable(() -> shredder.shred(objectMapper.readTree("[ 1, 2 ]")));
+      Throwable t = catchThrowable(() -> shredder.shred(objectMapper.readTree("[ 1, 2 ]"), ""));
 
       assertThat(t)
           .isNotNull()
@@ -318,7 +319,7 @@ public class ShredderTest {
     @Test
     public void docBadDocIdTypeArray() {
       Throwable t =
-          catchThrowable(() -> shredder.shred(objectMapper.readTree("{ \"_id\" : [ ] }")));
+          catchThrowable(() -> shredder.shred(objectMapper.readTree("{ \"_id\" : [ ] }"), ""));
 
       assertThat(t)
           .isNotNull()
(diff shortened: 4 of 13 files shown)


src/main/java/io/stargate/sgv2/jsonapi/api/v1/metrics/JsonApiMetricsConfig.java
```diff
@@ -26,6 +26,14 @@ public interface JsonApiMetricsConfig {
   @WithDefault("sort.type")
   String sortType();
 
+  @NotBlank
+  @WithDefault("serialization.json")
+  String serializationJson();
```
tatu-at-datastax, line 31: I understand there is need for one name configuration, but why 2?

src/main/java/io/stargate/sgv2/jsonapi/service/shredding/Shredder.java
```diff
@@ -70,8 +85,13 @@ public WritableShreddedDocument shred(JsonNode doc, UUID txId) {
     // Need to re-serialize document now that _id is normalized.
     // Also gets rid of pretty-printing (if any) and unifies escaping.
     try {
+      Timer.Sample sample = Timer.start(meterRegistry);
       // Important! Must use configured ObjectMapper for serialization, NOT JsonNode.toString()
       docJson = objectMapper.writeValueAsString(docWithId);
+      Tag commandTag = Tag.of(jsonApiMetricsConfig.command(), commandName);
+      Tag serializationTag = Tag.of(jsonApiMetricsConfig.serializationJson(), "true");
```
tatu-at-datastax, line 92: So, what is this tag used for? Is it just to be able to select metrics where JSON serialization timing is included?

src/main/java/io/stargate/sgv2/jsonapi/service/shredding/Shredder.java
```diff
@@ -70,8 +85,13 @@ public WritableShreddedDocument shred(JsonNode doc, UUID txId) {
...
+      Tags tags = Tags.of(commandTag, serializationTag);
+      sample.stop(meterRegistry.timer(jsonApiMetricsConfig.serializationMetricsName(), tags));
```
tatu-at-datastax, line 94: I think code for timing should probably start outside of `try` block, and `sample.stop()` called in `finally` block.
This way we would get timings for failing cases too?

I guess code here will get refactored if there is wrapper/callback.

src/main/java/io/stargate/sgv2/jsonapi/api/v1/metrics/JsonApiMetricsConfig.java
Hazel-Datastax: One is for the metrics name and one is for the tag that indicates its serialization or deserialization. I've renamed them to make it clear.

src/main/java/io/stargate/sgv2/jsonapi/api/v1/metrics/JsonShreddingMetricsFactory.java
```diff
@@ -0,0 +1,19 @@
...
+import jakarta.enterprise.inject.Produces;
+import jakarta.inject.Singleton;
+
+/**
+ * Configures the {@link JsonShreddingMetricsReporter} instance that going to be injectable and used
+ * in the app.
+ */
+public class JsonShreddingMetricsFactory {
+  /** Replaces the CDI producer for JsonShreddingMetricsReporter built into Quarkus. */
+  @Singleton
```
tatu-at-datastax, line 13: Probably cannot be singleton because we need new instance every time -- since it has state, `Timer.Sample`.

src/main/java/io/stargate/sgv2/jsonapi/service/shredding/Shredder.java
```diff
@@ -36,10 +37,38 @@ public class Shredder {
 
   private final DocumentLimitsConfig documentLimits;
 
+  private final JsonShreddingMetricsReporter jsonShreddingMetricsReporter;
```
tatu-at-datastax, line 40: No, cannot be assigned like this -- it is stateful object, we need new instance for every call.

src/main/java/io/stargate/sgv2/jsonapi/service/shredding/Shredder.java
```diff
@@ -36,10 +37,38 @@ public class Shredder {
...
+  }
+
+  /**
+   * Wrapper function for shred function to add serialization metrics
+   *
+   * @param doc {@link JsonNode} to shred.
+   * @param commandName command name
+   * @return WritableShreddedDocument
+   */
+  public WritableShreddedDocument shredWithMetrics(JsonNode doc, String commandName) {
```
tatu-at-datastax, line 59: These methods must take in `JsonShreddingMetricsReporter`

src/main/java/io/stargate/sgv2/jsonapi/api/v1/metrics/JsonApiMetricsConfig.java
```diff
@@ -26,6 +26,18 @@ public interface JsonApiMetricsConfig {
   @WithDefault("sort.type")
   String sortType();
 
+  @NotBlank
+  @WithDefault("json.shredding.type")
+  String jsonShreddingTypeTag();
+
+  @NotBlank
+  @WithDefault("json.string.length")
+  String jsonStringLength();
```
tatu-at-datastax, line 35: I think we need to figure out better names: we are not exposing String lengths: we do measure String length but report size of json document, for serialization.

So let's figure out good naming after resolving other issues

src/main/java/io/stargate/sgv2/jsonapi/api/v1/metrics/JsonShreddingMetricsFactory.java
```diff
@@ -0,0 +1,19 @@
...
+  @Produces
+  JsonShreddingMetricsReporter jsonShreddingMetricsReporter(
+      MeterRegistry meterRegistry, JsonApiMetricsConfig jsonApiMetricsConfig) {
+    return new JsonShreddingMetricsReporter(meterRegistry, jsonApiMetricsConfig);
```
tatu-at-datastax, line 17: Yes, this is better.

src/main/java/io/stargate/sgv2/jsonapi/api/v1/metrics/JsonShreddingMetricsReporter.java
```diff
@@ -0,0 +1,37 @@
...
+
+/** Reports the time and size of json shredding. */
+public class JsonShreddingMetricsReporter {
+
+  private final MeterRegistry meterRegistry;
+  private final JsonApiMetricsConfig jsonApiMetricsConfig;
+
+  private Timer.Sample sample;
+
+  @Inject
```
tatu-at-datastax, line 18: Remove this `@Inject`: instances are created by the factory, not injected by framework.

src/main/java/io/stargate/sgv2/jsonapi/api/v1/metrics/JsonShreddingMetricsReporter.java
```diff
@@ -0,0 +1,37 @@
...
+  public JsonShreddingMetricsReporter(
+      MeterRegistry meterRegistry, JsonApiMetricsConfig jsonApiMetricsConfig) {
+    this.meterRegistry = meterRegistry;
+    this.jsonApiMetricsConfig = jsonApiMetricsConfig;
+  }
+
+  public void startTimer() {
```
tatu-at-datastax, line 25: Let's call this `startMetrics()` (use of Timers is sort of implementation details from callers persepective)

src/main/java/io/stargate/sgv2/jsonapi/api/v1/metrics/JsonShreddingMetricsReporter.java
```diff
@@ -0,0 +1,37 @@
...
+    sample = Timer.start(meterRegistry);
+  }
+
+  public void stopTimer(WritableShreddedDocument doc, String commandName) {
```
tatu-at-datastax, line 29: and this `completeMetrics(...)`.

src/main/java/io/stargate/sgv2/jsonapi/service/shredding/Shredder.java
```diff
@@ -36,10 +37,38 @@ public class Shredder {
...
   private final DocumentLimitsConfig documentLimits;
 
+  private final JsonShreddingMetricsReporter jsonShreddingMetricsReporter;
+
   @Inject
-  public Shredder(ObjectMapper objectMapper, DocumentLimitsConfig documentLimits) {
+  public Shredder(
+      ObjectMapper objectMapper,
+      DocumentLimitsConfig documentLimits,
+      JsonShreddingMetricsReporter jsonShreddingMetricsReporter) {
```
tatu-at-datastax, line 46: I think that what needs to be injected is `JsonShreddingMetricsFactory`: and then shredder will hold a reference to that factory. Factory's reporter-creating factory method is then called when reporter instance is needed, later on.

src/main/java/io/stargate/sgv2/jsonapi/service/shredding/Shredder.java
```diff
@@ -36,10 +37,38 @@ public class Shredder {
...
+   * @param commandName command name
+   * @return WritableShreddedDocument
+   */
+  public WritableShreddedDocument shredWithMetrics(JsonNode doc, String commandName) {
+    return shredWithMetrics(doc, null, commandName);
+  }
+
+  public WritableShreddedDocument shredWithMetrics(JsonNode doc, UUID txId, String commandName) {
+    // Start the timer
+    jsonShreddingMetricsReporter.startTimer();
```
tatu-at-datastax, line 65: This is where we need to create reporter instance using metrics reporter factory object.

src/main/java/io/stargate/sgv2/jsonapi/api/v1/metrics/JsonApiMetricsConfig.java
```diff
@@ -26,6 +26,18 @@ public interface JsonApiMetricsConfig {
...
+
+  @NotBlank
+  @WithDefault("json.shredding.metrics")
+  String jsonShreddingMetricsName();
+
   @NotBlank
   @WithDefault("command.processor.process")
   String metricsName();
```
tatu-at-datastax, line 43: This is very generic name...

src/main/java/io/stargate/sgv2/jsonapi/api/v1/metrics/JsonApiMetricsConfig.java
```diff
@@ -26,6 +26,18 @@ public interface JsonApiMetricsConfig {
...
+  @WithDefault("json.shredding.type")
+  String jsonShreddingTypeTag();
+
+  @NotBlank
+  @WithDefault("json.string.length")
+  String jsonStringLength();
+
+  @NotBlank
+  @WithDefault("json.shredding.metrics")
+  String jsonShreddingMetricsName();
```
tatu-at-datastax, line 39: Maybe something like `jsonSerializationMetricsName()` with default of... `"json.shredding.serialization"`?

src/main/java/io/stargate/sgv2/jsonapi/api/v1/metrics/JsonApiMetricsConfig.java
```diff
@@ -26,6 +26,14 @@ public interface JsonApiMetricsConfig {
   @WithDefault("sort.type")
   String sortType();
 
+  @NotBlank
+  @WithDefault("json.serialized.size")
```
tatu-at-datastax, line 30: Should this also include `shredding`, so `json.shredding.serialization.size` ?
And then method like `jsonShreddingSerializedSize()`

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/model/impl/InsertManyCommandResolver.java
```diff
@@ -38,7 +38,12 @@ public Operation resolveCommand(CommandContext ctx, InsertManyCommand command) {
 
     final DocumentProjector projection = ctx.indexingProjector();
     final List<WritableShreddedDocument> shreddedDocuments =
-        command.documents().stream().map(doc -> shredder.shred(doc, null, projection)).toList();
+        command.documents().stream()
+            .map(
+                doc ->
+                    shredder.shredWithMetrics(
+                        doc, null, projection, command.getClass().getSimpleName()))
+            .toList();
```
tatu-at-datastax, line 46: I wonder... this will produce lots of individual metrics. I wonder if we should try to measure total time and size of all documents, instead of one by one?

Doing that would require more work here, to have `shredWithMetrics()` that takes `List` of documents, and then measures time over processing all, and counts total sum of size.

I think that would be more useful than having 20 different metrics point for a single command.

If we do that, it might also require reporting 3rd metric (number of Documents) -- which is actually useful metric on its own, I think. It'd be `1` for InsertOne, and then something else for `InsertMany`.

tatu-at-datastax (changes requested):
Looks good overall, but I realized that `insertMany()` metrics probably need to be reported as total over all documents, not separate for each.

We also need to finalize naming; I suggested one possible way that might work.

tatu-at-datastax:
Quick question: we shred during `UpdateXxx` and `FindOneAndXxx` commands too: would these not get reported?

Hazel-Datastax:
> Quick question: we shred during `UpdateXxx` and `FindOneAndXxx` commands too: would these not get reported?

`UpdateXxx` and `FindOneAndXxx` commands will return `ReadAndUpdateOperation` and use shred method in there.

src/main/java/io/stargate/sgv2/jsonapi/service/processor/MeteredCommandProcessor.java
```diff
@@ -285,7 +285,9 @@ public MeterFilter enableHistogram() {
       public DistributionStatisticConfig configure(
           Meter.Id id, DistributionStatisticConfig config) {
         if (id.getName().startsWith(jsonApiMetricsConfig.metricsName())
-            || id.getName().startsWith(HISTOGRAM_METRICS_NAME)) {
+            || id.getName().startsWith(HISTOGRAM_METRICS_NAME)
+            || id.getName()
+                .startsWith(jsonApiMetricsConfig.jsonSerializationPerformanceMetricsName())) {
```
tatu-at-datastax, line 290: Should this only apply to timing? Metrics?

src/main/java/io/stargate/sgv2/jsonapi/service/shredding/Shredder.java
```diff
@@ -37,10 +36,62 @@ public class Shredder {
...
+  public List<WritableShreddedDocument> shredMultipleJsonWithMetrics(
+      List<JsonNode> docs, DocumentProjector indexProjector, String commandName) {
+    // Start the metrics
+    JsonShreddingMetricsReporter jsonShreddingMetricsReporter =
+        jsonShreddingMetricsReporterFactory.jsonShreddingMetricsReporter();
+    jsonShreddingMetricsReporter.startMetrics();
+    // Perform the shredding operation
+    List<WritableShreddedDocument> result = new ArrayList<>(docs.size());
+    int totalJsonStringLen = 0;
+    int numberOfDocs = 0;
```
tatu-at-datastax, line 85: Could you just use `result.size()` instead of keeping count?

src/main/java/io/stargate/sgv2/jsonapi/api/v1/metrics/JsonShreddingMetricsReporter.java
```diff
@@ -0,0 +1,39 @@
...
+      MeterRegistry meterRegistry, JsonApiMetricsConfig jsonApiMetricsConfig) {
+    this.meterRegistry = meterRegistry;
+    this.jsonApiMetricsConfig = jsonApiMetricsConfig;
+  }
+
+  public void startMetrics() {
+    sample = Timer.start(meterRegistry);
+  }
+
+  public void completeMetrics(int totalJsonStringLen, int numberOfDocs, String commandName) {
```
tatu-at-datastax, line 26: Since we can now get multiple documents, should probably have

    long totalJsonStringLen

instead of `int`.

src/main/java/io/stargate/sgv2/jsonapi/service/shredding/Shredder.java
```diff
@@ -37,10 +36,62 @@ public class Shredder {
...
+  }
+
+  /**
+   * Wrapper function for shred function to add serialization metrics
+   *
+   * @param doc {@link JsonNode} to shred.
+   * @param commandName command name
+   * @return WritableShreddedDocument
+   */
+  public WritableShreddedDocument shredWithMetrics(JsonNode doc, String commandName) {
```
tatu-at-datastax, line 58: Is this still needed, when we have other 2 methods?

src/test/java/io/stargate/sgv2/jsonapi/service/shredding/ShredderTest.java
```diff
@@ -599,6 +600,90 @@ public void shredWithHugeNonIndexedString() throws Exception {
...
+      final String inputJson =
+          """
+                  {
+                    "name" : "Bob"
+                  }
+                  """;
+      final JsonNode inputDoc = objectMapper.readTree(inputJson);
+      List<JsonNode> listDocs = Arrays.asList(inputDoc, inputDoc);
+      shredder.shredMultipleJsonWithMetrics(
+          listDocs, DocumentProjector.identityProjector(), "testMultipleJsonCommend");
```
tatu-at-datastax, line 656: nit: Commend -> Command

src/main/java/io/stargate/sgv2/jsonapi/api/v1/metrics/JsonApiMetricsConfig.java
```diff
@@ -26,6 +26,18 @@ public interface JsonApiMetricsConfig {
...
+  @NotBlank
+  @WithDefault("json.shredding.serialization.total.documents")
+  String jsonShreddingSerializedTotalDocuments();
+
+  @NotBlank
+  @WithDefault("json.shredding.serialization.size")
+  String jsonShreddingSerializedSize();
+
+  @NotBlank
+  @WithDefault("json.shredding.serialization.performance")
```
tatu-at-datastax, line 38: I think "performance" is ambiguous. Could be "time.msecs" instead.

I'll ask @jeffreyscarpenter what he thinks would be nice and consistent naming for three things here (see above):

1. Number of documents serialized for operation (1 for insertOne(), 1-20 for insertMany())
2. Total serialized size of documents of operation (in Java chars)
3. Total time taken by serialization of all docs for operation

src/main/java/io/stargate/sgv2/jsonapi/service/shredding/Shredder.java
```diff
@@ -37,10 +36,62 @@ public class Shredder {
...
+
+  public List<WritableShreddedDocument> shredMultipleJsonWithMetrics(
+      List<JsonNode> docs, DocumentProjector indexProjector, String commandName) {
+    // Start the metrics
+    JsonShreddingMetricsReporter jsonShreddingMetricsReporter =
+        jsonShreddingMetricsReporterFactory.jsonShreddingMetricsReporter();
+    jsonShreddingMetricsReporter.startMetrics();
+    // Perform the shredding operation
+    List<WritableShreddedDocument> result = new ArrayList<>(docs.size());
+    int totalJsonStringLen = 0;
```
tatu-at-datastax, line 84: As per other note, may want to use `long` instead.

tatu-at-datastax (approved):
Looks good overall: couple of minor suggestions plus I'd like to have Jeff's feedback on metrics naming just to finalize those.

src/main/java/io/stargate/sgv2/jsonapi/api/v1/metrics/JsonApiMetricsConfig.java
```diff
@@ -26,6 +26,18 @@ public interface JsonApiMetricsConfig {
...
+  String jsonSerializationPerformanceMetricsName();
+
   @NotBlank
   @WithDefault("command.processor.process")
   String metricsName();
```
tatu-at-datastax, line 43: I know this is existing code, but I think name here is way too generic. It does not matter for actual metrics name selection, but I think it should be changed to something more meaningful. Like:

``
   String commandProcessorMetricsName();
```

Comment (jeffreyscarpenter) [reply src/main/java/io/stargate/sgv2/jsonapi/api/v1/metrics/JsonApiMetricsConfig.java:43]: > I know this is existing code, but I think name here is way too generic. It does not matter for actual metrics name selection, but I think it should be changed to something more meaningful. Like:
> 
> ```
>    String commandProcessorMetricsName();
> ```

I think we should leave this alone, we've already got downstream metrics processing and dashboards built off of the current name.

File: src/main/java/io/stargate/sgv2/jsonapi/api/v1/metrics/JsonApiMetricsConfig.java (hunk 23 above)
Comment (jeffreyscarpenter) [reply src/main/java/io/stargate/sgv2/jsonapi/api/v1/metrics/JsonApiMetricsConfig.java:38]: > 1. Number of documents serialized for operation (1 for insertOne(), 1-20 for insertMany())
I'm ok with the current name `json.shredding.serialization.total.documents`, but would also be ok with `json.shredding.serialization.documents`.

> 2. Total serialized size of documents of operation (in Java chars)
I'm ok with the current name `json.shredding.serialization.size`

> 3. Total time taken by serialization of all docs for operation
For other histograms we use in Cassandra and Stargate, I see terms like "latency" or "duration" used frequently. So `json.shredding.serialization.duration` seems good.

File: src/main/java/io/stargate/sgv2/jsonapi/api/v1/metrics/JsonApiMetricsConfig.java [hunk 26]
```diff
@@ -26,6 +26,18 @@ public interface JsonApiMetricsConfig {
   @WithDefault("sort.type")
   String sortType();
 
+  @NotBlank
+  @WithDefault("json.shredding.serialization.total.documents")
+  String jsonShreddingSerializedTotalDocuments();
```
jeffreyscarpenter, line 31: I assume this is a counter metric?

src/main/java/io/stargate/sgv2/jsonapi/api/v1/metrics/JsonApiMetricsConfig.java
```diff
@@ -26,6 +26,18 @@ public interface JsonApiMetricsConfig {
...
+
+  @NotBlank
+  @WithDefault("json.shredding.serialization.size")
+  String jsonShreddingSerializedSize();
```
jeffreyscarpenter, line 35: Is this also a counter?

jeffreyscarpenter: @amorton can you comment on what we were hoping to track with this? Does it make more sense as a counter or some other metric type?

src/main/java/io/stargate/sgv2/jsonapi/service/processor/MeteredCommandProcessor.java
Hazel-Datastax: Yeah... Since we also have size metrics, should I enable the histogram for it?

Hazel-Datastax:
Create another PR #829 to continue the metrics building

PR #760, opened 2024-01-03
`$not` operator

maheshrajamani:
**What this PR does**:

Fixes #<issue number>

**Checklist**
- [x] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/filter/ArrayComparisonOperator.java
```diff
@@ -3,7 +3,9 @@
 /** List of element level operator that can be used in Filter clause */
 public enum ArrayComparisonOperator implements FilterOperator {
   ALL("$all"),
-  SIZE("$size");
+  SIZE("$size"),
+  // Will not be supported from outside
+  NOTANY("$notany");
```
maheshrajamani, line 8: This is used only for negation for $all, not exposed as a filter operator

tatu-at-datastax: Maybe add this exact explanation in Java comment, as that is very helpful.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/filter/ComparisonExpression.java
```diff
@@ -33,6 +33,32 @@ public List<FilterOperation<?>> getFilterOperations() {
...
+
+  /**
+   * This method is used to flip the operand value when not operator is applied, e.g. $exists, $size
+   */
+  private JsonLiteral<?> getFlippedOperandValue(FilterOperator operator, JsonLiteral<?> operand) {
+    if (operator == ElementComparisonOperator.EXISTS) {
+      return new JsonLiteral<Boolean>(!((Boolean) operand.value()), operand.type());
+    } else if (operator == ArrayComparisonOperator.SIZE) {
+      return new JsonLiteral<BigDecimal>(
+          ((BigDecimal) operand.value()).multiply(new BigDecimal(-1)), operand.type());
```
tatu-at-datastax, line 56: If this is done to flip between 1 and -1, more efficient way would be to use `BigDecimal.negate()` (https://docs.oracle.com/javase/8/docs/api/java/math/BigDecimal.html#negate-- )

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/filter/LogicalExpression.java
```diff
@@ -25,10 +26,67 @@ public String getOperator() {
...
+    }
+    comparisonExpressions.clear();
+  }
+
+  private void flip() {
+    if (logicalRelation == LogicalOperator.AND) {
+      logicalRelation = LogicalOperator.OR;
+    } else if (logicalRelation == LogicalOperator.OR) {
+      logicalRelation = LogicalOperator.AND;
+    }
```
tatu-at-datastax, line 81: Is it possible this could be called for `NOT`? And if so, is "do nothing" what should be done (if it shouldn't be called, should probably throw `IllegalStateException` in `else` case.

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/model/impl/matcher/FilterableResolver.java
```diff
@@ -292,12 +295,27 @@ public static List<DBFilterBase> findDynamic(CaptureExpression captureExpression
...
       if (captureExpression.marker() == SIZE_GROUP) {
         BigDecimal bigDecimal = (BigDecimal) filterOperation.operand().value();
-        filters.add(new DBFilterBase.SizeFilter(captureExpression.path(), bigDecimal.intValue()));
+        // Flipping size operator will multiply the value by -1
+        // Negative means check array_size[?] != ?
+        int size = bigDecimal.intValue();
+        DBFilterBase.MapFilterBase.Operator operator;
+        if (size > 0) {
+          operator = DBFilterBase.MapFilterBase.Operator.MAP_EQUALS;
+        } else {
```
tatu-at-datastax, line 314: Ah. So here the earlier negative value for `$size` (multiply by -1) is converted to "not equals $size". Makes sense.

tatu-at-datastax (approved):
LGTM: added some questions, but nothing blocking merge.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/filter/LogicalExpression.java
maheshrajamani: We resolve depth first handled `NOT` by flipping the operators and moved the expression to the parent logical expression. Because of that flip() will not have any `NOT` logical expression below it.

tatu-at-datastax: Ok, thanks.

tatu-at-datastax (approved):
LGTM

vkarpov15 (commented):
I'm curious about why `$not` is an array. In MongoDB, $not is an object like the following:

```javascript
find({
  num: {
    $not: {
      $lt: 42
    }
  }
}); // Equivalent to `find({ num: { $gte: 42 } })`
```

Is there a reason for making $not an array?

maheshrajamani:
> I'm curious about why `$not` is an array. In MongoDB, $not is an object like the following:

I made it in line with `$and` and `$or` operator. It does `and` of conditions below it. Do we have to change this?

tatu-at-datastax:
> > I'm curious about why `$not` is an array. In MongoDB, $not is an object like the following:
> 
> I made it in line with `$and` and `$or` operator. It does `and` of conditions below it. Do we have to change this?

Hmmmh. Good point, thank you for pointing it out. It is interesting Mongo has this discrepancy: `$and` and `$or` take Array, `$not` for some reason Object. Go figure.

Earlier we wanted to be highly Mongo-compatible, and I would have suggested we follow its lead.
But now we are moving away from that.
Still, we are mostly extending things and not changing handling of features Mongo has.
So I am not 100% sure which way we should go.

I think we need to discuss this on our daily meeting @maheshrajamani see what Aaron thinks.

tatu-at-datastax:
Based on discussion with @amorton, I think the preferred option is to use JSON Object, same as how Mongo behaves.
I guess it also makes sense logically, come to think of it: `not` is typically unary operation whereas `and` and `or` are binary. So taking in Array would be more likely implicit "not and" (and possibly more confusing).

vkarpov15:
Another reason why we should make `$not` a unary operator is that [Mongoose's filter casting logic assumes `$not` is unary](https://github.com/Automattic/mongoose/blob/265741752a2834db806def4520091b3b318f73f4/lib/cast.js#L331-L350). In the following script, Mongoose throws a CastError `CastError: Cast to number failed for value "[ { '$gt': 42 } ]" (type Array) at path "age" for model "Test"`:

```javascript
'use strict';

const mongoose = require('mongoose');

mongoose.set('debug', true);

run().catch(err => console.log(err));

async function run() {
  await mongoose.connect('mongodb://127.0.0.1:27017/mongoose_test');

  const schema = new mongoose.Schema({ age: Number });
  const Test = mongoose.model('Test', schema);

  await Test.find({ age: { $not: [{ $gt: 42 }] } });
  console.log('Done');
}
```

However, `await Test.find({ age: { $not: { $gt: 42 } } });` (object rather than array for `$not`) casts correctly. So if we want to be compatible with what Mongoose expects, then `$not` should be unary.

maheshrajamani:
@vkarpov15 @tatu-at-datastax Have made the `$not` to accept one object as value. If you want to do another review of it.

tatu-at-datastax (approved):

src/test/java/io/stargate/sgv2/jsonapi/api/v1/FindIntegrationTest.java
```diff
@@ -1727,6 +1727,220 @@ public void NeSubdoc() {
...
+          .then()
+          .statusCode(200)
+          .body("status", is(nullValue()))
+          .body("errors", is(nullValue()))
+          .body("data.documents", not(containsInAnyOrder(expected)))
+          .body("data.documents", hasSize(5));
+    }
+
+    @Test
+    public void multipleNotCancelling() {
```
vkarpov15, line 1856: Nice, I'm glad you tested this. I checked whether multiple `$not` works in MongoDB and Mongoose: MongoDB supports it, but Mongoose doesn't. Something for Mongoose to fix.

PR #762, opened 2024-01-03
Fix #744: increase maximum property name length (48->100), add new path limit (250)

### LINKED ISSUE #744: Tune maximum property name length (from 48 to 100), add separate setting for max-total-length (250 or 400?)
Currently `DocumentLimitsConfig` has single setting -- `maxPropertyNameLength` -- defaulting to maximum of 48 characters per name "segment" (individual property name for Document or sub-Document property). This is too short for some use cases; for example for hex-encoded SHA-256 keys (64 characters). So we should probably increase that to, say, 100 characters.

But there is also a gap: there is no maximum total length for resulting dotted notation -- it is possible to theoretically get up to 8 times (max nesting) times 48 characters (8 * 48 + 7 == 391 characters).
It would make sense to add a new explicit constraint for this. We have two main options: either

1. Due to existing usage, we may need to allow 400 characters total length.
2. otherwise something like 250 would make more sense

tatu-at-datastax:
**What this PR does**:

Increased default limit for maximum property name to 100 (from 48); adds new "total" (path) length limit of 250.

Fixes #744

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/src/test/java/io/stargate/sgv2/jsonapi/api/v1/InsertIntegrationTest.java
+++ b/src/test/java/io/stargate/sgv2/jsonapi/api/v1/InsertIntegrationTest.java
@@ -569,18 +569,9 @@ public class InsertIntegrationTest extends AbstractCollectionIntegrationTestBase
       final String LONGEST_NAME = "a".repeat(DocumentLimitsConfig.DEFAULT_MAX_PROPERTY_NAME_LENGTH);
       ObjectNode doc = MAPPER.createObjectNode();
       doc.put(DocumentConstants.Fields.DOC_ID, "docWithLongName");
-      // Max property name: 48 characters
+      // Max property name: 100 characters
       doc.put(LONGEST_NAME, "stuff");
       _verifyInsert("docWithLongName", doc);
-
-      // But let's also ensure nested names longer than this are allowed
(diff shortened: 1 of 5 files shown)


vkarpov15 (commented):
LGTM so far

src/test/java/io/stargate/sgv2/jsonapi/api/v1/InsertIntegrationTest.java
```diff
@@ -569,25 +569,18 @@ public void insertLongestValidName() {
       final String LONGEST_NAME = "a".repeat(DocumentLimitsConfig.DEFAULT_MAX_PROPERTY_NAME_LENGTH);
       ObjectNode doc = MAPPER.createObjectNode();
       doc.put(DocumentConstants.Fields.DOC_ID, "docWithLongName");
-      // Max property name: 48 characters
+      // Max property name: 100 characters
       doc.put(LONGEST_NAME, "stuff");
       _verifyInsert("docWithLongName", doc);
-
```
tatu-at-datastax, line 575: Moved to separate test, so not needed here any more.

amorton (approved):
Looks good to me

Issue #766, opened 2024-01-04
[Indexing options] Add indexing option to cache


maheshrajamani:
Add the indexing option to CollectionProperties in the cache

tatu-at-datastax:
I think this is complete, closing

PR #771, opened 2024-01-05
Changes for indexing option - create collection and cache

### LINKED ISSUE #765: [Indexing options] Create collection changes to add indexing options
Create collection will support indexing options like below. 

`{
"createCollection": {
		"name": "monkey_research",
    		"options" : {
      		"indexing" : {
        			"allow" : [],
				"deny" : ["content"],
}
}}}
`

In the above example: 

- "indexing" is an optional field in the "options" object
- "indexing" only supports the fields "allow" and "deny"
- "allow" and "deny" can have the following values: 

1. undefined 
2. Null
3. An empty array
4. An array of strings

- Max length of the array is the max number of fields we allow in a document
- Documents stored in a collection configured like this will index all fields in a document other than one called "content"
- "options" and the "indexing" sub object are present when the collection definition is read using findCollections.

maheshrajamani:
**What this PR does**:
Added the option to provide fields for indexing - allowed or denied
Add the indexing option to cache

Fixes #765 #766 

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/CreateCollectionCommand.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/CreateCollectionCommand.java
@@ -104,14 +104,14 @@ public record CreateCollectionCommand(
           throw new JsonApiException(
               ErrorCode.INVALID_INDEXING_USAGE,
               ErrorCode.INVALID_INDEXING_USAGE.getMessage()
-                  + " - allow and deny cannot be used together");
+                  + " - `allow` and `deny` cannot be used together");
         }
 
         if (allow() == null && deny() == null) {
           throw new JsonApiException(
               ErrorCode.INVALID_INDEXING_USAGE,
               ErrorCode.INVALID_INDEXING_USAGE.getMessage()
-                  + " - allow or deny should be provided");
+                  + " - `allow` or `deny` should be provided");
         }
 
         if (allow() != null) {
@@ -120,7 +120,7 @@ public record CreateCollectionCommand(
             throw new JsonApiException(
                 ErrorCode.INVALID_INDEXING_USAGE,
                 ErrorCode.INVALID_INDEXING_USAGE.getMessage()
-                    + " - allow cannot contain duplicates");
+                    + " - `allow` cannot contain duplicates");
           }
         }
 
@@ -130,7 +130,7 @@ public record CreateCollectionCommand(
             throw new JsonApiException(
                 ErrorCode.INVALID_INDEXING_USAGE,
                 ErrorCode.INVALID_INDEXING_USAGE.getMessage()
-                    + " - deny cannot contain duplicates");
+                    + " - `deny` cannot contain duplicates");
           }
         }
       }
--- a/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/CreateCollectionCommand.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/CreateCollectionCommand.java
@@ -6,6 +6,7 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.annotation.JsonTypeName;
 import io.stargate.sgv2.jsonapi.api.model.command.NamespaceCommand;
 import jakarta.validation.constraints.*;
+import java.util.List;
 import javax.annotation.Nullable;
 import org.eclipse.microprofile.openapi.annotations.enums.SchemaType;
 import org.eclipse.microprofile.openapi.annotations.media.Schema;
@@ -40,7 +41,15 @@ public record CreateCollectionCommand(
               description = "Embedding api configuration to support `$vectorize`",
               type = SchemaType.OBJECT,
               implementation = VectorSearchConfig.class)
-          VectorizeConfig vectorize) {
+          VectorizeConfig vectorize,
+      @JsonInclude(JsonInclude.Include.NON_NULL)
+          @Nullable
+          @Schema(
+              description =
+                  "Optional indexing configuration to provide allow/deny list of fields for indexing",
+              type = SchemaType.OBJECT,
+              implementation = IndexingConfig.class)
+          IndexingConfig indexing) {
 
     public record VectorSearchConfig(
         @Positive(message = "dimension should be greater than `0`")
@@ -69,6 +78,22 @@ public record CreateCollectionCommand(
         this.metric = metric == null ? "cosine" : metric;
       }
     }
+
+    public record IndexingConfig(
+        @JsonInclude(JsonInclude.Include.NON_NULL)
+            @Schema(
+                description = "List of allowed indexing fields",
+                type = SchemaType.ARRAY,
+                implementation = String.class)
+            @Nullable
+            List<String> allow,
+        @JsonInclude(JsonInclude.Include.NON_NULL)
+            @Schema(
+                description = "List of denied indexing fields",
+                type = SchemaType.ARRAY,
+                implementation = String.class)
+            @Nullable
+            List<String> deny) {}
 
     public record VectorizeConfig(
         @NotNull
--- a/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/CreateCollectionCommand.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/CreateCollectionCommand.java
@@ -102,15 +102,15 @@ public record CreateCollectionCommand(
       public void validate() {
         if (allow() != null && deny() != null) {
           throw new JsonApiException(
-              ErrorCode.INVALID_INDEXING_USAGE,
-              ErrorCode.INVALID_INDEXING_USAGE.getMessage()
+              ErrorCode.INVALID_INDEXING_DEFINITION,
+              ErrorCode.INVALID_INDEXING_DEFINITION.getMessage()
                   + " - `allow` and `deny` cannot be used together");
         }
 
         if (allow() == null && deny() == null) {
           throw new JsonApiException(
-              ErrorCode.INVALID_INDEXING_USAGE,
-              ErrorCode.INVALID_INDEXING_USAGE.getMessage()
+              ErrorCode.INVALID_INDEXING_DEFINITION,
+              ErrorCode.INVALID_INDEXING_DEFINITION.getMessage()
                   + " - `allow` or `deny` should be provided");
         }
 
@@ -118,8 +118,8 @@ public record CreateCollectionCommand(
           Set<String> dedupe = new HashSet<>(allow());
           if (dedupe.size() != allow().size()) {
             throw new JsonApiException(
-                ErrorCode.INVALID_INDEXING_USAGE,
-                ErrorCode.INVALID_INDEXING_USAGE.getMessage()
+                ErrorCode.INVALID_INDEXING_DEFINITION,
+                ErrorCode.INVALID_INDEXING_DEFINITION.getMessage()
                     + " - `allow` cannot contain duplicates");
           }
         }
@@ -128,8 +128,8 @@ public record CreateCollectionCommand(
           Set<String> dedupe = new HashSet<>(deny());
           if (dedupe.size() != deny().size()) {
             throw new JsonApiException(
-                ErrorCode.INVALID_INDEXING_USAGE,
-                ErrorCode.INVALID_INDEXING_USAGE.getMessage()
+                ErrorCode.INVALID_INDEXING_DEFINITION,
+                ErrorCode.INVALID_INDEXING_DEFINITION.getMessage()
                     + " - `deny` cannot contain duplicates");
           }
         }
--- a/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/CreateCollectionCommand.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/CreateCollectionCommand.java
@@ -5,8 +5,12 @@ import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.annotation.JsonTypeName;
 import io.stargate.sgv2.jsonapi.api.model.command.NamespaceCommand;
+import io.stargate.sgv2.jsonapi.exception.ErrorCode;
+import io.stargate.sgv2.jsonapi.exception.JsonApiException;
 import jakarta.validation.constraints.*;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 import javax.annotation.Nullable;
 import org.eclipse.microprofile.openapi.annotations.enums.SchemaType;
 import org.eclipse.microprofile.openapi.annotations.media.Schema;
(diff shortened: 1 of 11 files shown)


src/main/java/io/stargate/sgv2/jsonapi/service/resolver/model/impl/CreateCollectionCommandResolver.java
```diff
@@ -51,44 +54,119 @@ public Class<CreateCollectionCommand> getCommandClass() {
...
-            ErrorCode.VECTOR_SEARCH_NOT_AVAILABLE.getMessage());
+
+    if (command.options() != null) {
+      String vectorize = null;
+      String indexing = null;
+      int vectorSize = 0;
+      String function = null;
+
+      // handling indexing options
+      if (command.options().indexing() != null) {
```
tatu-at-datastax, line 65: Would it make more sense to move this validation in `IndexingConfig` instance itself? Like `validate()` method or something -- this would encapsulate logic bit better.
Or maybe better yet, add `create()` method in `IndexingConfig` to encapsulate these, validate during construction?

Also: I think `allow` and `deny` are always both non-null since they are initialized as empty `Set`s?

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/CreateCollectionCommand.java
```diff
@@ -70,6 +79,22 @@ public VectorSearchConfig(Integer dimension, String metric) {
...
+                implementation = String.class)
+            @Nullable
+            List<String> allow,
+        @JsonInclude(JsonInclude.Include.NON_NULL)
+            @Schema(
+                description = "List of denied indexing fields",
+                type = SchemaType.ARRAY,
+                implementation = String.class)
+            @Nullable
+            List<String> deny) {}
```
tatu-at-datastax, line 96: As per my note below, I think validation could be encapsulated within config class itself; either in (canonical) constructor, or, maybe cleaner, in a `create(...)` method that verifies everything is good (and if not throws exception), then calls canonical constructor.

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/model/impl/CreateCollectionCommandResolver.java
```diff
@@ -51,44 +54,119 @@ public Class<CreateCollectionCommand> getCommandClass() {
...
+          }
+          if (vectorize != null) {
+            objectNode.put("vectorize", objectMapper.readTree(vectorize));
+          }
         } catch (JsonProcessingException e) {
-          // This should never happen because the object is extracted from json request
           throw new RuntimeException(e);
         }
+
+        comment = objectNode.toString();
```
tatu-at-datastax, line 150: Is there a reason we don't just use original JSON string?

tatu-at-datastax (commented):
Looks good overall, added some suggestions for refactoring.

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/model/impl/CreateCollectionCommandResolver.java
maheshrajamani: Original json is deserialized to object., here we needed the config as json to be stored in the comment

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/CollectionSettings.java
```diff
@@ -86,21 +103,18 @@ public static CollectionSettings getCollectionSettings(
...
+      final String comment = (String) table.getOptions().get(CqlIdentifier.fromInternal("comment"));
+      return createCollectionSettings(
           collectionName,
           vectorEnabled,
           0,
           CollectionSettings.SimilarityFunction.UNDEFINED,
-          null,
-          null);
+          comment,
+          objectMapper);
```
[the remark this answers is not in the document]
maheshrajamani: `mapper.readTree(comment)` needs to handle `JsonProcessingException`. Didn't want to handle this in multiple places., so handled it once inside `createCollectionSettings()`

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/model/impl/CreateCollectionCommandResolver.java
maheshrajamani: Moved it inside IndexingConfig. validate(). Have provided `@Nullable` for both allow and deny. So will be null in case they are not provided in the request.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/CreateCollectionCommand.java
```diff
@@ -70,6 +79,22 @@ public VectorSearchConfig(Integer dimension, String metric) {
       }
     }
 
+    public record IndexingConfig(
```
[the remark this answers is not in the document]
maheshrajamani: We created a different `IndexingConfig` inside `CollectionSetting`. This is done because we can change the one inside `CollectionSetting` for usage convenience. I have allow and deny as Set in here.

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/model/impl/CreateCollectionCommandResolver.java
maheshrajamani: > I think `allow` and `deny` are always both non-null since they are initialized as empty `Set`s

For cache I have different`IndexingConfig` inside `CollectionSetting` these fields are Set in there. Also we have not used the Command object in other places.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/CreateCollectionCommand.java
maheshrajamani: Thought about it., For this need to create separate validators to verify things like either one of the field can only be used. So decided to do it as part of resolver.

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/model/impl/CreateCollectionCommandResolver.java
tatu-at-datastax: Right but isn't it the same as original json String? So is there need to encode from Object again? It seemed like we were carrying json String along. If not, that's fine, not really problematic wrt performance (collection creation has many other costs).

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/CollectionSettings.java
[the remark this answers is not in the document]
tatu-at-datastax: it's in one place wherever that place is tho?

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/model/impl/CreateCollectionCommandResolver.java
tatu-at-datastax: I don't think `@Nullable` annotation will inject anything (they are mostly just for documentation?). But since tests pass, I guess check is ok.

But ok. I missed separate `IndexingConfig`; if they are different I guess it's ok.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/CreateCollectionCommand.java
tatu-at-datastax: By validator I didn't mean anything pluggable or related to Quarkus.

tatu-at-datastax:
Ok, I'll read through everything once more (in a bit, offline for an hour), to make sure I get the overall thing right.
I think I missed some aspects.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/CreateCollectionCommand.java
```diff
@@ -70,6 +83,59 @@ public VectorSearchConfig(Integer dimension, String metric) {
...
```
tatu-at-datastax, line 86: Ahh. I missed the fact we have 2 `IndexingConfig`s. That's fine then I think.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/CreateCollectionCommand.java
```diff
@@ -70,6 +83,59 @@ public VectorSearchConfig(Integer dimension, String metric) {
...
+                implementation = String.class)
+            @Nullable
+            List<String> deny) {
+
+      public void validate() {
+        if (allow() != null && deny() != null) {
+          throw new JsonApiException(
+              ErrorCode.INVALID_INDEXING_USAGE,
+              ErrorCode.INVALID_INDEXING_USAGE.getMessage()
+                  + " - allow and deny cannot be used together");
```
tatu-at-datastax, line 107: Nit: should also quotes (single or double?) around "allow" and "deny" to indicate they are properties.

src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCode.java
```diff
@@ -27,6 +27,8 @@ public enum ErrorCode {
 
   INVALID_REQUST("Request not supported by the data store"),
 
+  INVALID_INDEXING_USAGE("Invalid indexing usage"),
```
tatu-at-datastax, line 30: suggestion: usage -> definition

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/CollectionSettings.java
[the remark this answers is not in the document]
tatu-at-datastax: Not a big deal tho, so it's fine.

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/CollectionSettings.java
```diff
@@ -109,46 +123,63 @@ public static CollectionSettings getCollectionSettings(
...
+
+    if (comment == null || comment.isBlank()) {
+      return new CollectionSettings(
+          collectionName, vectorEnabled, vectorSize, function, null, null, null);
+    } else {
+      try {
+        String vectorizeServiceName = null;
+        String modelName = null;
+
+        JsonNode commentConfig = objectMapper.readTree(comment);
```
tatu-at-datastax, line 148: Fine to leave catch block further down; but alternate that I sometimes I use is to do

```
final JsonNode commentConfig;

try {
  commentConfig = objectMapper.readTree(comment);
} catch (JsonProcessingException e) {
  throw new RuntimeExcepion(e);
}
```

to make it clear where theoretical exception comes from.

This of course assumes there aren't multiple sources.

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/model/impl/CreateCollectionCommandResolver.java
tatu-at-datastax: Ok 2 instances of `IndexingConfig` makes sense, all good. :)

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/model/impl/CreateCollectionCommandResolver.java
tatu-at-datastax: Ah ok. I finally got it: JSON is bound by Jackson into pieces, but then we need to store the whole JSON into Table comment. So we cannot avoid parts.

However: I think we can avoid going back and forth multiple times: but I understand why code right now does it -- because `command.options().indexing()` is not `JsonNode`.
But you can actually add all kinds of POJOs (anything Jackson can serialize) with:

```
objectNode.put("indexing", objectMapper.readTree(indexing));

// can be replaced with:

objectNode.putPOJO("indexing", command.options.indexing());
```

which should work if I understand code correctly. If I missed something else, feel free to ignore it.

src/test/java/io/stargate/sgv2/jsonapi/service/resolver/model/impl/CreateCollectionCommandResolverTest.java
```diff
@@ -121,9 +124,80 @@ public void happyPathVectorizeSearch() throws Exception {
...
+      CreateCollectionCommand command = objectMapper.readValue(json, CreateCollectionCommand.class);
+      Throwable throwable = catchThrowable(() -> resolver.resolveCommand(commandContext, command));
+
+      assertThat(throwable)
+          .isInstanceOf(JsonApiException.class)
+          .satisfies(
+              e -> {
+                JsonApiException exception = (JsonApiException) e;
+                assertThat(exception.getMessage())
+                    .isEqualTo("Invalid indexing usage - allow and deny cannot be used together");
```
tatu-at-datastax, line 199: Maybe also verify `errorCode` since that's what client libraries will use.

src/test/java/io/stargate/sgv2/jsonapi/service/resolver/model/impl/CreateCollectionCommandResolverTest.java
```diff
@@ -121,9 +124,80 @@ public void happyPathVectorizeSearch() throws Exception {
...
+                assertThat(op.commandContext()).isEqualTo(commandContext);
+                assertThat(op.vectorSearch()).isEqualTo(true);
+                assertThat(op.vectorSize()).isEqualTo(4);
+                assertThat(op.vectorFunction()).isEqualTo("cosine");
+                assertThat(op.comment()).isEqualTo("{\"indexing\":{\"deny\":[\"comment\"]}}");
+              });
+    }
+
+    @Test
+    public void happyPathIndexingError() throws Exception {
```
tatu-at-datastax, line 170: nit: not a happy path if there's error?

tatu-at-datastax (approved):
Ok looks good, thank you for taking time to explain where I misunderstood code & constraints.

Added some suggestions as usual; nothing major.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/CreateCollectionCommand.java
```diff
@@ -70,6 +83,59 @@ public VectorSearchConfig(Integer dimension, String metric) {
...
+        }
+
+        if (allow() == null && deny() == null) {
+          throw new JsonApiException(
+              ErrorCode.INVALID_INDEXING_USAGE,
+              ErrorCode.INVALID_INDEXING_USAGE.getMessage()
+                  + " - allow or deny should be provided");
```
kathirsvn, line 114: Shouldn't it index all fields when both `allow` and `deny` are not specified?

src/test/java/io/stargate/sgv2/jsonapi/api/v1/CreateCollectionIntegrationTest.java
```diff
@@ -83,6 +83,79 @@ class CreateCollection {
...
+                  "createCollection": {
+                    "name": "simple_collection_error2",
+                    "options" : {
+                      "vector" : {
+                        "size" : 5,
+                        "function" : "cosine"
+                      },
+                      "indexing" : {
+                        "allow" : ["field1", "field2"],
+                        "deny" : ["field1", "field2"]
```
kathirsvn, line 152: Can we add tests for pattern matching (for example `address.city`) and `_id` fields?

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/model/impl/CreateCollectionCommandResolver.java
maheshrajamani: Made the change. Thanks for pointing out `objectNode.putPOJO`

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/CreateCollectionCommand.java
maheshrajamani: Yes it will index all fields., If "indexing" option is used it should have atleast one them.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/CreateCollectionIntegrationTest.java
maheshrajamani: The PR currently covers only adding fields. When we do PR for handling fields for indexing in Shredder will add those.

tatu-at-datastax (approved):

vkarpov15 (commented):
LGTM :+1:

PR #775, opened 2024-01-08
Lower limit for max-$vector dimension to highest backend supports (4,096) from earlier (16,000)

### LINKED ISSUE #772: Reduce max Vector dimension limit from 10,000 to 4,096 not to fail due to SAI/ANN limit
Currently JSON API limits maximum Vector (`$vector`) dimension to rather high limit -- 16,000 -- which was simply chosen as "big enough not to matter". But it appears that there is a true limit as per:

https://github.com/riptano/VECTOR-SEARCH/issues/73

which limits indexable entry to 16kB, and with `float`-vector (4 bytes for `float32`) it means actual upper max dimension of 4,096.
So let's align this limit so JSON API can validate creation without hitting DB guardrails (esp. since test DB may not even have guardrails enabled).

New limit should still be above sizes needed in practice (f.ex, OpenAI's "standard" 1,536 ).

tatu-at-datastax:
**What this PR does**:

Lowers default for guard-rail that limits maximum vector dimension so it aligns with maximum backend DB allows.

Fixes #772

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/config/DocumentLimitsConfig.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/config/DocumentLimitsConfig.java
@@ -52,6 +52,12 @@ public interface DocumentLimitsConfig {
   int DEFAULT_MAX_STRING_LENGTH_IN_BYTES = 8_000;
 
   /**
+   * Defines the maximum dimension allowed for {@code $vector} field allowed: defaults to 4096 (as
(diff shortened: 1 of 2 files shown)


maheshrajamani (approved):

Hazel-Datastax (approved):

vkarpov15 (commented):
LGTM :+1:

PR #779, opened 2024-01-09
[Indexing options] Changes for find collections to return indexing option

### LINKED ISSUE #769: [Indexing options] Add indexing options to findCollection response
When findCollections command is run with option {"explain" : true } return indexing option if specified.

maheshrajamani:
**What this PR does**:
Changes for find collections to return indexing option

Fixes #769 

**Checklist**
- [x] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/CreateCollectionCommand.java
```diff
@@ -34,7 +34,8 @@ public record CreateCollectionCommand(
   public record Options(
 
       // limit of returned documents
-      @Schema(
+      @JsonInclude(JsonInclude.Include.NON_NULL)
```
tatu-at-datastax, line 37: We could also use `JsonInclude.Include.NON_EMPTY` which would exclude empty `Collection`s.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/FindCollectionsOperation.java
```diff
@@ -109,14 +112,35 @@ public CommandResult get() {
...
                             new CreateCollectionCommand.Options.VectorSearchConfig(
                                 collectionProperty.vectorSize(),
                                 collectionProperty.similarityFunction().name().toLowerCase());
+                      }
+
+                      if (collectionProperty.indexingConfig() != null) {
+                        indexingConfig =
+                            new CreateCollectionCommand.Options.IndexingConfig(
+                                collectionProperty.indexingConfig().allowed() != null
+                                        && !collectionProperty.indexingConfig().allowed().isEmpty()
```
tatu-at-datastax, line 126: If using `NON_EMPTY` inclusion earlier, could avoid checks here I think? (so `null` and empty Collection would just not be serialized.

tatu-at-datastax (approved):

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/FindCollectionsOperation.java
maheshrajamani: The IndexingConfig object in the cache has the allow and deny set initialized to empty, I did it so we don't have to do null check when trying to do contains check. Here I need to do not empty check so indexing options don't show up in findCollections as below
```
"indexing": {
  "allow" : [],
  "deny" : []
}
```

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/CreateCollectionCommand.java
maheshrajamani: Ah., got it I will change it.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/FindCollectionsIntegrationTest.java
```diff
@@ -304,6 +309,77 @@ public void notExistingNamespace() {
...
+          .contentType(ContentType.JSON)
+          .body(json)
+          .when()
+          .post(NamespaceResource.BASE_PATH, namespaceName)
+          .then()
+          .statusCode(200)
+          .body("status.collections", hasSize(4))
+          .body(
+              "status.collections",
+              containsInAnyOrder(
```
Yuqi-Du, line 377: I thought we only expect 'expected4' here.
No need for expected1, expected2, expected3.

maheshrajamani: find collection returns all the collection created thus far. So add all 4.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/FindCollectionsOperation.java
tatu-at-datastax: Ok yes I realized it after reading later code. Yes, it simplifies later check so everything is fine as-is.

tatu-at-datastax (approved):
LGTM

vkarpov15 (commented):
LGTM :+1: Nice to know about the findCollections command

PR #780, opened 2024-01-09
Shredder refactoring to isolate parts for easier no-caching implement…

tatu-at-datastax:
**What this PR does**:

Minor refactoring to separate traversal from size validation.

N/A

**Checklist**
- [x] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/service/shredding/Shredder.java
```diff
@@ -67,6 +67,10 @@ public WritableShreddedDocument shred(JsonNode doc, UUID txId) {
     final DocumentId docId = DocumentId.fromJson(docWithId.get(DocumentConstants.Fields.DOC_ID));
     final String docJson;
 
+    // Now that we have both the traversable document and serialization, verify
```
maheshrajamani, line 70: If we are not indexing why need to validate the size as long as json size is less than the desired?

maheshrajamani (approved):

src/main/java/io/stargate/sgv2/jsonapi/service/shredding/Shredder.java
tatu-at-datastax: Right now this change is just changing ordering, not adding/removing/changing limits.
But we should probably re-evaluate some limits when no-indexing is added.

PR #781, opened 2024-01-09
Error out when using non indexed fields

### LINKED ISSUE #768: [Indexing options] Error using non indexed fields for filtering, sorting
Error requests if not indexed fields are used for filtering and sorting.

Hazel-Datastax:
**What this PR does**:
Error out when using non indexed fields

Fixes #768 

**Checklist**
- [x] Changes manually tested


src/test/java/io/stargate/sgv2/jsonapi/api/v1/IndexingConfigIntegrationTest.java
```diff
@@ -0,0 +1,345 @@
...
+  private static final String denyOneIndexingCollection = "deny_one_indexing_collection";
+
+  private static final String denyManyIndexingCollection = "deny_many_indexing_collection";
+
+  private static final String denyAllIndexingCollection = "deny_all_indexing_collection";
+
+  private static final String allowOneIndexingCollection = "allow_one_indexing_collection";
+
+  private static final String allowManyIndexingCollection = "allow_many_indexing_collection";
+
```
Hazel-Datastax, line 28: @maheshrajamani According to the [design doc](https://docs.google.com/document/d/1Q4F4XCW71VaPZ1UEfGzLuHu7quw_aOrutDi9in7RWoY/edit), the wildcat character "\*" can only be used with a deny list to deny all fields, its use in an allow list generates an error. So I didn't test "\*" in allow list. However, Your previous [PR](https://github.com/stargate/jsonapi/pull/771) didn't handle this case, should you add the error handling for this?

src/test/java/io/stargate/sgv2/jsonapi/api/v1/IndexingConfigIntegrationTest.java
```diff
@@ -0,0 +1,352 @@
...
+          .body(filterData2)
+          .when()
+          .post(CollectionResource.BASE_PATH, namespaceName, denyOneIndexingCollection)
+          .then()
+          .statusCode(200)
+          .body("status", is(nullValue()))
+          .body("errors", is(nullValue()))
+          .body("data.documents", hasSize(1));
+    }
+
```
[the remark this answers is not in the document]
Hazel-Datastax: Solved. Deny "address.city" will implicitly allow "address". User can use "address" and "address.street"

Hazel-Datastax:
May wait for Tatu's PR for further changes.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/CommandContext.java
```diff
@@ -22,13 +22,15 @@ public record CommandContext(
     String collection,
     boolean isVectorEnabled,
     CollectionSettings.SimilarityFunction similarityFunction,
-    EmbeddingService embeddingService) {
+    EmbeddingService embeddingService,
+    CollectionSettings.IndexingConfig indexingConfig) {
```
tatu-at-datastax, line 26: I am planning to change this to pass `CollectionSettings`, instead of just pieces.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/sort/SortClause.java
```diff
@@ -37,4 +38,8 @@ public boolean hasVectorizeSearchClause() {
             .path()
             .equals(DocumentConstants.Fields.VECTOR_EMBEDDING_TEXT_FIELD);
   }
+
+  public void validate(CollectionSettings.IndexingConfig indexingConfig) {
```
tatu-at-datastax, line 42: Needs to get and use `DocumentProjection` to determine, not `IndexingConfig`

src/test/java/io/stargate/sgv2/jsonapi/api/v1/IndexingConfigIntegrationTest.java
```diff
@@ -0,0 +1,689 @@
...
+          .when()
+          .post(CollectionResource.BASE_PATH, namespaceName, denyAllIndexingCollection)
+          .then()
+          .statusCode(200)
+          .body("status", is(nullValue()))
+          .body("data", is(nullValue()))
+          .body(
+              "errors[0].message",
+              endsWith(
+                  "The filter path ('_id') is not indexed, you can only use $eq or $in as the operator"))
```
vkarpov15, line 452: I'm a bit confused by this error message. The test on line 351 tests that you cannot query on a string field if the field isn't indexed. But this error message indicates that you should be able to query on a non-indexed field as long as you're using `$eq`. Can you please explain the difference?

vkarpov15 (commented):
Looking good so far, just one question to clarify the expected behavior

src/test/java/io/stargate/sgv2/jsonapi/api/v1/IndexingConfigIntegrationTest.java
Hazel-Datastax: If a field name is in the deny list or not in the allow list (implicitly denied), it cannot be used in `filter`. The exception is `_id`. If `_id` is denied, it can only be used with `$eq` or `$in` in `filter`. The reason is:
```
Denying _id
The Document Identity field, _id, can be denied from the index. This can be done by either listing it explicitly in the deny list, or using the wildcard pattern in the deny list. 

This is possible because the _id field has two processing paths in the API: 
1. It is the CQL Table primary key, stored as a tuple of (data_type, value). 
2. It is indexed through super shredding using the appropriate data type. 

Read queries decide which representation to use based on the filter: 
1. When _id is the only term in a filter, and the operation is either $eq or $in the CQL Table primary key is used as the fast path lookup to get the row. 
2. When _id is used with other terms, or other operations, the super shredding representation may be used . 

As a result, when _id is denied from the index it is still possible to filter documents by _id using either the $eq or $in operations.  Further, it is legal for a collection to deny all fields from the index using the wildcard match to create a collection that only supports reading by the document id. This would use significantly less disk space, cost less for the end user, and still support projections and updates on the JSON document.
```

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/filter/FilterClause.java
```diff
@@ -12,6 +20,112 @@
...
+                "%s: The filter path ('%s') is not indexed",
+                ErrorCode.UNINDEXED_FILTER_PATH.getMessage(), comparisonExpression.getPath()));
+      }
+    }
+  }
+
+  private void validateHashMap(
+      DocumentProjector indexingProjector, HashMap<?, ?> map, String currentPath) {
+    for (Map.Entry<?, ?> entry : map.entrySet()) {
+      if (entry.getKey() instanceof String) {
```
Hazel-Datastax, line 89: Can the key not be a String?

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/filter/FilterClause.java
```diff
@@ -12,6 +20,112 @@
...
+                  ErrorCode.UNINDEXED_FILTER_PATH.getMessage(), incrementalPath));
+        }
+        // continue build the incremental path if the value is a hashmap
+        if (entry.getValue() instanceof HashMap<?, ?> valueMap) {
+          validateHashMap(indexingProjector, valueMap, incrementalPath);
+        }
+        // continue build the incremental path if the value is a list
+        if (entry.getValue() instanceof List<?> list) {
+          validateList(indexingProjector, list, incrementalPath);
+        }
```
Hazel-Datastax, line 106: Is the value not allowed to be a List? I used this test case: 
```
{
    "find": {
        "filter": {
            "address": {
                "street": {
                    "$in" :["xxx"]
                }
            }
        }
    }
}
```
and got the error message:
```
{
    "errors": [
        {
            "message": "Unsupported NodeType OBJECT",
            "errorCode": "UNSUPPORTED_FILTER_DATA_TYPE"
        }
    ]
}
```
The error message is not clear as well.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/filter/FilterClause.java
```diff
@@ -12,6 +20,112 @@
...
+
+    JsonLiteral<?> operand = comparisonExpression.getFilterOperations().get(0).operand();
+    // If path is an object (like address), validate the incremental path (like address.city)
+    if (operand.type() == JsonType.ARRAY || operand.type() == JsonType.SUB_DOC) {
+      if (operand.value() instanceof HashMap<?, ?> map) {
+        validateHashMap(indexingProjector, map, path);
+      }
+      if (operand.value() instanceof List<?> list) {
+        validateList(indexingProjector, list, path);
+      }
```
Hazel-Datastax, line 73: Is it only possible to have a list or a map here?

src/test/java/io/stargate/sgv2/jsonapi/api/v1/IndexingConfigIntegrationTest.java
```diff
@@ -0,0 +1,893 @@
...
+          .header(HttpConstants.AUTHENTICATION_TOKEN_HEADER_NAME, getAuthToken())
+          .contentType(ContentType.JSON)
+          .body(filterData2)
+          .when()
+          .post(CollectionResource.BASE_PATH, namespaceName, allowManyIndexingCollection)
+          .then()
+          .statusCode(200)
+          .body("status", is(nullValue()))
+          .body("errors", is(nullValue()))
+          .body("data.documents", hasSize(1));
```
Hazel-Datastax, line 770: @Yuqi-Du Is this use case of $in allowed? I didn't get any errors and get any data

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/filter/FilterClause.java
```diff
@@ -12,6 +20,112 @@
     example = """
              {"name": "Aaron", "country": "US"}
               """)
-public record FilterClause(LogicalExpression logicalExpression) {}
+public record FilterClause(LogicalExpression logicalExpression) {
+  public void validate(CommandContext commandContext) {
+    DocumentProjector indexingProjector = commandContext.indexingProjector();
+    // If nothing specified, everything indexed
+    if (indexingProjector.equals(DocumentProjector.identityProjector())) {
```
tatu-at-datastax, line 27: I think `indexingProjector.isIdentityProjection()` would be better.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/IndexingConfigIntegrationTest.java
Yuqi-Du: Yes, this is allowed. Weird use case though.
You should have data return back.

Yuqi-Du: If you insert data like this, 
`{
    "insertOne": {
        "document": {
                    "address": [
                {
                    "city": "monkey town"
                }
            ]
} }`
I do not think deny or allow should work, since we actually do not have address.city field here.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/filter/FilterClause.java
Yuqi-Du: Not good indeed.
I guess it goes into $date filter logic.
Let me raise a ticket.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/filter/FilterClause.java
```diff
@@ -12,6 +20,112 @@
...
+            ErrorCode.UNINDEXED_FILTER_PATH,
+            String.format(
+                "%s: The filter path ('%s') is not indexed",
+                ErrorCode.UNINDEXED_FILTER_PATH.getMessage(), comparisonExpression.getPath()));
+      }
+    }
+  }
+
+  private void validateHashMap(
+      DocumentProjector indexingProjector, HashMap<?, ?> map, String currentPath) {
```
tatu-at-datastax, line 87: Let's change that to `Map`, no need to specify implementation.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/filter/FilterClause.java
```diff
@@ -12,6 +20,112 @@
...
+      } else {
+        // else, _id can be used, return
+        return;
+      }
+    }
+
+    JsonLiteral<?> operand = comparisonExpression.getFilterOperations().get(0).operand();
+    // If path is an object (like address), validate the incremental path (like address.city)
+    if (operand.type() == JsonType.ARRAY || operand.type() == JsonType.SUB_DOC) {
+      if (operand.value() instanceof HashMap<?, ?> map) {
```
tatu-at-datastax, line 68: Let's not check for `HashMap` but just `Map` -- in case we wanted to at some point use, say, `TreeMap` (I think actual type might even be `LinkedHashMap` if deserialized by Jackson).
And then cast to `Map<?, ?>`.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/filter/FilterClause.java
tatu-at-datastax: You can't use nesting like that; references need to be paths like `address.street`. And yes, exception is not very good, should figure out that `street` is not a valid Operator I think.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/filter/FilterClause.java
```diff
@@ -12,6 +20,112 @@
...
+        if (entry.getValue() instanceof List<?> list) {
+          validateList(indexingProjector, list, incrementalPath);
+        }
+      }
+    }
+  }
+
+  private void validateList(DocumentProjector indexingProjector, List<?> list, String currentPath) {
+    for (Object element : list) {
+      if (element instanceof HashMap<?, ?> map) {
```
tatu-at-datastax, line 113: HashMap -> Map

src/test/java/io/stargate/sgv2/jsonapi/api/v1/IndexingConfigIntegrationTest.java
tatu-at-datastax: There is a minor change to allow `[ "*" ]` for allow list as well, to provide symmetry.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/IndexingConfigIntegrationTest.java
```diff
@@ -0,0 +1,899 @@
...
+                    },
+                    "indexing" : {
+                      "deny" : ["name", "address"]
+                    }
+                  }
+                }
+              }
+                  """
+              .formatted(denyManyIndexingCollection);
+      // create collection
```
tatu-at-datastax, line 107: Since you create multiple collections, insert multiple entries, may make sense to extract helper methods.
`AbstractCollectionIntegrationTestBase` actually has `createCollection()` and `insertDoc()` if you could that base class instead -- we could maybe refactor things a bit and move `createCollection(String namespace, String collectionName)` into that class, for better reuse, reduced code duplication.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/filter/FilterClause.java
```diff
@@ -12,6 +20,112 @@
...
+      }
+    }
+  }
+
+  private void validateHashMap(
+      DocumentProjector indexingProjector, HashMap<?, ?> map, String currentPath) {
+    for (Map.Entry<?, ?> entry : map.entrySet()) {
+      if (entry.getKey() instanceof String) {
+        String subPath = (String) entry.getKey();
+        String incrementalPath = currentPath + "." + subPath;
```
tatu-at-datastax, line 91: Is this even necessary? I think filter clauses refer to full path and cannot be nested? /cc @maheshrajamani

src/test/java/io/stargate/sgv2/jsonapi/api/v1/IndexingConfigIntegrationTest.java
tatu-at-datastax: No, I don't think that is legal `filter`, wrt nesting. At main level there is always (full) path as key, and then JSON Object with operators as keys and so on.

Hazel-Datastax: My bad, I found a bug in my validation logic. Thank you Yuqi!

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/filter/FilterClause.java
Hazel-Datastax: `isIdentityProjection()` is a function inside the private class `PathCollector`. Do you want me to create a new function?

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/filter/FilterClause.java
Hazel-Datastax: Thank you!

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/filter/FilterClause.java
tatu-at-datastax: Hmmh. `isIdentityProjection()` should be member of `DocumentProjector`... I'll check.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/IndexingConfigIntegrationTest.java
vkarpov15: Thanks for clarifying, that makes sense

vkarpov15 (approved):

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/filter/FilterClause.java
```diff
@@ -12,6 +19,111 @@
...
+    }
+  }
+
+  public void validateComparisonExpression(
+      ComparisonExpression comparisonExpression, DocumentProjector indexingProjector) {
+    String path = comparisonExpression.getPath();
+    boolean isPathIndexed = indexingProjector.isPathIncluded(path);
+
+    // If path is "_id" and it's denied, the operator can only be $eq or $in
+    if (path.equals(DocumentConstants.Fields.DOC_ID) && !isPathIndexed) {
```
tatu-at-datastax, line 48: nit: reverse the order of checks (`!isPathIndexed` first), cheaper check first

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/filter/FilterClause.java
```diff
@@ -12,6 +19,111 @@
...
+      FilterOperator filterOperator = comparisonExpression.getFilterOperations().get(0).operator();
+      // if operator is not $eq or $in, throw error
+      if (!filterOperator.equals(ValueComparisonOperator.EQ)
```
tatu-at-datastax, line 51: Can simplify comparisons since `ValueComparisonOperator` is enum so no need for `equals()`.
So maybe:

```
if (filterOperator == ValueComparisonOperator.EQ || filterOperator == ValueComparisonOperator.IN) {
  return;
}
// otherwise throw JsonApiException
```

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/filter/FilterClause.java
```diff
@@ -12,6 +19,111 @@
...
+        validateList(indexingProjector, list, path);
+      }
+    }
+  }
+
+  private void validateMap(DocumentProjector indexingProjector, Map<?, ?> map, String currentPath) {
+    for (Map.Entry<?, ?> entry : map.entrySet()) {
+      if (entry.getKey() instanceof String) {
+        String subPath = (String) entry.getKey();
+        String incrementalPath = currentPath + "." + subPath;
```
tatu-at-datastax, line 89: Can simplify as:

```
  String incrementalPath = currentPath + "." + entry.getKey();
```

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/filter/FilterClause.java
```diff
@@ -12,6 +19,111 @@
...
+      }
+      if (operand.value() instanceof List<?> list) {
+        validateList(indexingProjector, list, path);
+      }
+    }
+  }
+
+  private void validateMap(DocumentProjector indexingProjector, Map<?, ?> map, String currentPath) {
+    for (Map.Entry<?, ?> entry : map.entrySet()) {
+      if (entry.getKey() instanceof String) {
```
tatu-at-datastax, line 87: No need to check if it's String; will always be.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/filter/FilterClause.java
```diff
@@ -12,6 +19,111 @@
...
+      }
+    }
+  }
+
+  private void validateList(DocumentProjector indexingProjector, List<?> list, String currentPath) {
+    for (Object element : list) {
+      if (element instanceof Map<?, ?> map) {
+        validateMap(indexingProjector, map, currentPath);
+      }
+      if (element instanceof List<?> sublList) {
```
tatu-at-datastax, line 114: `if` -> `else if` otherwise will check all cases even tho only one can match.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/filter/FilterClause.java
```diff
@@ -12,6 +19,111 @@
...
+        validateList(indexingProjector, sublList, currentPath);
+      }
+      if (element instanceof String) {
```
tatu-at-datastax, line 117: `if` -> `else if`

src/test/java/io/stargate/sgv2/jsonapi/api/v1/AbstractCollectionIntegrationTestBase.java
```diff
@@ -49,6 +49,35 @@ protected void createCollection(String collectionToCreate) {
         .statusCode(200);
   }
 
+  protected void createCollection(String collectionToCreate, String indexingConfig) {
```
tatu-at-datastax, line 52: Almost, but I think it should be passed JSON Object that is value of `"createCollection"`, otherwise hard-codes vector settings. So instead of "name", pass "collectionJson" String or so.

tatu-at-datastax (approved):
Looks pretty good; added some suggestions

PR #785, opened 2024-01-10
Changes to have ddl and count profiles

### LINKED ISSUE #783: Configurable CQL query timeouts by query type
Currently the JSON API only supports a single configurable value for CQL query timeouts, which defaults to 10s in the standard config file that is checked in. 

At high scale usage in cloud environments, we've observed the need for more flexible controls for different types of queries. For example, DDL queries like create/delete table have different behavior and can take longer in cloud environments. 

JSON API can make use of the execution profile feature offered by the CQL in order to create profiles for each type of query, with separate timeouts for each profile. This will allow users to have more flexible control over timeouts with a simple config file change.

Suggested profiles:
- default (use for DML that doesn't include LWT)
- lwt (use for DML that includes LWT)
- schema (DDL such as create/delete table)
- count (use for SELECT COUNT queries)

maheshrajamani:
**What this PR does**:
Changes to have ddl and count profiles

Fixes #783 

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/src/main/resources/application.conf
+++ b/src/main/resources/application.conf
@@ -43,10 +43,18 @@ datastax-java-driver {
       show-stack-traces = true
     }
   }
-  basic.request.timeout = 10 seconds
+  basic.request.timeout = 20 seconds
   profiles {
     slow {
       basic.request.timeout = 30 seconds
+    }
+
+    ddl {
+      basic.request.timeout = 30 seconds
+    }
+
+    count {
+      basic.request.timeout = 60 seconds
     }
   }
 }
--- a/src/test/java/io/stargate/sgv2/jsonapi/api/v1/InsertIntegrationTest.java
+++ b/src/test/java/io/stargate/sgv2/jsonapi/api/v1/InsertIntegrationTest.java
@@ -1312,7 +1312,7 @@ public class InsertIntegrationTest extends AbstractCollectionIntegrationTestBase
           .post(CollectionResource.BASE_PATH, namespaceName, collectionName)
           .then()
           .statusCode(200)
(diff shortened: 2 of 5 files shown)


src/test/java/io/stargate/sgv2/jsonapi/api/v1/InsertIntegrationTest.java
```diff
@@ -1152,7 +1152,10 @@ public void orderedDuplicateDocumentNoNamespace() {
                   "_id": "doc5",
                   "username": "user5"
                 }
-              ]
+              ],
```
maheshrajamani, line 1155: Side effect of a previous merged PR. Need to refactor the test case.

jeffreyscarpenter (approved):

jeffreyscarpenter:
@tatu-at-datastax can you have a look as well?

src/main/resources/application.conf
```diff
@@ -43,10 +43,18 @@ datastax-java-driver {
...
     slow {
       basic.request.timeout = 30 seconds
     }
+
+    ddl {
+      basic.request.timeout = 30 seconds
+    }
+
+    count {
+      basic.request.timeout = 60 seconds
```
tatu-at-datastax, line 57: This sounds bit high (as does basic request timeout of 20 secs), but easy enough to change later on I suppose.

tatu-at-datastax (approved):
LGTM: actual timeout levels chosen might be good to discuss but that can be done later on (lower timeouts can prevent server-side overload/meltdown by load shedding).

tatu-at-datastax:
> @tatu-at-datastax can you have a look as well?

Done.

PR #787, opened 2024-01-11
set the application name when creating CQL session

### LINKED ISSUE #387: JSON API should identify itself to bridge as jsonapi
JSON API is currently identifying itself as `rest` when connecting to the bridge. It should be identifying itself as a unique new API `jsonapi`.

This primarily has implications for deploying JSON API in AstraDB. The current state would have JSON API leveraging on the authorizations that users would typically get when creating the "default token" for their database (the "REST API" permission is automatically enabled). 

It might be a good idea to allow the value to be configurable to reduce the need to change multiple repos in lockstep. The Stargate repo should probably also have the value `jsonapi` added to the [SourceAPI](https://github.com/stargate/stargate/blob/main/coordinator/authnz/src/main/java/io/stargate/auth/SourceAPI.java) enumeration.

jeffreyscarpenter:
JSON API was updated to use the CQL Driver instead of the bridge to communicate with the coordinator. In order to allow the coordinator to distinguish CQL traffic that comes from the coordinator, we should set the application name on the CqlSession. We can reuse the Quarkus application name for this.

Fixes #387 

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/CQLSessionCache.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/CQLSessionCache.java
@@ -17,6 +17,7 @@ import java.time.Duration;
 import java.util.List;
 import java.util.Objects;
 import java.util.stream.Collectors;
+import org.eclipse.microprofile.config.inject.ConfigProperty;
 import org.slf4j.Logger;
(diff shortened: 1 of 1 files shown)


Yuqi-Du (approved):

kathirsvn (commented):
LGTM

PR #792, opened 2024-01-12
Fixes #767: Second part of "no-index" changes: allow large non-indexed String values, implicit allow for "$vector"

### LINKED ISSUE #767: [Indexing options] Shredder changes to index fields
Shredder changes to use the indexing configuration to allow or deny indexing of a field

tatu-at-datastax:
**What this PR does**:

Completes changes needed by #767 by:

1. Adding more testing for Shredder
2. Fix the issue wrt dropping indexing of "$vector" (needs handling similar to `_id` for query projections, i.e. always index)
3. Changing validation to ignore length of non-Indexed Strings (only limited by max doc size)

These are resolved by refactoring validation logic into smaller parts, run at different times during shredding process.

Fixes #767

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/shredding/Shredder.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/shredding/Shredder.java
@@ -226,6 +226,19 @@ public class Shredder {
     callback.shredVector(path, arr);
   }
 
+  private void validateDocumentSize(DocumentLimitsConfig limits, String docJson) {
+    // First: is the resulting document size (as serialized) too big?
+    if (docJson.length() > limits.maxSize()) {
+      throw new JsonApiException(
+          ErrorCode.SHRED_DOC_LIMIT_VIOLATION,
+          String.format(
+              "%s: document size (%d chars) exceeds maximum allowed (%d)",
+              ErrorCode.SHRED_DOC_LIMIT_VIOLATION.getMessage(),
+              docJson.length(),
+              limits.maxSize()));
+    }
+  }
+
   private void validateDocumentStructure(DocumentLimitsConfig limits, ObjectNode doc) {
     // Second: traverse to check for other constraints
     AtomicInteger totalProperties = new AtomicInteger(0);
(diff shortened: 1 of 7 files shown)


src/main/java/io/stargate/sgv2/jsonapi/service/shredding/Shredder.java
```diff
@@ -88,14 +88,21 @@ public WritableShreddedDocument shred(JsonNode doc, UUID txId, DocumentProjector
     // And then we can validate the document size
     validateDocumentSize(documentLimits, docJson);
 
-    // Before shredding, may need to drop "non-indexed" fields:
+    final WritableShreddedDocument.Builder b =
```
tatu-at-datastax, line 91: We need document builder immediately, to build index incrementally: in particular, `$vector` indexing should precede "no index" Projection that would typically drop it.

vkarpov15 (commented):
Looks good so far :+1:

tatu-at-datastax:
Remaining issue: when returning `after` Document for `findOneAndXxx()` need to retain it without pruning for "no-index", may need to create a deep copy.

src/main/java/io/stargate/sgv2/jsonapi/service/shredding/Shredder.java
maheshrajamani: Any reason why it needs to be if `allow` is specified, can't we set up required indexing fields with `$vector` in it?

maheshrajamani (approved):
Overall looks good except for the clarification on `$vector` field

tatu-at-datastax:
Sounds like implicit addition of `$vector` for non-empty `"allow"` -- similar to `_id` for filter Projection -- is what we'll be going with. Users can still add it to `"deny"` list if they really want to exclude it from being indexed; but that seems like an odd use case (why add `$vector` in Collection at all then?).

tatu-at-datastax:
Need to make change for auto-inclusion of `$vector`, moving back to Draft.

src/main/java/io/stargate/sgv2/jsonapi/service/shredding/Shredder.java
tatu-at-datastax: Will proceed with this idea, change code here.

tatu-at-datastax:
Ready for re-review.

tatu-at-datastax:
@maheshrajamani "$vector" handling changed as discussed: any other concerns? (I know you have approved but wanted to double-check before merging)

maheshrajamani:
> @maheshrajamani "$vector" handling changed as discussed: any other concerns? (I know you have approved but wanted to double-check before merging)

@tatu-at-datastax Changes LGTM

vkarpov15 (commented):
LGTM

PR #804, opened 2024-01-17
Add validation of paths used for indexing "allow"/"deny" lists.

### LINKED ISSUE #790: Validate paths used for `indexing` options for `CreateCollection`
Currently the only validation done for indexing options of CreateCollections is to make sure that only one of arrays ("allow", "deny") is non-empty.

But we should be able to further validate that all paths included are valid paths for JSON API, same as we do for paths used for shredding and filtering.

For example, currently user could use:

```json
"indexing" : {
   "deny" : [ "_id", "data*" ]
}
```

expecting `data*` to be regexp used: it won't and we'd take that to refer to hypothetical value like:

```json
{
   "data*" : 42 
}
```

but we do NOT allow such properties to be used (character '*' not allowed in path segments).

Adding validation would prevent users from creating Indexing settings that cannot work as intended.

tatu-at-datastax:
**What this PR does**:

Adds validation of "indexing" options paths (in "allow" / "deny" lists), tests to verify.

Fixes #790

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/src/test/java/io/stargate/sgv2/jsonapi/api/v1/CreateCollectionIntegrationTest.java
+++ b/src/test/java/io/stargate/sgv2/jsonapi/api/v1/CreateCollectionIntegrationTest.java
@@ -418,6 +418,7 @@ class CreateCollectionIntegrationTest extends AbstractNamespaceIntegrationTestBa
       given()
           .header(HttpConstants.AUTHENTICATION_TOKEN_HEADER_NAME, getAuthToken())
           .contentType(ContentType.JSON)
+          // Brackets not allowed in field names
           .body(
               """
                     {
@@ -425,7 +426,7 @@ class CreateCollectionIntegrationTest extends AbstractNamespaceIntegrationTestBa
                         "name": "collection_with_bad_allows",
                         "options" : {
                           "indexing" : {
-                            "allow" : ["field", "$eq"]
+                            "allow" : ["valid-field", "address[1].street"]
                           }
                         }
                       }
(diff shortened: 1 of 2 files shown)


src/test/java/io/stargate/sgv2/jsonapi/api/v1/CreateCollectionIntegrationTest.java
```diff
@@ -329,6 +330,20 @@ public void duplicateVectorCollectionNameWithDiffSetting() {
...
+          .body(deleteCollectionJson.formatted("simple_collection"))
+          .when()
+          .post(NamespaceResource.BASE_PATH, namespaceName)
+          .then()
+          .statusCode(200)
+          .body("status.ok", is(1));
+    }
+
+    @Test
+    public void createCollectionWithIndexing() {
```
tatu-at-datastax, line 346: Split existing long test method in two parts since they were testing different aspects.

maheshrajamani (approved):

src/test/java/io/stargate/sgv2/jsonapi/api/v1/CreateCollectionIntegrationTest.java
```diff
@@ -396,6 +411,70 @@ public void duplicateVectorCollectionNameWithDiffSetting() {
...
+          .header(HttpConstants.AUTHENTICATION_TOKEN_HEADER_NAME, getAuthToken())
+          .contentType(ContentType.JSON)
+          .body(
+              """
+                    {
+                      "createCollection": {
+                        "name": "collection_with_bad_allows",
+                        "options" : {
+                          "indexing" : {
+                            "allow" : ["field", "$eq"]
```
maheshrajamani, line 428: Can we add a test field with array index fails? `address.[0].city` should throws error?

tatu-at-datastax: Good idea, will add.

tatu-at-datastax:
@maheshrajamani Allowing `$vector` now, modified tests to include that, as well as check that brackets not allowed.

vkarpov15 (commented):
LGTM :+1:

PR #805, opened 2024-01-17
Fix to #790 implementation to ensure ["*"] is accepted as "allow" or "deny" list

### LINKED ISSUE #790: Validate paths used for `indexing` options for `CreateCollection`

tatu-at-datastax:
**What this PR does**:

Improves "createCollection" validation, testing, to ensure single "*" is accepted as expected

Fixes #790

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/src/test/java/io/stargate/sgv2/jsonapi/api/v1/CreateCollectionIntegrationTest.java
+++ b/src/test/java/io/stargate/sgv2/jsonapi/api/v1/CreateCollectionIntegrationTest.java
@@ -25,14 +25,6 @@ class CreateCollectionIntegrationTest extends AbstractNamespaceIntegrationTestBa
   @Nested
   @Order(1)
   class CreateCollection {
-    String deleteCollectionJson =
-        """
-            {
-              "deleteCollection": {
-                "name": "simple_collection"
-              }
-            }
-            """;
     String createNonVectorCollectionJson =
         """
             {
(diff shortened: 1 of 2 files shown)


src/test/java/io/stargate/sgv2/jsonapi/api/v1/CreateCollectionIntegrationTest.java
```diff
@@ -84,79 +76,6 @@ class CreateCollection {
                 }
                 """;
 
-    String createCollectionWithAllowIndexing =
```
tatu-at-datastax, line 87: Moved these to within tests that need them, no need to separate out.

Hazel-Datastax (approved):
LGTM

vkarpov15 (commented):
LGTM

PR #806, opened 2024-01-17
Do not return HTTP 504 Gateway Timeout on DB timeout 

Yuqi-Du:
when we get a timeout from the DB we are returning a HTTP 504 gateway timeout. This is incorrect, we should return a status 200 and an error, which we do in many cases.

Jira:  https://datastax.jira.com/browse/C2-3212

**Checklist**
- [x] Changes manually tested


amorton (changes requested):
please also handle the case where it is returning BAD_GATEWAY on UNAVAILABLE

do the tests confirm that when these things happen we return HTTP 200 and the error list has the appropriate errors ?

Yuqi-Du:
> BAD_GATEWAY

Yep, this is from the old bridge logic, remove that part now.

maheshrajamani (approved):
LGTM based on yesterday's discussion.

tatu-at-datastax (approved):
Looks good. Would be great if this could be tested with IT but not sure how that could be done.

vkarpov15 (commented):
LGTM :+1:

PR #811, opened 2024-01-19
Count optimization changes

### LINKED ISSUE #809: Count command optimization
Count command in the API uses cassandra's count function. Cassandra coordinator fetches all the filtered rows from storage and increments the counter for each valid rows. Because of this approach, counting cassandra table with large number of rows potentially takes longer time and eventually starts timing out.

Below is the proposed change for count operation in json api.

- A configurable limit will be specified to suggest what is maximum count that api will return. Default will be 1000.
- Count command will not use cassandra's count aggregation function.
- Read primary key (_id) column from DB for given filters
- Select query will be specified with `limit` <configured limit + 1>
- If total primary keys read from DB is equal to <configured limit +1>, then return count as <configured limit> and `more_data` to true. Else return the count of keys retrieved.

Translated CQL for count command will look as:
`select key from table where filters limit configured limit +1`

maheshrajamani:
**What this PR does**:
Count command changes to read primary key data and count in the client side. Number of keys read is controlled by configuration 
Fixes #809 

**Checklist**
- [x] Changes manually tested


CONFIGURATION.md
```diff
@@ -52,6 +52,9 @@ Here are some Stargate-relevant property groups that are necessary for correct s
...
 | `stargate.jsonapi.operations.database-config.session-cache-ttl-seconds` | `int` | `300`    | The amount of seconds that the cql session will be kept in memory after last access.                                                                                                                |
 | `stargate.jsonapi.operations.database-config.session-cache-max-size`    | `int` | `50`     | The maximum number of cql sessions that will be kept in memory.                                                                                                                                     |
+| `stargate.jsonapi.operations.default-count-page-size`                   | `int` | `100`    | The default Cassandra page size used for read queries that are used for counting by key purposes.                                                                                                   |
+| `stargate.jsonapi.operations.max-count-limit`                           | `int` | `1000`   | The default max count response when resolved using keys. Set to -1 use cassandra count function.                                                                                                    |
```
tatu-at-datastax, line 56: Maybe reword slightly... "default maximum number of rows to read for count operation" or something? Just to make it clear it limits reads, and not just response (is obvious to us, but someone reading without having background might not be)

src/main/java/io/stargate/sgv2/jsonapi/config/OperationsConfig.java
```diff
@@ -100,6 +100,23 @@ public interface OperationsConfig {
   @WithDefault("1000")
   int maxVectorSearchLimit();
 
+  /**
+   * @return Maximum size of keys read from database to return count, Setting it to -1 will use
```
tatu-at-datastax, line 104: Maximum size -> Maximum number ?

src/main/java/io/stargate/sgv2/jsonapi/config/OperationsConfig.java
```diff
@@ -100,6 +100,23 @@ public interface OperationsConfig {
...
+   *     Cassandra's count function. Default is 1000 <code>1000</code> command.
+   */
+  @Max(10000)
```
tatu-at-datastax, line 107: Is this needed? As in, should we limit what can be used as override?

src/main/java/io/stargate/sgv2/jsonapi/config/OperationsConfig.java
```diff
@@ -100,6 +100,23 @@ public interface OperationsConfig {
...
+  @WithDefault("1000")
+  int maxCountLimit();
+
+  /**
+   * @return Defines the default page size for count operation, having separate config because count
+   *     will more keys per page, defaults to <code>100</code>.
+   */
+  @Max(500)
```
tatu-at-datastax, line 115: Same here, is there need to limit the max one can use for overrides?

src/main/java/io/stargate/sgv2/jsonapi/config/OperationsConfig.java
```diff
@@ -100,6 +100,23 @@ public interface OperationsConfig {
...
+   * @return Maximum size of keys read from database to return count, Setting it to -1 will use
+   *     Cassandra's count function. Default is 1000 <code>1000</code> command.
+   */
+  @Max(10000)
+  @WithDefault("1000")
+  int maxCountLimit();
+
+  /**
+   * @return Defines the default page size for count operation, having separate config because count
+   *     will more keys per page, defaults to <code>100</code>.
```
tatu-at-datastax, line 113: nit: "will more" -> "will read more". Also separate from which setting? (add link/ref to the other setting)

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/CountOperation.java
```diff
@@ -20,15 +20,32 @@
  * Operation that returns count of documents based on the filter condition. Written with the
  * assumption that all variables to be indexed.
  */
-public record CountOperation(CommandContext commandContext, LogicalExpression logicalExpression)
+public record CountOperation(
```
tatu-at-datastax, line 23: Would it make sense to pass `OperationsConfig` instead of separate fields? (not sure, just wondering)

tatu-at-datastax: Actually, probably better this way for use, n/m.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/CountOperation.java
```diff
@@ -20,15 +20,32 @@
...
-        .transform(docs -> new CountOperationPage(docs.count()));
+        .transform(
+            docs -> {
+              if (limit == -1) {
+                return new CountOperationPage(docs.count(), false);
+              } else {
+                boolean moreData = docs.count() > limit();
+                return new CountOperationPage(
+                    docs.count() == 0
+                        ? 0
```
tatu-at-datastax, line 44: Is there need to separately check for `0`? Seems like that'd work without extra check.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/ReadOperation.java
```diff
@@ -363,6 +393,17 @@ default Uni<CountResponse> countDocuments(
             });
   }
 
+  private void getCount(AsyncResultSet rs, Throwable error, AtomicLong counter) {
+    if (error != null) {
+      throw new RuntimeException(error);
```
tatu-at-datastax, line 398: Would it be possible/better to throw JsonApiException from here, with extra information we may have about count query? (or at least indicate context as failure of COUNT query, with separate `ErrorCode`? Could indicate how many rows have been counted at minimum; page size used etc, making debugging of failures easier)

src/test/java/io/stargate/sgv2/jsonapi/api/v1/CountIntegrationTest.java
```diff
@@ -137,6 +147,7 @@ public void noFilter() {
           .then()
           .statusCode(200)
           .body("status.count", is(5))
+          .body("status.moreData", is(true))
```
tatu-at-datastax, line 150: Would be nice to mention in test class Javadoc that the default settings are overridden by DseTestResource, so max count is 5, max count page size 2? (was first surprised by this but it makes sense with overrides)

tatu-at-datastax (approved):
Looks good -- functionality fine, only suggestions are wrt naming and documentation/comments (to help whoever reads code).

src/main/java/io/stargate/sgv2/jsonapi/config/OperationsConfig.java
maheshrajamani: This needed to limit coordinator pressure. Think should be there.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/CountOperation.java
maheshrajamani: All other commands like find, delete etc did this way, so following same approach.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/CountOperation.java
maheshrajamani: Good Catch, fixed it.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/CountOperation.java
```diff
@@ -38,16 +55,29 @@ private SimpleStatement buildSelectQuery() {
     if (expressions != null && !expressions.isEmpty() && expressions.get(0) != null) {
       collect = ExpressionBuilder.getExpressionValuesInOrder(expressions.get(0));
```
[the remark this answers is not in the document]
maheshrajamani: Not sure why this is done that way. I will create a issue to handle all expressions,

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/CountOperationPage.java
```diff
@@ -5,9 +5,13 @@
 import java.util.Map;
 import java.util.function.Supplier;
 
-public record CountOperationPage(long count) implements Supplier<CommandResult> {
+public record CountOperationPage(long count, boolean moreData) implements Supplier<CommandResult> {
```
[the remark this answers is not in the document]
maheshrajamani: Used it because DeleteMany and UpdateMany use this falg.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/CountIntegrationTest.java
```diff
@@ -335,7 +351,7 @@ public void withExistFalseOperator() {
           .post(CollectionResource.BASE_PATH, namespaceName, collectionName)
           .then()
           .statusCode(200)
-          .body("status.count", is(4))
+          .body("status.count", is(5))
```
vkarpov15, line 354: Why did this change? Doesn't seem like a perf optimization should change the result here

maheshrajamani: @vkarpov15  Have inserted one more document for testing `moreData` flag. Because of the additional record count increased by 1 for the provided filter condition in this test case.

vkarpov15: Makes sense, thanks for clarifying.

PR #832, opened 2024-01-24
error out with invalid option when createCollection

### LINKED ISSUE #828: Add validation of `"options"` property of `createCollection` command to avoid silent failures
Currently there are 2 valid sub-properties to use with `createCollection` command: "vector" and "indexing".
Their contents are validated, and if found good, used for adding (or not) of `$vector` columnd (for first) and changing indexing aspects.

But if users adds any other option child property -- possibly (but not only) misspelling one of 2 valid names, like "Vector" (values are case-sensitive) or "index" (instead of "indexing"), collection creation will effectively just ignore such options.

We should validate properties of "options" and report a failure for unknown option (with new `ErrorCode`, similar to `COMMAND_ACCEPTS_NO_OPTIONS`). This is useful for indicating problem back to user.

Yuqi-Du:
When user provides an invalid option when createCollection,
currently, we have three options (case sensitive): vector, vectorize, indexing

Example:
```
{
    "createCollection": {
        "name": "tableEx",
        "options": {
            "Yuqi": {}
        }
    }
}
```

```
{
    "errors": [
        {
            "message": "The provided options is invalid: No option "Yuqi" found as createCollectionCommand option,  (known options: "indexing", "vector")",
            "errorCode": "INVALID_CREATE_COLLECTION_OPTIONS"
        }
    ]
}
```

Fixes #828 

**Checklist**
- [x] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/api/configuration/CommandObjectMapperHandler.java
```diff
@@ -1,13 +1,34 @@
...
+  public boolean handleUnknownProperty(
+      DeserializationContext ctxt,
+      JsonParser p,
+      JsonDeserializer<?> deserializer,
+      Object beanOrClass,
+      String propertyName)
+      throws IOException {
+    if (deserializer.handledType().toString().endsWith("CreateCollectionCommand$Options")) {
+      throw ErrorCode.INVALID_CREATE_COLLECTION_OPTIONS.toApiException(
+          "No option \"%s\" found as createCollectionCommand option", propertyName);
```
tatu-at-datastax, line 25: This is helpful: but even more helpful I think would be listing 2 existing options in addition?
Something like:

```
"No option \"%s\" found as createCollectionCommand option (known options: \"indexing\", \"vector\")",
```

src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCode.java
```diff
@@ -5,6 +5,7 @@ public enum ErrorCode {
   /** Command error codes. */
   COUNT_READ_FAILED("Unable to count documents"),
   COMMAND_NOT_IMPLEMENTED("The provided command is not implemented."),
+  INVALID_CREATE_COLLECTION_OPTIONS("The provided options is invalid"),
```
tatu-at-datastax, line 8: nit: OPTIONS -> OPTION

src/main/java/io/stargate/sgv2/jsonapi/api/configuration/CommandObjectMapperHandler.java
tatu-at-datastax: (not sure if `"vectorize"` should be included; it is technically allowed, but not publicly supported.

Yuqi-Du: solved, thanks

vkarpov15 (commented):
LGTM

PR #836, opened 2024-01-25
Fixes #835: ensure "createCollection()" still idempotent with "no-index" options

### LINKED ISSUE #835: [regression] - Adding `indexing` options in the `createCollection` remove idempotence
Previously this request could be executed multiple times without errors:
```
{"createCollection":{"name":"collection_vector","options":{"vector":{"dimension":14,"metric":"cosine"}}}}
```
Will get you all the time.
```
{"status":{"ok":1}}
```

With the indexing options the behaviour is now different:

```
{"createCollection":{"name":"collection_vector","options":{"vector":{"dimension":14,"metric":"cosine"},"indexing":{"deny":["blob_body"]}}}}

=> {"status":{"ok":1}}

{"createCollection":{"name":"collection_vector","options":{"vector":{"dimension":14,"metric":"cosine"},"indexing":{"deny":["blob_body"]}}}}

=> {"errors":[{"message":"The provided collection name 'collection_vector' already exists with a different vector setting.","errorCode":"INVALID_COLLECTION_NAME"}]}
```

tatu-at-datastax:
**What this PR does**:

Adds tests to verify idempotency of "createCollection" command when using new "no-index" options; and fixes issues found, if any

Fixes #835

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/src/test/java/io/stargate/sgv2/jsonapi/api/v1/CreateCollectionIntegrationTest.java
+++ b/src/test/java/io/stargate/sgv2/jsonapi/api/v1/CreateCollectionIntegrationTest.java
@@ -236,56 +236,75 @@ class CreateCollectionIntegrationTest extends AbstractNamespaceIntegrationTestBa
     }
 
     @Test
-    public void happyCreateCollectionWithIndexing() {
+    public void happyCreateCollectionWithIndexingAllow() {
+      final String createCollectionRequest =
+          """
+              {
+                "createCollection": {
+                  "name": "simple_collection_allow_indexing",
+                  "options" : {
+                    "vector" : {
+                      "size" : 5,
+                      "function" : "cosine"
+                    },
+                    "indexing" : {
+                      "allow" : ["field1", "field2", "address.city", "_id", "$vector"]
+                    }
+                  }
+                }
+              }
+              """;
+
       // create vector collection with indexing allow option
       given()
           .header(HttpConstants.AUTHENTICATION_TOKEN_HEADER_NAME, getAuthToken())
           .contentType(ContentType.JSON)
-          .body(
-              """
-                      {
-                        "createCollection": {
-                          "name": "simple_collection_allow_indexing",
-                          "options" : {
-                            "vector" : {
-                              "size" : 5,
-                              "function" : "cosine"
-                            },
-                            "indexing" : {
-                              "allow" : ["field1", "field2", "address.city", "_id", "$vector"]
-                            }
-                          }
-                        }
-                      }
-                      """)
+          .body(createCollectionRequest)
           .when()
           .post(NamespaceResource.BASE_PATH, namespaceName)
           .then()
           .statusCode(200)
           .body("status.ok", is(1));
-      deleteCollection("simple_collection_allow_indexing");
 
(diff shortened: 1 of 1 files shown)


src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/CollectionSettings.java
```diff
@@ -80,6 +81,17 @@ public static IndexingConfig fromJson(JsonNode jsonNode) {
       }
       return new IndexingConfig(allowed, denied);
     }
+
+    // Need to override to prevent comparison of the supplier
+    @Override
+    public boolean equals(Object o) {
```
tatu-at-datastax, line 87: This is the "real fix".

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/CreateCollectionOperation.java
```diff
@@ -112,45 +112,17 @@ public Uni<Supplier<CommandResult>> execute(QueryExecutor queryExecutor) {
...
-                new JsonApiException(
-                    ErrorCode.INVALID_COLLECTION_NAME,
-                    "The provided collection name '%s' already exists with a vector setting."
-                        .formatted(name)));
-      } else {
-        // if existing table is a non-vector collection, continue
-        return executeCollectionCreation(queryExecutor);
-      }
+    // if table exists we have to choices:
+    // (1) trying to create with same options -> ok, proceed
```
Yuqi-Du, line 116: proceed, but essentially nothing changes. 
Am i understanding it right?

Yuqi-Du (approved):
LGTM!

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/CreateCollectionOperation.java
tatu-at-datastax: @Yuqi-Du  In the sense that this is what used to happen before changes too, yes.

But I think it will now try to re-create Collection, which will be no-op due to "IF NOT EXISTS" used in CQL.
Except! Important case is that if we failed to create one or more of SAIs, these will now be created, which is why retry of a timed out creation could fix things.

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/CollectionSettings.java
```diff
@@ -80,6 +81,17 @@ public static IndexingConfig fromJson(JsonNode jsonNode) {
...
+      if (this == o) return true;
+      if (o instanceof IndexingConfig other) {
```
vkarpov15, line 89: I might be missing something but do we also need to check vector config here?

tatu-at-datastax: This is within nested class, `IndeingConfig` and not for `CollectionSettings` -- `CollectionSettings` uses default, record-provided field-by-field equality check. So that will cover `vectorEnabled` and other fields.

PR #845, opened 2024-01-30
Changes for adding retry delay

maheshrajamani:
**What this PR does**:

Fixes #<issue number>

**Checklist**
- [x] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/QueryExecutor.java
```diff
@@ -164,26 +166,40 @@ private Uni<AsyncResultSet> executeSchemaChange(SimpleStatement boundStatement,
...
-                                  .setIdempotent(true)
-                                  .setSerialConsistencyLevel(
-                                      operationsConfig
-                                          .queriesConfig()
-                                          .consistency()
-                                          .schemaChanges())));
+                  .item(throwable)
+                  .onItem()
+                  .delayIt()
+                  .by(Duration.ofSeconds(1))
```
tatu-at-datastax, line 182: Would it make sense to make this configurable? (and if so, specify in milliseconds, not seconds, to allow things like `200` msecs)

maheshrajamani: Made it configurable. But from what we saw this needs delay in seconds.

tatu-at-datastax: I am not sure I follow... why in seconds? I mean, default value being 1 second is probably fine and makes sense, but is it never useful to configure it in fractions (1.5 seconds etc)?

tatu-at-datastax (approved):

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/QueryExecutor.java
maheshrajamani: Changed to millis as suggested.

jeffreyscarpenter (approved):

PR #851, opened 2024-02-01
(incremental) Catch and re-throw Java CQL Driver fail exceptions as JsonApiException

### LINKED ISSUE #751: Catch and re-throw Java CQL Driver fail exceptions as `JsonApiException` with appropriate `errorCode`s
Currently failures caught by JSON API itself are reported using `JsonApiException`, which takes one of `ErrorCode` enum values, as well as regular Exception `message`. Error code gets included in response message as `errorCode` property, and allows clients to classify failures without having to decode possibly varying (including over time) textual message.
This is useful thing to have.

But same is not currently being done for Exceptions thrown by Java CQL driver which is used to talk to the storage backend (via Coordinator now, possibly directly in future): these exceptions result in error messages, but without `errorCode`.
We need to start catching these exceptions and throwing `JsonApiException`s with appropriate `errorCode`s -- or equivalent handling that results in `errorCode` being included in response message returned to caller.

Yuqi-Du:
authentication failure from driver -> JsonApiException UNAUTHORIZED_REQUEST, 401
invalid query (e.g. vector dimension mismatch) -> JsonApiException INVALID_QUERY, 200
DriverTimeout, WriteTimeout, ReadTimeout -> JsonApiException OPERATION_TIMEOUT, 200

**What this PR does**:
partially fix #751 

This PR should be actively discussed, since driver exception like NoNodeAvailableException, is tricky to locate and map into JsonApiException

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCode.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCode.java
@@ -130,7 +130,11 @@ public enum ErrorCode {
   VECTORIZE_SERVICE_TYPE_UNAVAILABLE("Vectorize service unavailable : "),
   VECTORIZE_USAGE_ERROR("Vectorize search can't be used with other sort clause"),
 
-  VECTORIZECONFIG_CHECK_FAIL("Internal server error: VectorizeConfig check fail");
+  VECTORIZECONFIG_CHECK_FAIL("Internal server error: VectorizeConfig check fail"),
+
+  UNAUTHORIZED_REQUEST("UNAUTHENTICATED: Invalid token"),
+  INVALID_QUERY("INVALID QUERY"),
+  OPERATION_TIMEOUT("OPERATION_TIMEOUT");
 
   private final String message;
 
@@ -144,5 +148,9 @@ public enum ErrorCode {
 
   public JsonApiException toApiException(String format, Object... args) {
     return new JsonApiException(this, message + ": " + String.format(format, args));
+  }
+
+  public JsonApiException toApiException() {
+    return new JsonApiException(this, message);
   }
 }
--- a/src/main/java/io/stargate/sgv2/jsonapi/exception/JsonApiException.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/exception/JsonApiException.java
@@ -48,7 +48,7 @@ public class JsonApiException extends RuntimeException implements Supplier<Comma
     }
 
     // construct and return
-    CommandResult.Error error = getCommandResultError(message);
+    CommandResult.Error error = getCommandResultError(message, Response.Status.OK);
 
     // handle cause as well
     Throwable cause = getCause();
@@ -60,7 +60,7 @@ public class JsonApiException extends RuntimeException implements Supplier<Comma
     }
   }
 
-  public CommandResult.Error getCommandResultError(String message) {
+  public CommandResult.Error getCommandResultError(String message, Response.Status status) {
     Map<String, Object> fieldsForMetricsTag =
         Map.of("errorCode", errorCode.name(), "exceptionClass", this.getClass().getSimpleName());
     SmallRyeConfig config = ConfigProvider.getConfig().unwrap(SmallRyeConfig.class);
@@ -72,7 +72,7 @@ public class JsonApiException extends RuntimeException implements Supplier<Comma
             ? Map.of(
                 "errorCode", errorCode.name(), "exceptionClass", this.getClass().getSimpleName())
             : Map.of("errorCode", errorCode.name());
-    return new CommandResult.Error(message, fieldsForMetricsTag, fields, Response.Status.OK);
+    return new CommandResult.Error(message, fieldsForMetricsTag, fields, status);
   }
 
   public ErrorCode getErrorCode() {
--- a/src/main/java/io/stargate/sgv2/jsonapi/exception/mappers/ThrowableToErrorMapper.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/exception/mappers/ThrowableToErrorMapper.java
@@ -30,7 +30,7 @@ public final class ThrowableToErrorMapper {
       (throwable, message) -> {
         // if our own exception, shortcut
         if (throwable instanceof JsonApiException jae) {
-          return jae.getCommandResultError(message);
+          return jae.getCommandResultError(message, Response.Status.OK);
         }
         // Override response error code
         SmallRyeConfig config = ConfigProvider.getConfig().unwrap(SmallRyeConfig.class);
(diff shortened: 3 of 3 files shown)


Yuqi-Du:
currently identified:

| Exception                   | Commands                                               |
|-----------------------------|--------------------------------------------------------|
| DriverTimeoutException      | CreateCollectionCommand                                |
| NoNodeAvailableException    | CreateCollectionCommand                                |
|                             | FindCommand                                             |
| AllNodesFailedException     | CreateCollectionCommand                                |
|                             | FindCollectionsCommand                                 |
|                             | FindNamespacesCommand                                   |
| NoNodeAvailableException    | DeleteCollectionCommand                                |
|                             | FindOneCommand                                          |
|                             | InsertManyCommand                                       |
|                             | CreateCollectionCommand                                |
|                             | FindCommand                                             |
| InvalidQueryException       | CountDocumentsCommand                                   |
|                             | CreateCollectionCommand                                |
|                             | DeleteCollectionCommand                                |
|                             | FindCommand                                             |
|                             | InsertManyCommand                                       |
|                             | InsertOneCommand                                        |
| ClosedConnectionException   | CountDocumentsCommand                                   |
|                             | CreateCollectionCommand                                |
|                             | DeleteCollectionCommand                                |
|                             | FindCommand                                             |
| ReadFailureException         | CountDocumentsCommand                                   |
|                             | FindCommand                                             |
| UnauthorizedException        | CreateCollectionCommand                                |
| ServerError                  | CreateCollectionCommand                                |

src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCode.java
```diff
@@ -130,7 +130,13 @@ public enum ErrorCode {
   VECTORIZE_SERVICE_TYPE_UNAVAILABLE("Vectorize service unavailable : "),
   VECTORIZE_USAGE_ERROR("Vectorize search can't be used with other sort clause"),
 
-  VECTORIZECONFIG_CHECK_FAIL("Internal server error: VectorizeConfig check fail");
+  VECTORIZECONFIG_CHECK_FAIL("Internal server error: VectorizeConfig check fail"),
+
+  UNAUTHORIZED_REQUEST("UNAUTHENTICATED: Invalid token"),
```
tatu-at-datastax, line 135: Wrong enum: "UNAUTHORIZED" should be "UNAUTHENTICATED"?

src/main/java/io/stargate/sgv2/jsonapi/exception/mappers/ThrowableToErrorMapper.java
```diff
@@ -30,71 +32,90 @@ public final class ThrowableToErrorMapper {
...
             || throwable
                 instanceof com.datastax.oss.driver.api.core.servererrors.UnauthorizedException) {
-          return new CommandResult.Error(
-              "UNAUTHENTICATED: Invalid token",
-              fieldsForMetricsTag,
-              fields,
-              Response.Status.UNAUTHORIZED);
+          return ErrorCode.UNAUTHORIZED_REQUEST
+              .toApiException()
+              .getCommandResultError(message, Response.Status.UNAUTHORIZED);
```
tatu-at-datastax, line 53: I am not sure: should this be converted to `JsonApiException`, if it needs to produce non-200 HTTP Response code?

I guess we just need to decide which way to go. I guess there are some benefits in making everything `JsonApiException`, but not necessarily HTTP 200 responses.

src/main/java/io/stargate/sgv2/jsonapi/exception/mappers/ThrowableToErrorMapper.java
```diff
@@ -30,71 +32,90 @@ public final class ThrowableToErrorMapper {
...
-                                .getMessage()
-                                .contains(
-                                    "Provided username token and/or password are incorrect")))) {
-                  return new CommandResult.Error(
-                      "UNAUTHENTICATED: Invalid token",
-                      fieldsForMetricsTag,
-                      fields,
-                      Response.Status.UNAUTHORIZED);
-                }
+          return ErrorCode.DRIVER_TIMEOUT
```
tatu-at-datastax, line 66: We really should get more information from exception (e.getMessage()?) to pass along. Otherwise it's 

    Driver timeout on

with nothing describing what actually happens?

src/main/java/io/stargate/sgv2/jsonapi/exception/mappers/ThrowableToErrorMapper.java
```diff
@@ -30,71 +32,90 @@ public final class ThrowableToErrorMapper {
...
+              // connect to oss cassandra throws AuthenticationException for invalid credentials
+              // connect to AstraDB throws IllegalArgumentException for invalid token/credentials
+              if (error instanceof AuthenticationException
+                  || (error instanceof IllegalArgumentException
+                      && (error.getMessage().contains("AUTHENTICATION ERROR")
+                          || error
+                              .getMessage()
+                              .contains(
+                                  "Provided username token and/or password are incorrect")))) {
+                return ErrorCode.UNAUTHORIZED_REQUEST
```
tatu-at-datastax, line 96: As per earlier comment, `ErrorCode` should "UNAUTHENTICATED_".

src/main/java/io/stargate/sgv2/jsonapi/exception/mappers/ThrowableToErrorMapper.java
```diff
@@ -30,71 +32,90 @@ public final class ThrowableToErrorMapper {
...
+                    .toApiException()
+                    .getCommandResultError(message, Response.Status.UNAUTHORIZED);
```
tatu-at-datastax, line 98: ... that despite HTTP status has to be "UNAUTHORIZED" :)

src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCode.java
Yuqi-Du: Thanks

src/main/java/io/stargate/sgv2/jsonapi/exception/mappers/ThrowableToErrorMapper.java
Yuqi-Du: Yeah, that is why the above method has status added.
I think currently design for Authentication Failure is to return JsonApiException with 401.

src/main/java/io/stargate/sgv2/jsonapi/exception/JsonApiException.java
```diff
@@ -60,7 +60,7 @@ public CommandResult get() {
     }
   }
 
-  public CommandResult.Error getCommandResultError(String message) {
+  public CommandResult.Error getCommandResultError(String message, Response.Status status) {
```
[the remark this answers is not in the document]
Yuqi-Du: resolved for now to make the page neat, will change it if we decide 200s for all JsonApiExceptions

src/main/java/io/stargate/sgv2/jsonapi/exception/mappers/ThrowableToErrorMapper.java
tatu-at-datastax: This may make sense, as it is more of a "hard" failure, and matches what HTTP 401 response is about.

src/main/java/io/stargate/sgv2/jsonapi/exception/mappers/ThrowableToErrorMapper.java
tatu-at-datastax: Ok. As long as that is retained in log output that's fine. You could also pass it with

   .toApiException(message)

but since we do not actually use exception for anything I guess it does not matter.

tatu-at-datastax (approved):

src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCode.java
```diff
@@ -130,7 +130,13 @@ public enum ErrorCode {
   VECTORIZE_SERVICE_TYPE_UNAVAILABLE("Vectorize service unavailable : "),
   VECTORIZE_USAGE_ERROR("Vectorize search can't be used with other sort clause"),
 
-  VECTORIZECONFIG_CHECK_FAIL("Internal server error: VectorizeConfig check fail");
+  VECTORIZECONFIG_CHECK_FAIL("Internal server error: VectorizeConfig check fail"),
+
+  UNAUTHENTICATED_REQUEST("UNAUTHENTICATED: Invalid token"),
+  INVALID_QUERY("Invalid query"),
+  DRIVER_TIMEOUT("Driver timeout on"),
```
jeffreyscarpenter, line 137: Why is the word "on" included?

src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCode.java
```diff
@@ -130,7 +130,13 @@ public enum ErrorCode {
...
+  DRIVER_CLOSED_CONNECTION("Driver request connection is closed due to an unrelated event"),
```
jeffreyscarpenter, line 138: I don't know what "due to an unrelated event" means, so I doubt a user would either. Are we sure we want to have this as something we report as a separate error? Is this a case where we should do a retry?

src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCode.java
jeffreyscarpenter: Also I'm not sure we should mention the driver? it's an implementation detail.

jeffreyscarpenter (commented):
I think we should discuss what makes sense to expose as its own error code. Seems to me we are getting too fine grained in the error codes with implementation-specific cases.

src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCode.java
tatu-at-datastax: I think we should re-phrase, or remove "due to an unrelated event". I get what it means -- we did not close it,  but something caused it to close (server-side, proxy) -- but I agree with @jeffreyscarpenter that it probably brings more questions than answers.

tatu-at-datastax (approved):

PR #863, opened 2024-02-06
Bug fix for array size filter for `0` length

### LINKED ISSUE #862: Filtering array element by $size: 0 doesn't seem to work
Sending the following query:

```
{
  "find": {
    "filter": {
      "tags": {
        "$size": 0
      }
    }
  }
}
```

returns all documents where `tags` is **not** an array of length 0. I first noticed this while working on test failures for this PR: https://github.com/stargate/stargate-mongoose/pull/183, looks like this changed between v1.0.0-BETA-6 and v1.0.0. Maybe with `$not` handling?

maheshrajamani:
**What this PR does**:
Bug fix for array size filter for `0` length

Fixes #862 

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -23,7 +23,6 @@ Here are some Stargate-relevant property groups that are necessary for correct s
 ## Document limits configuration
 *Configuration for document limits, defined by [DocumentLimitsConfig.java](src/main/java/io/stargate/sgv2/jsonapi/config/DocumentLimitsConfig.java).*
 
-<<<<<<< HEAD
 | Property                                                        | Type  | Default     | Description                                                                             |
 |-----------------------------------------------------------------|-------|-------------|-----------------------------------------------------------------------------------------|
 | `stargate.jsonapi.document.limits.max-size`                     | `int` | `4_000_000` | The maximum size of (in characters) a single document.                                  |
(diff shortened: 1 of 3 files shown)


Yuqi-Du (approved):
LGTM

tatu-at-datastax (approved):

tatu-at-datastax:
Quick question: do we know what happens if negative value is given? Is that valid argument?

maheshrajamani:
> Quick question: do we know what happens if negative value is given? Is that valid argument?

We don't support -ve number for `$size`. It's coded for it internally to support pushdown of logical `$not` operator.

tatu-at-datastax:
@maheshrajamani That makes sense. However, I wonder then why did the original report suggest that `$size` with `0` return everything BUT 0-sized array documents -- if it was translated to `$not` (which I think is what was happening due to wrong comparison), shouldn't it have returned nothing?

vkarpov15:
LGTM. I was concerned for a minute that this change meant that `{ $not: { $size: 0 } }` would no longer work, but it looks like we don't support `$not` with `$size` currently. So no new issues.

tatu-at-datastax:
@vkarpov15 I had the same question and @maheshrajamani explained how it worked so yeah, great minds think alike etc. :)

PR #864, opened 2024-02-06
Rename jsonapi -> dataapi

### LINKED ISSUE #860: Renaming to Data API
As part of marketing decisions around General Availability of the API, we renamed it to "Data API" when deploying in Astra. We should perform appropriate renaming on this open source repo, with the goal of having minimal impact to downstream dependencies. 

The following changes are proposed:

- rename this repo from `stargate/jsonapi` to `stargate/dataapi`. GitHub will handle forwarding to the new address.
- update documentation such as the README and spec
- update how the app identifies itself in quarkus
- note that the postman collections for the API have already been renamed
- docker image name to change from `stargateio/jsonapi` to `stargateio/dataapi`. We will tag images with both names on Docker Hub for some duration (2 months?) before retiring the `jsonapi` name.

The following references to `json` or `jsonapi` are proposed to not be changed:

- the URL path `/v1/json...`, as changing would break compatibility for many users and client libraries.
- metrics reporting as `sgv2-jsonapi`
- Source code structure such as 
  - base Java package name `io.stargate.sgv2.jsonapi` 
  - named classes such as `JsonApiException`, `JsonApiMetricsConfig`, etc. 
  - base property name `stargate.jsonapi`

We would target these changes for an early post GA (1.0.0) release such as 1.0.1 or 1.0.2.

Yuqi-Du:
Changes
- banner
- readme
- spec
- nosqlbench file, fallout
- docker image (publish another one image called dataapi, only for dockerhub)
- quarkus app identification: `quarkus: application: name: Stargate DATA API`
- META-INF/resources/index.html

**What this PR does**:

Fixes #860 

**Checklist**
- [x] Changes manually tested


.github/workflows/image-publish-dockerhub.yaml
```diff
@@ -0,0 +1,52 @@
+name: Publish Docker Image to docker hub
```
jeffreyscarpenter, line 1: I am not sure we want a manually triggered workflow for this, why isn't the current release workflow sufficient?

jeffreyscarpenter (commented):
We should also change the naming used in the swagger/openapi spec? src/main/java/io/stargate/sgv2/jsonapi/StargateJsonApi.java

Should probably also change

- docs/README.md
- nosqlbench/http-jsonapi-search-filter-sort.yaml
- src/main/resources/META-INF/resources/index.html

clun:
(minor), It feels more natural to call repository name and docker image `data-api` instead of of `dataapi`

.github/workflows/image-publish-dockerhub.yaml
Yuqi-Du: Besides the release workflow, ECR image could also be published by '[docker-image-publish.yaml'](https://github.com/stargate/jsonapi/blob/main/.github/workflows/docker-image-publish.yaml). 
I was thinking about making the similar one to dockerhub image also.

Yuqi-Du: Deleted.

PR #866, opened 2024-02-12
EstimatedDocumentCount operation

jeffreyscarpenter:
Adds a new operation EstimatedDocumentCount to the API.

**Checklist**
- [x] Changes manually tested


docs/jsonapi-spec.md
```diff
@@ -540,6 +540,32 @@ See [Multi-Document Failure Considerations](#multi-document-failure-consideratio
...
+#### estimatedDocumentCount Command Options
+
+The `estimatedDocumentCount` command does not support any options.
+
+#### estimatedDocumentCount Command Response
+
+| Response Elements | Description                                             |
+| ----------------- | ------------------------------------------------------- |
+| `data`            | Not present                                             |
+| `status`          | Preset with fields: `count: <zero-or-positive-integer>` |
```
tatu-at-datastax, line 564: Nit: preset -> present

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/ReadOperation.java
```diff
@@ -404,6 +409,55 @@ private void getCount(AsyncResultSet rs, Throwable error, AtomicLong counter) {
...
+        .onItem()
+        .transform(
+            rs -> {
+              return new CountResponse(counter.get());
+            });
+  }
+
+  private void getEstimatedCount(AsyncResultSet rs, Throwable error, AtomicLong counter) {
+    if (error != null) {
+      throw new JsonApiException(ErrorCode.COUNT_READ_FAILED);
```
tatu-at-datastax, line 441: Should include `error.getMessage()` in description? Something like

```
throw ErrorCode.COUNT_READ_FAILED.toApiException("root cause: %s", error.getMessage());
```

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/ReadOperation.java
```diff
@@ -53,6 +53,11 @@ public interface ReadOperation extends Operation {
           "query_null_values['%s']",
           "query_timestamp_values['%s']");
   int SORT_INDEX_COLUMNS_SIZE = sortIndexColumns.size();
+
+  double TOTAL_TOKEN_RANGE = Math.pow(2, 64);
+  double MAX_TOKEN = Math.pow(2, 63) - 1;
```
tatu-at-datastax, line 58: Does it make sense to have these as `double`s instead of as `long`s?

long MAX_TOKEN = Long.MAX_VALUE;
long MIN_TOKEN = Long.MIN_VALUE;

(I think)

considering they are compared to `long`/`Long` values later on.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/ReadOperation.java
```diff
@@ -404,6 +409,55 @@ private void getCount(AsyncResultSet rs, Throwable error, AtomicLong counter) {
...
+    if (error != null) {
+      throw new JsonApiException(ErrorCode.COUNT_READ_FAILED);
+    } else {
+
+      for (Row row : rs.currentPage()) {
+        Long rangeStart = Long.valueOf(row.getString("range_start"));
+        Long rangeEnd = Long.valueOf(row.getString("range_end"));
+        double rangeSize = 0;
+        if (rangeStart > rangeEnd) {
+          rangeSize = (MAX_TOKEN - rangeStart) + (rangeEnd - MIN_TOKEN);
```
tatu-at-datastax, line 449: this is where I wonder if `long` vs `double` might not work well: they have same number of bits but accuracy of `double` actually only covers subset of integer positions I think (causing issues in Javascript where 64-bit `double` is the only numeric type, and cannot accurately express all of 64-bit `long` range)

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/ReadOperation.java
jeffreyscarpenter: We went back and forth on this but as @Yuqi-Du identified, we actually need to perform calculations that can potentially exceed Long.MAX_VALUE so it's safer to treat as doubles.

tatu-at-datastax: Ok. Just be aware of `double` NOT being able to accurately represent all `long` values (only up to 2^53). This is why Javascript has all the "fun" issues wrt serializing of 64-bit long values.

I doubt this will be problematic either way, and implementation can be changed if we do see problems.

src/main/resources/application.yaml
```diff
@@ -68,6 +68,9 @@ quarkus:
   banner:
     path: banner/banner.txt
 
+  log:
```
tatu-at-datastax, line 71: Is this intentional? I think we try to keep settings in alphabetic order.
(I can see it was moved from being under `http` so that was necessary)

tatu-at-datastax (approved):
LGTM

src/main/resources/application.yaml
jeffreyscarpenter: I see. I moved it where it made logical sense to me, was not aware of alpha order convention

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/EstimatedDocumentCountCommand.java
```diff
@@ -0,0 +1,12 @@
+package io.stargate.sgv2.jsonapi.api.model.command.impl;
+
+import com.fasterxml.jackson.annotation.JsonTypeName;
+import io.stargate.sgv2.jsonapi.api.model.command.NoOptionsCommand;
+import io.stargate.sgv2.jsonapi.api.model.command.ReadCommand;
+import org.eclipse.microprofile.openapi.annotations.media.Schema;
+
+@Schema(
+    description =
+        "Command that returns estimated count of documents in a collection based on the collection.")
```
vkarpov15, line 10: "based on the collection." reads a bit strange - maybe "based on the collection metadata." or something similar?

tatu-at-datastax: Ok this was missed in code review; definitely accidentally weird. :)

PR #869, opened 2024-02-12
Increase default maximum HTTP body size at server level from 5MB to 20MB

### LINKED ISSUE #856: Increase current `quarkus.http.limits.max-body-size` from `5M` to `20M`
One of few service-protection settings we have is to limit maximum accepted request by Quarkus HTTP endpoints.
It was set at 5 megs a while ago; but with the Document Limits being raised to allow 4 meg documents, and insert batches allowing 20 documents, this limit is too low.

We may not want to allow absolute maximum batch size (for which we'd need to allow almost 100 meg requests), but should at least raise it to, say, 20 megs. And document the limit as well.

tatu-at-datastax:
**What this PR does**:

Increases Quarkus HTTP limit for maximum body size to 20 megs, to better align with increased maximum document size (4M) allowed at API level

Fixes #856

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/src/main/resources/application.yaml
+++ b/src/main/resources/application.yaml
@@ -84,10 +84,11 @@ quarkus:
           policy: authenticated
 
     limits:
-      # Let's limit low-level maximum HTTP request size to 5 megs: stricter limit (1 meg)
-      # is applied at the JSON API level. Low-level limits may result in EPIPE/413 errors
+      # Let's limit low-level maximum HTTP request size to 20 megs: stricter
+      # limit (4 meg per document) is applied at the JSON API level.
+      # Low-level limits may result in EPIPE/413 errors
       # whereas at higher level we can use regular JSON API error responses
-      max-body-size: 5M
+      max-body-size: 20M
 
     port: 8181
 
--- a/src/main/resources/application.yaml
+++ b/src/main/resources/application.yaml
@@ -88,7 +88,7 @@ quarkus:
       # limit (4 meg per document) is applied at the JSON API level.
       # Low-level limits may result in EPIPE/413 errors
       # whereas at higher level we can use regular JSON API error responses
-      max-body-size: 20M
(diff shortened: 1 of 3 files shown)


Hazel-Datastax (approved):
LGTM

src/main/resources/application.yaml
```diff
@@ -84,10 +84,11 @@ quarkus:
...
 
     limits:
-      # Let's limit low-level maximum HTTP request size to 5 megs: stricter limit (1 meg)
-      # is applied at the JSON API level. Low-level limits may result in EPIPE/413 errors
+      # Let's limit low-level maximum HTTP request size to 20 megs: stricter
+      # limit (4 meg per document) is applied at the JSON API level.
+      # Low-level limits may result in EPIPE/413 errors
       # whereas at higher level we can use regular JSON API error responses
-      max-body-size: 5M
+      max-body-size: 20M
```
maheshrajamani, line 91: Can you add this to the CONFIGURATION.md file?

maheshrajamani: Do we need to check limit on each document in insertMany? What if user add one document of 10 MB size with no index set. We will try to store the data into single doc_json field. Is that the intended change?

maheshrajamani (approved):
LGTM. Asked for a clarification

src/main/resources/application.yaml
tatu-at-datastax: Document sizes are still checked by shredder separately, and 4 meg limit is for document regardless of indexing. This limit is at lower, HTTP level, mostly to prevent someone from sending say 1 gig request. If so, we'd read it all in memory, do processing, and only later on hit per-document limit.

So this is just for DoS reasons. Specific value to use is somewhat arbitrary; could use larger value if we want to (f.ex to allow larger payloads for insertMany)

PR #871, opened 2024-02-13
circular reference fix and add operationId for each path

Yuqi-Du:
One issue we have for data api OpenAPI spec is that logicalExpression has circular reference, which will cause OpenAPI spec validation failure.
Also add operationId for each path.
- /v1
- /v1/{namespace}
- /v1/{namespace}/{collection}

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/filter/FilterClause.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/filter/FilterClause.java
@@ -19,7 +19,7 @@ import org.eclipse.microprofile.openapi.annotations.media.Schema;
     example = """
              {"name": "Aaron", "country": "US"}
               """)
-public record FilterClause(LogicalExpression logicalExpression) {
+public record FilterClause(@Schema(hidden = true) LogicalExpression logicalExpression) {
   public void validate(CommandContext commandContext) {
     DocumentProjector indexingProjector = commandContext.indexingProjector();
     // If nothing specified, everything indexed
--- a/src/main/java/io/stargate/sgv2/jsonapi/api/v1/CollectionResource.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/api/v1/CollectionResource.java
@@ -72,7 +72,8 @@ public class CollectionResource {
 
   @Operation(
       summary = "Execute command",
-      description = "Executes a single command against a collection.")
+      description = "Executes a single command against a collection.",
+      operationId = "collection operation")
   @Parameters(
       value = {
         @Parameter(name = "namespace", ref = "namespace"),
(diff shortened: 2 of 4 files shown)


src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/filter/FilterClause.java
```diff
@@ -19,7 +19,7 @@
     example = """
              {"name": "Aaron", "country": "US"}
               """)
-public record FilterClause(LogicalExpression logicalExpression) {
```
tatu-at-datastax, line 22: Does this need to be hidden to prevent cyclic definition or... ?

Yuqi-Du: Exactly

kathirsvn: Will it have any negative impact? (like not exposing our API schema as intended?)

Yuqi-Du:
```
      "FilterClause": {
        "type": "object",
        "properties": {
          "logicalExpression": {
            "$ref": "#/components/schemas/LogicalExpression"
          }
        },
        "example": {
          "name": "Aaron",
          "country": "US"
        }
      },
```

For example, previous we have FilterClause schema like this.
Now it will be 

  ```
    "FilterClause" : {
        "type" : "object",
        "example" : {
          "name" : "Aaron",
          "country" : "US"
        }
      },
```

Yuqi-Du: Is there a need to expose logicalExpression property?

kathirsvn: I believe yes, since its part of the API definition. I see it is used in `DeleteManyCommand` and `DeleteOneCommand` as well apart from the `FilterClause`.

Yuqi-Du: Sure, let's discuss it.
So, we certainly has the circular reference in our design for logicalExpression, that is for the logical operator like $or and $and.

Regarding the openAPI world, the criteria for the circular reference is different. For example, swaggerUI will work with a circular reference spec. However, for example, GPT action validator will error with a circular reference spec.

Some references: 
https://apitools.dev/swagger-parser/docs/#:~:text=//%20Error%0A%7D-,Circular%20%24Refs,-Swagger%20APIs%20can
https://github.com/stoplightio/prism/issues/1456

kathirsvn: 👍 I understand that part. I'm not sure if hiding that field from the schema is the solution here. Should we work with the OpenAI team to understand on how this kind of schema can be validated (because having recursive seems in a OpenAPI spec seems normal to me).

PR #879, opened 2024-02-15
index creation failure msg, usage with collection not fully indexed

### LINKED ISSUE #812: Invalid request / allow filtering error on find command
User reported the following error when running a find command:

```
{"errors":[{"message":"Cannot execute this query as it might involve data filtering and thus may have unpredictable performance. If you want to execute this query despite the performance unpredictability, use ALLOW FILTERING","errorCode":"INVALID_REQUST"}],"command":{"find":{"filter":{"tags":{"$in":["javascript"]}},"projection":{"$vector":0},"options":{"limit":20}}}}
```

The find command was something like:

```
{"find":{"filter":{"tags":{"$in":["tools"]}},"projection":{"$vector":0},"options":{"limit":20}}}}
```

Yuqi-Du:
There are chances of the collection creation is successful, but not fully indexed. 
Even though we now get less failures because index creation changes into serial execution.
This PR will throw a JsonApiException, when there is table creation failure or index creation failure.
Also, if user is using a not fully indexed collection, also prompt with proper error msgs.

Fixes #812

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCode.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCode.java
@@ -137,8 +137,7 @@ public enum ErrorCode {
   DRIVER_TIMEOUT("Driver timeout"),
   DRIVER_CLOSED_CONNECTION("Driver request connection is closed"),
   NO_NODE_AVAILABLE("No node was available to execute the query"),
-  NO_INDEX_ERROR(
-      "Collection creation failure (unable to create indexes). Recommend re-creating the collection"),
+  NO_INDEX_ERROR("Faulty collection (missing indexes). Recommend re-creating the collection"),
   COLLECTION_CREATION_ERROR(
       "Collection creation failure (unable to create table). Recommend re-creating the collection");
 
--- a/src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCode.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCode.java
@@ -138,8 +138,9 @@ public enum ErrorCode {
   DRIVER_CLOSED_CONNECTION("Driver request connection is closed"),
   NO_NODE_AVAILABLE("No node was available to execute the query"),
   NO_INDEX_ERROR(
-      "The collection has indexes creation failure, recommend to recreate the collection"),
-  COLLECTION_CREATION_ERROR("Collection creation failure, recommend to recreate the collection");
+      "Collection creation failure (unable to create indexes). Recommend re-creating the collection"),
+  COLLECTION_CREATION_ERROR(
+      "Collection creation failure (unable to create table). Recommend re-creating the collection");
 
   private final String message;
 
--- a/src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCode.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCode.java
@@ -130,7 +130,13 @@ public enum ErrorCode {
   VECTORIZE_SERVICE_TYPE_UNAVAILABLE("Vectorize service unavailable : "),
   VECTORIZE_USAGE_ERROR("Vectorize search can't be used with other sort clause"),
 
-  VECTORIZECONFIG_CHECK_FAIL("Internal server error: VectorizeConfig check fail");
+  VECTORIZECONFIG_CHECK_FAIL("Internal server error: VectorizeConfig check fail"),
+
+  UNAUTHENTICATED_REQUEST("UNAUTHENTICATED: Invalid token"),
+  INVALID_QUERY("Invalid query"),
+  DRIVER_TIMEOUT("Driver timeout"),
+  DRIVER_CLOSED_CONNECTION("Driver request connection is closed"),
+  NO_NODE_AVAILABLE("No node was available to execute the query");
 
   private final String message;
 
@@ -144,5 +150,9 @@ public enum ErrorCode {
 
   public JsonApiException toApiException(String format, Object... args) {
     return new JsonApiException(this, message + ": " + String.format(format, args));
+  }
+
+  public JsonApiException toApiException() {
+    return new JsonApiException(this, message);
   }
 }
--- a/src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCode.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCode.java
@@ -61,9 +61,9 @@ public enum ErrorCode {
 
   SHRED_BAD_EJSON_VALUE("Bad EJSON value"),
 
-  SHRED_BAD_VECTOR_SIZE("$vector field can't be empty"),
+  SHRED_BAD_VECTOR_SIZE("$vector value can't be empty"),
 
-  SHRED_BAD_VECTOR_VALUE("$vector search needs to be array of numbers"),
+  SHRED_BAD_VECTOR_VALUE("$vector value needs to be array of numbers"),
   SHRED_BAD_VECTORIZE_VALUE("$vectorize search needs to be text value"),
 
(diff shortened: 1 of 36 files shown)


src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCode.java
```diff
@@ -136,7 +136,10 @@ public enum ErrorCode {
   INVALID_QUERY("Invalid query"),
   DRIVER_TIMEOUT("Driver timeout"),
   DRIVER_CLOSED_CONNECTION("Driver request connection is closed"),
-  NO_NODE_AVAILABLE("No node was available to execute the query");
+  NO_NODE_AVAILABLE("No node was available to execute the query"),
+  NO_INDEX_ERROR(
+      "The collection has indexes creation failure, recommend to recreate the collection"),
```
jeffreyscarpenter, line 141: How about: "Collection creation failure (unable to create indexes). Recommend re-creating the collection"

src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCode.java
```diff
@@ -136,7 +136,10 @@ public enum ErrorCode {
...
+  COLLECTION_CREATION_ERROR("Collection creation failure, recommend to recreate the collection");
```
jeffreyscarpenter, line 142: How about: "Collection creation failure (unable to create table). Recommend re-creating the collection"

jeffreyscarpenter (commented):
minor suggestion

src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCode.java
Yuqi-Du: Thanks!

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/CreateCollectionOperation.java
```diff
@@ -171,7 +171,17 @@ private Uni<Supplier<CommandResult>> executeCollectionCreation(QueryExecutor que
...
                   }
                 });
-    return indexResult.onItem().transform(SchemaChangeResult::new);
+    return indexResult
+        .onItem()
+        .transform(
+            res -> {
+              if (!res) {
+                // table creation failure or index creation failure
+                return ErrorCode.COLLECTION_CREATION_ERROR.toApiException();
```
tatu-at-datastax, line 180: I think error message should contain name of Collection for which creation failed (should be available from context right?)

src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCode.java
```diff
@@ -136,7 +136,11 @@ public enum ErrorCode {
   INVALID_QUERY("Invalid query"),
   DRIVER_TIMEOUT("Driver timeout"),
   DRIVER_CLOSED_CONNECTION("Driver request connection is closed"),
-  NO_NODE_AVAILABLE("No node was available to execute the query");
+  NO_NODE_AVAILABLE("No node was available to execute the query"),
+  NO_INDEX_ERROR(
```
tatu-at-datastax, line 140: Since this is thrown at query time (if I understand it correctly), not creation time, it should probably be worded to indicate that Collection is missing indexes (possibly due to faulty creation), but not that operation failing was CreateCollection. To me latter would be confusing as a user.

tatu-at-datastax (commented):
I like improvements here, but I think we can probably improve on wording & inclusion of information (collection name should be included in all, I think). And for "missing index" case indicate it as such (instead of suggesting operation that failed was create-collection -- if I understood logic correctly).

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/CreateCollectionOperation.java
Yuqi-Du: yes, we can get from the context.
But not sure about this, since it is indeed the createCollectionOperation, and request path already has the collection name. 
Clients should be aware of the collection name they are creating?

src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCode.java
Yuqi-Du: thanks. make sense!

src/main/java/io/stargate/sgv2/jsonapi/exception/mappers/ThrowableToErrorMapper.java
```diff
@@ -57,6 +57,13 @@ public final class ThrowableToErrorMapper {
           if (message.contains("vector<float,")) {
             message = "Mismatched vector dimension";
           }
+          if (message.contains(
+                  "If you want to execute this query despite the performance unpredictability, use ALLOW FILTERING")
+              || message.contains("ANN ordering by vector requires the column to be indexed")) {
```
vkarpov15, line 62: Does this mean that any error message that contains "If you want to execute this query despite the performance unpredictability, use ALLOW FILTERING" will be replaced with "Faulty collection (missing indexes). Recommend re-creating the collection"?

I think the new error message may be missing some valuable information, like the fact that this error is due to vector search or a query on an unindexed field. Maybe we can add some more details, like some info on which index is missing? Or at least have separate error messages for vector search vs other missing indexes?

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/CreateCollectionOperation.java
vkarpov15: I'd recommend putting the collection name in the error message, it doesn't add too much bulk to the error message and it can help provide context in cases where the client isn't able to identify the offending collection from their logs.

src/main/java/io/stargate/sgv2/jsonapi/exception/mappers/ThrowableToErrorMapper.java
jeffreyscarpenter: I'm not sure that knowing which index was impacted has much benefit to the user, as it doesn't change the required action.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/CreateCollectionOperation.java
tatu-at-datastax: Right, context of call is not always super obvious from, say, logs. So a little bit of contextual information can be quite useful to know what individual response is about.

src/main/java/io/stargate/sgv2/jsonapi/exception/mappers/ThrowableToErrorMapper.java
tatu-at-datastax: As long as there is indication of missing index(es), makes sense to me.
Knowing which index(es) missing would be somewhat useful to us (maintainers) but not so much users, agree.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/CreateCollectionOperation.java
Yuqi-Du: Thanks guys! got it.

src/main/java/io/stargate/sgv2/jsonapi/exception/mappers/ThrowableToErrorMapper.java
```diff
@@ -57,6 +57,13 @@ public final class ThrowableToErrorMapper {
...
+            return ErrorCode.NO_INDEX_ERROR
```
tatu-at-datastax, line 63: I think this should still add Collection name (just not index name(s)).

PR #885, opened 2024-02-20
Fixes #884: unify JSON Object member reference to "property" (from "field" etc)

### LINKED ISSUE #884: Unify terms "field" and "property" in `ErrorCode` as "property"
(note: follow-up to #878)

Looks like we use terms "field" and "property" somewhat interchangeably in error messages, when referring content of Documents stored. In JSON, preferred term is "property" (although multiple other terms are sometimes used as well) so let's unify error messages to use that over field.

Note: term "field" is still to be used for arguments passed to Command -- for example "filter fields". This so that "property" should generally imply something within stored/retrieved Document, and field something else.

tatu-at-datastax:
**What this PR does**:

Unifies terms used to refer to key/value pairs in JSON Document by using "property", over alternate names like "field"

Fixes #884

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCode.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCode.java
@@ -61,9 +61,9 @@ public enum ErrorCode {
 
   SHRED_BAD_EJSON_VALUE("Bad EJSON value"),
 
-  SHRED_BAD_VECTOR_SIZE("$vector property can't be empty"),
+  SHRED_BAD_VECTOR_SIZE("$vector value can't be empty"),
 
-  SHRED_BAD_VECTOR_VALUE("$vector search needs to be array of numbers"),
+  SHRED_BAD_VECTOR_VALUE("$vector value needs to be array of numbers"),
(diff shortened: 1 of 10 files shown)


src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCode.java
```diff
@@ -61,9 +61,9 @@ public enum ErrorCode {
 
   SHRED_BAD_EJSON_VALUE("Bad EJSON value"),
 
-  SHRED_BAD_VECTOR_SIZE("$vector field can't be empty"),
+  SHRED_BAD_VECTOR_SIZE("$vector value can't be empty"),
```
tatu-at-datastax, line 64: Since this is more general reference (used from multiple places), cannot refer as "property", but that is implied by name `$vector` -- hence "value".

Yuqi-Du: make sence

Yuqi-Du (approved):

vkarpov15 (commented):
LGTM

PR #888, opened 2024-02-21
Support cohere embedding service api

maheshrajamani:
**What this PR does**:
Support cohere embedding service api

**Checklist**
- [x] Changes manually tested


Hazel-Datastax (approved):
LGTM

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/configuration/PropertyBasedEmbeddingServiceConfigStore.java
```diff
@@ -49,6 +49,14 @@ public EmbeddingServiceConfigStore.ServiceConfig getConfiguration(
         throw new JsonApiException(
             ErrorCode.VECTORIZE_SERVICE_TYPE_NOT_ENABLED,
             ErrorCode.VECTORIZE_SERVICE_TYPE_NOT_ENABLED.getMessage() + serviceName);
+      case "cohere":
+        if (config.cohere().enabled()) {
+          return ServiceConfig.provider(
+              serviceName, serviceName, config.cohere().apiKey(), config.cohere().url().toString());
+        }
+        throw new JsonApiException(
```
tatu-at-datastax, line 57: There's new method in `ErrorCode` to simplify throwing:

    throw ErrorCode.VECTORIZE_SERVICE_TYPE_NOT_ENABLED.toApiException(serviceName);

it will call `String.format()`on arguments, concatenate that message to base message defined for enum. Slightly less code to write.

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/CohereEmbeddingClient.java
```diff
@@ -0,0 +1,117 @@
...
+        @HeaderParam("Authorization") String accessToken, EmbeddingRequest request);
+  }
+
+  private record EmbeddingRequest(String[] texts, String model, String input_type) {}
+
+  private static class EmbeddingResponse {
+
+    public EmbeddingResponse() {}
+
+    @JsonIgnore private String id;
```
tatu-at-datastax, line 50: You can add `@JsonIgnore` here, although more common is to add annotation to getter method.
(both work as long as names are linked)..

tatu-at-datastax: Actually, if we do not actually need these values and just want to ignore them coming from server, can instead use `@JsonIgnoreProperties` on class declaration, and omit field/getter/setter?
I'll add note above to show how.

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/CohereEmbeddingClient.java
```diff
@@ -0,0 +1,117 @@
...
+    @POST
+    @Path("/embed")
+    @ClientHeaderParam(name = "Content-Type", value = "application/json")
+    Uni<EmbeddingResponse> embed(
+        @HeaderParam("Authorization") String accessToken, EmbeddingRequest request);
+  }
+
+  private record EmbeddingRequest(String[] texts, String model, String input_type) {}
+
+  private static class EmbeddingResponse {
```
tatu-at-datastax, line 46: If we want to ignore 2 fields we don't care about, can just add:

```
@JsonIgnoreProperties({ "meta", "response_type" })
```

and then no need to add setters, getters, or field.

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/CohereEmbeddingClient.java
```diff
@@ -0,0 +1,117 @@
...
+
+    public EmbeddingResponse() {}
```
tatu-at-datastax, line 48: If we don't need to call it ourself, ok to leave as package-protected (no modifier) or `protected`.
Jackson deserializer can force access as necessary.

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/configuration/PropertyBasedEmbeddingServiceConfig.java
```diff
@@ -56,6 +59,17 @@ public interface VertexAiConfig {
     String apiKey();
   }
 
+  public interface CohereConfig {
+    @WithDefault("false")
+    boolean enabled();
+
+    @WithDefault(" https://api.cohere.ai/v1/")
```
tatu-at-datastax, line 66: Since we have `application.yaml`, are these defaults actually needed?

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/configuration/PropertyBasedEmbeddingServiceConfig.java
```diff
@@ -56,6 +59,17 @@ public interface VertexAiConfig {
     String apiKey();
   }
 
+  public interface CohereConfig {
```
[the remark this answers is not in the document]
maheshrajamani: Thought of making it as one, but wasn't sure because some embedding services will have more that the 3 provided details. Eg: vertexAi may get location and project id, Amazon Titan may get additional embedding size property(supports dynamically passing it). Should we handle this when handling createCollection changes which to be discussed today? apiKey having "" will remove it.

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/configuration/PropertyBasedEmbeddingServiceConfigStore.java
maheshrajamani: Good change will fix it

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/CohereEmbeddingClient.java
maheshrajamani: Will change this

maheshrajamani: Is there a way to say ignore all fields other than the ones defined?

vkarpov15 (commented):
LGTM

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/configuration/PropertyBasedEmbeddingServiceConfig.java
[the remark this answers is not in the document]
tatu-at-datastax: Oh sorry, what I meant was a common base interface, not common to everything.
I 100% agree some will need more than others.

But there's no real hurry, we can always refactor once we have better idea of what all is needed.

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/CohereEmbeddingClient.java
tatu-at-datastax: There is:

    @JsonIgnoreProperties(ignoreUnknown = true)

which may make sense (but can sometimes hide problems too).

PR #890, opened 2024-02-23
remove some bridge related dependencies

Yuqi-Du:
- update old helm charts
- delete grpc retry related codes
- delete DataStoreConfig
- docker compose file
- remove quarkus.grpc.clients.bridge.host when start integration test container

stargate.data-store.ignore-bridge can be removed, but will get some complains during service startup, since we still have quarkus-common dependency, 

**Checklist**
- [x] Changes manually tested


helm/jsonapi/templates/stargate_jsonapi_deployment.yaml
```diff
@@ -1,62 +1,195 @@
...
+            - name: QUARKUS_SHUTDOWN_TIMEOUT
+              value: "PT30S"
+            - name: QUARKUS_SMALLRYE_OPENAPI_SERVERS
+              value: /api/json
+            - name: QUARKUS_HTTP_ACCESS_LOG_ENABLED
+              value: "true"
+            - name: QUARKUS_HTTP_ACCESS_LOG_PATTERN
+              value: "{{ .Values.accessLogPattern }}"
+            - name: QUARKUS_LOG_MIN_LEVEL
+              value: "TRACE"
```
tatu-at-datastax, line 60: Do we really want to set this to "TRACE"? That level is potentially very verbose isn't it? Default is, I think, "DEBUG".

kathirsvn: Setting MIN_LEVEL to TRACE really helps when we want to find something with trace enabled. Without this, when required, even though we set the level to TRACE for a package or class, it will only be go up to DEBUG level.

tatu-at-datastax: Ok so there's some pre-filtering done or something? I wasn't 100% sure what 2 separate levels meant.

kathirsvn: Yes, there are two separate levels. But I just noticed that `QUARKUS_LOG_MIN_LEVEL` is a build time property and so we will have to put in the application.yaml and passing it here as an environment generates a warning message on the start up.

https://quarkus.io/guides/all-config#quarkus-core_quarkus-log-min-level

tatu-at-datastax: I think that might explain logic:

1. Min level determines log statements that are actually compiled in (so can statically remove unused trace level statements, if so desired)
2. Actual level used will dynamically determine if log statement is executed at run time

so just include TRACE in minimum does not enable trace-level output but allows if configure.

So basically what you said @kathirsvn  :)

Yuqi-Du: Thanks, I have set log_min_level in configuration file, please check

helm/jsonapi/templates/stargate_jsonapi_deployment.yaml
```diff
@@ -1,62 +1,195 @@
...
+            - name: QUARKUS_LOG_CONSOLE_JSON_EXCEPTION_OUTPUT_TYPE
+              value: formatted
+            {{- range $key, $value := .Values.logParams }}
+            {{- if hasPrefix "QUARKUS_LOG_" $key }}
+            - name: {{ $key }}
+              value: {{ $value }}
+            {{- end }}
+            {{- end }}
+            - name: STARGATE_DATA_STORE_IGNORE_BRIDGE
```
jeffreyscarpenter, line 69: are we not able to get rid of this flag yet?

helm/jsonapi/templates/stargate_jsonapi_deployment.yaml
```diff
@@ -1,62 +1,195 @@
...
+              value: "true"
+            - name: STARGATE_DATA_STORE_SAI_ENABLED
+              value: "true"
+            - name: STARGATE_METRICS_TENANT_REQUEST_COUNTER_USER_AGENT_TAG_ENABLED
+              value: "true"
+            - name: STARGATE_MULTI_TENANCY_ENABLED
```
jeffreyscarpenter, line 75: not sure enabling multi-tenancy makes sense for this helm chart.

helm/jsonapi/templates/stargate_jsonapi_deployment.yaml
```diff
@@ -1,62 +1,195 @@
...
+            - name: STARGATE_MULTI_TENANCY_TENANT_RESOLVER_SUBDOMAIN_REGEX
+              value: "[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}"
+            - name: JAVA_MAX_MEM_RATIO
+              value: "75"
+            - name: JAVA_INITIAL_MEM_RATIO
+              value: "50"
+            - name: GC_CONTAINER_OPTIONS
+              value: "-XX:+UseG1GC"
+            - name : STARGATE_JSONAPI_OPERATIONS_DATABASE_CONFIG_TYPE
+              value: astra
```
jeffreyscarpenter, line 90: really?

jeffreyscarpenter: did we test this helm chart?

jeffreyscarpenter:
@Yuqi-Du the description says that the docker compose file is updated, but I don't see that in the changes?

helm/jsonapi/templates/stargate_jsonapi_deployment.yaml
Yuqi-Du: No, I referenced from what we have in serverless-platform-charts repo.
Let me refine it.

jeffreyscarpenter (approved):

PR #894, opened 2024-02-26
Fixes #893: add IT for InsertMany failure due to doc count exceeding limit

### LINKED ISSUE #893: Add IT for `InsertMany` error case (of too many documents) to verify error message
Looks like we only have a unit test for trying to verify handling of case of too many documents being inserted, but no integration test. Latter is important in verifying that not only is the limit validated but also that the error message contains relevant information.

tatu-at-datastax:
**What this PR does**:

Adds an integration test to verify that error message for attempts to insert more documents than allowed results in proper exception message

Fixes #893

**Checklist**
- [ ] Changes manually tested


vkarpov15 (commented):
LGTM :+1:

PR #896, opened 2024-02-26
Add findVectorProviders command

Hazel-Datastax:
**What this PR does**:
Add findVectorProviders command

Fixes [Jira C2-3318](https://datastax.jira.com/browse/C2-3318)

**Checklist**
- [x] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/service/embedding/configuration/PropertyBasedEmbeddingServiceConfig.java
```diff
@@ -37,6 +44,60 @@ public interface ServiceConfig {
...
+    interface ParameterConfig {
+      @JsonProperty
+      String type();
+
+      @JsonProperty
+      boolean required();
+
+      @JsonProperty
+      @WithName("default")
+      Optional<String> default_();
```
maheshrajamani, line 75: why do we need to have the variable with `_` at the end?

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/configuration/PropertyBasedEmbeddingServiceConfig.java
```diff
@@ -37,6 +44,60 @@ public interface ServiceConfig {
...
+      Optional<Integer> maxRetries();
+
+      @JsonProperty
+      Optional<Integer> retryDelayMs();
+
+      @JsonProperty
+      Optional<Integer> timeoutMs();
+
+      @JsonProperty
+      Optional<Integer> maxTokens();
```
maheshrajamani, line 89: I don't think we have a valid logic to check for token size. Would keep it to maxFieldSize().

src/main/resources/application.yaml
```diff
@@ -58,6 +58,31 @@ stargate:
           url: https://us-central1-aiplatform.googleapis.com
           api-key: YOUR_API_KEY
 
+        google:
```
maheshrajamani, line 61: Why is it a different section, should be under `vertexai`

src/main/resources/application.yaml
```diff
@@ -58,6 +58,31 @@ stargate:
...
+          enabled: true
+          url: "https://us-central1-aiplatform.googleapis.com/v1/projects/{PROJECT_ID}/locations/us-central1/publishers/google/models/{MODEL}:predict"
+          supported-authentication:
+            - "HEADER"
+            - "SHARED_SECRETS"
+          parameters:
+            PROJECT_ID:
```
maheshrajamani, line 68: We may need to create provider conjugates, can't we have PROJECT as a field value like name: "PROJECT_ID"

src/main/resources/application.yaml
```diff
@@ -58,6 +58,31 @@ stargate:
...
+              type: "string"
+              required: true
+          properties:
+            max-retries: 3
+            retry-delay-ms: 100
+            timeout-ms: 1000
+          models:
+            textembedding-gecko@003:
```
maheshrajamani, line 76: Again we should see if this can be made as name : "textembedding-gecko@003"

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/FindVectorProvidersOperation.java
```diff
@@ -0,0 +1,46 @@
...
+  @Override
+  public Uni<Supplier<CommandResult>> execute(QueryExecutor queryExecutor) {
+    return Uni.createFrom()
+        .item(
+            () -> {
+              // TODO: get all available vector providers
+              Map<String, PropertyBasedEmbeddingServiceConfig.VectorProviderConfig>
+                  vectorProviders = new HashMap<>();
+              if (propertyBasedEmbeddingServiceConfig.google().enabled()) {
+                vectorProviders.put("google", propertyBasedEmbeddingServiceConfig.google());
```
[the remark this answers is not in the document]
tatu-at-datastax: Agreed. And probably not "google" only (could be "google/vertexai" or something, but not just company name IMO).

src/main/resources/application.yaml
Hazel-Datastax: I think there will be a problem if we have multiple parameters:
```
  parameters:
    name: PROJECT_ID
    type: STRING
    required: true
    name: Another_parameter
    type: STRING
    required: true
```

maheshrajamani: You can have it as array.
parameters:
   - name: PROJECT_ID
      type: STRING
      required: true
    - name: Another_parameter
      type: STRING
      required: true

src/main/resources/application.yaml
Hazel-Datastax: Because I am not sure about the final structure and just added another one temporarily. Which URL should I use? If I change to `https://us-central1-aiplatform.googleapis.com/v1/projects/{PROJECT_ID}/locations/us-central1/publishers/google/models/{MODEL}:predict`, will it affect the service now? Should I remove the `api-key`?

Hazel-Datastax: Also, should I check every other model’s docs to add more config details now so I can remove `ServiceConfig`?

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/FindVectorProvidersOperation.java
```diff
@@ -0,0 +1,74 @@
...
+          vectorProviderConfig.supportedAuthentication(),
+          vectorProviderConfig.parameters(),
+          modelsRemoveProperties);
+    }
+  }
+
+  private record ReturnModelConfig(
+      String name,
+      List<PropertyBasedEmbeddingServiceConfig.VectorProviderConfig.ParameterConfig> parameters) {}
+}
```
Hazel-Datastax, line 74: As discussed in the meeting, we want to remove the properties config. So I didn't directly use the `VectorProviderConfig` to construct the return body. 
 The current return body looks like:
 ```
{
    "status": {
        "vectorProviders": {
            "vertexai": {
                "url": "https://us-central1-aiplatform.googleapis.com/v1/projects/{PROJECT_ID}/locations/us-central1/publishers/google/models/{MODEL}:predict",
                "supportedAuthentication": [
                    "HEADER",
                    "SHARED_SECRETS"
                ],
                "parameters": [
                    {
                        "name": "PROJECT_ID",
                        "type": "STRING",
                        "defaultValue": null,
                        "help": null,
                        "required": true
                    }
                ],
                "models": [
                    {
                        "name": "textembedding-gecko@003",
                        "parameters": [
                            {
                                "name": "autoTruncate",
                                "type": "BOOLEAN",
                                "defaultValue": "true",
                                "help": null,
                                "required": false
                            }
                        ]
                    }
                ]
            }
        }
    }
}
```
If we want to include all the configs, the return body will look like:
```
{
    "status": {
        "vectorProviders": {
            "vertexai": {
                "url": "https://us-central1-aiplatform.googleapis.com/v1/projects/{PROJECT_ID}/locations/us-central1/publishers/google/models/{MODEL}:predict",
                "enabled": true,
                "properties": {
                    "max-retries": "3",
                    "retry-delay-ms": "100",
                    "timeout-ms": "1000"
                },
                "parameters": [
                    {
                        "name": "PROJECT_ID",
                        "type": "STRING",
                        "defaultValue": null,
                        "help": null,
                        "required": true
                    }
                ],
                "supportedAuthentication": [
                    "HEADER",
                    "SHARED_SECRETS"
                ],
                "models": [
                    {
                        "name": "textembedding-gecko@003",
                        "properties": {
                            "max-tokens": "3072"
                        },
                        "parameters": [
                            {
                                "name": "autoTruncate",
                                "type": "BOOLEAN",
                                "defaultValue": "true",
                                "help": null,
                                "required": false
                            }
                        ]
                    }
                ]
            }
        }
    }
}

```

**Comment by Hazel-Datastax** [comment]
To separate the vector providers configs from `application.yaml`, there are two ways:
1. Create `application.properties` file and Quarkus will handle everything automatically. 
2. Create a customized yaml file `vectorProvider.yaml` file. Implement a custom`ConfigSourceProvider` in `VectorProvidersConfig` to read the yaml file. Register the custom `VectorProvidersConfig` in `META-INF/services`

Not sure which one is better. I had both ways work in my PR. Once decided, I will remove the other one.

File: src/main/java/io/stargate/sgv2/jsonapi/StargateJsonApi.java [hunk 8]
```diff
@@ -673,6 +673,83 @@
                         }
                       }
                       """),
+              @ExampleObject(
+                  name = "findVectorProviders",
+                  summary = "`findVectorProviders` command result",
+                  value =
+                      """
+                        {
+                            "status": {
```
tatu-at-datastax, line 682: "status"? That seems like an odd choice for JSON key here. Is that actually used or left from copying another example?

Hazel-Datastax: The example is the actual return. I use this by following the [spec](https://docs.google.com/document/d/1XkCLUI1YgMuqqUU0KYMfyhgBgRmdC4aSVQXZrodV5T0/edit#heading=h.qitn2whv4gd3) example. Do we need to change it to a better name?

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/configuration/VectorProvidersConfig.java
```diff
@@ -0,0 +1,26 @@
...
+public class VectorProvidersConfig
+    implements org.eclipse.microprofile.config.spi.ConfigSourceProvider {
+  @Override
+  public Iterable<ConfigSource> getConfigSources(ClassLoader forClassLoader) {
+    try {
+      URL resourceURL = forClassLoader.getResource("vectorProviders.yaml");
+      if (resourceURL != null) {
+        YamlConfigSource configSource = new YamlConfigSource(resourceURL);
+        return Collections.singletonList(configSource);
+      }
```
tatu-at-datastax, line 20: Should also throw an exception if `resourceURL` is `null`?

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/FindVectorProvidersOperation.java
tatu-at-datastax: I think that `"status"` is just wrong: and I am not sure we even need `"vectorProviders"` level there. So why not return a JSON Object with name as key (so remove 2 layers)?

Or, if we have to conform to shared structure, maybe main level key should be `"data"`?

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/configuration/VectorProvidersConfig.java
Hazel-Datastax: Yes, throw error when resourceURL is null

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/FindVectorProvidersOperation.java
Hazel-Datastax: Solved, `status` is from the convention of `findCollections` and `findNamespaces`, we will follow this

Hazel-Datastax:
For the config separation, we will choose the custom config file: 
1. It keeps other general config settings more compact (we have tons of providers to configure)
2. We may need something similar for other configuration in the future

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/configuration/PropertyBasedEmbeddingServiceConfig.java
```diff
@@ -1,40 +1,94 @@
...
+      String name();
+
+      @JsonProperty
+      ParameterType type();
+
+      @JsonProperty
+      boolean required();
+
+      @JsonProperty
+      Optional<String> defaultValue();
```
maheshrajamani, line 81: Can you mark the fields that can be Nullable

src/main/java/io/stargate/sgv2/jsonapi/StargateJsonApi.java
```diff
@@ -673,6 +673,83 @@
...
+                                "vectorProviders": {
+                                    "cohere": {
```
tatu-at-datastax, line 684: I think it'd make sense to use bit shorter example, remove all but 2 providers?

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/configuration/PropertyBasedEmbeddingServiceConfig.java
```diff
@@ -1,40 +1,94 @@
...
+import java.util.Map;
 import java.util.Optional;
 import org.eclipse.microprofile.config.spi.Converter;
 
-@ConfigMapping(prefix = "stargate.jsonapi.embedding.service")
+@ConfigMapping(prefix = "jsonapi.vector.provider")
 public interface PropertyBasedEmbeddingServiceConfig {
 
   @Nullable
-  ServiceConfig openai();
```
tatu-at-datastax, line 15: It may be late to ask this, but why wouldn't it just be something like:

    Map<String, ServiceConfig> providers

instead of hard-coding provider names as fields?

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/configuration/VectorProvidersConfig.java
```diff
@@ -0,0 +1,27 @@
...
+import io.quarkus.runtime.annotations.StaticInitSafe;
+import io.smallrye.config.source.yaml.YamlConfigSource;
+import java.io.IOException;
+import java.net.URL;
+import java.util.Collections;
+import org.eclipse.microprofile.config.spi.ConfigSource;
+
+@StaticInitSafe
+public class VectorProvidersConfig
+    implements org.eclipse.microprofile.config.spi.ConfigSourceProvider {
```
maheshrajamani, line 12: Can you move it the class path to import?

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/configuration/PropertyBasedEmbeddingServiceConfig.java
```diff
@@ -1,40 +1,94 @@
...
+
+    @JsonProperty
+    List<String> supportedAuthentication();
+
+    @Nullable
+    @JsonProperty
+    List<ParameterConfig> parameters();
+
+    @Nullable
+    @JsonProperty
```
tatu-at-datastax, line 52: What is the difference between `parameters` and `properties`? I think Javadocs would make sense here to explain.

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/configuration/VectorProvidersConfig.java
```diff
@@ -0,0 +1,27 @@
...
+  @Override
+  public Iterable<ConfigSource> getConfigSources(ClassLoader forClassLoader) {
+    try {
+      URL resourceURL = forClassLoader.getResource("vector-providers-config.yaml");
+      if (resourceURL != null) {
+        YamlConfigSource configSource = new YamlConfigSource(resourceURL);
+        return Collections.singletonList(configSource);
+      } else {
+        throw new RuntimeException("Failed to load vector provider config, resource is null");
```
maheshrajamani, line 21: Is this loaded on start up? What is handling this exception? Believe it will never happen since we have the file part of source code.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/FindVectorProvidersOperation.java
```diff
@@ -0,0 +1,97 @@
...
+
+    @Override
+    public CommandResult get() {
+      Map<CommandStatus, Object> statuses =
+          Map.of(CommandStatus.EXISTING_VECTOR_PROVIDERS, vectorProviders);
+      return new CommandResult(statuses);
+    }
+  }
+
+  private record ReturnVectorProviderConfig(
```
tatu-at-datastax, line 73: Why couldn't we use the same objects that we read as provider configuration?

maheshrajamani (approved):
LGTM

vkarpov15 (approved):
LGTM

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/configuration/PropertyBasedEmbeddingServiceConfig.java
Hazel-Datastax: I am not sure I understand the question correctly. I think that's the way for mapping configuration to objects. In the configuration yaml, I have something like:
```
jsonapi:
  vector:
    provider:
       openai:
       huggingface:
```
In the interface, I use `@ConfigMapping(prefix = "jsonapi.vector.provider")` to locate the configuration, and the field name `VectorProviderConfig openai()` will be mapped to `openai` in the configuration file.

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/configuration/PropertyBasedEmbeddingServiceConfig.java
Hazel-Datastax: I think `parameters` are used by the user to construct the URL or to customize the model. (external) `properties` are used by us to contact the real model. (internal) The example below gives the full configuration, you could see the difference that `parameters` are for the user to customize their model and `properties` are for us to actually use the model.

```
{
    "status": {
        "vectorProviders": {
            "vertexai": {
                "url": "https://us-central1-aiplatform.googleapis.com/v1/projects/{PROJECT_ID}/locations/us-central1/publishers/google/models/{MODEL}:predict",
                "enabled": true,
                "properties": {
                    "max-retries": "3",
                    "retry-delay-ms": "100",
                    "timeout-ms": "1000"
                },
                "parameters": [
                    {
                        "name": "PROJECT_ID",
                        "type": "STRING",
                        "defaultValue": null,
                        "help": null,
                        "required": true
                    }
                ],
                "supportedAuthentication": [
                    "HEADER",
                    "SHARED_SECRETS"
                ],
                "models": [
                    {
                        "name": "textembedding-gecko@003",
                        "properties": {
                            "max-tokens": "3072"
                        },
                        "parameters": [
                            {
                                "name": "autoTruncate",
                                "type": "BOOLEAN",
                                "defaultValue": "true",
                                "help": null,
                                "required": false
                            }
                        ]
                    }
                ]
            }
        }
    }
}
```

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/FindVectorProvidersOperation.java
Hazel-Datastax: As I explained in the previous comment, `parameters` are for external use and `properties` are for internal use. We don't want to expose `properties` so I made a new object here.

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/configuration/PropertyBasedEmbeddingServiceConfig.java
Hazel-Datastax: Sure, but `Optional` has already made this field Nullable

PR #899, opened 2024-02-27
Fixes #895: truncate reported value that fails constraint violation if too long

### LINKED ISSUE #895: Truncate value length included by `ConstraintViolationExceptionMapper` into configurable max length
Currently `ConstraintViolationExceptionMapper` which handles Quarkus/Bean Validation API annotated constraints will include full value of property in exception message. This is often fine for scalar values like numbers of enumerated text, but can be problematic when referencing full JSON payloads -- for example, Array of Documents to insert.
This can be lead to huge error descriptions, problematic for a few reasons:

1. Loss of information when client truncates error messages (see #893) -- often most useful details are at the end of message, after embedded value -- so client libraries may lose valuable information
2. Logging systems may end up logging megabytes of content per single failed call, adding to noise
3. (more importantly) Inclusion of huge error payload could be a (minor, maybe) DoS vector

To resolve these issues, let's add configurable limit to maximum value length to include without truncation; starting with maximum length of 4096 units (characters or bytes depending on kind of value), and truncate values longer than that for error reporting purposes.

tatu-at-datastax:
**What this PR does**:

Truncates reported failed value in ConstraintsViolation exception, if value length exceeds limit (1000 characters).

Fixes #895

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/exception/mappers/ConstraintViolationExceptionMapper.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/exception/mappers/ConstraintViolationExceptionMapper.java
@@ -1,5 +1,6 @@
 package io.stargate.sgv2.jsonapi.exception.mappers;
 
+import com.fasterxml.jackson.databind.JsonNode;
 import io.stargate.sgv2.jsonapi.api.model.command.CommandResult;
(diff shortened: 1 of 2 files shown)


maheshrajamani (approved):

vkarpov15 (commented):
LGTM

PR #903, opened 2024-02-28
Error if `$vectorize` is used on table for which the feature is not enabled.

### LINKED ISSUE #901: Was able to use $vectorize used with non vectorize collection
Related to [https://datastax.jira.com/browse/C2-3338](https://datastax.jira.com/browse/C2-3338|smart-link) 

I had not actually changed the createCollection to use enable vectorize() BUT I was able to:

* use $vectorize in an insert , and now I think about it when I looked at the docs in the portal they did not have $vector
* use $vectorize in a find and got the error in the linked issue 

Can we check that if $vectorize is used the collection must support it ? And return appropriate error if it is not supported

maheshrajamani:
**What this PR does**:
Error if `$vectorize` is used on table for which the feature is not enabled.

Fixes #901 

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/src/test/java/io/stargate/sgv2/jsonapi/service/resolver/model/impl/CommandResolverWithVectorizerTest.java
+++ b/src/test/java/io/stargate/sgv2/jsonapi/service/resolver/model/impl/CommandResolverWithVectorizerTest.java
@@ -165,9 +165,10 @@ public class CommandResolverWithVectorizerTest {
                 JsonApiException exception = (JsonApiException) e;
                 assertThat(exception.getMessage())
                     .isEqualTo(
-                        "Unable to vectorize data, embedding service not configured for the collection");
+                        "Unable to vectorize data, embedding service not configured for the collection : "
+                            + VECTOR_COMMAND_CONTEXT.collection());
                 assertThat(exception.getErrorCode())
-                    .isEqualTo(ErrorCode.UNAVAILABLE_EMBEDDING_SERVICE);
+                    .isEqualTo(ErrorCode.EMBEDDING_SERVICE_NOT_CONFIGURED);
               });
     }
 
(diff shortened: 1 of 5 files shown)


Hazel-Datastax (approved):
LGTM

src/test/java/io/stargate/sgv2/jsonapi/service/resolver/model/impl/CommandResolverWithVectorizerTest.java
```diff
@@ -118,6 +138,39 @@ public void find() throws Exception {
...
+              .getFailure();
+
+      assertThat(throwable)
+          .isInstanceOf(JsonApiException.class)
+          .satisfies(
+              e -> {
+                JsonApiException exception = (JsonApiException) e;
+                assertThat(exception.getMessage())
+                    .isEqualTo(
+                        "Unable to vectorize data, embedding service not configured for the collection");
```
tatu-at-datastax, line 168: Should include Collection name in message?

src/test/java/io/stargate/sgv2/jsonapi/service/resolver/model/impl/CommandResolverWithVectorizerTest.java
```diff
@@ -118,6 +138,39 @@ public void find() throws Exception {
...
+                assertThat(exception.getErrorCode())
+                    .isEqualTo(ErrorCode.UNAVAILABLE_EMBEDDING_SERVICE);
```
tatu-at-datastax, line 170: I guess names can be whatever, but "UNAVAILABLE" sounds like availability issue -- so should it rather be something like "EMBEDDING_SERVICE_NOT_CONFIGURED" or "NO_EMBEDDING_SERVICE_FOR_COLLECTION"?

tatu-at-datastax (approved):
LGTM, added some suggestions.

vkarpov15 (commented):
LGTM

PR #908, opened 2024-02-29
bug fix: valid vector metric

### LINKED ISSUE #907: Collection gets created when specified a wrong vector metric
If user specifies a wrong vector function when creating collection, let’s say `Cosinnee` .
The collection will still be created, but with the vector index missing, since there is no function in C* called `Cosinnee` .
But since we do explain collections (findCollections), only from seeing if a table has the vector column, which it does in this case, so user will see this collection is there and with vector enabled by using findCollections command.
Only when they do a ANN search, we returned a `NO_INDEX_ERROR` error.

Yuqi-Du:
add @valid annotation to enable cascade validation for createCollectionCommand

Fixes #907

**Checklist**
- [x] Changes manually tested


tatu-at-datastax (approved):

Yuqi-Du:
Thanks!

vkarpov15 (commented):
LGTM

PR #913, opened 2024-03-04
Fixes #898: improve error message for "multiple _id filters" case

### LINKED ISSUE #898: Confusing error message when trying to find by multiple `_id` values
I'm trying to execute something like:
```
const docs = await collection.find({
    $or: [{_id: 1}, {_id: 2}, {_id: 3}]
  }).toArray()
```

And here is the error message I'm getting:

```
errors: [
    {
      message: 'Should only have one _id filter, document id cannot be restricted by more than one relation if it includes an Equal',
      errorCode: 'FILTER_MULTIPLE_ID_FILTER'
    }
  ]
```

First, I'm not sure why this operation isn't allowed if I can `$or` on other properties of the document.

Second, I would update this error message to point developers to the preferred method of using `$in`.

tatu-at-datastax:
**What this PR does**:

Improves error message for case where multiple equals filter clauses used for doc id field

Fixes #898

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCode.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCode.java
@@ -29,7 +29,7 @@ public enum ErrorCode {
   FILTER_UNRESOLVABLE("Unable to resolve the filter"),
 
   FILTER_MULTIPLE_ID_FILTER(
-      "Should only have one _id filter, document id cannot be restricted by more than one relation if it includes an Equal"),
+      "Cannot have more than one _id equals filter clause: use $in operator instead"),
 
   FILTER_FIELDS_LIMIT_VIOLATION("Filter fields size limitation violated"),
 
--- a/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/deserializers/FilterClauseDeserializer.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/deserializers/FilterClauseDeserializer.java
@@ -135,8 +135,7 @@ public class FilterClauseDeserializer extends StdDeserializer<FilterClause> {
 
   private void validate(LogicalExpression logicalExpression) {
     if (logicalExpression.getTotalIdComparisonExpressionCount() > 1) {
-      throw new JsonApiException(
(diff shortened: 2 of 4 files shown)


Hazel-Datastax (approved):

src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCode.java
```diff
@@ -29,7 +29,7 @@ public enum ErrorCode {
   FILTER_UNRESOLVABLE("Unable to resolve the filter"),
 
   FILTER_MULTIPLE_ID_FILTER(
-      "Should only have one _id filter, document id cannot be restricted by more than one relation if it includes an Equal"),
+      "Cannot have more than one _id equals filter clause: use $in operator instead"),
```
vkarpov15, line 32: Would you get the same error message if you did something like `{ _id: { $in: ['foo', 'bar'] }, $or: [{ _id: { $in: ['baz', 'qux'] } }] }`?

tatu-at-datastax: Hmmh. Good question; I do not know for sure... but I don't think you would.

@Yuqi-Du WDYT?

vkarpov15: I checked and yep, same error message if you do the following `Person.findOne({ _id: { $in: [john._id] }, $or: [{ _id: { $in: [bill._id] } }] })`

Maybe better to replace `use $in operator instead` with `use a single top-level $in operator instead`? Admittedly this is a bit of an edge case, but the error message looks a bit silly if you're already using `$in`.

tatu-at-datastax: @vkarpov15 agreed, this is still not great. Would you mind filing another issue, referencing this PR, for further work? (something stating that only one `$eq` or `$in` clause can reference `_id` field)

I try not to re-open issues when there may be a release containing (partial) fix, release note generation tends to get things wrong in such cases.

vkarpov15: @tatu-at-datastax done #950

tatu-at-datastax: Thank you @vkarpov15 !

PR #916, opened 2024-03-04
Vectorize by input type

maheshrajamani:
**What this PR does**:
Some embedding provider accepts input type for vectorizing. Added it for Cohere and NVidia.

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/EmbeddingService.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/EmbeddingService.java
@@ -9,5 +9,6 @@ import java.util.Optional;
  * of chosen model.
  */
 public interface EmbeddingService {
-  Uni<List<float[]>> vectorize(List<String> texts, Optional<String> apiKeyOverride);
+  Uni<List<float[]>> vectorize(
+      List<String> texts, Optional<String> apiKeyOverride, boolean isWrite);
 }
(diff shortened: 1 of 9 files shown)


vkarpov15 (approved):
LGTM

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/EmbeddingService.java
```diff
@@ -9,5 +9,6 @@
  * of chosen model.
  */
 public interface EmbeddingService {
-  Uni<List<float[]>> vectorize(List<String> texts, Optional<String> apiKeyOverride);
+  Uni<List<float[]>> vectorize(
```
tatu-at-datastax, line 12: Needs Javadoc to explain "isWrite" at least?

tatu-at-datastax: Actually, I think name "isWrite" is probably not very descriptive, not sure what'd be better: I think it means "used for vectorizing document to insert"? (vs "read" meaning "used for queries"?).

Is there likelihood we might need more than 2 types in future?

tatu-at-datastax (approved):
Looks good - I have concerns about naming & type of parameter, but I think you can figure out what the best approach is.

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/EmbeddingService.java
maheshrajamani: Make sense, changed it to a enum with values `INDEX` vs `SEARCH`

tatu-at-datastax (approved):
LGTM

PR #929, opened 2024-03-06
collection limit rollback change

### LINKED ISSUE #912: Rollback on CreateCollection failure
There is logic in the CreateCollection operation to check the number of indexes and collections against configured limits before creating a new collection. However, it is still possible to experience a "too many indexes" failure when creating indexes on a "collection" Cassandra table. This situation could occur if additional indexes were created on the database between the time the API makes the check and when it tries to create an index, and the database enforces the index limit.

To handle this case more cleanly, the CreateCollection operation should attempt to delete the Cassandra table (which will also remove any indexes on the table) and return the "too many indexes" error. If the deletion fails, this can be noted in the response body so that the user is made aware of how to clean up manually.

Yuqi-Du:
This is a rollback change for Astra collection limit relaxation, and part of the short term solution we have for db collection limit.
In general, before violates Data API collection guardrail, if tenant hit the astra side guardrail and has indexes creation failure, Data api will also drop the table(plus indexes)

if index creation violates DB index limit and collection not existed before, and rollback is enabled, then drop the collection

Fixes #912
**Checklist**
- [x] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/config/DatabaseLimitsConfig.java
```diff
@@ -30,7 +30,7 @@ public interface DatabaseLimitsConfig {
    * Default setting for {@link #maxCollections()}}: maximum number of Collections allowed to be
    * created per Database, regardless of whether enough Indexes are available.
    */
-  int DEFAULT_MAX_COLLECTIONS = 5;
+  int DEFAULT_MAX_COLLECTIONS = 10;
```
jeffreyscarpenter, line 33: Sorry for not being clear, my intent was not that we update the defaults here but that we override configured values for specific environments.

src/main/java/io/stargate/sgv2/jsonapi/config/DatabaseLimitsConfig.java
```diff
@@ -42,7 +42,7 @@ public interface DatabaseLimitsConfig {
    * Default setting for {@link #indexesAvailablePerDatabase()}: number of SAIs per Database that
    * can be created (and needed for Collection creation)
    */
-  int DEFAULT_INDEXES_AVAILABLE_PER_DATABASE = 50;
+  int DEFAULT_INDEXES_AVAILABLE_PER_DATABASE = 100;
```
jeffreyscarpenter, line 45: Same as previous comment.

jeffreyscarpenter (changes requested):

src/main/java/io/stargate/sgv2/jsonapi/config/DatabaseLimitsConfig.java
Yuqi-Du: Thanks, reverted.

src/main/java/io/stargate/sgv2/jsonapi/config/DatabaseLimitsConfig.java
Yuqi-Du: Thanks, reverted.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/CreateCollectionOperation.java
```diff
@@ -182,7 +186,33 @@ private Uni<Supplier<CommandResult>> executeCollectionCreation(QueryExecutor que
...
               }
-            });
+            })
+        .onFailure(
+            error ->
+                error instanceof InvalidQueryException
+                    && error
+                        .getMessage()
+                        .matches(
+                            ".*Cannot have more than \\d+ indexes, failed to create index on table.*"))
```
tatu-at-datastax, line 196: This may be necessary, but I fear that error message may vary across different versions of backend; we saw this a lot with Stargate REST where C-3.11/4.0/DSE had sometimes small, sometimes big differences in fail messages.
If there was error code of some kind it'd better to check that.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/CreateCollectionOperation.java
```diff
@@ -182,7 +186,33 @@ private Uni<Supplier<CommandResult>> executeCollectionCreation(QueryExecutor que
...
+        .recoverWithUni(
+            error ->
+                deleteCollectionOperation
```
jeffreyscarpenter, line 199: Are there possible side effects to reusing the delete collection operation here?

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/CreateCollectionOperation.java
```diff
@@ -31,6 +32,7 @@
 
 public record CreateCollectionOperation(
     CommandContext commandContext,
+    DeleteCollectionOperation deleteCollectionOperation,
```
tatu-at-datastax, line 35: Instead of passing via constructor, this could/should be constructed as needed, as `commandContext` and collection name are known within code.

tatu-at-datastax (commented):
Looks reasonable, suggested some changes.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/CreateCollectionOperation.java
tatu-at-datastax: It seems to me there shouldn't; operations being relatively simple containers of functionality?
(but asking @maheshrajamani to confirm).

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/CreateCollectionOperation.java
```diff
@@ -182,7 +186,33 @@ private Uni<Supplier<CommandResult>> executeCollectionCreation(QueryExecutor que
...
+                            ErrorCode.TOO_MANY_INDEXES.toApiException(
+                                "cannot create a new collection; need %d indexes to create the collection;",
+                                dbLimitsConfig.indexesNeededPerCollection()))
+                    .onFailure()
+                    .recoverWithItem(
+                        e -> {
+                          // This is unlikely to happen for delete collection, return with
+                          // TOO_MANY_INDEXES exception
+                          return ErrorCode.TOO_MANY_INDEXES.toApiException(
+                              "collection created with indexes creation failure; need %d indexes to create the collection;",
```
tatu-at-datastax, line 213: Should include collection name somewhere. And maybe reword as "collection creation failed due to index creation failing; need ...  to create a collection"

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/CreateCollectionOperation.java
maheshrajamani: Looking at the logic very concerned with dropping table for invalid query. Should we do read on table with limit 1 and delete collection only on a successful empty result set. Dropping collection because of a failed query need more proof checking. @amorton @jeffreyscarpenter Thoughts?

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/CreateCollectionOperation.java
maheshrajamani: Drop collection sometimes produces timeouts.

Yuqi-Du: I mean, we have precedent, for example, deleteOperation will reuse findOperation?

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/CreateCollectionOperation.java
Yuqi-Du: Make Sense, thanks!

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/CreateCollectionOperation.java
Yuqi-Du: Thanks, solved

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/CreateCollectionOperation.java
jeffreyscarpenter: @maheshrajamani the intent is that we only do the rollback in the case of a too many indexes error, which it appears this is attempting to do. To @tatu 's point, do we not have an error code yet at this point in the flow that we can check?

jeffreyscarpenter:
Do we need to add a new test case or two to make sure this is working as expected @Yuqi-Du ?

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/CreateCollectionOperation.java
maheshrajamani: My concern was around what if some user tries to recreate existing collection and index creation statement fails and we end up dropping the table.

Yuqi-Du: table dropping will only occurs when DB guardrail is violated.
`If a user recreates an existing collection`, I assume you are saying this existing collection do not have all the sai indexes? In this case, the collection is invalid in our logic, if ending up dropping the table, it would be reasonable, IMO.

Yuqi-Du: Regarding the error code, unfortunately no.
The only thing driver has is error msg.

```

Exception in thread "main" com.datastax.oss.driver.api.core.servererrors.InvalidQueryException: Cannot have more than 50 indexes, failed to create index on table with CREATE CUSTOM INDEX IF NOT EXISTS text5_index1 ON collection_test1 (text5) USING 'StorageAttachedIndex'
        at com.datastax.oss.driver.api.core.servererrors.InvalidQueryException.copy(InvalidQueryException.java:48)
        at com.datastax.oss.driver.internal.core.util.concurrent.CompletableFutures.getUninterruptibly(CompletableFutures.java:149)
        at com.datastax.oss.driver.internal.core.cql.CqlRequestSyncProcessor.process(CqlRequestSyncProcessor.java:53)
        at com.datastax.oss.driver.internal.core.cql.CqlRequestSyncProcessor.process(CqlRequestSyncProcessor.java:30)
        at com.datastax.oss.driver.internal.core.session.DefaultSession.execute(DefaultSession.java:230)
        at com.datastax.oss.driver.api.core.cql.SyncCqlSession.execute(SyncCqlSession.java:54)
        at com.datastax.oss.driver.api.core.cql.SyncCqlSession.execute(SyncCqlSession.java:78)
        at org.example.VectorTest.main(VectorTest.java:64)
```

Yuqi-Du:
I have added a 'feature flag' to disable the rollback.
Currently, it is if `index creation violates DB index limit and collection not existed before and rollback is enabled`, then drop the collection

vkarpov15 (approved):
LGTM :+1:

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/CreateCollectionOperation.java
```diff
@@ -46,6 +47,9 @@ public record CreateCollectionOperation(
   // shared matcher instance used to tell Collections from Tables
   private static final JsonapiTableMatcher COLLECTION_MATCHER = new JsonapiTableMatcher();
 
+  // boolean flag for collection creation rollback
+  private static final boolean TOO_MANY_INDEXES_ROLLBACK = true;
```
jeffreyscarpenter, line 51: Don't you have to annotate this somehow to make it a configurable property?

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/CreateCollectionOperation.java
```diff
@@ -107,16 +114,14 @@ public Uni<Supplier<CommandResult>> execute(QueryExecutor queryExecutor) {
     }
     TableMetadata table = findTableAndValidateLimits(allKeyspaces, currKeyspace, name);
 
-    // table doesn't exist, continue
+    // if table doesn't exist, continue to create collection
     if (table == null) {
-      return executeCollectionCreation(queryExecutor);
+      return executeCollectionCreation(queryExecutor, true);
```
maheshrajamani, line 119: `collectionExisted` flag should be sent as false if table ==null.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/CreateCollectionOperation.java
```diff
@@ -127,8 +132,8 @@ public Uni<Supplier<CommandResult>> execute(QueryExecutor queryExecutor) {
     // if table exists we have to choices:
     // (1) trying to create with same options -> ok, proceed
     // (2) trying to create with different options -> error out
-    if (collectionSettings.equals(collectionSettings_cur)) {
-      return executeCollectionCreation(queryExecutor);
+    if (existedCollectionSettings.equals(newCollectionSettings)) {
+      return executeCollectionCreation(queryExecutor, false);
```
maheshrajamani, line 136: Here `collectionExisted` should be sent true from here.

maheshrajamani (commented):
The flags sent to the create collection doesn't look correct. 
If possible can you add unit tests?

maheshrajamani (approved):
LGTM

PR #933, opened 2024-03-06
Error out when `skip` is used with empty sort `sort: {}` in the `find` command

### LINKED ISSUE #917: Error out when `skip` is used with empty sort `sort={}` in the `find` command
The error only appears when sort clause is null:
```
{
    "find": {
        "options": {
            "skip": 1
        }
    }
}
```
```
{
    "errors": [
        {
            "message": "Request invalid: field 'command.options.skip' value \"FindCommand[filterClause=null, projectionDefinition=null, sortClause=null, options=Options[limit=null, skip=1, pageState=null, includeSimilarity=false]]\" not valid. Problem: skip options should be used with sort clause.",
            "exceptionClass": "ConstraintViolationException"
        }
    ]
}
```
When the sort clause is empty, there is no error and the `skip` is set to 0 implicitly. 
```
{
    "find": {
        "sort": {},
        "options": {
            "skip": 1
        }
    }
}
```

Hazel-Datastax:
**What this PR does**:
Error out when `skip` is used with empty sort `sort={}` in the `find` command

Fixes #917 

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/validation/FindOptionsValidation.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/validation/FindOptionsValidation.java
@@ -31,6 +31,15 @@ public class FindOptionsValidation implements ConstraintValidator<CheckFindOptio
       return false;
     }
 
+    if (options.skip() != null && value.sortClause().isEmpty()) {
+      context
+          .buildConstraintViolationWithTemplate(
+              "skip options should be used with non empty sort clause")
+          .addPropertyNode("options.skip")
+          .addConstraintViolation();
+      return false;
+    }
+
     if (options.skip() != null
         && value.sortClause() != null
         && value.sortClause().hasVsearchClause()) {
--- a/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/validation/FindOptionsValidation.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/validation/FindOptionsValidation.java
@@ -23,18 +23,9 @@ public class FindOptionsValidation implements ConstraintValidator<CheckFindOptio
     if (options == null) return true;
 
     context.disableDefaultConstraintViolation();
-    if (options.skip() != null && value.sortClause() == null) {
+    if (options.skip() != null && (value.sortClause() == null || value.sortClause().isEmpty())) {
       context
           .buildConstraintViolationWithTemplate("skip options should be used with sort clause")
-          .addPropertyNode("options.skip")
-          .addConstraintViolation();
-      return false;
-    }
-
-    if (options.skip() != null && value.sortClause().isEmpty()) {
-      context
-          .buildConstraintViolationWithTemplate(
-              "skip options should be used with non empty sort clause")
           .addPropertyNode("options.skip")
           .addConstraintViolation();
       return false;
--- a/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/validation/FindOptionsValidation.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/validation/FindOptionsValidation.java
@@ -25,7 +25,7 @@ public class FindOptionsValidation implements ConstraintValidator<CheckFindOptio
     context.disableDefaultConstraintViolation();
     if (options.skip() != null && (value.sortClause() == null || value.sortClause().isEmpty())) {
       context
-          .buildConstraintViolationWithTemplate("skip options should be used with sort clause")
+          .buildConstraintViolationWithTemplate("skip option should be used with sort clause")
           .addPropertyNode("options.skip")
           .addConstraintViolation();
       return false;
@@ -35,8 +35,7 @@ public class FindOptionsValidation implements ConstraintValidator<CheckFindOptio
         && value.sortClause() != null
         && value.sortClause().hasVsearchClause()) {
       context
-          .buildConstraintViolationWithTemplate(
-              "skip options should not be used with vector search")
+          .buildConstraintViolationWithTemplate("skip option should not be used with vector search")
           .addPropertyNode("options.skip")
           .addConstraintViolation();
       return false;
--- a/src/main/java/io/stargate/sgv2/jsonapi/exception/mappers/ConstraintViolationExceptionMapper.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/exception/mappers/ConstraintViolationExceptionMapper.java
@@ -2,6 +2,7 @@ package io.stargate.sgv2.jsonapi.exception.mappers;
 
 import com.fasterxml.jackson.databind.JsonNode;
 import io.stargate.sgv2.jsonapi.api.model.command.CommandResult;
+import io.stargate.sgv2.jsonapi.api.model.command.impl.FindCommand;
 import io.stargate.sgv2.jsonapi.exception.ErrorCode;
 import io.stargate.sgv2.jsonapi.exception.JsonApiException;
 import jakarta.validation.ConstraintViolation;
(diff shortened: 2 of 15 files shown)


src/main/java/io/stargate/sgv2/jsonapi/api/model/command/validation/FindOptionsValidation.java
```diff
@@ -31,6 +31,15 @@ public boolean isValid(FindCommand value, ConstraintValidatorContext context) {
       return false;
     }
 
+    if (options.skip() != null && value.sortClause().isEmpty()) {
+      context
+          .buildConstraintViolationWithTemplate(
+              "skip options should be used with non empty sort clause")
```
tatu-at-datastax, line 37: should -> can ?

src/test/java/io/stargate/sgv2/jsonapi/api/model/command/impl/FindCommandTest.java
```diff
@@ -162,6 +162,30 @@ public void invalidOptionsSkipNoSort() throws Exception {
...
+              }
+              """;
+
+      FindCommand command = objectMapper.readValue(json, FindCommand.class);
+      Set<ConstraintViolation<FindCommand>> result = validator.validate(command);
+
+      assertThat(result)
+          .isNotEmpty()
+          .extracting(ConstraintViolation::getMessage)
+          .contains("skip options should be used with sort clause");
```
vkarpov15, line 186: Minor nitpick that I know wasn't introduced with this PR, but can we change "skip options" -> "skip option"? There's only one skip option.

vkarpov15 (approved):
Minor copy nitpick, LGTM overall :+1:

src/test/java/io/stargate/sgv2/jsonapi/api/model/command/impl/FindCommandTest.java
Hazel-Datastax: Good catch, thanks!

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/validation/FindOptionsValidation.java
Hazel-Datastax: yes, combined with previous check

src/main/java/io/stargate/sgv2/jsonapi/exception/mappers/ConstraintViolationExceptionMapper.java
```diff
@@ -76,12 +77,24 @@ private static String valueDescription(Object rawValue) {
...
+      if (propertyPath.contains("options.pageState")) {
+        return String.format("%s", ((FindCommand) rawValue).options().pageState());
+      }
+    }
+
     String valueDesc = rawValue.toString();
     if (valueDesc.length() <= MAX_VALUE_LENGTH_TO_INCLUDE) {
-      return String.format("\"%s\"", valueDesc);
+      return String.format("%s", valueDesc);
     }
```
Hazel-Datastax, line 95: The return JSON body will contain the escape character `\`
```
{
    "errors": [
        {
            "message": "Request invalid: field 'command.options.skip' value \"FindCommand[filterClause=null, projectionDefinition=null, sortClause=null, options=Options[limit=null, skip=1, pageState=null, includeSimilarity=false]]\" not valid. Problem: skip options should be used with sort clause.",
            "errorCode": "COMMAND_FIELD_INVALID"
        }
    ]
}
```
so I removed it

src/main/java/io/stargate/sgv2/jsonapi/exception/mappers/ConstraintViolationExceptionMapper.java
```diff
@@ -76,12 +77,24 @@ private static String valueDescription(Object rawValue) {
...
+        return String.format("%s", ((FindCommand) rawValue).options().skip());
+      }
+      if (propertyPath.contains("options.limit")) {
+        return String.format("%s", ((FindCommand) rawValue).options().limit());
+      }
+      if (propertyPath.contains("options.pageState")) {
+        return String.format("%s", ((FindCommand) rawValue).options().pageState());
+      }
+    }
+
```
Hazel-Datastax, line 91: There is nowhere for me to know why the validation fails, so I pass the `propertyPath` and check it here. @tatu-at-datastax Any better idea?

src/main/java/io/stargate/sgv2/jsonapi/exception/mappers/ConstraintViolationExceptionMapper.java
Hazel-Datastax: Oh, some tests include this character `\`. @tatu-at-datastax Do we add `\` intentionally?

tatu-at-datastax: Those quotes were there on purpose, to indicate `String` value. JSON will need to have escapes, that is fine.
We could use single quotes instead, but it is nice to know type of value and sometimes to separate it from message itself (without quotes it is sometimes hard to tell what is value, what is message).

Backslash is needed/added by JSON on serialization, but Java source code also requires escaping.
We do not add it otherwise than those.

src/main/java/io/stargate/sgv2/jsonapi/exception/mappers/ConstraintViolationExceptionMapper.java
```diff
@@ -76,12 +77,24 @@ private static String valueDescription(Object rawValue) {
...
     String valueDesc = rawValue.toString();
     if (valueDesc.length() <= MAX_VALUE_LENGTH_TO_INCLUDE) {
-      return String.format("\"%s\"", valueDesc);
+      return String.format("%s", valueDesc);
     }
     return String.format(
-        "\"%s\"...[TRUNCATED from %d to %d characters]",
+        "%s...[TRUNCATED from %d to %d characters]",
```
tatu-at-datastax, line 97: Same here, let's leave either double or single quotes to contain String value.

src/main/java/io/stargate/sgv2/jsonapi/exception/mappers/ConstraintViolationExceptionMapper.java
tatu-at-datastax: You could just include all values in a description, without trying to check which are relevant?

src/main/java/io/stargate/sgv2/jsonapi/exception/mappers/ConstraintViolationExceptionMapper.java
Hazel-Datastax: I understand we need to use quotes to separate the value from messages. But using double quotes will need a backslash to escape and the backslash will somehow show in the return body. I think it would be better to use single quotes to avoid the backslash appearing in the return body?

src/main/java/io/stargate/sgv2/jsonapi/exception/mappers/ConstraintViolationExceptionMapper.java
Hazel-Datastax: So, we could have a message like this to include all the values in `options`? 
```
{
    "errors": [
        {
            "message": "Request invalid: field 'command.options.skip' value 'Options[limit=null, skip=10, pageState=null, includeSimilarity=false]' not valid. Problem: skip option should be used with sort clause.",
            "errorCode": "COMMAND_FIELD_INVALID"
        }
    ]
}
```

src/main/java/io/stargate/sgv2/jsonapi/exception/mappers/ConstraintViolationExceptionMapper.java
tatu-at-datastax: Backslash is needed for double-quote if included in JSON; it is not added by us (and will be "removed" when reading JSON by code) but handled automatically.
But I'm fine going with single-quote to make it easier to read when enclosed in JSON.

Hazel-Datastax:
Close PR. We decided to support skip for non sort

PR #939, opened 2024-03-08
Paginated updateMany

### LINKED ISSUE #937: Support pagination to `UpdateMany` command
Support pagination to `UpdateMany` command. Following are the changes:
- The command will return a `nextPagingState` in case read operation done internally for this command has one. 
- The command will accept a `pagingState` which is used along with the find operation.
- Continue to support `moreData` flag, for backward compatibility but this will be just `nextPagingState != null` and may return a false positive.

maheshrajamani:
**What this PR does**:
Support pagination in updateMany command
Fixes #937 

**Checklist**
- [x] Changes manually tested


vkarpov15:
Can we add an option to JSON API for the JSON API layer to handle pagination in updateMany? I suppose we can handle this in the client layer as well, but it would be cleaner if JSON API could handle that.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/UpdateManyCommand.java
```diff
@@ -27,5 +29,13 @@ public record Options(
...
               defaultValue = "false")
-          boolean upsert) {}
+          boolean upsert,
+      @Nullable
+          @Schema(
+              description = "Next page state for pagination.",
+              type = SchemaType.STRING,
+              implementation = String.class)
+          @JsonProperty("pageState")
+          @JsonAlias("pagingState") // old name, 1.0.0-BETA-3 and prior
```
tatu-at-datastax, line 39: Do we need this also here? I know it is used for older cases. I guess that makes sense for consistency, and we need not document it (i.e. only include main new option in descriptions).

src/test/java/io/stargate/sgv2/jsonapi/api/v1/UpdateManyIntegrationTest.java
```diff
@@ -284,6 +289,80 @@ public void limitMoreDataFlag() {
...
+              .body("status.matchedCount", is(20))
+              .body("status.modifiedCount", is(20))
+              .body("status.moreData", is(true))
+              .body("status.nextPageState", notNullValue())
+              .body("errors", is(nullValue()))
+              .extract()
+              .body()
+              .path("status.nextPageState");
+
+      System.out.println("nextPageState : " + nextPageState);
```
Yuqi-Du, line 322: Could we delete this print.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/UpdateOperationPage.java
```diff
@@ -61,7 +61,10 @@ public CommandResult get() {
     updateStatus.put(CommandStatus.MATCHED_COUNT, matchedCount());
     updateStatus.put(CommandStatus.MODIFIED_COUNT, modifiedCount());
 
-    if (moreDataFlag) updateStatus.put(CommandStatus.MORE_DATA, moreDataFlag);
+    if (pagingState != null) {
+      updateStatus.put(CommandStatus.MORE_DATA, true);
```
Yuqi-Du, line 65: This is neat! for backwards compatibility.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/UpdateManyIntegrationTest.java
```diff
@@ -217,7 +220,8 @@ public void limit() {
           .statusCode(200)
           .body("status.matchedCount", is(20))
           .body("status.modifiedCount", is(20))
-          .body("status.moreData", nullValue())
+          .body("status.moreData", is(true))
+          .body("status.nextPageState", notNullValue())
```
tatu-at-datastax, line 224: I know we can't compare to specific value (as it's generated token), but maybe check it's a non-empty String (length above 0)?

tatu-at-datastax (approved):

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/UpdateManyCommand.java
maheshrajamani: Just used what's available in `find` command.

tatu-at-datastax: Right, figured. That's fine.

vkarpov15 (approved):
LGTM

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/ReadAndUpdateOperation.java
```diff
@@ -55,46 +54,26 @@ public record ReadAndUpdateOperation(
...
-                if (matchedCount.get() + docs.size() <= updateLimit) {
-                  matchedCount.addAndGet(docs.size());
-                  return Multi.createFrom().items(docs.stream());
-                } else {
-                  int needed = updateLimit - matchedCount.get();
-                  matchedCount.addAndGet(needed);
-                  moreDataFlag.set(true);
-                  return Multi.createFrom().items(findResponse.docs().subList(0, needed).stream());
-                }
+                matchedCount.addAndGet(docs.size());
```
vkarpov15, line 71: I noticed while working on stargate/stargate-mongoose#199 that this is a breaking change. In v1.0.1, `moreData` would be false if exactly 20 documents were updated. But now `moreData` is true if exactly 20 documents are updated.

maheshrajamani: Cassandra returns a page state if page size of records retrieved are retrieved . Reading with this pageState won't not return any more data. Yes, we would need to deprecate `moreData` flag. The looping should be based on `pageState != null`

PR #982, opened 2024-03-20
fix $not with $size 0

Yuqi-Du:
This filter should find all the documents with ages array size is not 0, or simply do not have ages array.

```
{
    "find": {
        "filter": {
            "$not": {
                "ages": {
                    "$size": 0
                }
            }
        }
    }
}
```
since we pushed down the $not logic by negating the size number, so, 0 * -1 still -1
FilterableResolver can not know the intention of doing a map_not_equal.

Methodology here, is to add validation first, that $size value should be integer.
Then, since we want to push down the $not info to the FilterableResolver, we need a flag to say this is negating 0.
I choose -1.5 as the flag, since 
- it is a negative number, meaning we have $not logic in it.
- it is not an integer, so will never be user's input, and we can reserve it as a flag indicating negating 0

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/filter/JsonLiteral.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/filter/JsonLiteral.java
@@ -1,17 +1,9 @@
 package io.stargate.sgv2.jsonapi.api.model.command.clause.filter;
 
 // Literal value to use as RHS operand in the query
-
-import java.math.BigDecimal;
-
 /**
  * @param value Literal value to use as RHS operand in the query
  * @param type Literal value type of the RHS operand in the query
  * @param <T> Data type of the object
  */
-public record JsonLiteral<T>(T value, JsonType type) {
-
-  // for negating 0, will set BigDecimal -0.5 as flag
-  // this is used when pushing down "$not" to {"$size":0}
-  public static final BigDecimal NEGATE_ZERO = new BigDecimal("-1.5");
-}
+public record JsonLiteral<T>(T value, JsonType type) {}
--- a/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/filter/JsonLiteral.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/filter/JsonLiteral.java
@@ -2,9 +2,16 @@ package io.stargate.sgv2.jsonapi.api.model.command.clause.filter;
 
 // Literal value to use as RHS operand in the query
 
+import java.math.BigDecimal;
+
 /**
  * @param value Literal value to use as RHS operand in the query
  * @param type Literal value type of the RHS operand in the query
  * @param <T> Data type of the object
  */
-public record JsonLiteral<T>(T value, JsonType type) {}
+public record JsonLiteral<T>(T value, JsonType type) {
+
+  // for negating 0, will set BigDecimal -0.5 as flag
+  // this is used when pushing down "$not" to {"$size":0}
+  public static final BigDecimal NEGATE_ZERO = new BigDecimal("-1.5");
+}
--- a/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/deserializers/FilterClauseDeserializer.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/deserializers/FilterClauseDeserializer.java
@@ -230,6 +230,11 @@ public class FilterClauseDeserializer extends StdDeserializer<FilterClause> {
                   ErrorCode.INVALID_FILTER_EXPRESSION,
                   "$size operator must have integer value >= 0");
             }
+            // Check if the value is an integer by comparing its scale.
+            if (i.stripTrailingZeros().scale() > 0) {
+              throw new JsonApiException(
+                  ErrorCode.INVALID_FILTER_EXPRESSION, "$size operator must have an integer value");
+            }
           } else {
             throw new JsonApiException(
                 ErrorCode.INVALID_FILTER_EXPRESSION, "$size operator must have integer");
--- a/src/test/java/io/stargate/sgv2/jsonapi/api/model/command/deserializers/FilterClauseDeserializerTest.java
+++ b/src/test/java/io/stargate/sgv2/jsonapi/api/model/command/deserializers/FilterClauseDeserializerTest.java
@@ -430,12 +430,82 @@ public class FilterClauseDeserializerTest {
     }
 
     @Test
+    public void mustHandleIntegerWithTrailingZeroSize() throws Exception {
+      String json = """
+          {"sizePath" : {"$size": 0.0}}
+        """;
+      final ComparisonExpression expectedResult =
+          new ComparisonExpression(
+              "sizePath",
+              List.of(
+                  new ValueComparisonOperation(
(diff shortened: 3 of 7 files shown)


src/main/java/io/stargate/sgv2/jsonapi/api/model/command/deserializers/FilterClauseDeserializer.java
```diff
@@ -230,6 +230,11 @@ private void validate(
                   ErrorCode.INVALID_FILTER_EXPRESSION,
                   "$size operator must have integer value >= 0");
             }
+            // Check if the value is an integer by comparing its scale.
+            if (i.stripTrailingZeros().scale() > 0) {
+              throw new JsonApiException(
+                  ErrorCode.INVALID_FILTER_EXPRESSION, "$size operator must have an integer value");
```
maheshrajamani, line 236: Why this is needed, the previous line validation should have taken care of this?

maheshrajamani (approved):
LGTM.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/filter/JsonLiteral.java
```diff
@@ -2,9 +2,16 @@
...
+
 /**
  * @param value Literal value to use as RHS operand in the query
  * @param type Literal value type of the RHS operand in the query
  * @param <T> Data type of the object
  */
-public record JsonLiteral<T>(T value, JsonType type) {}
+public record JsonLiteral<T>(T value, JsonType type) {
+
+  // for negating 0, will set BigDecimal -0.5 as flag
```
tatu-at-datastax, line 14: Comment and value differ here...

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/deserializers/FilterClauseDeserializer.java
```diff
@@ -230,6 +230,11 @@ private void validate(
                   ErrorCode.INVALID_FILTER_EXPRESSION,
                   "$size operator must have integer value >= 0");
             }
+            // Check if the value is an integer by comparing its scale.
+            if (i.stripTrailingZeros().scale() > 0) {
```
tatu-at-datastax, line 234: I don't think we have suitable `NumberUtil` class yet, but in future we might want to create something like `ValueUtil` to collect helper methods like this one here (I did check and this is indeed a good way for checking).

src/test/java/io/stargate/sgv2/jsonapi/api/model/command/deserializers/FilterClauseDeserializerTest.java
```diff
@@ -429,6 +429,49 @@ public void mustHandleSize() throws Exception {
           .isEqualTo(expectedResult.getPath());
     }
 
+    @Test
+    public void mustHandleIntegerWithTrailingZeroSize() throws Exception {
+      String json = """
+          {"sizePath" : {"$size": 0.0}}
```
tatu-at-datastax, line 435: Is this actually required to be legal? (curious, maybe Mongo allows it?)

tatu-at-datastax:
Ok, aside from smaller questions I had, my main question is this: can we avoid using `BigDecimal` marker, and use either `Boolean` or `String`? Of these, `Boolean.TRUE` might be safest (not the default value of `Boolean.FALSE`, nor same as explicit number value)?

There may be reason this is not doable, so just asking.

tatu-at-datastax (approved):
LGTM; had some questions but nothing blocking.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/deserializers/FilterClauseDeserializer.java
Yuqi-Du: This is for user inputs that are not Integer.

{
    "find": {
        "filter": {
                "ages": {
                    "$size":  2.43
                }
        }
    }
}

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/deserializers/FilterClauseDeserializer.java
Yuqi-Du: Thanks

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/filter/JsonLiteral.java
Yuqi-Du: fixed, thanks

src/test/java/io/stargate/sgv2/jsonapi/api/model/command/deserializers/FilterClauseDeserializerTest.java
Yuqi-Du: I checked Mongo Side. They allow it.

Yuqi-Du:
> Ok, aside from smaller questions I had, my main question is this: can we avoid using `BigDecimal` marker, and use either `Boolean` or `String`? Of these, `Boolean.TRUE` might be safest (not the default value of `Boolean.FALSE`, nor same as explicit number value)?
> 
> There may be reason this is not doable, so just asking.

Nice suggestion.
Fixed!

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/filter/JsonLiteral.java
```diff
@@ -1,7 +1,6 @@
 package io.stargate.sgv2.jsonapi.api.model.command.clause.filter;
 
 // Literal value to use as RHS operand in the query
```
tatu-at-datastax, line 3: nit: this comment probably needs to be removed too?

tatu-at-datastax (approved):

src/test/java/io/stargate/sgv2/jsonapi/api/v1/FindIntegrationTest.java
```diff
@@ -929,6 +929,37 @@ public void withSizeOperatorNoMatch() {
...
     }
 
+    @Test
+    public void withSizeOperatorNotZero() {
+      String json =
+          """
+                  {
+                      "find": {
+                          "filter": {
+                              "$not": {
```
vkarpov15, line 939: The $not syntax that Mongoose supports is `tags: { $not: { $size: 0 } }`, not `{ $not: { tags: { $size: 0 } } }`. Does the `tags: { $not: { $size: 0 } }` syntax work here?

src/test/java/io/stargate/sgv2/jsonapi/api/v1/FindIntegrationTest.java
```diff
@@ -929,6 +929,37 @@ public void withSizeOperatorNoMatch() {
...
+          .contentType(ContentType.JSON)
+          .body(json)
+          .when()
+          .post(CollectionResource.BASE_PATH, namespaceName, collectionName)
+          .then()
+          .statusCode(200)
+          .body("status", is(nullValue()))
+          .body("errors", is(nullValue()))
+          // should have all the documents
+          .body("data.documents", hasSize(6));
```
vkarpov15, line 960: For completeness, would be great to have a document with `tags: []` and test that's excluded from the output

PR #984, opened 2024-03-20
Fix #983: prevent serialization of "empty" defaultId in CollectionSet…

### LINKED ISSUE #983: ObjectId: do not include `defaultId` entry in `findCollections` if user provided no `type` setting
The initial implementation of type id discovery in `CollectionSettings` returns empty String for case where user did not provide any value (or provided `null` or empty String), when requested with `findCollections`. Instead we should simply not include it (suppress).

Further, since it is the only property of `defaultId`, we should also suppress `defaultId` Object itself.

tatu-at-datastax:
**What this PR does**:

Filters out "empty" value for `defaultId`, for cases where user did not provide mechanism (backwards compatible)

Fixes #983

**Checklist**
- [ ] Changes manually tested


Yuqi-Du (approved):

vkarpov15 (commented):
LGTM

PR #998, opened 2024-03-25
Stop tagging CQL metrics by session id

jeffreyscarpenter:
When deployed in a multi tenant environment with lots of sessions being created, tagging the CQL driver metrics by session ID can lead to an explosion in metric cardinality, making it difficult to scrape metrics.


jeffreyscarpenter:
I thought about adding a property to make tagging by session optional, but not sure there is use for that.

Hazel-Datastax (approved):

Issue #1005, opened 2024-03-27
Exclude $vector and $vectorize from the default Projection


tatu-at-datastax:
(note: depends on #1001)

Since fields `$vector` (and soon to be added `$vectorize`) are rarely needed by most users and (equally importantly) are generally large and heavy-weight to process, let's make them "exclude-by-default" -- reverse of `_id` "included-by-default".

This means that without Projection definition, they would not be included in output but instead need to be explicitly included (except for #1001 which will add a short-cut for forced inclusion).

PR #1008, opened 2024-03-29
Add support for "star" inclusion/exclusion (include all/exclude all)

### LINKED ISSUE #1001: Support wildcard "include EVERYTHING" and "include NOTHING" for Projection clause
Due to planned changes for auto-exclusion of certain pseudo-virtual fields (namely, `$vector`), we need mechanism to allow specifying "include everything", like so:

```
{"projection": {
	"*" : 1
}}
```

and for symmetry, might as well add its counterpart:

```
{"projection": {
	"*" : 0
}}
```

for "include nothing".

In both cases this would affect all fields stored in underlying `"doc_json"` as well as otherwise special key property, `"_id"`.

tatu-at-datastax:
**What this PR does**:

Adds support for 2 special query projections:

```
{ "*": 1} // include everything
{"*": 0} // include nothing
```

as per #1001, as convenience shortcuts that ignore special field (_id, $vector, $vectorize) logic regarding default inclusion/exclusion.

Also prepares for bigger fix, #1005.

Fixes #1001

**Checklist**
- [ ] Changes manually tested


src/test/java/io/stargate/sgv2/jsonapi/api/v1/VectorSearchIntegrationTest.java
```diff
@@ -588,6 +588,69 @@ public void happyPath() {
 
     @Test
     @Order(3)
+    public void happyPathWithIncludeAll() {
+      String json =
+          """
+                          {
+                            "find": {
+                              "sort" : {"$vector" : [0.15, 0.1, 0.1, 0.35, 0.55]},
+                              "projection" : {"*" : 1},
```
kathirsvn, line 597: Can we add a test case with non-zero and non-one value as well? (to make sure it behaves as true)

kathirsvn (approved):
LGTM

src/test/java/io/stargate/sgv2/jsonapi/api/v1/VectorSearchIntegrationTest.java
tatu-at-datastax: @kathirsvn Sounds good, will add in suitable place.

tatu-at-datastax: Added more unit tests to cover non-0/non-1 numbers, booleans.

vkarpov15 (commented):
LGTM

PR #1013, opened 2024-04-02
Refactor `ThrowableToErrorMapper` class

Hazel-Datastax:
**What this PR does**:
Refactor `ThrowableToErrorMapper` class

NA

**Checklist**
- [x] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCode.java
```diff
@@ -155,17 +155,25 @@ public enum ErrorCode {
...
       "Collection creation failure (unable to create table). Recommend re-creating the collection"),
   INVALID_SCHEMA_VERSION(
       "Collection has invalid schema version. Recommend re-creating the collection"),
-  INVALID_ID_TYPE("Invalid Id type");
+  INVALID_ID_TYPE("Invalid Id type"),
+
+  // Driver failure codes
+  /** Error codes related to driver exceptions. */
+  SERVER_NO_NODE_AVAILABLE("No node was available to execute the query"),
+  SERVER_INVALID_QUERY("Invalid query"),
```
jeffreyscarpenter, line 167: Isn't this a user error? According to the error mapping code we return 200 OK. Should remove `SERVER_` prefix here.

src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCode.java
```diff
@@ -155,17 +155,25 @@ public enum ErrorCode {
...
+  SERVER_NO_INDEX_ERROR(
```
jeffreyscarpenter, line 168: Same here. We return 200, so it is not a server error.

src/main/java/io/stargate/sgv2/jsonapi/exception/mappers/ThrowableToErrorMapper.java
```diff
@@ -45,95 +42,158 @@ public final class ThrowableToErrorMapper {
...
         }
 
+        // handle all driver exceptions
+        if (throwable instanceof DriverException) {
+          return handleDriverException(
+              (DriverException) throwable, message, fields, fieldsForMetricsTag);
+        }
+
+        // handle all other exceptions
         return new CommandResult.Error(message, fieldsForMetricsTag, fields, Response.Status.OK);
```
jeffreyscarpenter, line 60: Does this mean we are not returning a JsonApiException if we get to here? If it's an exception we don't recognize in any of our logic, should we really be returning HTTP OK?

src/main/java/io/stargate/sgv2/jsonapi/exception/mappers/ThrowableToErrorMapper.java
```diff
@@ -45,95 +42,158 @@ public final class ThrowableToErrorMapper {
...
+                  ErrorCode.UNAUTHENTICATED_REQUEST.getMessage(), Response.Status.UNAUTHORIZED);
+          // Driver NoNodeAvailableException -> ErrorCode.NO_NODE_AVAILABLE
+        } else if (error instanceof NoNodeAvailableException) {
+          return ErrorCode.SERVER_NO_NODE_AVAILABLE
+              .toApiException()
+              .getCommandResultError(message, Response.Status.INTERNAL_SERVER_ERROR);
+        }
+      }
+    }
+    return new CommandResult.Error(message, fieldsForMetricsTag, fields, Response.Status.OK);
```
jeffreyscarpenter, line 194: Same comment as above "Does this mean we are not returning a JsonApiException if we get to here? If it's an exception we don't recognize in any of our logic, should we really be returning HTTP OK?"

jeffreyscarpenter (changes requested):

src/main/java/io/stargate/sgv2/jsonapi/exception/mappers/ThrowableToErrorMapper.java
jeffreyscarpenter: This return statement is our fallback in case there is an exception we do not recognize. We want to make a big deal out of this so that we can highlight the error for ourselves. Let's map it to `JsonApiException` with a new error code `SERVER_UNHANDLED_ERROR` and `Response.Status.INTERNAL_SERVER_ERROR` (aka 500).

jeffreyscarpenter (approved):

PR #1016, opened 2024-04-02
Add support for default exclusion of $vector and $vectorize

### LINKED ISSUE #1005: Exclude $vector and $vectorize from the default Projection

tatu-at-datastax:
**What this PR does**:

Will change default projection inclusion to exclude `$vector` and `$vectorize`, needing explicit inclusion.

Fixes #1005

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/projection/DocumentProjector.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/projection/DocumentProjector.java
@@ -55,6 +55,10 @@ public class DocumentProjector {
     return DefaultProjectorWrapper.defaultProjector();
   }
 
+  public static DocumentProjector includeAllProjector() {
+    return INCLUDE_ALL_PROJECTOR;
+  }
+
   DocumentProjector withIncludeSimilarity(boolean includeSimilarityScore) {
     if (this.includeSimilarityScore == includeSimilarityScore) {
       return this;
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/projection/DocumentProjector.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/projection/DocumentProjector.java
@@ -140,7 +140,7 @@ public class DocumentProjector {
       // In either case, we may need to add similarity score if present
       if (includeSimilarityScore && similarityScore != null) {
         ((ObjectNode) document)
-            .put(DocumentConstants.Fields.VECTOR_FUNCTION_PROJECTION_FIELD, similarityScore);
+            .put(DocumentConstants.Fields.VECTOR_FUNCTION_SIMILARITY_FIELD, similarityScore);
       }
       return;
     }
@@ -151,7 +151,7 @@ public class DocumentProjector {
     }
     if (includeSimilarityScore && similarityScore != null) {
       ((ObjectNode) document)
-          .put(DocumentConstants.Fields.VECTOR_FUNCTION_PROJECTION_FIELD, similarityScore);
+          .put(DocumentConstants.Fields.VECTOR_FUNCTION_SIMILARITY_FIELD, similarityScore);
     }
   }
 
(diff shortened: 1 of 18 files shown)


src/main/java/io/stargate/sgv2/jsonapi/service/projection/DocumentProjector.java
```diff
@@ -191,36 +228,32 @@ static PathCollector collectPaths(JsonNode def, boolean includeSimilarity) {
     }
 
     public DocumentProjector buildProjector() {
-      if (isDefaultProjection()) {
```
tatu-at-datastax, line 194: Caller will use the short-cut, no need to check whether explicit definition happens to work like default projection.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/FindOneAndUpdateNoIndexIntegrationTest.java
```diff
@@ -126,8 +126,7 @@ public void byIdAfterUpdate() {
                         "name": "Joe",
                         "age": 42,
                         "enabled": true,
-                        "value": -1,
-                        "$vector" : [ 0.5, -0.25 ]
```
tatu-at-datastax, line 130: Simplifying the test since $vector not really needed for test (but new default projection would drop)

vkarpov15 (commented):
LGTM so far

src/main/java/io/stargate/sgv2/jsonapi/service/projection/DocumentProjector.java
```diff
@@ -191,36 +230,36 @@ static PathCollector collectPaths(JsonNode def, boolean includeSimilarity) {
...
       // One more thing: do we need to add document id?
       if (inclusions > 0) { // inclusion-based projection
-        // doc-id included unless explicitly excluded
         return new DocumentProjector(
-            ProjectionLayer.buildLayersNoOverlap(paths, slices, !Boolean.FALSE.equals(idInclusion)),
+            ProjectionLayer.buildLayersForProjection(
+                paths,
+                slices,
+                // doc-id included unless explicitly excluded
+                !Boolean.FALSE.equals(idInclusion),
```
Yuqi-Du, line 240: This is interesting.
So by default, idInclusion will be null, which will result as **addDocId** as
- true for inclusion-based projection
- false for exclusion-based projection

So essentially the meaning of **addDocId** will be flipped, in terms of what projection is used `inclusion-based` or `exclusion-based`. The variable name addDocId (similarly add$vector, add$vectorize) gives some confusion to me initially, but it makes sense afterwards.

Yuqi-Du (approved):

src/main/java/io/stargate/sgv2/jsonapi/service/projection/DocumentProjector.java
tatu-at-datastax: Good point, naming is hard. Instead of "addDocId" could use "includeDocId" but that would be even more confusing.
If you (or anyone else) has better name suggestion would be happy to rename.

tatu-at-datastax:
> if we understand correctly, `"projection": {"name" : true, "age":false}` will error out, since projection logic thinks this is inclusion-based but since $vector and $vectorize are special cases and handled differently(not by List paths), so we still allow something like this ? `"projection": { "$vector": 1, "$vectorize" : 0}`

Correct: `_id`, `$vector`, `$vectorize` (and any future `$`-prefixed special fields) are special in that they can be used for any inclusion/exclusion case.

PR #1023, opened 2024-04-11
Username/Password credential based token format

maheshrajamani:
**What this PR does**:
Changed the token format for username/password credential to `"Cassandra:"+Base64(username)+Base64(password)`

**Checklist**
- [x] Changes manually tested


Yuqi-Du (approved):
LGTM

vkarpov15 (commented):
Do clients need to do anything with this change? I don't see anywhere where we use the "username/password" syntax in stargate-mongoose

tatu-at-datastax:
@vkarpov15 this is not needed when connecting to Astra, only for local testing for now, with DSE 7. So probably not a problem in short term for stargate-mongoose.

vkarpov15:
@tatu-at-datastax  What about for local development? Because I'm trying stargate-mongoose with data-api@1.0.7 and every single request now fails with an HTTP 401, but tests pass with data-api@1.0.6. Is there anything I need to fix in stargate-mongoose, or should I just avoid data-api@1.0.7 for now?

tatu-at-datastax:
@vkarpov15 That is unexpected, I think -- nothing in 1.0.7-SNAPSHOT should break existing usage, as far as I know. But I don't know the details of changes, I may be missing something.

@maheshrajamani WDYT?

vkarpov15:
That's strange, because right now when I run v1.0.7 using [this script](https://github.com/stargate/stargate-mongoose/blob/main/bin/start_data_api.sh), I get back an auth token from the coordinator:

```
{ authToken: '79314c19-8e0e-474a-b0e3-bd524c22b135' }
```

I send the following HTTP request:

```
{
  url: 'http://localhost:8181/v1',
  method: 'POST',
  data: { createNamespace: { name: 'testks1' } },
  headers: { 'X-Cassandra-Token': '79314c19-8e0e-474a-b0e3-bd524c22b135' }
}

```

And I get back an HTTP 401 with the following errors:

```
[
  {
    message: 'Invalid credentials format, expected `Cassandra:Base64(username):Base64(password)`',
    errorCode: 'UNAUTHENTICATED_REQUEST'
  }
]

```

tatu-at-datastax:
Ahhhhh! That does sounds familiar -- could you bring this up on Slack -- there are changes like this.

PR #1026, opened 2024-04-15
Add test to show #1000 no longer applies

### LINKED ISSUE #1000: Upsertion with `projection: { _id: 0 }` appears broken
When doing `findOneAndReplace` with `projection: { _id: 0 }`, resulting in an upsert, the following is returned:
```ts
{
  data: {},
  errors: [{ message: 'Cannot invoke "com.fasterxml.jackson.databind.JsonNode.isArray()" because "subtree" is null' }],
}
```
where the data is interestingly non-null and the error field being quite... "internal"

Command sent:
```ts
{
  findOneAndReplace: {
    filter: { 'address.city': 'nyc' },
    replacement: {
      username: 'jimr',
      human: true,
      age: 52,
      password: 'gasxaq==',
      address: [Object]
    },
    options: { returnDocument: 'before', upsert: true },
    projection: { _id: 0 }
  }
}
```

The only document in the collection at the time (causing an upsert)
```ts
{
  username: 'aaron',
  human: true,
  age: 47,
  password: null,
  address: {
    number: 86,
    street: 'monkey street',
    suburb: 'not null',
    city: 'big banana',
    is_office: false,
  },
}
```

This may or may not happen with other commands as well, haven't tested.

tatu-at-datastax:
**What this PR does**:

Tries to reproduce issue #1000 but instead passes -- looks like issue has been resolved with a release since report (probably as part of #1005 )

Fixes #1000

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/src/test/java/io/stargate/sgv2/jsonapi/api/v1/FindOneAndReplaceIntegrationTest.java
+++ b/src/test/java/io/stargate/sgv2/jsonapi/api/v1/FindOneAndReplaceIntegrationTest.java
@@ -685,65 +685,41 @@ public class FindOneAndReplaceIntegrationTest extends AbstractCollectionIntegrat
           """
               {
                 "_id": "docProjBeforeNoId",
-                "username": "aaron",
-                "human": true,
-                "age": 47,
-                "password": null,
-                "address": {
-                  "number": 86,
-                  "street": "monkey street",
-                  "is_office": false
-                }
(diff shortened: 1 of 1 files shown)


tatu-at-datastax:
Quick note: was able to reproduce again 1.0.5 version; no longer fails against current head of `main` which will be released as 1.0.6.

PR #1028, opened 2024-04-16
Fix #1042: Increase `DEFAULT_MAX_DOCUMENT_INSERT_COUNT` to 100 (from 20)

### LINKED ISSUE #1042: Increase DEFAULT_MAX_DOCUMENT_INSERT_COUNT to 100 (from 20)
As part of re-evaluation of processing limits, based on feedback, it was decided that we should allow bigger batches for `InsertMany`, if and when performance testing confirms this is safe to do (wrt likelihood of causing cascading failures on server-side due to timeouts).

Initial performance analysis confirms that:

1. Increase to 100 for insertMany appears safe for tested cases (vector/non-vector; ordered/unordered)
2. For maximum throughput, increase of batch size up to ~50 may be beneficial for some cases (in particular small non-vector documents) -- no further benefits observed beyond this point, and for many cases (vector documents) peak throughput with batch size as low as 15

so we can apply an increase to this limit, uniformly. Further increase would likely consider different limits for ordered/unordered case as as possibly vector/non-vector case -- this may get difficult (esp. latter, with "vectorize" functionality).

tatu-at-datastax:
**What this PR does**:

Increases `DEFAULT_MAX_DOCUMENT_INSERT_COUNT` which limits maximum number of documents allowed for "insertMany()" to 100, from its current (v1.0.5) value of 20.

N/A

**Checklist**
- [x] Changes manually tested


tatu-at-datastax:
Quick note: initial performance analysis confirms that:

1. Increase to 100 for `insertMany` appears safe for tested cases (vector/non-vector; ordered/unordered)
2. For maximum throughput, increase of batch size up to ~50 may be beneficial for some cases (in particular small non-vector documents) -- no further benefits observed beyond this point, and for many cases (vector documents) peak throughput with batch size as low as 15

so can proceed with this PR.

tatu-at-datastax:
Actually, will close and re-create to link to #1042 properly.

PR #1029, opened 2024-04-16
Remove histograms from command processor metrics

jeffreyscarpenter:
Remove histograms from `command_processor_process` metrics (i.e. no more `command_processor_process_seconds_bucket`) since this results in very high cardinality, especially in multi-tenant deployments. The latency metrics produced are of little additional value since they are so close to the latencies that are available via `http_server_requests_seconds_bucket` (the HTTP layer is very efficient and adds little overhead on top of the command processor)


jeffreyscarpenter:
Sample output of metrics endpoint obtained from local test run.

[metrics_without_histogram2.txt](https://github.com/stargate/data-api/files/15269839/metrics_without_histogram2.txt)

Yuqi-Du (approved):

PR #1044, opened 2024-04-19
Add Data API integration Test logs

Yuqi-Du:
**Checklist**
- [x] Changes manually tested


tatu-at-datastax (approved):

jeffreyscarpenter:
Do we want to do this? This makes it so that the API is run via forked JVM process rather than as a container. I was hoping to get the logs from the container.

jeffreyscarpenter:
I think we should close this, as I'd prefer we run ITs with the docker container. We could add a note to the README to describe how to run ITs without container for debugging purposes.

PR #1047, opened 2024-04-29
Added support for embedding gateway

maheshrajamani:
**What this PR does**:

Fixes #<issue number>

**Checklist**
- [x] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/api/v1/CollectionResource.java
```diff
@@ -181,7 +181,14 @@ else if (error instanceof JsonApiException jsonApiException) {
                       embeddingProviderFactory.getConfiguration(
                           dataApiRequestInfo.getTenantId(),
                           collectionProperty.vectorConfig().vectorizeConfig().provider(),
-                          collectionProperty.vectorConfig().vectorizeConfig().modelName());
+                          collectionProperty.vectorConfig().vectorizeConfig().modelName(),
```
tatu-at-datastax, line 184: nit: I know use of tmp variables is frowned upon, but here it might make sense to use:

```
var vectorizeConfig = collectionProperty.vectorConfig().vectoriConfig();
```

to shorten code to be more readable

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/gateway/EmbeddingGatewayClient.java
```diff
@@ -0,0 +1,162 @@
...
+                  EmbeddingGateway.ProviderEmbedRequest.EmbeddingRequest.ParameterValue.newBuilder()
+                      .setFloatValue((Float) value)
+                      .build());
+            else if (value instanceof Boolean)
+              grpcVectorizeServiceParameter.put(
+                  key,
+                  EmbeddingGateway.ProviderEmbedRequest.EmbeddingRequest.ParameterValue.newBuilder()
+                      .setBoolValue((Boolean) value)
+                      .build());
+          });
```
tatu-at-datastax, line 109: should also have one more `else` to throw exception for unrecognized `value` type?

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/gateway/EmbeddingGatewayClient.java
```diff
@@ -0,0 +1,162 @@
...
+                embeddingRequestType == EmbeddingRequestType.INDEX
+                    ? EmbeddingGateway.ProviderEmbedRequest.EmbeddingRequest.InputType.INDEX
+                    : EmbeddingGateway.ProviderEmbedRequest.EmbeddingRequest.InputType.SEARCH)
+            .addAllInputs(texts)
+            .build();
+
+    final EmbeddingGateway.ProviderEmbedRequest.ProviderContext.Builder builder =
+        EmbeddingGateway.ProviderEmbedRequest.ProviderContext.newBuilder()
+            .setProviderName(provider)
+            .setTenantId(tenant.orElse("default"));
```
tatu-at-datastax, line 126: Maybe define constant for `DEFAULT_TENANT_ID`?

Not sure where this should be documented, but would be also good to explain briefly meaning -- does this mean defaulting or that tenant-id is not available and not used (which I think is the case)?
(could also use different token here -- empty String, or `"missing"` or something -- unless there is some actual default tenant-id that is used)

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/gateway/EmbeddingGatewayClient.java
```diff
@@ -0,0 +1,162 @@
...
+                    ErrorCode.valueOf(resp.getError().getErrorCode()),
+                    resp.getError().getErrorMessage());
+              }
+              if (resp.getEmbeddingsList() == null) {
+                return Collections.emptyList();
+              }
+              return resp.getEmbeddingsList().stream()
+                  .map(
+                      data -> {
+                        Float[] embedding = new Float[data.getEmbeddingList().size()];
```
tatu-at-datastax, line 155: `Float[]` is quite inefficient (wrapper over `float`) -- is there any chance native `float[]` could be used instead? This would avoid construction of throw-away `Float` values if input is as `float`s.

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/CohereEmbeddingClient.java
```diff
@@ -33,16 +34,19 @@ public class CohereEmbeddingClient implements EmbeddingProvider {
   private String modelName;
   private String baseUrl;
   private final CohereEmbeddingProvider embeddingProvider;
+  private Map<String, Object> vectorizeServiceParameter;
 
   public CohereEmbeddingClient(
```
tatu-at-datastax, line 39: Asking offline; but question is whether concrete impl clients are needed on Data API side -- perhaps this is only needed temporarily? Otherwise it seems like impl differences should only be handed on Embedding Gateway service side.

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/CohereEmbeddingClient.java
```diff
@@ -33,16 +34,19 @@ public class CohereEmbeddingClient implements EmbeddingProvider {
...
       EmbeddingProviderConfigStore.RequestProperties requestProperties,
       String baseUrl,
       String apiKey,
-      String modelName) {
+      String modelName,
+      Map<String, Object> vectorizeServiceParameter) {
```
tatu-at-datastax, line 44: nit: `parameter` -> `parameters`?

tatu-at-datastax (approved):
LGTM: added some comments but not blocking merge.

vkarpov15 (commented):
LGTM

PR #1048, opened 2024-04-30
Update Embedding Providers Config 

Hazel-Datastax:
**What this PR does**:
1. Change the config yaml file according to the recent design changes
    1.  Change `supported-authentication` from list to map
    2. `vector-dimension` can be null and the value could be in parameters
    3. Remove `api-key`
    4. Lowercase parameter type: `STRING` -> `string`
    5. Add more embedding providers (only in config yaml) - azureOpenAI, jinaAI, voyageAI, mistral

TODO in the following PR:
1. Changes in config file break most verification code in `CreateCollectionCommandResolver`, fix them in the following PR.
2. `voyageAI` properties config in yaml - rethink the design of `RequestProperties` - more properties in the future?
3. Fix the uppercase `NUMERIC_RANGE` in the `findEmbeddingProviders` return body - cause by enum

Fixes #<issue number>

**Checklist**
- [x] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/service/embedding/configuration/PropertyBasedEmbeddingProviderConfig.java
```diff
@@ -29,8 +29,46 @@ interface EmbeddingProviderConfig {
...
 
+    /**
+     * A map of supported authentications. HEADER, SHARED_SECRET and NONE are the only techniques
+     * the DataAPI supports (i.e. the key of map can only be HEADER, SHARED_SECRET or NONE).
+     *
+     * @return
+     */
     @JsonProperty
-    List<String> supportedAuthentication();
+    Map<String, AuthenticationConfig> supportedAuthentication();
```
tatu-at-datastax, line 39: Could we use an `enum` of known auth methods as the key instead of String?

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/configuration/PropertyBasedEmbeddingProviderConfig.java
```diff
@@ -84,6 +129,25 @@ interface ParameterConfig {
...
+       * <p>`options` if present is an array of valid options the user must select from, for example
+       * if a model supports 3 different dimensions. If options are present the only allowed values
+       * for the parameter are those in the options list. If not present, null, or an empty array
+       * any value of the correct type is accepted.
+       *
+       * @return
+       */
+      @Nullable
+      @JsonProperty
+      Map<String, List<Integer>> validation();
```
tatu-at-datastax, line 149: Could we use enum as the key here too? (for `numericRange`, `option`)

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/configuration/PropertyBasedEmbeddingProviderConfig.java
tatu-at-datastax, line 39: nit: should maybe be "supportedAuthentications" (plural) since it can contain multiple entries, not just one

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/configuration/PropertyBasedEmbeddingProviderConfig.java
```diff
@@ -29,8 +29,46 @@ interface EmbeddingProviderConfig {
...
+    interface AuthenticationConfig {
+      @JsonProperty
+      boolean enabled();
+
+      @JsonProperty
+      List<TokenConfig> tokens();
+    }
+
+    /**
+     * For the HEADER technique the `accepted` value is the name of the header the client should
```
tatu-at-datastax, line 58: Why don't we use something like "headerName" instead of "accepted" if this is related to HEADER technique?
And similarly "secretName" instead of "forwarded" when SHARED_SECRET option?

tatu-at-datastax: Ok so basically this is because both "accepted" and "forwarded" are accepted for both HEADER and SHARED_SECRET. Make sense.

tatu-at-datastax (approved):

vkarpov15 (commented):
LGTM

PR #1054, opened 2024-05-02
Move embedding-service class under embedding package

tatu-at-datastax:
**What this PR does**:

Moves embedding-gateway - related class to proper Java package

N/A

**Checklist**
- [ ] Changes manually tested


tatu-at-datastax:
@maheshrajamani I realized that this may or may not make sense. Basically:

1. If this is for use by Data API itself, current location makes sense
2. If this is for use by Embedding gateway (separate package), should be moved as with this PR.

Just LMK either way so I can either close PR (if (1)), otherwise merge pending approval (if (2)).

maheshrajamani (approved):

maheshrajamani:
> * If this is for use by Data API itself, current location makes sense

This is used in data api. But can be moved because the header is for calling embedding provider service

PR #1056, opened 2024-05-02
Add support for Azure OpenAI embedding service

tatu-at-datastax:
**What this PR does**:

Adds support Azure OpenAI embedding provider

N/A

**Checklist**
- [x] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/AzureOpenAiEmbeddingClient.java
```diff
@@ -0,0 +1,111 @@
...
+import org.eclipse.microprofile.rest.client.annotation.RegisterProvider;
+import org.eclipse.microprofile.rest.client.inject.RegisterRestClient;
+
+public class AzureOpenAiEmbeddingClient implements EmbeddingProvider {
+  private EmbeddingProviderConfigStore.RequestProperties requestProperties;
+  private String modelName;
+  private int dimension;
+  private final OpenAiEmbeddingProvider embeddingProvider;
+
+  public AzureOpenAiEmbeddingClient(
```
maheshrajamani, line 34: Can you add the docs?

maheshrajamani (approved):
LGTM

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/EmbeddingProviderFactory.java
```diff
@@ -31,12 +31,14 @@ EmbeddingProvider create(
...
         String baseUrl,
         String modelName,
+        int dimension,
         Map<String, Object> vectorizeServiceParameter);
   }
 
   private static final Map<String, ProviderConstructor> providersMap =
       Map.ofEntries(
           Map.entry(ProviderConstants.OPENAI, OpenAiEmbeddingClient::new),
+          Map.entry(ProviderConstants.AZURE_OPENAI, AzureOpenAiEmbeddingClient::new),
```
vkarpov15, line 41: Minor nitpick: maybe rename `AzureOpenAiEmbeddingClient` -> `AzureOpenAIEmbeddingClient` for consistency with `VertexAIEmbeddingClient` below?

src/main/resources/embedding-providers-config.yaml
```diff
@@ -74,6 +75,28 @@ stargate:
...
+          models:
+            - name: text-embedding-3-small
+              parameters:
+                - name: vectorDimension
+                  type: number
+                  required: true
+                  default-value: 512
+                  validation:
+                    numeric-range: [2, 1536]
+                  help: "Vector dimension to use in the database and when calling Open AI."
```
vkarpov15, line 87: "OpenAI" is one word

src/main/resources/embedding-providers-config.yaml
```diff
@@ -74,6 +75,28 @@ stargate:
               help: "The API version to use for this operation. This follows the YYYY-MM-DD format."
           properties:
             max-input-length: 16
+          models:
+            - name: text-embedding-3-small
+              parameters:
+                - name: vectorDimension
+                  type: number
+                  required: true
+                  default-value: 512
```
vkarpov15, line 84: Are we hard-coding defaults ourselves? [This page on the OpenAI docs](https://platform.openai.com/docs/guides/embeddings/what-are-embeddings) indicates that text-embedding-3-small's default dimension is 1536

image

vkarpov15 (commented):
A couple of minor comments but LGTM overall

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/EmbeddingProviderFactory.java
tatu-at-datastax: Tough call: I follow `OpenAiEmbeddingClient` here... :-/

src/main/resources/embedding-providers-config.yaml
tatu-at-datastax: Fixed in new commit

src/main/resources/embedding-providers-config.yaml
tatu-at-datastax: I copied this from non-Azure OpenAI so not sure about reasoning.

@maheshrajamani @Hazel-Datastax WDYT?

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/EmbeddingProviderFactory.java
tatu-at-datastax: @maheshrajamani @Hazel-Datastax I can change names of both if you prefer "AI" capitalization, but wanted to check before doing that

src/main/resources/embedding-providers-config.yaml
tatu-at-datastax: I _think_ default in these cases would be what is given as the baseline in UI (Portal); neither Data API nor Embedding gateway would implement defaulting logic, but always pass explicit value used with `createCollection()` call.

Hazel-Datastax: I copied the config from Aaron's design docs and didn't check the openai's docs :(
Yes, Tatu is correct. No matter what the default value is, users have to give one when they create the collection. But we can still change the default value to align with the openai's docs now, it will not cause any trouble.

tatu-at-datastax: I asked on team Slack on which values to use: we do not have to use same value as OpenAI, if we think users would be served better by lower value. But this is sort of more Product question and I am not sure of the right answer myself. Will see if anyone has strong opinions.

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/AzureOpenAiEmbeddingClient.java
tatu-at-datastax: I did not see Javadocs for other clients, so not sure what to describe here? Link to Azure OpenAI docs maybe?

src/main/resources/embedding-providers-config.yaml
tatu-at-datastax: Leaving as-is for (near-)future improvements.

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/AzureOpenAiEmbeddingClient.java
tatu-at-datastax: Adding link from class Javadoc.

PR #1061, opened 2024-05-06
update workflows to publish jar

jeffreyscarpenter:
Update workflows that publish Data API as jar:

- update Publish Jar workflow to run on every commit to main (this will update SNAPSHOT versions)
- update Release workflow to publish non-SNAPSHOT versions

(also added event dispatch to release workflow)

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -347,3 +347,19 @@ jobs:
           branch-suffix: "short-commit-hash"
           base: "main"
 
+  # dispatch release event
+  dispatch:
+    name: Dispatch release event
+    needs: [ 'resolve-tag', 'publish-jar', 'publish-image-dockerhub', 'publish-image-ecr']
+    if: ${{ always() }}
+    continue-on-error: true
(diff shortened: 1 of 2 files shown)


tatu-at-datastax (approved):

.github/workflows/release.yaml
```diff
@@ -347,3 +347,19 @@ jobs:
           branch-suffix: "short-commit-hash"
           base: "main"
 
+  # dispatch release event
+  dispatch:
+    name: Dispatch release event
+    needs: [ 'resolve-tag', 'publish-jar', 'publish-image-dockerhub', 'publish-image-ecr']
+    if: ${{ always() }}
+    continue-on-error: true
```
tatu-at-datastax, line 355: Ok why is this? Should it be sent even if release fails? (maybe add a comment above explaining the reason)

tatu-at-datastax (approved):
LGTM: only have one question on why run always, even in case of failed step(s)?

PR #1105, opened 2024-05-21
Handle empty string `$vectorize` field

maheshrajamani:
**What this PR does**:
Resolve empty string `$vectorize` field value to json null value for `$vector` field.

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/embedding/DataVectorizer.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/embedding/DataVectorizer.java
@@ -88,7 +88,14 @@ public class DataVectorizer {
                     + (position + 1));
           }
 
-          vectorizeTexts.add(jsonNode.asText());
+          String vectorizeData = jsonNode.asText();
+          if (vectorizeData.isEmpty()) {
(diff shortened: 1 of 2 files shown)


Hazel-Datastax (approved):

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/DataVectorizer.java
```diff
@@ -88,7 +88,14 @@ public Uni<Boolean> vectorize(List<JsonNode> documents) {
                     + (position + 1));
           }
 
-          vectorizeTexts.add(jsonNode.asText());
+          String vectorizeData = jsonNode.asText();
+          if (vectorizeData.isEmpty()) {
```
tatu-at-datastax, line 92: Too bad I missed this. I think we may want to further improve on this:

1. `isBlank()` would cover white space - only cases
2. Since punctuation-only has likely same problem, it'd make sense to use regexp to look for alpha-numeric character (like maybe `"\\d"`) for non-empty.

But I guess it is fine and we can iterate on this.

PR #1112, opened 2024-05-22
Fix #1107: log unmapped exceptions in addition to mapping

### LINKED ISSUE #1107: Data API exception mapper (`ThrowableToErrorMapper`) does not log error for unmapped server errors (`ErrorCode.SERVER_UNHANDLED_ERROR`)
Currently `ThrowableToErrorMapper` has fallback mapping for unrecognized/unmapped cases:

```
        return ErrorCode.SERVER_UNHANDLED_ERROR
            .toApiException("root cause: (%s) %s", throwable.getClass().getName(), message)
            .getCommandResultError(Response.Status.INTERNAL_SERVER_ERROR);
```

which makes sense. But since this is a failure to map, we want to get rid of all such cases: for this, we'd need to actually log these cases.
So let's add log statement.

tatu-at-datastax:
**What this PR does**:

Adds error logic for unmapped exceptions by Data API.

Fixes #1107

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/exception/mappers/ThrowableToErrorMapper.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/exception/mappers/ThrowableToErrorMapper.java
@@ -17,12 +17,16 @@ import java.util.List;
 import java.util.Map;
 import java.util.function.BiFunction;
 import java.util.function.Function;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * Simple mapper for mapping {@link Throwable}s to {@link CommandResult.Error}, with a default
  * implementation.
  */
 public final class ThrowableToErrorMapper {
+  private static final Logger logger = LoggerFactory.getLogger(ThrowableToErrorMapper.class);
+
   private static final BiFunction<Throwable, String, CommandResult.Error> MAPPER_WITH_MESSAGE =
       (throwable, message) -> {
         // if our own exception, shortcut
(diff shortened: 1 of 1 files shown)


src/main/java/io/stargate/sgv2/jsonapi/exception/mappers/ThrowableToErrorMapper.java
```diff
@@ -46,6 +50,10 @@ public final class ThrowableToErrorMapper {
         }
 
         // handle all other exceptions
+        logger.error(
+            String.format(
+                "Unrecognized exception caught, mapped to SERVER_UNHANDLED_ERROR: %s", message),
+            throwable);
```
tatu-at-datastax, line 56: Will log, including stack trace, when `Throwable` passed.

Yuqi-Du (approved):

src/main/java/io/stargate/sgv2/jsonapi/exception/mappers/ThrowableToErrorMapper.java
jeffreyscarpenter: possible for this to be tied to tenant ID?

tatu-at-datastax: Will try to see: cannot pass directly (one of things I dislike about Dependency Injection style), but perhaps there's a way with context (MDC?)

tatu-at-datastax: Looks like we do set `tenantId` entry of MDC in `MeteredCommandProcessor`. And assuming the new logger uses default formatting, and that is includes contents of MDC under `"mdc"` key (which I _think_) happens, I think it should be automatically included.

PR #1117, opened 2024-05-24
Create collection mixed casing fix

### LINKED ISSUE #1110: Index creation needs to be wrapped with quote
`"CREATE CUSTOM INDEX IF NOT EXISTS %s_query_bool_values ON \"%s\".\"%s\" (entries(query_bool_values)) USING 'StorageAttachedIndex'";`
%s_query_bool_values does not have quoto, which means c* will treat it as case-insensitive. 
It will be a problem when people have collections with same name but different cases. 
For example, openaitable and openAItable.

maheshrajamani:
**What this PR does**:
Create collection mixed casing fix

Fixes #1110 

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/CreateCollectionOperation.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/CreateCollectionOperation.java
@@ -174,7 +174,7 @@ public record CreateCollectionOperation(
                 res -> {
                   if (res.wasApplied()) {
                     final List<SimpleStatement> indexStatements =
-                        getIndexStatements(commandContext.namespace(), name);
+                        getIndexStatements(commandContext.namespace(), name, collectionExisted);
                     Multi<AsyncResultSet> indexResultMulti;
                     /*
                     CI will override ddlDelayMillis to 0 using `-Dstargate.jsonapi.operations.database-config.ddl-delay-millis=0`
@@ -222,10 +222,11 @@ public record CreateCollectionOperation(
             error ->
                 // InvalidQueryException(DB index limit violation)
                 error instanceof InvalidQueryException
-                    && error
-                        .getMessage()
-                        .matches(
-                            ".*Cannot have more than \\d+ indexes, failed to create index on table.*"))
+                    && (error
+                            .getMessage()
+                            .matches(
+                                ".*Cannot have more than \\d+ indexes, failed to create index on table.*")
+                        || error.getMessage().matches("Index .* already exists")))
         .recoverWithUni(
             error -> {
               // if index creation violates DB index limit and collection not existed before,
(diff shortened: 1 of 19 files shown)


Yuqi-Du:
LGTM

Yuqi-Du (approved):

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/CreateCollectionOperation.java
```diff
@@ -235,12 +236,21 @@ private Uni<Supplier<CommandResult>> executeCollectionCreation(
...
-                  .item(
-                      () ->
-                          ErrorCode.TOO_MANY_INDEXES.toApiException(
-                              "collection \"%s\" creation failed due to index creation failing; need %d indexes to create the collection;",
-                              name, dbLimitsConfig.indexesNeededPerCollection()));
+              if (error.getMessage().matches("Index .* already exists")) {
+                return Uni.createFrom()
+                    .item(
+                        () ->
+                            ErrorCode.INDEXES_CREATION_FAILED.toApiException(
```
amorton, line 243: -1 on this error message, it hides information the API already knows. 

The API knows the reason the index failed to be created was that it already existed, we should say that in the message. In general avoid messages that only say "it did not work" 

Also, the "check schema" part - what should the user do ? Check it for what ? We should give actionable advice.  

Same on line 250 below.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/CreateCollectionOperation.java
```diff
@@ -400,48 +410,60 @@ public SimpleStatement getCreateTable(String keyspace, String table) {
...
-  public List<SimpleStatement> getIndexStatements(String keyspace, String table) {
+  public List<SimpleStatement> getIndexStatements(
+      String keyspace, String table, boolean collectionExisted) {
     List<SimpleStatement> statements = new ArrayList<>(10);
+    String apprender =
+        collectionExisted ? "CREATE CUSTOM INDEX IF NOT EXISTS" : "CREATE CUSTOM INDEX";
     if (!indexingDenyAll()) {
       String existKeys =
-          "CREATE CUSTOM INDEX IF NOT EXISTS %s_exists_keys ON \"%s\".\"%s\" (exist_keys) USING 'StorageAttachedIndex'";
+          apprender
```
amorton, line 420: Add a comment to explain why we are using quoted identifiers, comments like this make it less likely another person will change it later. 

I would also consider a helper function or lambda to quote the identifier rather than have lots of `\"` in a string which can easily become unbalanced. 

Comment also for the create table statement. 

Example "Using CQL quoted identifiers to preserve the case of the identifiers in the CQL schema, and enforce case sensitive comparisons. This is expected behaviour for the Data API."

PR #1122, opened 2024-05-28
[WIP]Micro batching for embedding clients

maheshrajamani:
**What this PR does**:

Fixes #<issue number>

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/embedding/configuration/EmbeddingProviderConfigStore.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/embedding/configuration/EmbeddingProviderConfigStore.java
@@ -31,15 +31,22 @@ public interface EmbeddingProviderConfigStore {
       int retryDelayInMillis,
       int timeoutInMillis,
       Optional<String> requestTypeQuery,
-      Optional<String> requestTypeIndex) {
+      Optional<String> requestTypeIndex,
+      int batchSize) {
     public static RequestProperties of(
         int maxRetries,
         int retryDelayInMillis,
         int timeoutInMillis,
         Optional<String> requestTypeQuery,
-        Optional<String> requestTypeIndex) {
+        Optional<String> requestTypeIndex,
+        int batchSize) {
       return new RequestProperties(
-          maxRetries, retryDelayInMillis, timeoutInMillis, requestTypeQuery, requestTypeIndex);
+          maxRetries,
+          retryDelayInMillis,
+          timeoutInMillis,
+          requestTypeQuery,
+          requestTypeIndex,
+          batchSize);
     }
   }
 
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/embedding/configuration/EmbeddingProvidersConfig.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/embedding/configuration/EmbeddingProvidersConfig.java
@@ -220,6 +220,10 @@ public interface EmbeddingProvidersConfig {
 
       @Nullable
       Optional<String> taskTypeRead();
+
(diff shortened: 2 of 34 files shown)


src/main/java/io/stargate/sgv2/jsonapi/service/embedding/configuration/EmbeddingProviderConfigStore.java
```diff
@@ -31,15 +31,22 @@ record RequestProperties(
...
+      Optional<String> requestTypeIndex,
+      int batchSize) {
     public static RequestProperties of(
         int maxRetries,
         int retryDelayInMillis,
         int timeoutInMillis,
         Optional<String> requestTypeQuery,
-        Optional<String> requestTypeIndex) {
+        Optional<String> requestTypeIndex,
+        int batchSize) {
```
tatu-at-datastax, line 42: Should this be named "maxBatchSize" where it defines maximum batch provider allows?

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/configuration/EmbeddingProvidersConfig.java
```diff
@@ -220,6 +220,10 @@ interface RequestProperties {
 
       @Nullable
       Optional<String> taskTypeRead();
+
+      /** Batch size supported by the provider. */
```
tatu-at-datastax, line 224: Same here: adding "max(imum)" would make sense to me wrt semantics.

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/configuration/EmbeddingProvidersConfig.java
```diff
@@ -220,6 +220,10 @@ interface RequestProperties {
...
+      @WithDefault("2048")
```
tatu-at-datastax, line 225: Do we want to use defaults in this case? I'd guess every provider must have their specific maximum; many have lower than 2048.

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/gateway/EmbeddingGatewayClient.java
```diff
@@ -12,6 +12,7 @@
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import org.apache.commons.lang3.tuple.Pair;
```
tatu-at-datastax, line 15: I am ok wrt using `Pair` in many cases, but for readability it'd be better to use a Record I think: mostly just so that the first number is clearer (Batch id, not batch size).

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/MeteredEmbeddingProvider.java
```diff
@@ -55,18 +61,54 @@ public Uni<List<float[]>> vectorize(
...
+            })
+        .merge()
+        .collect()
+        .asList()
+        .onItem()
+        .transform(
+            vectorizedBatches -> {
+              // sort it by batch id, this is required because the merge() run the calls in parallel
+              Collections.sort(
+                  vectorizedBatches, (a, b) -> Integer.compare(a.getLeft(), b.getLeft()));
```
tatu-at-datastax, line 92: this is (f.ex) where `getLeft()` is confusing & Record would give more context (we are sorting by batch id)

tatu-at-datastax (approved):
Functionally makes sense, looks good. I would prefer naming changes (more importantly actual Record type over `Pair`; less important "maxBatchSize" over "batchSize"), but let you decide if it's worth it.

PR #1129, opened 2024-05-30
Changes for using dse 6.9  for CI

maheshrajamani:
**What this PR does**:
Changes for using dse 6.9 for CI testing profile dse-it

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/src/test/java/io/stargate/sgv2/jsonapi/testresource/StargateTestResource.java
+++ b/src/test/java/io/stargate/sgv2/jsonapi/testresource/StargateTestResource.java
@@ -61,6 +61,7 @@ public abstract class StargateTestResource
       } else {
         propsBuilder = this.startWithoutContainerNetwork(reuse);
       }
+
       if (useCoordinator()) {
         Integer authPort = this.stargateContainer.getMappedPort(8081);
         String token = this.getAuthToken(this.stargateContainer.getHost(), authPort);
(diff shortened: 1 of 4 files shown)


src/test/java/io/stargate/sgv2/jsonapi/testresource/StargateTestResource.java
```diff
@@ -159,25 +160,33 @@ public void stop() {
 
   private GenericContainer<?> baseCassandraContainer(boolean reuse) {
     String image = this.getCassandraImage();
+    String startupMessage =
+        this.isDse() ? ".*DSE startup complete.*\\n" : ".*Created default superuser role.*\\n";
```
jeffreyscarpenter, line 164: I would like to see if we can change this to be based on cqlsh access, as we have done in docker-compose files. We will probably need to come back and do that later as the initial version of the DSE 6.9 container we are currently testing with doesn't have cqlsh working yet.

src/test/resources/dse.yaml
```diff
@@ -0,0 +1,1217 @@
...
+#                authentication attempts will be logged in as the anonymous user
+#   normal     - If credentials are passed they are authenticated. If the
+#                authentication is successful then the user is logged in, otherwise
+#                the user is logged in as anonymous. If no credentials are passed,
+#                then the user is logged in as anonymous
+#   strict     - If credentials are passed they are authenticated. If the
+#                authentication is successful, the user is logged in. If the
+#                authentication fails, an authentication error is returned. If no
+#                credentials are passed, the user is logged in as anonymous
+authentication_options:
```
jeffreyscarpenter, line 66: My guess is that this is the section that we are modifying from the default `dse.yaml`, is that correct? If so, can we put a big block comment in highlighting that? This will make it easier to maintain when we have to update this yaml in the future due to new DSE version.

jeffreyscarpenter (approved):
I have one suggestion for now, and one for a future PR. See my comments

src/test/resources/dse.yaml
maheshrajamani: I took the file from the image and enabled the authentication_option. Not able to find env variable option to enable these. Something to explore with dse team. That way we don't need this yaml override.

src/test/java/io/stargate/sgv2/jsonapi/testresource/StargateTestResource.java
maheshrajamani: @jeffreyscarpenter Have to check how it can be done. Can that be done as separate PR?

PR #1139, opened 2024-06-03
Fix null header exception from embedding provider's response

### LINKED ISSUE #1131: Jina AI timeouts when model is cold
**Jina embedding errors**
From Stefano reported,
```
the HEADER method for the jinaAI model jina-embeddings-v2-base-de consistently fails with
Cannot invoke "jakarta.ws.rs.core.MultivaluedMap.getFirst(Object)" because "headers" is null
This happens only with the de model: all other models, passing the (header) secret in the very same way, work just fine
```
From Cedrick reported,
`jina => error header but only for service zh though`

**Cohere, voyage rate limited**

Hazel-Datastax:
**What this PR does**:
Null header response was first found in Jina models. They normally take a longer time (~15s) for the first vectorize call, probably due to model initialization. This cause a timeout exception in Quarkus (`NoStackTraceTimeoutException`) with the message `The timeout period of 10000ms has been exceeded while executing POST /v1/embeddings for server api.jina.ai:443` and no response. This PR adds a status checker in the interceptor to handle this scenario.

Fixes one of the issue in #1131 - null header exception 

**Checklist**
- [x] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/service/embedding/configuration/EmbeddingProviderResponseValidation.java
```diff
@@ -31,6 +31,11 @@ public class EmbeddingProviderResponseValidation implements ClientResponseFilter
   @Override
   public void filter(ClientRequestContext requestContext, ClientResponseContext responseContext)
       throws JsonApiException {
+    // Check if the header is null
```
maheshrajamani, line 34: When timed out, is it not throwing any http error code? Should we run this filter only when the response http status code is in 2xx?

tatu-at-datastax: I would think true timeout would throw an exception and not server `null` headers... so it almost seems like an odd error condition by rest client we use.

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/VoyageAIEmbeddingClient.java
```diff
@@ -98,9 +99,10 @@ public Uni<Response> vectorize(
             .embed("Bearer " + apiKeyOverride.get(), request)
             .onFailure(
                 throwable -> {
-                  return (throwable.getCause() != null
-                      && throwable.getCause() instanceof JsonApiException jae
-                      && jae.getErrorCode() == ErrorCode.EMBEDDING_PROVIDER_TIMEOUT);
+                  return ((throwable.getCause() != null
+                          && throwable.getCause() instanceof JsonApiException jae
+                          && jae.getErrorCode() == ErrorCode.EMBEDDING_PROVIDER_TIMEOUT)
+                      || throwable instanceof TimeoutException);
```
Hazel-Datastax, line 105: It's annoying to repeat the changes to every provider, and probably will have another PR to create a "BaseEmbeddingClient"

src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCode.java
```diff
@@ -183,6 +183,7 @@ public enum ErrorCode {
   SERVER_COORDINATOR_FAILURE("Coordinator failed"),
   SERVER_FAILURE("Driver failed"),
   SERVER_UNHANDLED_ERROR("Server failed"),
+  SERVER_EMBEDDING_PROVIDER_TIMEOUT("Embedding provider timed out, please try again later"),
```
maheshrajamani, line 186: I think we should only report it as "Embedding provider timed out"

maheshrajamani (approved):
LGTM., Added minor comments

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/AzureOpenAIEmbeddingClient.java
```diff
@@ -95,9 +96,10 @@ public Uni<Response> vectorize(
             .embed(apiKeyOverride.get(), request)
             .onFailure(
                 throwable -> {
-                  return (throwable.getCause() != null
-                      && throwable.getCause() instanceof JsonApiException jae
-                      && jae.getErrorCode() == ErrorCode.EMBEDDING_PROVIDER_TIMEOUT);
+                  return ((throwable.getCause() != null
```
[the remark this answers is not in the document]
Hazel-Datastax: I want to do the refactor as a whole in the next PR and will move it to util method or "BaseEmbeddingClient"

PR #1146, opened 2024-06-04
Add more request timeout and retry delay for Jina

Hazel-Datastax:
**What this PR does**:
This is a continued PR for #1139, add more time for timeout and retry delay for a potential long response from Jina

Fixes one of the issue in https://github.com/stargate/data-api/issues/1131 - Jina null header exception

**Checklist**
- [x] Changes manually tested


src/main/resources/embedding-providers-config.yaml
```diff
@@ -239,6 +239,8 @@ stargate:
                   forwarded: Authorization
           properties:
             max-batch-size: 32
+            request-timeout-millis: 15000
```
tatu-at-datastax, line 242: Maybe add a comment to reference the Github issue we have for this change, as background for increasing default timeout?

tatu-at-datastax (approved):
LGTM

src/main/resources/embedding-providers-config.yaml
Hazel-Datastax: Since we got another similar error report from OpenAI, @maheshrajamani suggests changing the overall default value. What do you think?

Hazel-Datastax:
After discussing with the team, we want to make more changes to the config and new changes are in #1154

PR #1156, opened 2024-06-10
Fix #1038: improve error message for unknown dollar-starting field

### LINKED ISSUE #1038: Misleading error message about projection having $fields
Considering that

```# These projections:
    projection={"$uuid": True}
    projection={"$objectId": True}
    projection={"$similarity": True}
    projection={"$argyroneta": True}  # i.e. arbitrary $something...

# yield the error:
    {
        "errors": [
            {
                "errorCode": "UNSUPPORTED_PROJECTION_PARAM",
                "message": "Unsupported projection parameter: path cannot start with '$' (no root-level operators)"
            }
        ]
    }
```

but

```
# These work as intended:
    projection={"$vector": True}
    projection={"$vectorize": True}
```

my point is that the text of the error message is a bit misleading. I suggest it should say something to the effect of,

> Unsupported projection parameter: `$vector[/ize]` are the only allowed path that can start with `$`

or something similar to convey the idea that yes, actual fields from the document which start with $ are projectable.

====

[Tatu]:

Agreed... it is bit tricky because `$` is overloaded for us now in some ways: usually it means operator. But sometimes like as with Projection there are "special" fields.

but I agree that error is misleading. So [...] **we can change wording, leaving error code as is**

([source](https://datastax.slack.com/archives/C05PTSGBYL9/p1713444202915699))

tatu-at-datastax:
**What this PR does**:

Improves failure message for case where user tries to project on key starting with "$" (other than $vector or $vectorize), to indicate accepted keys (instead of referring to operators)

Fixes #1038

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/projection/DocumentProjector.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/projection/DocumentProjector.java
@@ -294,7 +294,7 @@ public class DocumentProjector {
             throw new JsonApiException(
                 ErrorCode.UNSUPPORTED_PROJECTION_PARAM,
                 ErrorCode.UNSUPPORTED_PROJECTION_PARAM.getMessage()
-                    + ": path cannot start with '$' (no root-level operators)");
+                    + ": '$vector'/'$vectorize' are the only allowed paths that can start with '$'");
           }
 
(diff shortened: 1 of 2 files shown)


Yuqi-Du (approved):
LGTM!

PR #1157, opened 2024-06-10
Add huggingface dedicated provider support

Yuqi-Du:
since provider huggingfaceDedicated does not require model, modelName field in createCollectionCommand will be changed as allow null. Other providers still need model specified, so validations move down to createCollectionCommandResolver.

For huggingFace dedicated, the endpoint we use is: 
- url: https://{endpointName}.{regionName}.{cloudName}.endpoints.huggingface.cloud/embeddings
- Note, this is an openAI compatible endpoint
- https://huggingface.github.io/text-embeddings-inference/#/Text%20Embeddings%20Inference/embed

We do not require modelName, but dimension is still needed, and will error out is not provided.
Also, parameters: endpointName, regionName, cloudName are needed to compute the url.

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/CreateCollectionCommand.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/CreateCollectionCommand.java
@@ -122,8 +122,7 @@ public record CreateCollectionCommand(
                   implementation = String.class)
               @JsonProperty("provider")
               String provider,
-          @NotNull
-              @Schema(
+          @Schema(
                   description = "Registered Embedding service model",
                   type = SchemaType.STRING,
                   implementation = String.class)
--- a/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/CreateCollectionCommand.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/CreateCollectionCommand.java
@@ -7,6 +7,7 @@ import com.fasterxml.jackson.annotation.JsonTypeName;
 import io.stargate.sgv2.jsonapi.api.model.command.NamespaceCommand;
 import io.stargate.sgv2.jsonapi.config.constants.DocumentConstants;
 import io.stargate.sgv2.jsonapi.exception.ErrorCode;
+import io.stargate.sgv2.jsonapi.service.embedding.configuration.ProviderConstants;
 import jakarta.validation.Valid;
 import jakarta.validation.constraints.*;
 import java.util.HashSet;
@@ -143,7 +144,25 @@ public record CreateCollectionCommand(
                   type = SchemaType.OBJECT)
               @JsonProperty("parameters")
               @JsonInclude(JsonInclude.Include.NON_NULL)
-              Map<String, Object> parameters) {}
+              Map<String, Object> parameters) {
+
+        public VectorizeConfig(
+            String provider,
+            String modelName,
+            Map<String, String> authentication,
+            Map<String, Object> parameters) {
+          this.provider = provider;
+          // HuggingfaceDedicated does not need user to specify model
+          // use endpoint-defined-model as placeholder
+          if (provider.equals(ProviderConstants.HUGGINGFACE_DEDICATED)) {
+            this.modelName = "endpoint-defined-model";
+          } else {
+            this.modelName = modelName;
+          }
+          this.authentication = authentication;
+          this.parameters = parameters;
+        }
+      }
     }
 
     public record IndexingConfig(
--- a/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/CreateCollectionCommand.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/CreateCollectionCommand.java
@@ -155,7 +155,7 @@ public record CreateCollectionCommand(
           // HuggingfaceDedicated does not need user to specify model
           // use endpoint-defined-model as placeholder
           if (provider.equals(ProviderConstants.HUGGINGFACE_DEDICATED)) {
-            this.modelName = "endpoint-defined-model";
+            this.modelName = ProviderConstants.HUGGINGFACE_DEDICATED_DEFINED_MODEL;
           } else {
             this.modelName = modelName;
           }
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/resolver/model/impl/CreateCollectionCommandResolver.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/resolver/model/impl/CreateCollectionCommandResolver.java
@@ -259,8 +259,10 @@ public class CreateCollectionCommandResolver implements CommandResolver<CreateCo
     // Check secret name for shared secret authentication, if applicable
     validateAuthentication(userConfig, providerConfig);
 
-    // Validate the model and its vector dimension, if userVectorDimension is null, return value
-    // will be the config/default value
+    // Validate the model and its vector dimension:
+    //   huggingFaceDedicated: must have vectorDimension specified
+    //   other providers: must have model specified, and default dimension when dimension not
+    // specified
     Integer vectorDimension =
         validateModelAndDimension(userConfig, providerConfig, userVectorDimension);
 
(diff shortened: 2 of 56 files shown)


src/main/java/io/stargate/sgv2/jsonapi/service/resolver/model/impl/CreateCollectionCommandResolver.java
```diff
@@ -419,7 +422,7 @@ private void validateUserParameters(
             userParamName -> {
               if (!expectedParamNames.contains(userParamName)) {
                 throw ErrorCode.INVALID_CREATE_COLLECTION_OPTIONS.toApiException(
-                    "Unexpected parameter '%s' for the provider '%s' provided",
+                    "Unexpected parameter '%s' for the provider '%s' provider",
```
tatu-at-datastax, line 425: I think the original was correct, change is wrong :)

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/model/impl/CreateCollectionCommandResolver.java
```diff
@@ -497,6 +500,26 @@ private Integer validateModelAndDimension(
       EmbeddingProvidersConfig.EmbeddingProviderConfig providerConfig,
       Integer userVectorDimension) {
     // Find the model configuration by matching the model name
+    // 1. huggingfaceDedicated does not require model, but requires dimension
```
tatu-at-datastax, line 503: For missing model: instead of adding hard-coded logic, would it be better to allow missing model if configuration specifies no models (which is the case for HuggingFace Dedicated)?
That way it is more configuration driven.

Similarly would be good to make "dimensions required" a property of configuration setting, but maybe that is more difficult do do.

src/main/resources/embedding-providers-config.yaml
```diff
@@ -128,7 +128,47 @@ stargate:
...
+                - accepted: x-embedding-api-key
+                  forwarded: Authorization
+            SHARED_SECRET:
+              enabled: true
+              tokens:
+                - accepted: providerKey
+                  forwarded: Authorization
+          properties:
+            max-batch-size: 32
+          models:
```
tatu-at-datastax, line 151: Since we have no models, this could be used (as I suggested earlier) as indication that caller need not (... since they can not) provide a valid model id. And UI should also not ask for model given there is none to choose from.

src/main/resources/embedding-providers-config.yaml
```diff
@@ -128,7 +128,47 @@ stargate:
...
+              required: true
+              help: "The region your Hugging Face dedicated endpoint is deployed to, the second part of the Endpoint URL."
+            - name: "cloudName"
+              type: string
+              required: true
+              help: "The cloud your Hugging Face dedicated endpoint is deployed to, the third part of the Endpoint URL."
+            # This parameter helps to validate user dimension input
+            - name: vectorDimension
+              type: number
+              required: false
```
tatu-at-datastax, line 168: But isn't this specifically required to be given by user?

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/model/impl/CreateCollectionCommandResolver.java
Yuqi-Du: oh, my bad

src/main/resources/embedding-providers-config.yaml
Yuqi-Du: yes, that is the agreement with UI.
`The Portal should not ask for the model name, the API will return an empty list of options.`

src/main/resources/embedding-providers-config.yaml
Yuqi-Du: Aaron writes the yaml 
required as true. I changed it as false, because:
this is in section of parameters, user actually specifies dimension in vector level.
And we validates all parameters, if anything that is required, but missing, will error out, that is why I set it as false.
```
{
    "createCollection": {
        "name": "hf3",
        "options": {
            "vector": {
                "metric": "cosine",
                "dimension": 384,
                "service": {
                    "provider": "huggingfaceDedicated",
                      "parameters": {
                        "endpointName": "uk6hf0dyowqozmgc",
                        "regionName": "us-east-1",
                        "cloudName": "aws"
                    }
                } // end of service
            } // end of vector
        }
    }
}
```

amorton: dimensions shoudl be required because we need it to create the table

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/CreateCollectionCommand.java
```diff
@@ -144,7 +144,25 @@ public record VectorizeConfig(
...
+        public VectorizeConfig(
+            String provider,
+            String modelName,
+            Map<String, String> authentication,
+            Map<String, Object> parameters) {
+          this.provider = provider;
+          // HuggingfaceDedicated does not need user to specify model
+          // use endpoint-defined-model as placeholder
+          if (provider.equals(ProviderConstants.HUGGINGFACE_DEDICATED)) {
+            this.modelName = "endpoint-defined-model";
```
tatu-at-datastax, line 158: Should this be extracted into constant in provider constants as well?

Yuqi-Du: good point!

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/model/impl/CreateCollectionCommandResolver.java
```diff
@@ -497,6 +498,18 @@ private Integer validateModelAndDimension(
...
+    if (userConfig.provider().equals(ProviderConstants.HUGGINGFACE_DEDICATED)) {
+      if (userVectorDimension == null) {
+        throw ErrorCode.INVALID_CREATE_COLLECTION_OPTIONS.toApiException(
```
tatu-at-datastax, line 504: Use templating, `ProviderConstants.HUGGINGFACE_DEDICATED` for error.
Also quotes around `dimension`?

tatu-at-datastax: Also, is it `'dimension'` or `'vectorDimension'`?

tatu-at-datastax (approved):
LGTM, added minor comments.

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/model/impl/CreateCollectionCommandResolver.java
Yuqi-Du: thanks. fixed
yeah, user input is dimension

a-effort:
Create collection

@Yuqi-Du , here is content for the Hugging Face — Dedicated provider to be pulled in for form fields in the UI for the create-collection flow: 

Label: User access token
Placeholder: Select user access token

Label: Embedding model
Placeholder: Select embedding model
Helper text under form field: Model to use to generate embeddings

Label: Endpoint name
Placeholder: Add endpoint name
Helper text under form field: Add the first part of the dedicated endpoint URL

Label: Region name
Placeholder: Add region name
Helper text under form field: Add the second part of the dedicated endpoint URL

Label: Cloud provider the dedicated endpoint is deployed to
Placeholder: Add cloud name
Helper text under form field: Add the third part of the dedicated endpoint URL

Label: Dimensions
Placeholder: Add a number between 2 and 3072
Helper text under form field: Vector length in your dataset (between 2 and 3072)

a-effort:
@Yuqi-Du , here is content for the form fields for Hugging Face — Dedicated create collection: 

Label: User access token
Placeholder: Select user access token

Label: Embedding model
Placeholder: Select embedding model
Helper text under form field: Model to use to generate embeddings

Label: Endpoint name
Placeholder: Add endpoint name
Helper text under form field: Add the first part of the dedicated endpoint URL

Label: Region name
Placeholder: Add region name
Helper text under form field: Add the second part of the dedicated endpoint URL

Label: Cloud provider the dedicated endpoint is deployed to
Placeholder: Add cloud name
Helper text under form field: Add the third part of the dedicated endpoint URL

Label: Dimensions
Placeholder: Add a number between 2 and 3072
Helper text under form field: Vector length in your dataset (between 2 and 3072)

HuggingFace - dedicated

PR #1159, opened 2024-06-11
Improve Error Messages from Embedding Providers

Hazel-Datastax:
**What this PR does**:
Improve Error Messages from Embedding Providers

Fixes [Jira C2-3551](https://datastax.jira.com/browse/C2-3551)

**Checklist**
- [x] Changes manually tested


Hazel-Datastax:
Will modify and add ITs in EGW repo

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/JinaAIEmbeddingClient.java
```diff
@@ -54,7 +57,53 @@ Uni<EmbeddingResponse> embed(
...
+      try {
+        ObjectMapper mapper = new ObjectMapper();
+        JsonNode rootNode = mapper.readTree(responseBody);
+        // Extract the "detail" node
+        JsonNode detailNode = rootNode.path("detail");
+        // Assuming the error message is within the "msg" part of the detail
+        if (!detailNode.isMissingNode()) {
+          String detailText = detailNode.asText();
+          // Use regex to extract the message part from the detail text
+          java.util.regex.Pattern pattern = java.util.regex.Pattern.compile("'msg':\\s*'([^']*)'");
```
vkarpov15, line 92: Minor detail but is it possible that the `'msg': '...'` will contain an escaped single quote like `'msg': 'Single text can\'t exceed 8192 tokens'`? If so, likely worthwhile to handle that here, but I haven't taken a look at Jina AI's docs so I wouldn't know.

vkarpov15 (commented):
One minor comment but looks good otherwise

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/AzureOpenAIEmbeddingClient.java
```diff
@@ -67,7 +70,40 @@ Uni<EmbeddingResponse> embed(
...
+     * }
+     * </pre>
+     *
+     * @param response The response body as a String.
+     * @return The error message extracted from the response body.
+     */
+    private static String getErrorMessage(jakarta.ws.rs.core.Response response) {
+      String responseBody = response.readEntity(String.class);
+      try {
+        ObjectMapper mapper = new ObjectMapper();
```
tatu-at-datastax, line 96: This should be statically created instance: constructing new `ObjectMapper` is very expensive so we do not want to create one per each error, throw it away.
So having `private static ObjectMapper MAPPER = new ObjectMapper();` is much preferred.

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/CohereEmbeddingClient.java
```diff
@@ -63,7 +66,50 @@ Uni<EmbeddingResponse> embed(
...
+     * </pre>
+     *
+     * @param response The response body as a String.
+     * @return The error message extracted from the response body.
+     */
+    private static String getErrorMessage(jakarta.ws.rs.core.Response response) {
+      String responseBody = response.readEntity(String.class);
+      try {
+        ObjectMapper mapper = new ObjectMapper();
+        JsonNode rootNode = mapper.readTree(responseBody);
```
tatu-at-datastax, line 95: If all/most provider client require `ObjectMapper`, should probably have `protected static` instance in base class.

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/JinaAIEmbeddingClient.java
tatu-at-datastax, line 92: Aside from regexp usage, it's also more efficient to create and compile `Pattern` just once (and keep in a static variable), instead of create and compile every time it is needed: building state machine (DFA?) is heavy-weight operation and since it never changes there's no point in re-creating it every time.

Hazel-Datastax: Aaron pointed it out as well. After discussing it with him, he suggested removing the regex. It's hard to handle the problem Val mentioned above, and it's fine to return the whole body `ValidationError(model='TextDoc', errors=[{'loc': ('text',), 'msg': 'Single text cannot exceed 8192 tokens. 10454 tokens given.', 'type': 'value_error'}])`, even though it looks wired.

tatu-at-datastax: Yes that makes sense; matching error messages is fragile as they can change more easily than error ids.

src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCode.java
```diff
@@ -27,20 +27,11 @@ public enum ErrorCode {
...
-
-  EMBEDDING_PROVIDER_INVALID_REQUEST(
-      "The configured Embedding Provider for this collection refused to process the request, response was"),
-
-  EMBEDDING_PROVIDER_INVALID_RESPONSE(
-      "The configured Embedding Provider for this collection return an invalid response"),
+  EMBEDDING_PROVIDER_CLIENT_ERROR("The Embedding Provider returned a HTTP client error"),
+  EMBEDDING_PROVIDER_SERVER_ERROR("The Embedding Provider returned a HTTP server error"),
+  EMBEDDING_PROVIDER_RATE_LIMITED("The Embedding Provider rate limited the request"),
+  EMBEDDING_PROVIDER_TIMEOUT("The configured Embedding Provider timed out"),
```
tatu-at-datastax, line 33: Maybe remove "configured" here too since others don't have it

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/configuration/EmbeddingProviderResponseValidation.java
```diff
@@ -36,19 +45,35 @@ public void filter(ClientRequestContext requestContext, ClientResponseContext re
     if (responseContext.getStatus() == 0) {
       return;
     }
+    // log the whole error body here
+    String responseBody = null;
```
tatu-at-datastax, line 49: Is this called only for error responses? From name it'd sound it was called for every response, error and non-error: if so, we should only collect and log response in fail case (4xx or 5xx response).

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/configuration/EmbeddingProviderResponseValidation.java
```diff
@@ -36,19 +45,35 @@ public void filter(ClientRequestContext requestContext, ClientResponseContext re
...
+    try {
+      // Read the entity stream into a byte array
+      InputStream originalEntityStream = responseContext.getEntityStream();
+      ByteArrayOutputStream baos = new ByteArrayOutputStream();
+      originalEntityStream.transferTo(baos);
+      responseBody = baos.toString();
```
tatu-at-datastax, line 55: Should specify encoding (StandardCharsets.UTF8).

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/configuration/EmbeddingProviderResponseValidation.java
Hazel-Datastax: Yes, I was moving too fast and forgot this

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/AzureOpenAIEmbeddingClient.java
```diff
@@ -67,7 +70,41 @@ Uni<EmbeddingResponse> embed(
...
+     *   }
+     * }
+     * </pre>
+     *
+     * @param response The response body as a String.
+     * @return The error message extracted from the response body.
+     */
+    private static String getErrorMessage(jakarta.ws.rs.core.Response response) {
+      // Get the whole response body
+      String responseBody = response.readEntity(String.class);
```
tatu-at-datastax, line 95: Would it not work to directly use

    JsonNode rootNode = response.readEntity(JsonNode.class)

?

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/AzureOpenAIEmbeddingClient.java
```diff
@@ -67,7 +70,41 @@ Uni<EmbeddingResponse> embed(
...
+      JsonNode rootNode;
+      try {
+        rootNode = OBJECT_MAPPER.readTree(responseBody);
+      } catch (JsonProcessingException e) {
+        // should not go here, already check json format in EmbeddingProviderResponseValidation.
+        // if it happens, return the whole response body
+        return responseBody;
+      }
+      // Extract the "message" node from the "error" node
+      JsonNode messageNode = rootNode.path("error").path("message");
```
tatu-at-datastax, line 105: This works; but can also use JSON Pointer:

JsonNode messageNode = rootNode.at("/error/message");

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/AzureOpenAIEmbeddingClient.java
```diff
@@ -67,7 +70,41 @@ Uni<EmbeddingResponse> embed(
...
+      // Return the text of the "message" node, or the whole response body if it is missing
+      return messageNode.isMissingNode() ? responseBody : messageNode.asText();
```
tatu-at-datastax, line 107: Good check with `isMissingNode()`!

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/UpstageAIEmbeddingClient.java
```diff
@@ -59,7 +62,59 @@ Uni<EmbeddingResponse> embed(
...
+      // Check if the root node contains a "message" field
+      JsonNode messageNode = rootNode.path("message");
+      if (!messageNode.isMissingNode()) {
+        return messageNode.asText();
+      }
+      // If the "message" field is not found, check for the nested "error" object
+      JsonNode errorNode = rootNode.path("error");
+      if (!errorNode.isMissingNode()) {
+        JsonNode errorMessageNode = errorNode.path("message");
+        if (!errorMessageNode.isMissingNode()) {
```
tatu-at-datastax, line 112: In this case use JSON Pointer would simplify code a bit:

JsonNode errorMessageNode = rootNode.at("/error/message");
if (!errorMessageNode.isMissingNode()) {

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/EmbeddingProvider.java
```diff
@@ -9,6 +10,8 @@
  * of chosen model.
  */
 public interface EmbeddingProvider {
```
tatu-at-datastax, line 12: This may be better idea for follow-up PR, not here, but I think that instead of interface, we should make this an abstract class that takes `String providerId`, defines abstract method `String getErrorMessage(response)` and can then implement `mapException()` in this class.
This would reduce copied code quite bit.

But need not be done within this PR, just mentioning it (I could do PR once this is merged).

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/error/HttpResponseErrorMessageMapper.java
```diff
@@ -1,47 +1,54 @@
...
       return ErrorCode.EMBEDDING_PROVIDER_SERVER_ERROR.toApiException(
-          "Error Code : %s response description : %s",
-          response.getStatus(), response.getStatusInfo().getReasonPhrase());
+          "Provider: %s; HTTP Status: %s; Error Message: %s",
+          providerName, response.getStatus(), message);
     }
 
     // All other errors, Should never happen as all errors are covered above
-    return ErrorCode.EMBEDDING_PROVIDER_UNAVAILBLE.toApiException();
+    return ErrorCode.EMBEDDING_PROVIDER_UNEXPECTED_RESPONSE.toApiException(
```
tatu-at-datastax, line 50: Good! I like changes in this class to unify formatting, information.

tatu-at-datastax (commented):
Looks good: added some comments/suggestions.

But I do have one question about `EmbeddingProviderResponseValidation`, if method is called for all response, or just errors? Handling should really only be applied to errors, and it looked (although I might be wrong) to be applied to all responses.

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/EmbeddingProvider.java
Hazel-Datastax: Created a new issue about this: https://github.com/stargate/data-api/issues/1173

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/configuration/EmbeddingProviderResponseValidation.java
```diff
@@ -21,6 +29,8 @@ public class EmbeddingProviderResponseValidation implements ClientResponseFilter
 
   static final MediaType MEDIATYPE_TEXT_JSON = new MediaType("text", "json");
 
+  Logger logger = LoggerFactory.getLogger(EmbeddingProvider.class);
```
tatu-at-datastax, line 32: Usually should be the class itself, here `EmbeddingProviderResponseValidation.class`

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/configuration/EmbeddingProviderResponseValidation.java
```diff
@@ -36,20 +46,33 @@ public void filter(ClientRequestContext requestContext, ClientResponseContext re
...
+    if (!responseContext.hasEntity()) {
+      throw EMBEDDING_PROVIDER_UNEXPECTED_RESPONSE.toApiException(
+          "No response body from the embedding provider");
+    }
+
     // Check the Content-Type of the response
     MediaType contentType = responseContext.getMediaType();
     if (contentType == null
         || !(MediaType.APPLICATION_JSON_TYPE.isCompatible(contentType)
             || MEDIATYPE_TEXT_JSON.isCompatible(contentType))) {
```
tatu-at-datastax, line 59: Just to make sure I understand: is response only read in case it is not declared as JSON?
If so, makes sense. But let's change above "Check the Content-Type of the response" to something that indicates what and why is done here.
(basically: response should always be JSON; if not, error out, include raw response message for trouble-shooting purposes)

Hazel-Datastax: Yes, correct. Will update the comments

tatu-at-datastax (approved):
Looks good, only minor additional suggestions.

PR #1169, opened 2024-06-14
Vectorize as needed (findOneAndUpdate, updateOne, updateMany, findOneAndReplace)

Yuqi-Du:
For these update commands: findOneAndUpdate, updateOne, updateMany, findOneAndReplace
We need to vectorize the update clause or replacement document as needed, that is when:
- we have document found by find operation. (For update, only vectorize when there is a diff)
OR
- upsert, will vectorize no matter what

Straightforwardly, we can not continue to do vectorization for these four commands at commandResolver level.
Since the pre-requisite is findOperation's execution. 
Postpone the vectorization in ReadAndUpdateOperation, utilize the documentUpdater, only vectorize as needed.

**Checklist**
- [x] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/service/updater/DocumentUpdater.java
```diff
@@ -105,9 +145,32 @@ private boolean replace(ObjectNode docToUpdate, boolean docInserted) {
...
 
+  /**
+   * vectorize replacementDocument as needed, only when document is found will be used by
+   * findOneAndReplace
+   *
+   * @param dataVectorizer
+   */
+  public void vectorizeTheReplacementDocument(DataVectorizer dataVectorizer) {
+    try {
+      dataVectorizer
```
maheshrajamani, line 156: Same code as above method can be merge the methods?

src/main/java/io/stargate/sgv2/jsonapi/service/updater/DocumentUpdater.java
```diff
@@ -70,6 +87,29 @@ private boolean update(ObjectNode docToUpdate, boolean docInserted) {
     return modified;
   }
 
+  /**
+   * vectorize UpdateClause as needed, only when documents are found or upsert will be used by
```
maheshrajamani, line 91: The request was not vectorize only if document is found and `$vectorize` field value is different.

maheshrajamani (changes requested):
I am not OK with this change because it does vectorize call in blocking mode in operation class. It will create threading issues. Can this be discussed in our team meeting? After the last meeting we decided @aaron will do a write up on the approach. Only thing we decided to do was to de-dupe text in case of `InsertManyCommand`.

src/main/java/io/stargate/sgv2/jsonapi/service/updater/DocumentUpdater.java
```diff
@@ -70,6 +87,29 @@ private boolean update(ObjectNode docToUpdate, boolean docInserted) {
...
+          .runSubscriptionOn(Infrastructure.getDefaultWorkerPool())
+          .subscribeAsCompletionStage()
+          .get();
+    } catch (Exception e) {
+      if (e instanceof ExecutionException exception) {
+        if (exception.getCause() instanceof JsonApiException jsonApiException) {
+          throw jsonApiException;
+        }
+      }
+      throw new RuntimeException(e);
```
tatu-at-datastax, line 109: minor: should not wrap if already a `RuntimeException`.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/VectorizeSearchIntegrationTest.java
```diff
@@ -872,7 +872,61 @@ public void findOneAndUpdate() {
...
     @Order(2)
-    public void updateOne() {
+    public void findOneAndUpdate_updateClause() {
+      insertVectorDocuments();
+      String json =
+          """
+                  {
+                    "findOneAndUpdate": {
+                      "sort" : {"$vectorize" : "A deep learning display that controls your mood"},
+                      "update" : {"$set" : {"status" : "active","$vectorize":"An AI quilt to help you sleep forever"}},
```
vkarpov15, line 882: We should probably also test `$vectorize` being set to the exact same value. Is there a good way to test that vectorize is not called in that case?

src/main/java/io/stargate/sgv2/jsonapi/service/updater/DocumentUpdater.java
```diff
@@ -70,6 +87,29 @@ private boolean update(ObjectNode docToUpdate, boolean docInserted) {
...
 
+  /**
+   * vectorize UpdateClause as needed, only when documents are found or upsert will be used by
+   * updateOne, findOneAndUpdate, updateMany
+   *
+   * @param dataVectorizer
+   */
+  public void vectorizeUpdateClause(DataVectorizer dataVectorizer) {
+    try {
+      dataVectorizer
```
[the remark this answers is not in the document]
Yuqi-Du: Thanks, changed to uni

src/main/java/io/stargate/sgv2/jsonapi/service/updater/DocumentUpdater.java
Yuqi-Du: method body is different though, leave as it since it is more clearer

src/main/java/io/stargate/sgv2/jsonapi/service/updater/DocumentUpdater.java
Yuqi-Du: Added logic in documentUpdater,
Check if there is any $vectorize diff If there are docs found to update or doc to replace, then this is a necessary condition to proceed vectorization

src/main/java/io/stargate/sgv2/jsonapi/service/updater/DocumentUpdater.java
Yuqi-Du: deleted, thanks

src/test/java/io/stargate/sgv2/jsonapi/api/v1/VectorizeSearchIntegrationTest.java
Yuqi-Du: Thanks, I have added some tests in [DocumentUpdaterTest.java](https://github.com/stargate/data-api/pull/1169/files#diff-504d091a2e59c884abcfc39b63f95a7dd5e223348cc06369e6cd637fafdd9582).
This is to check if there is any diff for $vectorize.

src/main/java/io/stargate/sgv2/jsonapi/service/updater/DocumentUpdater.java
```diff
@@ -23,7 +35,11 @@ public record DocumentUpdater(
    * @return
    */
   public static DocumentUpdater construct(UpdateClause updateDef) {
-    return new DocumentUpdater(updateDef.buildOperations(), null, null, UpdateType.UPDATE);
+    // try to build operations but do not save the result
+    // this is for validating the UpdateClause, for example, updator path conflict
+    // error out before update operation's execution
+    updateDef.buildOperations();
```
tatu-at-datastax, line 41: Why is the result discarded? Do we build operations twice?

src/main/java/io/stargate/sgv2/jsonapi/service/updater/DocumentUpdater.java
```diff
@@ -62,6 +78,7 @@ public DocumentUpdaterResponse apply(JsonNode readDocument, boolean docInserted)
    */
   private boolean update(ObjectNode docToUpdate, boolean docInserted) {
     boolean modified = false;
+    List<UpdateOperation> updateOperations = updateClause.buildOperations();
```
tatu-at-datastax, line 81: As per earlier comment, should just pass `updateClause.buildOperations()` by caller, since it has been done earlier (it's not a cheap thing to do)

src/main/java/io/stargate/sgv2/jsonapi/service/updater/DocumentUpdater.java
```diff
@@ -105,9 +132,65 @@ private boolean replace(ObjectNode docToUpdate, boolean docInserted) {
     return true;
   }
 
+  /**
+   * vectorize replacementDocument as needed, only when document is found will be used by
+   * findOneAndReplace
+   *
+   * @param dataVectorizer
+   */
+  public Uni<Boolean> vectorizeTheReplacementDocument(DataVectorizer dataVectorizer) {
```
tatu-at-datastax, line 141: nit: we don't usually use articles in method names, so `vectorizeReplacementDocument()`.

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/DataVectorizerService.java
```diff
@@ -47,6 +45,17 @@ public DataVectorizerService(
    */
   public Uni<Command> vectorize(
       DataApiRequestInfo dataApiRequestInfo, CommandContext commandContext, Command command) {
+    final DataVectorizer dataVectorizer =
+        constructDataVectorizer(dataApiRequestInfo, commandContext);
+    return vectorizeSortClause(dataVectorizer, commandContext, command)
+        .onItem()
+        .transformToUni(flag -> vectorizeDocument(dataVectorizer, commandContext, command))
```
amorton, line 52: Why is the flag value ignored ? Write a comment or refactor to use the value

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/ReadAndUpdateOperation.java
```diff
@@ -64,10 +68,39 @@ public Uni<Supplier<CommandResult>> execute(
...
     return docsToUpdate
         .onItem()
-        .transformToMulti(
+        .transformToUni(
             findResponse -> {
               pageStateReference.set(findResponse.pageState());
               final List<ReadDocument> docs = findResponse.docs();
+
+              // vectorize the updateClause or ReplacementDocument as needed
+              Uni<Boolean> vectorization = Uni.createFrom().item(false);
```
amorton, line 77: comment to explain why false is passed here

src/main/java/io/stargate/sgv2/jsonapi/service/updater/DocumentUpdater.java
```diff
@@ -105,9 +132,65 @@ private boolean replace(ObjectNode docToUpdate, boolean docInserted) {
...
+      // extract $vectorize if updateClause set operator has it
+      final ObjectNode setNode = updateClause.updateOperationDefs().get(UpdateOperator.SET);
+      if (setNode != null) {
+        final JsonNode updateClauseVectorizeTextJsonNode =
+            setNode.get(DocumentConstants.Fields.VECTOR_EMBEDDING_TEXT_FIELD);
+        if (updateClauseVectorizeTextJsonNode != null) {
+          vectorizeTextUpdate = updateClauseVectorizeTextJsonNode.asText();
+        }
+      }
+    } else if (updateType().equals(DocumentUpdater.UpdateType.REPLACE)) {
```
tatu-at-datastax, line 163: I wonder if we should always just Vectorize if the whole doc is replaced...

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/ReadAndUpdateOperation.java
```diff
@@ -64,10 +68,39 @@ public Uni<Supplier<CommandResult>> execute(
...
+              final DataVectorizer dataVectorizer =
+                  dataVectorizerService.constructDataVectorizer(dataApiRequestInfo, commandContext);
+
+              // 1. UpdateCommand(updateOne, findOneAndUpdate, updateMany):
+              if (documentUpdater.updateType() == DocumentUpdater.UpdateType.UPDATE) {
```
amorton, line 82: move this whole test into the documentUpdater, make it handle the empty list, and use isEmpty when testing if a list is empty

e.g. DocumentUpdater.hasVectorizeDiff(docs, bool isUpsert). 

if docs list empty if the matchedCount  is zero do not need to pass, otherwise will need to pass the match count as well

src/main/java/io/stargate/sgv2/jsonapi/service/operation/model/impl/ReadAndUpdateOperation.java
```diff
@@ -64,10 +68,39 @@ public Uni<Supplier<CommandResult>> execute(
...
+                }
+              }
+              return vectorization
+                  .onItem()
+                  .transformToUni(vectorized -> Uni.createFrom().item(docs));
+            })
+        .onItem()
+        .transformToMulti(
+            docs -> {
               if (upsert() && docs.size() == 0 && matchedCount.get() == 0) {
```
amorton, line 104: nit: while here, use list .isEmpty()

src/main/java/io/stargate/sgv2/jsonapi/service/updater/DocumentUpdater.java
```diff
@@ -2,17 +2,29 @@
...
+ *
+ * @param updateClause - vectorize it as needed before building update operations
+ * @param replaceDocument - replaceDocument to replace the one read from DB
+ * @param replaceDocumentId - documentId from replaceDocument
+ * @param updateType - UPDATE/REPLACE
+ */
 public record DocumentUpdater(
-    List<UpdateOperation> updateOperations,
+    // buildOperations will be executed when apply to update
+    UpdateClause updateClause,
```
amorton, line 27: Why does this now get the clause and not the operations ? That feels like a big change

amorton: also, if all the constructions is via static factory methods we should make the ctor private

src/main/java/io/stargate/sgv2/jsonapi/service/updater/DocumentUpdater.java
```diff
@@ -70,6 +87,16 @@ private boolean update(ObjectNode docToUpdate, boolean docInserted) {
...
   }
 
+  /**
+   * vectorize UpdateClause as needed, only when documents are found or upsert will be used by
+   * updateOne, findOneAndUpdate, updateMany
+   *
+   * @param dataVectorizer
+   */
+  public Uni<Boolean> vectorizeUpdateClause(DataVectorizer dataVectorizer) {
+    return dataVectorizer.vectorizeUpdateClause(updateClause);
```
amorton, line 97: Can we go through why the update clause is needed for the here, I would have though the previous and current versions of the document would be the best way to go

src/main/java/io/stargate/sgv2/jsonapi/service/updater/DocumentUpdater.java
```diff
@@ -105,9 +132,65 @@ private boolean replace(ObjectNode docToUpdate, boolean docInserted) {
...
+    return dataVectorizer.vectorize(List.of(replaceDocument));
+  }
+
+  /**
+   * Check if there is any $vectorize diff If there are docs found to update or doc to replace, then
+   * this is a necessary condition to proceed vectorization
+   *
+   * @param foundDocs
+   */
+  public boolean hasVectorizeDiff(List<ReadDocument> foundDocs) {
```
amorton, line 151: per above, some of this logic is also outside in the ReadAndUpdateOpeation , we should encapsulate all the logic in this function

src/main/java/io/stargate/sgv2/jsonapi/service/updater/DocumentUpdater.java
```diff
@@ -105,9 +132,65 @@ private boolean replace(ObjectNode docToUpdate, boolean docInserted) {
...
+      // extract $vectorize if replaceDocument has it
+      final JsonNode replaceDocumentVectorizeTextJsonNode =
+          replaceDocument.get(DocumentConstants.Fields.VECTOR_EMBEDDING_TEXT_FIELD);
+      if (replaceDocumentVectorizeTextJsonNode != null) {
+        vectorizeTextUpdate = replaceDocumentVectorizeTextJsonNode.asText();
+      }
+    }
+
+    // if there is no $vectorize to update or replace, then no diff.
+    if (vectorizeTextUpdate == null) {
```
amorton, line 173: possible bug ? What is the value that is being set is null ?

src/main/java/io/stargate/sgv2/jsonapi/service/updater/DocumentUpdater.java
```diff
@@ -105,9 +132,65 @@ private boolean replace(ObjectNode docToUpdate, boolean docInserted) {
...
+    }
+
+    // iterate foundDocs, see if there is any diff for $vectorize
+    for (ReadDocument foundDoc : foundDocs) {
+      final JsonNode foundDocVectorizeTextJsonNode =
+          foundDoc.document().get(DocumentConstants.Fields.VECTOR_EMBEDDING_TEXT_FIELD);
+      if (foundDocVectorizeTextJsonNode != null) {
+        if (!foundDocVectorizeTextJsonNode.asText().equals(vectorizeTextUpdate)) {
+          // There is a diff
+          return true;
```
amorton, line 184: what if there are 10 docs, and only the first was changed ? 

Also want to check if we can just compare the docs before and after, that feels a safer and easier that extracting something from the update clause

amorton (changes requested):
some changes, we should have a chat to make it easier

src/main/java/io/stargate/sgv2/jsonapi/service/updater/DocumentUpdater.java
Yuqi-Du: Thanks! fixed

Issue #1173, opened 2024-06-17
Change `EmbeddingProvider` from interface to abstract class


Hazel-Datastax:
It's from Tatu's comment in #1159

tatu-at-datastax:
And specifically this is done to allow adding configuration state, shared functionality (interfaces can only have public methods, and no fields)

PR #1189, opened 2024-06-24
Refactor embedding providers classes

### LINKED ISSUE #1152: Refactor Embedding Providers
As mentioned in #1139, we would like to refactor the embedding providers to avoid duplicated codes

Hazel-Datastax:
**What this PR does**:
Refactor embedding providers based class:
1. Change based class `EmbeddingProvider` from `interface` to `abstract`
2. Add duplicated fields and constructor in `EmbeddingProvider`
3. Extract the redundant retry logic in each provider to `applyRetry` method

Fixes #1152, #1173

**Checklist**
- [x] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/AzureOpenAIEmbeddingClient.java
```diff
@@ -32,10 +28,8 @@
  * href="https://learn.microsoft.com/en-us/azure/ai-services/openai/reference">API reference</a> for
  * details of REST API being called.
  */
-public class AzureOpenAIEmbeddingClient implements EmbeddingProvider {
-  private EmbeddingProviderConfigStore.RequestProperties requestProperties;
-  private String modelName;
-  private int dimension;
+public class AzureOpenAIEmbeddingClient extends EmbeddingProvider {
+  private static final String providerId = ProviderConstants.AZURE_OPENAI;
```
Hazel-Datastax, line 32: Got a suggestion to extract `providerId` and `getErrorMessage()` to abstract class `EmbeddingProvider` in previous PR: https://github.com/stargate/data-api/pull/1159#discussion_r1640393432
But found it was hard since both `providerId` and `getErrorMessage()` would be used in the interface `XXXXEmbeddingProvider` in each class and they should be static. What do you think @tatu-at-datastax ?

Hazel-Datastax:
Since we've changed `EmbeddingProvider` to abstract, could I remove `EmbeddingUtil` and have all the methods in `EmbeddingProvider`? @tatu-at-datastax

tatu-at-datastax:
> Since we've changed `EmbeddingProvider` to abstract, could I remove `EmbeddingUtil` and have all the methods in `EmbeddingProvider`? @tatu-at-datastax

That does sound like a good idea.

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/AzureOpenAIEmbeddingClient.java
tatu-at-datastax: Typically sub-class would pass `providerId` constant in constructor to super class; and super class would have a field set, using that for its own processing.
I'll have to see what was the thing with `getErrorMessage()`.

It is also fine for abstract base class to define methods for sub-classes to implement, call those methods for its needs.

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/EmbeddingProvider.java
```diff
@@ -1,17 +1,68 @@
...
+public abstract class EmbeddingProvider {
+  protected static final Logger logger = LoggerFactory.getLogger(EmbeddingProvider.class);
+  protected EmbeddingProviderConfigStore.RequestProperties requestProperties;
+  protected String baseUrl;
+  protected String modelName;
+  protected int dimension;
+  protected Map<String, Object> vectorizeServiceParameters;
+
+  /** Default constructor */
+  protected EmbeddingProvider() {}
```
tatu-at-datastax, line 28: Is this actually needed?

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/EmbeddingProvider.java
```diff
@@ -1,17 +1,68 @@
...
 
 /**
  * Interface that accepts a list of texts that needs to be vectorized and returns embeddings based
  * of chosen model.
  */
-public interface EmbeddingProvider {
-  Logger logger = LoggerFactory.getLogger(EmbeddingProvider.class);
+public abstract class EmbeddingProvider {
+  protected static final Logger logger = LoggerFactory.getLogger(EmbeddingProvider.class);
+  protected EmbeddingProviderConfigStore.RequestProperties requestProperties;
```
tatu-at-datastax, line 21: Could these be made `final`, if they are only set in constructor?

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/EmbeddingProvider.java
```diff
@@ -33,21 +84,67 @@ Uni<Response> vectorize(
...
+  }
 
   /**
    * Record to hold the batchId and embedding vectors
    *
    * @param batchId - Sequence number for the batch to order the vectors.
    * @param embeddings - Embedding vectors for the given text inputs.
    */
-  record Response(int batchId, List<float[]> embeddings) {
+  public record Response(int batchId, List<float[]> embeddings) {
```
tatu-at-datastax, line 141: Can also be `static`? (in interfaces, public and static are implied by default I think)

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/AzureOpenAIEmbeddingClient.java
```diff
@@ -44,12 +37,15 @@ public AzureOpenAIEmbeddingClient(
       String modelName,
       int dimension,
       Map<String, Object> vectorizeServiceParameters) {
-    this.requestProperties = requestProperties;
-    this.modelName = modelName;
     // One special case: legacy "ada-002" model does not accept "dimension" parameter
-    this.dimension = EmbeddingUtil.acceptsOpenAIDimensions(modelName) ? dimension : 0;
-    String actualUrl = EmbeddingUtil.replaceParameters(baseUrl, vectorizeServiceParameters);
+    super(
```
tatu-at-datastax, line 41: As per earlier comment, should probably pass `providerId` constant via constructor as well, for the abstract super class to use.

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/AzureOpenAIEmbeddingClient.java
```diff
@@ -44,12 +37,15 @@ public AzureOpenAIEmbeddingClient(
...
+        requestProperties,
+        baseUrl,
+        modelName,
+        EmbeddingProvider.acceptsOpenAIDimensions(modelName) ? dimension : 0,
```
tatu-at-datastax, line 45: Since it's method in super-class, no need to use `EmbeddingProvider.` prefix.

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/EmbeddingProvider.java
```diff
@@ -1,17 +1,68 @@
...
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
  * Interface that accepts a list of texts that needs to be vectorized and returns embeddings based
  * of chosen model.
  */
-public interface EmbeddingProvider {
-  Logger logger = LoggerFactory.getLogger(EmbeddingProvider.class);
+public abstract class EmbeddingProvider {
```
tatu-at-datastax, line 19: Since all concrete subtypes are Clients, should this also be `EmbeddingClient` instead of provider?

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/CohereEmbeddingClient.java
```diff
@@ -30,23 +26,18 @@
  * Interface that accepts a list of texts that needs to be vectorized and returns embeddings based
  * of chosen Cohere model.
  */
-public class CohereEmbeddingClient implements EmbeddingProvider {
-  private EmbeddingProviderConfigStore.RequestProperties requestProperties;
-  private String modelName;
-  private String baseUrl;
+public class CohereEmbeddingClient extends EmbeddingProvider {
+  private static final String providerId = ProviderConstants.COHERE;
```
tatu-at-datastax, line 30: As per my other notes, can be moved to super-class as field, passed by subtype constructor.
And then referenced like other `protected` fields.

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/MeteredEmbeddingProvider.java
```diff
@@ -18,7 +18,7 @@
  * collect and report various metrics, such as the duration of vectorization calls and the size of
  * input texts.
  */
-public class MeteredEmbeddingProvider implements EmbeddingProvider {
+public class MeteredEmbeddingProvider extends EmbeddingProvider {
```
tatu-at-datastax, line 21: Ah. So the default constructor in `EmbeddingProvider` needed for this sub-class.

tatu-at-datastax (commented):
I think this is going in the right direction: only some minor comments. Biggest is just adding `protected String providerId` in the abstract base class.
And maybe make other protected fields final.

tatu-at-datastax (approved):
As per our discussion, looks good, no need to change providerId handling

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/EmbeddingProvider.java
Hazel-Datastax: I got a warning message `Modifier 'static' is redundant for inner records ` from IDE, probably no need to add it?

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/MeteredEmbeddingProvider.java
Hazel-Datastax: Yes, and `TestEmbeddingProvider`

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/AzureOpenAIEmbeddingClient.java
tatu-at-datastax: Based on our discussion, this does not quite work like I thought, so we can forget about it for now.

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/EmbeddingProvider.java
tatu-at-datastax: Ok. Wasn't sure, then it's same as within an `interface`. So no need to add.

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/CohereEmbeddingClient.java
tatu-at-datastax: (as per earlier note, can't be done)

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/EmbeddingProvider.java
Hazel-Datastax: It looks a little bit odd after I change `EmbeddingProvider` to `EmbeddingClient`. We use `EmbeddingProvider` in many different places like using it to reference subclasses and we also expose this noun (like `findEmbeddingProviders`) to users. I think it may be better to change the subclass to `Provider` like `OpenAIEmbeddingProvider extends EmbeddingProvider` and change the inner interface to `Client` like `public interface OpenAIEmbeddingClient`

tatu-at-datastax: Yes, this sounds like a very good plan: let's go with it.
Thank you for thinking it through.

Hazel-Datastax:
Passed manual ITs in Embedding Gateway repo, merge!

PR #1149, opened 2024-06-04
Bridge-Removal - miscellaneous cleanups

Yuqi-Du:
Background: one step further to the bridge-removal

# Dependencies related
- quarkus-smallrye-openapi, quarkus-smallrye-fault-tolerance (openapi related)
- quarkus-security (used by principle-header auth)
- quarkus-grpc (quarkus grpc support)
- quarkus-config-yaml (This is the one we use to resolve embedding-providers-config.yaml)
- add jakarta.validation dependency
- add quarkus-resteasy-reactive (this is for injecting jakarta.ws.rs.core.SecurityContext in DataAPIRequestInfo)
- and quarkus-resteasy-reactive-jackson (this is to remove the warning msg noise: `[WARNING] [io.quarkus.resteasy.reactive.server.deployment.QuarkusServerEndpointIndexer] Quarkus detected the use of JSON in JAX-RS method 'io.stargate.sgv2.jsonapi.api.v1.NamespaceResource#postCommand' but no JSON extension has been added. Consider adding 'quarkus-resteasy-reactive-jackson' (recommended) or 'quarkus-resteasy-reactive-jsonb'.`)
- caffeine (cache)
- jbool_expressions dependency (used when building cql expressions)
- add quarkus-hibernate-validator dependency (Enables validation of configuration, beans, REST endpoints, etc.)

# other cleanups
- add loggingConstants, add log-console format
- disable quarkus http basic auth,
- add quarkus-application ui-header
- add ReservedKeywords class, which is used by ColumnUtils
- add http access log pattern, "%h %l %t \"%r\" %s %b" 
- add openAPI configuration (checked generated openAPI files by fileDiff)
- add swaggerUi configuration (checked http://localhost:8183/swagger-ui/, also added stargate logo.png to META-INF)
- Delete CreateCollectionResolverVectorDisabledTest, DisableVectorSearchProfile, since they rely on stargate data-store config, and to check if vectorSearch is enabled. This will be deprecated after we remove quarkus-common module dependency, and this is just a redundant check because we only deploy data-api in vector kubes.

**Checklist**
- [x] Changes manually tested
- [x] Automated Tests added/updated
- [x] Documentation added/updated
- [x] CLA Signed: [DataStax CLA](https://cla.datastax.com/)

PR #1191, opened 2024-06-25
Bridge-Removal - Detach the quarkus-common-module dependency (OFF WE GO!)

Yuqi-Du:
This PR mainly focuses on `restoring the metrics components` after the removal of quarkus-common-module dependency. 
But as you can see, there are other cleanups in this PR, this is because some cleanups can not be done in this PR Bridge-[Removal - miscellaneous cleanups](https://github.com/stargate/data-api/pull/1149), as long as the quarkus-common-module dependency remains, there are some bean production and injection required, that is why these cleanups must be done here.

# Detach quarkus-common-module dependency
Remove quarkus-common-module dependency in Data API pom file.
As for stargate version in Data API pom file, @maheshrajamani may need your confirmation, we could remove stargate versions, and just separate HCD, DSE versions out to make IT work?

# Metrics 
- Add Metrics config. (I still keep the configMapping path starts with stargate, this is because we still have these paths used helm chart, but we can do deep cleanups in the future).
- Add MicrometerConfiguration, this produces a globalTagsMeterFilter, which will add `sgv2-jsonapi` as global tag to all metrics entries.
- Add prometheus metric export path, same as before. `http://localhost:8181/metrics`

# Other cleanups
- Remove DataStoreConfig, this is coupling with Bridge calls, and only used to see if backend is vectorSearchEnabled(redundant).
- Remove RetriableStargateBridgeProvider
- Remove JsonApiGrpcRetryPolicy, and corresponding tests

# README.md, CONFIGURATION.md
Cleanup quarkus-common keywords in markdown file.
Detaching quarkus-common-module also means we lose some shared documentation there. Created a ticket for this.
[Documentation revisit after quarkus-common-module removal](https://github.com/stargate/data-api/issues/1193)

**Checklist**
- [x] Changes manually tested


tatu-at-datastax (approved):
LGTM

PR #1202, opened 2024-06-27
Fixes #1201: enable inclusion of source JSON snippet on parsing failures

### LINKED ISSUE #1201: Enable `StreamReadFeature.INCLUDE_SOURCE_IN_LOCATION` to get JSON source (snippet) logged
Troubleshooting issues #1199 and #1200, I realized that we get following line instead of source location:

```
at [Source: REDACTED (`StreamReadFeature.INCLUDE_SOURCE_IN_LOCATION` disabled); line: 1, column: 1]
```

and we really should enable said feature so we get better idea of input that triggered the exception.
(this is relatively new Jackson setting, disabled for Jackson 2.16)

tatu-at-datastax:
**What this PR does**:

Enables inclusion of JSON document (source) snippet on Jackson exception messages.
This setting was changed to `false` in Jackson 2.16; this PR enables it so we get more information in logs.

Fixes #1201

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/api/configuration/ObjectMapperConfiguration.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/api/configuration/ObjectMapperConfiguration.java
@@ -34,23 +34,17 @@ public class ObjectMapperConfiguration {
     // Number token limit handled by lower-level parser factory, need to construct first:
     JsonFactory jsonFactory =
         JsonFactory.builder()
+            // Configure maximum number length (against DoS)
             .streamReadConstraints(
                 StreamReadConstraints.builder().maxNumberLength(maxNumLen).build())
+            // Enable fast(er) floating-point number handling
             .enable(StreamReadFeature.USE_FAST_DOUBLE_PARSER)
             .enable(StreamReadFeature.USE_FAST_BIG_NUMBER_PARSER)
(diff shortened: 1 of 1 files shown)


Hazel-Datastax (approved):
LGTM

jeffreyscarpenter:
@tatu-at-datastax this sounds like a great change. Can you provide sample output so we can see what it will look like?

tatu-at-datastax:
Ok so the old location info is like:

```
com.fasterxml.jackson.core.JsonParseException: Unexpected character (',' (code 44)): expected a value
 at [Source: REDACTED (`StreamReadFeature.INCLUDE_SOURCE_IN_LOCATION` disabled); line: 1, column: 36]
```

but the new one depends on type of input source passed. If a `String` or `byte[]` is given we get a snippet of up to 1024 characters.
Unfortunately tho I think Quarkus will usually pass `InputStream` so we will only get something like:

```
 at [Source: `java.io.InputStream`; line: 1, column: 36]```
```

but I will double-check what tests show now.

tatu-at-datastax:
Ok: specifically we seemed to be getting:

```
[Source: (ByteArrayInputStream); line: 4, column: 4] 
```

as location, for main parsing (of incoming value bound by Quarkus).

Shredding would give different source, however, as we get `doc_json` as `String`.
I'll see if I can figure what is shown for those cases.

PR #1208, opened 2024-06-28
Fixes #1207: handle case of unknown "createCollection.options.indexing" fields

tatu-at-datastax:
**What this PR does**:

Adds explicit handling for another "CreateCollection" options mismatch

**Which issue(s) this PR fixes**:
Fixes #1207

**Checklist**
- [x] Changes manually tested
- [ ] Automated Tests added/updated
- [ ] Documentation added/updated
- [x] CLA Signed: [DataStax CLA](https://cla.datastax.com/)

PR #1213, opened 2024-06-28
Fixes #1199: map `UnrecognizedPropertyException` to INVALID_REQUEST_UNKNOWN_FIELD

### LINKED ISSUE #1199: Invalid JSON structure throwing `UnrecognizedPropertyException ` results in `SERVER_UNHANDLED_ERROR` 500 failure
Looking at unmapped exceptions, there is one logged as:

```
Unrecognized exception caught, mapped to SERVER_UNHANDLED_ERROR: Unrecognized
field "title" (class io.stargate.sgv2.jsonapi.api.model.command.impl.InsertOneCommand),
not marked as ignorable (2 known properties: "document", "options"])

com.fasterxml.jackson.databind.exc.UnrecognizedPropertyException: Unrecognized field
"title" (class io.stargate.sgv2.jsonapi.api.model.command.impl.InsertOneCommand),
not marked as ignorable (2 known properties: "document", "options"]
```

which we should be able to map better.

tatu-at-datastax:
**What this PR does**:

Adds explicit mapping for `UnrecognizedPropertyException` so it no longer results in `ErrorCode. SERVER_UNHANDLED_ERROR` but gives more useful information and regular HTTP 200 JsonApiException response

Fixes #1199

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/src/test/java/io/stargate/sgv2/jsonapi/api/v1/CollectionResourceIntegrationTest.java
+++ b/src/test/java/io/stargate/sgv2/jsonapi/api/v1/CollectionResourceIntegrationTest.java
@@ -83,6 +83,35 @@ class CollectionResourceIntegrationTest extends AbstractNamespaceIntegrationTest
     }
 
     @Test
+    public void unknownCommandField() {
+      String json =
+          """
+              {
+                "findOne": {
+                    "unknown": "value"
+                }
+              }
+              """;
+
+      given()
+          .headers(getHeaders())
+          .contentType(ContentType.JSON)
+          .body(json)
+          .when()
+          .post(CollectionResource.BASE_PATH, namespaceName, collectionName)
+          .then()
+          .statusCode(200)
+          .body("errors", hasSize(1))
+          .body("errors[0].errorCode", is("INVALID_REQUEST_UNKNOWN_FIELD"))
+          .body("errors[0].exceptionClass", is("JsonApiException"))
+          .body("errors[0].message", startsWith("Request invalid, unrecognized JSON field"))
+          .body("errors[0].message", containsString("\"unknown\" not one of known fields"))
+          .body(
+              "errors[0].message",
+              containsString("(\"filter\", \"options\", \"projection\", \"sort\")"));
+    }
+
+    @Test
     public void invalidNamespaceName() {
       String json =
           """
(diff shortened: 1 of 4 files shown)


src/test/java/io/stargate/sgv2/jsonapi/api/v1/CollectionResourceIntegrationTest.java
```diff
@@ -82,6 +82,35 @@ public void unknownCommand() {
...
+          .body(json)
+          .when()
+          .post(CollectionResource.BASE_PATH, namespaceName, collectionName)
+          .then()
+          .statusCode(200)
+          .body("errors", hasSize(1))
+          .body("errors[0].errorCode", is("INVALID_REQUEST_UNKNOWN_FIELD"))
+          .body("errors[0].exceptionClass", is("JsonApiException"))
+          .body("errors[0].message", startsWith("Request invalid, unrecognized JSON field"))
+          .body("errors[0].message", containsString("\"unknown\" not one of known fields"))
```
vkarpov15, line 108: Just checking but I presume this error message will also contain `at 'findOne'`? Having the full path in the error message similar to #1208 would be ideal.

tatu-at-datastax: Unfortunately it depends on location (in JSON doc) of mismatch. #1208 is special-casing matching and same could be done elsewhere but general handling added in this case is invoked if no special handler is matched.
I suspect `findOne` is not included as it is type identifier used to match specific command subtype. Other than that path is indeed indicated.

We can add special handling for commonly encountered cases but effort to do it for everything is prohibitive

I'll see what is the exact message being given just to make sure.

tatu-at-datastax: Ok, so, "findOne" is kind of indicated: 

```
Request invalid, unrecognized JSON field: \"unknown\" 
....
 at 'io.stargate.sgv2.jsonapi.api.model.command.impl.FindOneCommand["unknown"]'
```

which is not ideal (as it refers to Java class names), but is the path we can get out of exception.

Hazel-Datastax (approved):

PR #1217, opened 2024-07-01
Refactoring to allow fixing #1216 by making `JsonApiException` accept HTTP status code, propagate

### LINKED ISSUE #1216: Change "unparseable JSON" error to HTTP 400 (follow-up to #1200 / #1212)
Initial mapping of `JsonParseException` -- for invalid JSON that cannot be parsed -- is to HTTP 200 with `ErrorCode.INVALID_REQUEST_NOT_JSON`.
But it would make more sense to actually indicate it as HTTP 400 (same `ErrorCode`) since we cannot actually process is at all, the criteria for returning 200/OK.

tatu-at-datastax:
**What this PR does**:

Changes plumbing wrt `ErrorCode` and `JsonApiException` to allow passing HTTP response status along with exception, to indicate status other than 200 as necessary.

Meant to help resolve #1216 but does not quite achieve that.

N/A

**Checklist**
- [ ] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/api/model/command/CommandResult.java
```diff
@@ -194,11 +194,9 @@ public record Error(
    * @return
    */
   public RestResponse map() {
-    if (null != this.errors() && !this.errors().isEmpty()) {
+    if (null != this.errors()) {
```
tatu-at-datastax, line 197: Not necessary to check for empty, simplify code.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/CommandResult.java
```diff
@@ -194,11 +194,9 @@ public record Error(
...
       final Optional<Error> first =
-          this.errors().stream()
-              .filter(error -> error.status().getStatusCode() != Response.Status.OK.getStatusCode())
-              .findFirst();
+          this.errors().stream().filter(error -> error.status() != Response.Status.OK).findFirst();
```
tatu-at-datastax, line 199: Can do direct equality comparison of Enum values

src/main/java/io/stargate/sgv2/jsonapi/exception/JsonApiException.java
```diff
@@ -23,6 +23,8 @@ public class JsonApiException extends RuntimeException implements Supplier<Comma
 
   private final ErrorCode errorCode;
 
+  private final Response.Status httpStatus;
```
tatu-at-datastax, line 26: Will try propagating HTTP response code along with the exception so it's neither lost nor need to be carried separately.

vkarpov15 (commented):
Looks good so far, I like this change

tatu-at-datastax:
> Looks good so far, I like this change

Yeah. If only it actually worked... having hard time figuring out where the status code is lost. :-/
(IT fails, response is stubbornly 200)

src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCode.java
```diff
@@ -188,11 +190,21 @@ public JsonApiException toApiException(String format, Object... args) {
     return new JsonApiException(this, message + ": " + String.format(format, args));
   }
 
+  public JsonApiException toApiException(
```
tatu-at-datastax, line 193: Factory method for passing HTTP status code along: needs to be the first argument due to varargs needed to pass format String arguments.

src/main/java/io/stargate/sgv2/jsonapi/exception/mappers/ThrowableToErrorMapper.java
```diff
@@ -219,7 +223,7 @@ private static CommandResult.Error handleJsonProcessingException(
           .toApiException(
               "\"%s\" not one of known fields (%s) at '%s'",
               upe.getPropertyName(), knownDesc, upe.getPathReference())
-          .getCommandResultError(Response.Status.OK);
+          .getCommandResultError();
```
jeffreyscarpenter, line 226: so this defaults to 200?

jeffreyscarpenter (approved):
looks good, just left one question

src/main/java/io/stargate/sgv2/jsonapi/exception/mappers/ThrowableToErrorMapper.java
tatu-at-datastax: Yes; it uses `Response.Status` from `JsonApiException`, which in turn defaults to `Response.Status.OK`.

PR #1220, opened 2024-07-01
Add Amazon bedrock

maheshrajamani:
**What this PR does**:
Added aws Bedrock as a provider

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -57,7 +57,7 @@
       <dependency>
         <groupId>software.amazon.awssdk</groupId>
         <artifactId>bom</artifactId>
-        <version>2.25.67</version>
+        <version>2.26.12</version>
         <type>pom</type>
         <scope>import</scope>
       </dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -54,6 +54,13 @@
         <type>pom</type>
         <scope>import</scope>
       </dependency>
+      <dependency>
+        <groupId>software.amazon.awssdk</groupId>
+        <artifactId>bom</artifactId>
+        <version>2.25.67</version>
+        <type>pom</type>
+        <scope>import</scope>
+      </dependency>
     </dependencies>
   </dependencyManagement>
   <dependencies>
@@ -85,6 +92,14 @@
     <dependency>
       <groupId>com.google.guava</groupId>
       <artifactId>guava</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>software.amazon.awssdk</groupId>
+      <artifactId>bedrockruntime</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>software.amazon.awssdk</groupId>
+      <artifactId>sts</artifactId>
     </dependency>
     <dependency>
       <groupId>com.datastax.oss</groupId>
--- a/src/main/java/io/stargate/sgv2/jsonapi/api/request/HeaderBasedEmbeddingApiKeysResolver.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/api/request/HeaderBasedEmbeddingApiKeysResolver.java
@@ -5,27 +5,26 @@ import io.vertx.ext.web.RoutingContext;
 import java.util.Optional;
 
 /** Functional interface to resolve the embedding api key from the request header. */
-public class HeaderBasedEmbeddingApiKeyResolver implements EmbeddingCredentialResolver {
+public class HeaderBasedEmbeddingApiKeysResolver implements EmbeddingCredentialResolver {
   private final String tokenHeaderName;
   private final String accessIdHeaderName;
   private final String secretIdHeaderName;
 
-  public HeaderBasedEmbeddingApiKeyResolver(
+  public HeaderBasedEmbeddingApiKeysResolver(
       String tokenHeaderName, String accessIdHeaderName, String secretIdHeaderName) {
     this.tokenHeaderName = tokenHeaderName;
     this.accessIdHeaderName = accessIdHeaderName;
     this.secretIdHeaderName = secretIdHeaderName;
   }
 
-  public Optional<EmbeddingCredential> resolveEmbeddingCredential(RoutingContext context) {
+  public EmbeddingCredential resolveEmbeddingCredential(RoutingContext context) {
     HttpServerRequest request = context.request();
     String headerValue = request.getHeader(this.tokenHeaderName);
     String accessId = request.getHeader(this.accessIdHeaderName);
     String secretId = request.getHeader(this.secretIdHeaderName);
-    return Optional.ofNullable(
-        new EmbeddingCredential(
-            Optional.ofNullable(headerValue),
-            Optional.ofNullable(accessId),
-            Optional.ofNullable(secretId)));
+    return new EmbeddingCredential(
+        Optional.ofNullable(headerValue),
+        Optional.ofNullable(accessId),
+        Optional.ofNullable(secretId));
   }
 }
--- a/src/main/java/io/stargate/sgv2/jsonapi/api/request/HeaderBasedEmbeddingCredentialResolver.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/api/request/HeaderBasedEmbeddingCredentialResolver.java
@@ -2,9 +2,12 @@ package io.stargate.sgv2.jsonapi.api.request;
 
 import io.vertx.core.http.HttpServerRequest;
 import io.vertx.ext.web.RoutingContext;
+import java.util.Objects;
 import java.util.Optional;
 
-/** Functional interface to resolve the embedding api key from the request header. */
+/**
+ * Implementation to resolve the embedding api key, access id and secret id from the request header.
+ */
 public class HeaderBasedEmbeddingCredentialResolver implements EmbeddingCredentialResolver {
   private final String tokenHeaderName;
   private final String accessIdHeaderName;
@@ -12,6 +15,9 @@ public class HeaderBasedEmbeddingCredentialResolver implements EmbeddingCredenti
 
   public HeaderBasedEmbeddingCredentialResolver(
       String tokenHeaderName, String accessIdHeaderName, String secretIdHeaderName) {
+    Objects.requireNonNull(tokenHeaderName, "Token header name cannot be null");
+    Objects.requireNonNull(accessIdHeaderName, "Access Id header name cannot be null");
+    Objects.requireNonNull(secretIdHeaderName, "Secret Id header name cannot be null");
     this.tokenHeaderName = tokenHeaderName;
     this.accessIdHeaderName = accessIdHeaderName;
     this.secretIdHeaderName = secretIdHeaderName;
--- a/src/main/java/io/stargate/sgv2/jsonapi/api/request/HeaderBasedEmbeddingCredentialResolver.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/api/request/HeaderBasedEmbeddingCredentialResolver.java
@@ -5,12 +5,12 @@ import io.vertx.ext.web.RoutingContext;
 import java.util.Optional;
 
 /** Functional interface to resolve the embedding api key from the request header. */
-public class HeaderBasedEmbeddingApiKeysResolver implements EmbeddingCredentialResolver {
+public class HeaderBasedEmbeddingCredentialResolver implements EmbeddingCredentialResolver {
   private final String tokenHeaderName;
   private final String accessIdHeaderName;
   private final String secretIdHeaderName;
 
-  public HeaderBasedEmbeddingApiKeysResolver(
+  public HeaderBasedEmbeddingCredentialResolver(
       String tokenHeaderName, String accessIdHeaderName, String secretIdHeaderName) {
     this.tokenHeaderName = tokenHeaderName;
     this.accessIdHeaderName = accessIdHeaderName;
--- a/src/main/java/io/stargate/sgv2/jsonapi/config/constants/HttpConstants.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/config/constants/HttpConstants.java
@@ -22,19 +22,19 @@ public interface HttpConstants {
   String EMBEDDING_AUTHENTICATION_SECRET_ID_HEADER_NAME = "x-embedding-secret-Id";
 
   /**
-   * @return Embedding service header name for token<code>20</code>.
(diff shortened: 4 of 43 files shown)


pom.xml
```diff
@@ -54,6 +54,13 @@
         <type>pom</type>
         <scope>import</scope>
       </dependency>
+      <dependency>
+        <groupId>software.amazon.awssdk</groupId>
+        <artifactId>bom</artifactId>
+        <version>2.25.67</version>
```
tatu-at-datastax, line 60: Is there specific reason to use this version, instead of later/latest (https://mvnrepository.com/artifact/software.amazon.awssdk/bom), 2.25.70 / 2.26.12 ?

src/main/java/io/stargate/sgv2/jsonapi/api/request/HeaderBasedEmbeddingApiKeysResolver.java
```diff
@@ -0,0 +1,30 @@
+package io.stargate.sgv2.jsonapi.api.request;
+
+import io.vertx.core.http.HttpServerRequest;
+import io.vertx.ext.web.RoutingContext;
+import java.util.Optional;
+
+/** Functional interface to resolve the embedding api key from the request header. */
+public class HeaderBasedEmbeddingApiKeysResolver implements EmbeddingCredentialResolver {
```
tatu-at-datastax, line 8: Should this be name "HeaderBased[Embedding]CredentialResolver" ?

src/main/java/io/stargate/sgv2/jsonapi/config/constants/HttpConstants.java
```diff
@@ -15,9 +15,27 @@ public interface HttpConstants {
...
 
+  /** JSON API Embedding serive access id header name. */
+  String EMBEDDING_AUTHENTICATION_ACCESS_ID_HEADER_NAME = "x-embedding-access-Id";
+
+  /** JSON API Embedding serive secret id header name. */
+  String EMBEDDING_AUTHENTICATION_SECRET_ID_HEADER_NAME = "x-embedding-secret-Id";
+
   /**
-   * @return Embedding service header name <code>20</code>.
+   * @return Embedding service header name for token<code>20</code>.
```
tatu-at-datastax, line 25: I know it was already there but.... what does "20" here mean?

src/main/resources/embedding-providers-config.yaml
```diff
@@ -104,6 +104,48 @@ stargate:
                   help: "Vector dimension to use in the database and when calling Azure OpenAI."
             - name: text-embedding-ada-002
               vector-dimension: 1536
+        bedrock:
+          # see https://learn.microsoft.com/en-us/azure/ai-services/openai/reference
```
tatu-at-datastax, line 108: Probably not right URL?

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/AwsBedrockEnbeddingProvider.java
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  @Override
+  public Uni<Response> vectorize(
+      int batchId,
+      List<String> texts,
+      EmbeddingProvider.Credentials credentials,
+      EmbeddingRequestType embeddingRequestType) {
+    if (!credentials.accessKeyId().isPresent() || !credentials.secretAccessKey().isPresent()) {
+      throw new RuntimeException("AccessId and SecretId are required for AWS Bedrock");
```
tatu-at-datastax, line 51: Not sure how we do it with other clients; is it not possible to throw `JsonApiException`?

pom.xml
maheshrajamani: Just copied it from the bedrock examples. Will change it to newer version.

src/main/java/io/stargate/sgv2/jsonapi/config/constants/HttpConstants.java
maheshrajamani: Copy/paste issue. Resolved.

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/AwsBedrockEnbeddingProvider.java
maheshrajamani: Yes., will change it. Wanted the first version to work, resolving this.

src/main/resources/embedding-providers-config.yaml
maheshrajamani: Yes, URL is not an optional field. Both empty string and null not accepted. Need a separate PR to handle this since there is no URL for Amazon Bedrock.

maheshrajamani:
I will add few IT's for this similar to OpenAI which will run based on keys available as system property,

src/main/resources/embedding-providers-config.yaml
```diff
@@ -104,6 +104,47 @@ stargate:
                   help: "Vector dimension to use in the database and when calling Azure OpenAI."
             - name: text-embedding-ada-002
               vector-dimension: 1536
+        bedrock:
+          display-name: Amazon Bedrock
+          enabled: true
+          url: " "
```
maheshrajamani, line 110: This needs to be handled separately as URL field is mandatory in data-api and there is no endpoint url available for bedrock since we use the aws sdk.

src/main/java/io/stargate/sgv2/jsonapi/config/constants/HttpConstants.java
```diff
@@ -15,9 +15,27 @@ public interface HttpConstants {
   /** JSON API Embedding serive Authentication token header name. */
   String EMBEDDING_AUTHENTICATION_TOKEN_HEADER_NAME = "x-embedding-api-key";
 
+  /** JSON API Embedding serive access id header name. */
+  String EMBEDDING_AUTHENTICATION_ACCESS_ID_HEADER_NAME = "x-embedding-access-Id";
+
+  /** JSON API Embedding serive secret id header name. */
+  String EMBEDDING_AUTHENTICATION_SECRET_ID_HEADER_NAME = "x-embedding-secret-Id";
+
```
Hazel-Datastax, line 23: Can we make the name more aligned with the field names `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY`

tatu-at-datastax: Not sure about that -- if the idea is that other providers might use same 2 concepts as well?

maheshrajamani: Not sure if any other provider will be supporting 2 key format, so didn't use AWS.

src/main/java/io/stargate/sgv2/jsonapi/api/request/HeaderBasedEmbeddingCredentialResolver.java
```diff
@@ -0,0 +1,30 @@
+package io.stargate.sgv2.jsonapi.api.request;
+
+import io.vertx.core.http.HttpServerRequest;
+import io.vertx.ext.web.RoutingContext;
+import java.util.Optional;
+
+/** Functional interface to resolve the embedding api key from the request header. */
```
tatu-at-datastax, line 7: nit: not an interface, implementation class.

src/main/java/io/stargate/sgv2/jsonapi/config/constants/HttpConstants.java
```diff
@@ -15,9 +15,27 @@ public interface HttpConstants {
   /** JSON API Embedding serive Authentication token header name. */
   String EMBEDDING_AUTHENTICATION_TOKEN_HEADER_NAME = "x-embedding-api-key";
 
+  /** JSON API Embedding serive access id header name. */
+  String EMBEDDING_AUTHENTICATION_ACCESS_ID_HEADER_NAME = "x-embedding-access-Id";
```
tatu-at-datastax, line 19: Although header names are case-sensitive, I think recommendation is to use all lower-case, so:

   x-embedding-access-id

(f.ex https://scrapfly.io/blog/what-case-should-http-headers-be/)

src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCode.java
```diff
@@ -27,6 +27,8 @@ public enum ErrorCode {
       "The replace document and document resolved using filter have different _id"),
 
   /** Embedding provider service error codes. */
+  EMBEDDING_PROVIDER_AUTHENTICATIN_KEYS_NOT_PROVIDED(
```
tatu-at-datastax, line 30: `AUTHENTICATIN` -> `AUTHENTICATION`

src/main/java/io/stargate/sgv2/jsonapi/api/request/HeaderBasedEmbeddingCredentialResolver.java
```diff
@@ -0,0 +1,30 @@
...
+public class HeaderBasedEmbeddingCredentialResolver implements EmbeddingCredentialResolver {
+  private final String tokenHeaderName;
+  private final String accessIdHeaderName;
+  private final String secretIdHeaderName;
+
+  public HeaderBasedEmbeddingCredentialResolver(
+      String tokenHeaderName, String accessIdHeaderName, String secretIdHeaderName) {
+    this.tokenHeaderName = tokenHeaderName;
```
tatu-at-datastax, line 15: Could verify there are no `null`s with `Objects.requireNonNull()`?

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/AwsBedrockEnbeddingProvider.java
```diff
@@ -0,0 +1,158 @@
...
+import software.amazon.awssdk.regions.Region;
+import software.amazon.awssdk.services.bedrockruntime.BedrockRuntimeAsyncClient;
+import software.amazon.awssdk.services.bedrockruntime.model.BedrockRuntimeException;
+import software.amazon.awssdk.services.bedrockruntime.model.InvokeModelResponse;
+
+/** Provider implementation for AWS Bedrock. To start we support only Titan embedding models. */
+public class AwsBedrockEnbeddingProvider extends EmbeddingProvider {
+
+  private static final String providerId = ProviderConstants.BEDROCK;
+  private static ObjectWriter ow = new ObjectMapper().writer().withDefaultPrettyPrinter();
```
tatu-at-datastax, line 29: We probably do not want to use pretty-printing as it increases message sizes by ~30%.

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/AwsBedrockEnbeddingProvider.java
```diff
@@ -0,0 +1,158 @@
...
+    BedrockRuntimeAsyncClient client =
+        BedrockRuntimeAsyncClient.builder()
+            .credentialsProvider(StaticCredentialsProvider.create(awsCreds))
+            .region(Region.of(vectorizeServiceParameters.get("region").toString()))
+            .build();
+    final CompletableFuture<InvokeModelResponse> invokeModelResponseCompletableFuture =
+        client.invokeModel(
+            request -> {
+              try {
+                String inputJson =
```
tatu-at-datastax, line 69: Since we need UTF-8 bytes, instead use `writeValueAsBytes()` to avoid separate conversion (Jackson directly outputs encoded content avoid both String allocation and encoding step)

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/AwsBedrockEnbeddingProvider.java
```diff
@@ -0,0 +1,158 @@
...
+                    ow.writeValueAsString(new EmbeddingRequest(texts.get(0), dimension));
+                request.body(SdkBytes.fromUtf8String(inputJson)).modelId(modelName);
+              } catch (JsonProcessingException e) {
+                throw new RuntimeException(e);
```
tatu-at-datastax, line 73: Would it make sense to instead create and throw `JsonApiException`?

tatu-at-datastax (approved):
Added some notes & suggestions; looks good overall!

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/AwsBedrockEnbeddingProvider.java
maheshrajamani: These should never happen. Anyway will add something

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/AwsBedrockEnbeddingProvider.java
```diff
@@ -0,0 +1,158 @@
...
+                  providerId,
+                  bedrockRuntimeException.statusCode(),
+                  bedrockRuntimeException.getMessage());
+            });
+  }
+
+  private record EmbeddingRequest(
+      String inputText, @JsonInclude(value = JsonInclude.Include.NON_DEFAULT) int dimensions) {}
+
+  @JsonIgnoreProperties(ignoreUnknown = true)
```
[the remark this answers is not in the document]
maheshrajamani: I think its not needed as both model response is same.

src/main/java/io/stargate/sgv2/jsonapi/config/constants/HttpConstants.java
tatu-at-datastax: Makes sense to me.

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/AwsBedrockEnbeddingProvider.java
tatu-at-datastax: I wouldn't assume that: due to routing etc it is still possible for return message may be malformed, truncated, or text (due to proxies for example).

But I agree that no point in spending lots of effort for rare/unlikely case.
So maybe simple wrapping makes sense initially; then address if it becomes an issue (but only then).

PR #1233, opened 2024-07-03
Fix #1232: unify/rename 2 different ErrorCodes for similar thing

### LINKED ISSUE #1232: Unify `ErrorCode`s `COMMAND_NOT_IMPLEMENTED` and `NO_COMMAND_MATCHED` into `COMMAND_UNKNOWN`
(alternative: unify as existing `COMMAND_NOT_IMPLEMENTED` -- I like `COMMAND_` as prefix)
(another alternative: `COMMAND_UNRECOGNIZED`)

Currently there are 2 separate `ErrorCode`s used essentially for the same issue: case of API request referring to Command that we do not recognized (unknown, unmapped, mis-spelled).
Neither name is quite right:

* `COMMAND_NOT_IMPLEMENTED` -- this leads to assumption that such command might be added in future, which may or (more likely) may not be the case
* `NO_COMMAND_MATCHED` -- slightly better, but should start with `COMMAND`. Also, not sure "matching" conveys exact meaning.

Neither response is widely used: for past 48 hours there are not a single instance of either (out of ~5,000 error responses) so renaming should be safe to do at this point.

**EDIT**: Actually `COMMAND_NOT_IMPLEMENTED` is an internal failure for known command without matching resolver. Needs to be changed to new `SERVER_INTERNAL_ERROR`

tatu-at-datastax:
**What this PR does**:

Improves `ErrorCode` usage by removing a duplicate, improving code used for Unknown Command.

Fixes #1232

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/api/configuration/CommandObjectMapperHandler.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/api/configuration/CommandObjectMapperHandler.java
@@ -61,11 +61,10 @@ public class CommandObjectMapperHandler extends DeserializationProblemHandler {
     // CollectionCommand
     //            interface io.stargate.sgv2.jsonapi.api.model.command.GeneralCommand ->
     // GeneralCommand
-    final String rawCommandClassString = baseType.getRawClass().toString();
+    final String rawCommandClassString = baseType.getRawClass().getName();
     final String baseCommand =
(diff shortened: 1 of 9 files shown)


Hazel-Datastax (approved):

vkarpov15 (commented):
LGTM :+1:

PR #1246, opened 2024-07-09
Update Quarkus Data API uses to 3.9.5 (from parent 3.6.x)

### LINKED ISSUE #1245: Try to speed up slow(est) "unit" tests -- ones for CQL Session caching -- somehow (possibly change to Integration Tests)
Currently there are couple of "unit tests" that require running of C* backend, to test CQL Session caching aspects.
Tests are in `io.stargate.sgv2.jsonapi.service.cqldriver` package. Since they run very slowly -- in fact, taking about 50% of TOTAL time for ALL unit tests. It'd be good to speed them up, if possible; or if not, 
change them to integration tests.
This could take end-to-end unit test time from 8 to 4 minutes, making it more practical to run the full unit test suite locally as sanity check and rely on CI more wrt Integration Tests.

If changing type, the only change that should be needed would be renaming tests to contains `IntegrationTest` instead of `Test` as that's what surefire plug-in is configured to look for.

tatu-at-datastax:
**What this PR does**:

Incrementally upgrades Quarkus Data API uses from 3.6.x (from Stargate parent) to 3.9.5, fixing 3 tests that would be failing. This is only half-way through to latest (3.12) but done to minimize risks from upgrade.

N/A

**Checklist**
- [ ] Changes manually tested


src/test/java/io/stargate/sgv2/jsonapi/service/embedding/operation/EmbeddingProviderErrorMessageTest.java
```diff
@@ -175,7 +175,7 @@ public void testIncorrectContentTypeXML() {
               .withSubscriber(UniAssertSubscriber.create())
               .awaitFailure()
               .getFailure();
-      assertThat(exception.getCause())
```
tatu-at-datastax, line 178: This is strange: Quarkus version seems to have changed wrapping of exceptions, affecting 3 test methods in this test class, but nowhere else.

Hazel-Datastax (approved):
LGTM

PR #1251, opened 2024-07-09
Small fix: Improve error message when not providing provider key through `x-embedding-api-key`

### LINKED ISSUE #1250: Improve the error message when users call vectorize but don't provide the provider key through `x-embedding-api-key`
The current error message is:
```
{
    "errors": [
        {
            "message": "Server failed: root cause: (java.util.NoSuchElementException) No value present",
            "errorCode": "SERVER_UNHANDLED_ERROR"
        }
    ]
}
```
and it happens in `apiKeyOverride.get()` 
https://github.com/stargate/data-api/blob/dd4248507fc901c0f172e0e669f48e4d322764bd/src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/JinaAIEmbeddingProvider.java#L105

We need to add a checker before calling `get()`

Hazel-Datastax:
**What this PR does**:
Small fix: Improve error message when not providing provider key through `x-embedding-api-key`

Fixes #1250

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/AzureOpenAIEmbeddingProvider.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/AzureOpenAIEmbeddingProvider.java
@@ -115,6 +115,7 @@ public class AzureOpenAIEmbeddingProvider extends EmbeddingProvider {
       List<String> texts,
       Optional<String> apiKeyOverride,
       EmbeddingRequestType embeddingRequestType) {
+    checkEmbeddingApiKeyHeader(providerId, apiKeyOverride);
     String[] textArray = new String[texts.size()];
     EmbeddingRequest request = new EmbeddingRequest(texts.toArray(textArray), modelName, dimension);
 
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/AzureOpenAIEmbeddingProvider.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/AzureOpenAIEmbeddingProvider.java
@@ -113,15 +113,15 @@ public class AzureOpenAIEmbeddingProvider extends EmbeddingProvider {
   public Uni<Response> vectorize(
       int batchId,
       List<String> texts,
-      Optional<String> apiKeyOverride,
+      Optional<String> apiKey,
       EmbeddingRequestType embeddingRequestType) {
-    checkEmbeddingApiKeyHeader(providerId, apiKeyOverride);
+    checkEmbeddingApiKeyHeader(providerId, apiKey);
     String[] textArray = new String[texts.size()];
     EmbeddingRequest request = new EmbeddingRequest(texts.toArray(textArray), modelName, dimension);
 
     // NOTE: NO "Bearer " prefix with API key for Azure OpenAI
     Uni<EmbeddingResponse> response =
-        applyRetry(openAIEmbeddingProviderClient.embed(apiKeyOverride.get(), request));
+        applyRetry(openAIEmbeddingProviderClient.embed(apiKey.get(), request));
 
     return response
         .onItem()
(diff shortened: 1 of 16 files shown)


src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/AzureOpenAIEmbeddingProvider.java
```diff
@@ -115,6 +115,7 @@ public Uni<Response> vectorize(
       List<String> texts,
       Optional<String> apiKeyOverride,
       EmbeddingRequestType embeddingRequestType) {
+    checkEmbeddingApiKeyHeader(providerId, apiKeyOverride);
```
tatu-at-datastax, line 118: Is override really always needed? (maybe it's just odd name for parameter and means "apiKey"?)

tatu-at-datastax:
As long as `Optional` apiKeyOverride is really always needed, makes sense.
But confused about "override" part of the name -- if that's really just the key that is needed, fine, otherwise override would be optional I think.

Also come to think of that... if it's required, why would we use `Optional<>` at all?
Shouldn't be non-null value?

Hazel-Datastax:
The name "override" is a legacy term. I've updated it accordingly.

I think changing `Optional<String>` to `String` might be too aggressive. All `apiKey` values come from `getEmbeddingApiKey()` in `DataApiRequestInfo` (https://github.com/stargate/data-api/blob/a9cb4360f78a83521d3c02714f9c268cac2a3fec/src/main/java/io/stargate/sgv2/jsonapi/api/request/DataApiRequestInfo.java#L54). Instead of adding a checker in each embedding provider class, we could modify `getEmbeddingApiKey()` as follows:

```
  public String getEmbeddingApiKey() {
    if (embeddingApiKey.isEmpty()) {
      throw EMBEDDING_PROVIDER_API_KEY_MISSING.toApiException(
              "header value `%s` is missing in the request.",
              EMBEDDING_AUTHENTICATION_TOKEN_HEADER_NAME);
    }
    return embeddingApiKey.get();
  }
```

However, I think this change would limit the flexibility of the `getEmbeddingApiKey()` function, even though it is only used to pass the header value to each embedding provider class.

@tatu-at-datastax WDYT? Adding checker in each class or changing `getEmbeddingApiKey()`

tatu-at-datastax (approved):

tatu-at-datastax:
> The name "override" is a legacy term. I've updated it accordingly.
> 
> I think changing `Optional<String>` to `String` might be too aggressive. All `apiKey` values come from `getEmbeddingApiKey()` in `DataApiRequestInfo` (
> 
> https://github.com/stargate/data-api/blob/a9cb4360f78a83521d3c02714f9c268cac2a3fec/src/main/java/io/stargate/sgv2/jsonapi/api/request/DataApiRequestInfo.java#L54
> 
> ). Instead of adding a checker in each embedding provider class, we could modify `getEmbeddingApiKey()` as follows:
> ```
>   public String getEmbeddingApiKey() {
>     if (embeddingApiKey.isEmpty()) {
>       throw EMBEDDING_PROVIDER_API_KEY_MISSING.toApiException(
>               "header value `%s` is missing in the request.",
>               EMBEDDING_AUTHENTICATION_TOKEN_HEADER_NAME);
>     }
>     return embeddingApiKey.get();
>   }
> ```
> 
> However, I think this change would limit the flexibility of the `getEmbeddingApiKey()` function, even though it is only used to pass the header value to each embedding provider class.
> 
> @tatu-at-datastax WDYT? Adding checker in each class or changing `getEmbeddingApiKey()`

I like centralized validation: but you can split this in 2 -- "plain" getter that simply returns value (letting caller validate), then "getAndValidateEmbeddingApiKey()" that validates and gives error? Second method calls first to get value, then checks validity.

vkarpov15 (commented):
LGTM :+1:

PR #1270, opened 2024-07-12
refine log levels to eliminate unwanted messages

jeffreyscarpenter:
set default log level to warn and other levels for driver, cache, and stargate/data api. 

**Checklist**
- [x] Changes manually tested
- [ ] Automated Tests added/updated
- [ ] Documentation added/updated
- [x] CLA Signed: [DataStax CLA](https://cla.datastax.com/)

PR #1279, opened 2024-07-16
fix stargate log level to info

jeffreyscarpenter:
Fix error in log configuration for `io.stargate` classes introduced in #1270 

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/src/main/resources/application.yaml
+++ b/src/main/resources/application.yaml
@@ -111,7 +111,7 @@ quarkus:
         level: ERROR
       'com.datastax.oss.driver':
         level: WARN
-      'io.stargate.*':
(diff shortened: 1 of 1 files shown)


src/main/resources/application.yaml
```diff
@@ -111,7 +111,7 @@ quarkus:
         level: ERROR
       'com.datastax.oss.driver':
         level: WARN
-      'io.stargate.*':
+      'io.stargate':
```
tatu-at-datastax, line 114: Ahhhhh. Yeah, no need to include star (and it won't work if we do).

tatu-at-datastax (approved):

PR #1298, opened 2024-07-23
415 for invalid Content-Type header, remove redundant Token Filter

Yuqi-Du:
1. Previously if no Content-Type header is provided or an invalid one, we returned a weird Jarkada exception, and it is not a standard 415 Media Type not supported.
2. Token Filter is adding a redundant check, same logic is there is tokenResolver.

**Checklist**
- [x] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/exception/mappers/ThrowableToErrorMapper.java
```diff
@@ -69,6 +70,17 @@ public final class ThrowableToErrorMapper {
           return handleDriverException((DriverException) throwable, message);
         }
 
+        // handle an invalid Content-Type header
+        if (throwable instanceof NotSupportedException
+            && throwable
+                .getMessage()
+                .contains("The content-type header value did not match the value")) {
```
tatu-at-datastax, line 77: Is this check needed? Wouldn't we always convert this exception to 415?

src/main/java/io/stargate/sgv2/jsonapi/exception/mappers/ThrowableToErrorMapper.java
```diff
@@ -69,6 +70,17 @@ public final class ThrowableToErrorMapper {
...
+          // validate the Content-Type header, 415 if failed
+          return ErrorCode.INVALID_CONTENT_TYPE_HEADER
+              .toApiException()
+              .getCommandResultError(Response.Status.UNSUPPORTED_MEDIA_TYPE);
```
tatu-at-datastax, line 81: It's too bad actual incorrect Content-Type seems not available from `NotSupportedException` -- if it was, would be good to include in exception message.

src/main/java/io/stargate/sgv2/jsonapi/exception/mappers/WebApplicationExceptionMapper.java
```diff
@@ -42,6 +43,12 @@ public RestResponse<CommandResult> webApplicationExceptionMapper(WebApplicationE
     if (e instanceof NotFoundException) {
       return RestResponse.status(RestResponse.Status.NOT_FOUND, commandResult);
     }
+    // Return 415 for invalid Content-Type
+    if (e instanceof NotSupportedException
```
tatu-at-datastax, line 47: It's odd we have to do this in 2 different places -- do you know which one is necessary for the test to pass? Is it possible to test both handlers?

tatu-at-datastax (approved):
LGTM: added couple of questions, comments, but approved as-is so you can decide if there's anything worth changing.

src/main/java/io/stargate/sgv2/jsonapi/exception/mappers/ThrowableToErrorMapper.java
Yuqi-Du: Yes, let me change that.
I was just try to make sure.

src/main/java/io/stargate/sgv2/jsonapi/exception/mappers/ThrowableToErrorMapper.java
Yuqi-Du: Yeah, that is not included, unfortunately.

src/main/java/io/stargate/sgv2/jsonapi/exception/mappers/WebApplicationExceptionMapper.java
Yuqi-Du: This is because our current Exception structure design.
So this failure happens before our command processing, and it is captured by this WebApplicationExceptionMapper, but we still need want to convert the Throwable to CommandResult, so the same logic goes to ThrowableToErrorMapper. 
Here we just want the status 415 to be added.

Issue #1283, opened 2024-07-19
Re-factor: use `ErrorCode.toApiException()` everywhere instead of `new JsonApiException`


tatu-at-datastax:
Recently added convenience method `ErrorCode.toApiException` simplifies construction (less/more compact code) and ensures use of static base description before variable (and optional) detailed description. New method is used 133 cases; we should convert remaining ~100 cases to the new usage.

This is a pre-req step for bigger Error reporting re-vamp, clean up step.

PR #1301, opened 2024-07-24
Fix #1287: convert internal assertion exceptions to use JsonApiException.SERVER_INTERNAL_ERROR

### LINKED ISSUE #1287: Convert internal "assert"-style `RuntimeException` into `JsonApiException` with `ErrorCode.SERVER_INTERNAL_ERROR`
There are a few cases of plain `RuntimeException` (and `UnsupportedOperationException`s, `IllegalArgumentException`s, `IllegalStateException`s) thrown to basically assert that something that should never happen, never happens. While we can map these in various places to `JsonApiException`, there seems no real benefit to not just throwing with

    throw ErrorCode.SERVER_INTERNAL_ERROR.toApiException(message); // possibly templated

so let's clean code base; there are about 63 of these all in all and not all need conversion.

tatu-at-datastax:
**What this PR does**:

Converts most of usage of `RuntimeException` and sub-classes (`IllegalArgumentException`, `UnsupportedOperationException`) to use `JsonApiException` with `ErrorCode.SERVER_INTERNAL_ERROR` for better diagnostics

Fixes #1287

**Checklist**
- [ ] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/exception/JsonApiException.java
```diff
@@ -25,23 +25,24 @@ public class JsonApiException extends RuntimeException implements Supplier<Comma
 
   private final Response.Status httpStatus;
 
-  public JsonApiException(ErrorCode errorCode) {
+  protected JsonApiException(ErrorCode errorCode) {
```
tatu-at-datastax, line 28: These changes were accidentally missing from PR for #1283

src/main/java/io/stargate/sgv2/jsonapi/service/operation/filters/collection/AllCollectionFilter.java
```diff
@@ -66,6 +67,7 @@ public List<BuiltCondition> getAll() {
 
   @Override
   public BuiltCondition get() {
-    throw new UnsupportedOperationException("For $all filter we always use getAll() method");
+    throw ErrorCode.SERVER_INTERNAL_ERROR.toApiException(
+        "For $all filter we always use getAll() method");
```
Yuqi-Du, line 71: Memory comes back, if we can not use get() for all filters, that means we need to rethink about the interface design. I will add a todo.

Yuqi-Du (approved):

PR #1303, opened 2024-07-24
Create table command implementation

maheshrajamani:
**What this PR does**:
Create table command implementation

**Checklist**
- [x] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/api/v1/NamespaceResource.java
```diff
@@ -41,11 +49,12 @@
 public class NamespaceResource {
```
maheshrajamani, line 49: Can't add 2 resource for the same path, so added the createTable command to NamespaceResource. Not sure if this needs to be renamed to something else.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/tables/CreateTableOperation.java
```diff
@@ -0,0 +1,109 @@
...
+      List<String> partitionKeys,
+      List<OrderingKey> clusteringKeys,
+      Map<String, ColumnType> columnTypes) {
+    CreateTable createTable = null;
+    for (String partitionKey : partitionKeys) {
+      if (createTable == null) {
+        createTable =
+            create.withPartitionKey(partitionKey, columnTypes.get(partitionKey).getCqlType());
+      } else {
+        createTable =
```
maheshrajamani, line 75: Need to be reassigned, not working when chained.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/deserializers/ColumnDefinitionDeserializer.java
```diff
@@ -0,0 +1,30 @@
...
+  public ColumnDefinitionDeserializer() {
+    super(ColumnDefinition.class);
+  }
+
+  @Override
+  public ColumnDefinition deserialize(
+      JsonParser jsonParser, DeserializationContext deserializationContext)
+      throws IOException, JacksonException {
+    JsonNode definition = deserializationContext.readTree(jsonParser);
+    ColumnType type = ColumnType.fromString(definition.get("type").asText());
```
tatu-at-datastax, line 27: nit: `get()` -> `path()` to avoid NPE for missing "type".
(or maybe throw explicit exception for missing)

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/CreateTableCommand.java
```diff
@@ -0,0 +1,60 @@
...
+import jakarta.annotation.Nullable;
+import jakarta.validation.Valid;
+import jakarta.validation.constraints.NotNull;
+import jakarta.validation.constraints.Pattern;
+import jakarta.validation.constraints.Size;
+import java.util.Map;
+import org.eclipse.microprofile.openapi.annotations.enums.SchemaType;
+import org.eclipse.microprofile.openapi.annotations.media.Schema;
+
+@Schema(description = "Command that creates a api table.")
```
tatu-at-datastax, line 17: nit: "a api" -> "an api" ?

src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCode.java
```diff
@@ -185,6 +185,10 @@ public enum ErrorCode {
 
   // API Table Error Codes
   API_TABLE_FEATURE_NOT_ENABLED("API Table feature is not enabled"),
```
tatu-at-datastax, line 187: Just realized that this should probably not have `API_` prefix -- so if you change new ones to start with `TABLE_`, could change this one as well.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/column/definition/ColumnDefinition.java
```diff
@@ -0,0 +1,19 @@
+package io.stargate.sgv2.jsonapi.api.model.command.column.definition;
```
Yuqi-Du, line 1: Why don't we need the folder definition, if we just have one class ColumnDefinition.java.
Just being curious, like to know how you think this type of questions

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/column/definition/datatype/PrimitiveTypes.java
```diff
@@ -0,0 +1,25 @@
...
+    @Override
+    public DataType getCqlType() {
+      return DataTypes.TEXT;
+    }
+  }
+
+  private static class Int implements ColumnType {
+    @Override
+    public DataType getCqlType() {
+      return DataTypes.TEXT;
```
Yuqi-Du, line 22: TEXT?

src/main/java/io/stargate/sgv2/jsonapi/api/v1/NamespaceResource.java
Yuqi-Du: Interesting, don't we want to have some prefix to distinguish the path
Like
"/v1/collection/{namespace}"
"/v1/table/{namespace}"

Just thinking loud.

But it can also explain that KeyspaceCommand and NamepsaceCommand are all SchemaCommand, so they use the same Resouce.
If we are gonna use one Resource, perhaps not call it NamespaceResource anymore.

src/main/java/io/stargate/sgv2/jsonapi/api/v1/NamespaceResource.java
```diff
@@ -102,4 +128,30 @@ public Uni<RestResponse<CommandResult>> postCommand(
         // map to 2xx unless overridden by error
         .map(commandResult -> commandResult.map());
   }
```
Yuqi-Du, line 130: Looks like the only difference for these two method keyspaceCommand and namespaceCommand is 
how to construct the schemaObject?

Possible to exact that out?

src/main/java/io/stargate/sgv2/jsonapi/api/v1/NamespaceResource.java
tatu-at-datastax: Changing path for existing Collection creation would be major incompatibility, fwtw. I guess we could add new one, deprecate old, but that would end up us supporting 2 paths practically forever.
So it may be best to add new path segment for "Table" case and leave "Collection" one the way it is, if possible.

src/main/java/io/stargate/sgv2/jsonapi/api/v1/NamespaceResource.java
Yuqi-Du: I see.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/column/definition/ColumnDefinition.java
Yuqi-Du: Thanks, as long as you have a reason for it.

src/main/java/io/stargate/sgv2/jsonapi/api/v1/NamespaceResource.java
maheshrajamani: @Yuqi-Du As I make changes for new commands I don't changes in these method, so made it back to single method as suggested.

src/main/java/io/stargate/sgv2/jsonapi/api/v1/NamespaceResource.java
maheshrajamani: As per @amorton , the path needs to be same as collections. As for as name of the class he wants it to be left as is until further discussion done around it.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/tables/CreateTableOperation.java
```diff
@@ -0,0 +1,108 @@
...
+    createWithOptions = addClusteringOrder(createWithOptions, clusteringKeys);
+    final SimpleStatement statement = createWithOptions.build();
+    final Uni<AsyncResultSet> resultSetUni =
+        queryExecutor.executeCreateSchemaChange(dataApiRequestInfo, statement);
+    return resultSetUni.onItem().transform(rs -> new SchemaChangeResult(true));
+  }
+
+  private CreateTable addColumnsAndKeys(
+      CreateTableStart create,
+      List<String> partitionKeys,
```
amorton, line 65: three of these params are object state and do not need to be passed as params and the local name hides the class level name

src/test/java/io/stargate/sgv2/jsonapi/api/v1/NamespaceResourceIntegrationTest.java
```diff
@@ -68,7 +68,7 @@ public void unknownCommand() {
           .body(
               "errors[0].message",
               startsWith(
-                  "Provided command unknown: \"unknownCommand\" not one of \"NamespaceCommand\"s: known commands are [createCollection"));
```
amorton, line 71: revert this , lets hold off of changing the Namespace term

src/main/java/io/stargate/sgv2/jsonapi/StargateJsonApi.java
```diff
@@ -416,6 +416,63 @@
                         }
                       }
                       """),
+              @ExampleObject(
+                  name = "createTableCompositePrimaryKey",
```
amorton, line 420: lets discuss the implications of including this, the code is only alpha and it's going to change. If we go down this path how many commands would we need to keep up to date ?

amorton (changes requested):
i am going through to learn in some detail to learn how this gets done, so want to check a few things. 

but want to change this PR so it does not introduce the idea of Keyspace in the API

src/main/java/io/stargate/sgv2/jsonapi/StargateJsonApi.java
amorton: talking - ok to remove and we will fix swagger when we make things beta (or decide to drop swagger specs)

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/KeyspaceCommand.java
```diff
@@ -0,0 +1,10 @@
+package io.stargate.sgv2.jsonapi.api.model.command;
+
+import com.fasterxml.jackson.annotation.JsonSubTypes;
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+import io.stargate.sgv2.jsonapi.api.model.command.impl.CreateTableCommand;
+
+/** Interface for all commands executed against a keyspace. */
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, include = JsonTypeInfo.As.WRAPPER_OBJECT)
+@JsonSubTypes({@JsonSubTypes.Type(value = CreateTableCommand.class)})
+public interface KeyspaceCommand extends SchemaCommand {}
```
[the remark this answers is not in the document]
maheshrajamani: Renamed it to KeyspaceCommand->TableOnlyCommand and NamespaceCommand->CollectionOnlyCommand

amorton (approved):
looks good to me

vkarpov15 (commented):
LGTM :+1:

PR #1309, opened 2024-07-25
Added ValidatableCommandClause POC

amorton:
**What this PR does**:
Encapsulates behaviour for clauses that can be validated and better supports schema object types.

This is a POC for adding this, just to the level needed to make table code work. 

I will create a ticket to centralize all of the validation, based on this approach, and add here. 

See https://github.com/stargate/data-api/issues/1310 

NOTE : THIS PR DOES NOT CLOSE THE ISSUE ABOVE ^^^^

Fixes #<issue number>

**Checklist**
- [*] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/api/model/command/ValidatableCommandClause.java
```diff
@@ -0,0 +1,128 @@
...
+   * methods on the interface directly.
+   *
+   * @param commandContext The context the command is running against, including the {@link
+   *     SchemaObject}
+   * @param validatable An object that implements {@link ValidatableCommandClause}, may be null
+   * @param <T> Type of the {@link SchemaObject}
+   */
+  static <T extends SchemaObject> void maybeValidate(
+      CommandContext<T> commandContext, ValidatableCommandClause validatable) {
+    if (validatable == null) {
```
tatu-at-datastax, line 34: Is `null` legal? Assumption being that missing Clause is always valid thing?

tatu-at-datastax (approved):
Ok, I have some questions about the big picture, but details are fine. We can go back and consider this -- approving since it's a pre-req for other changes.

Yuqi-Du:
Reviewed, approach is clear. 
+1 that we may need to make some exceptions, since not every validation is easy to do or good to do in the resolver level.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/ValidatableCommandClause.java
Yuqi-Du: yes, I think so.

PR #1318, opened 2024-07-30
 POC for updateOne on a table

amorton:
**What this PR does**:

Continues to refactor how we build a CQL statement using the Java
driver, see the CQLClause and it's subclasses for the structure.

UpdateOneCommandResolver works with tables, creates a
TableUpdateOperation and re-uses the where clause from select.

Only supports the $set and $unset operations for now.

Fixes #<issue number>

**Checklist**
- [*] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/service/operation/tables/UpdateTableOperation.java
```diff
@@ -0,0 +1,91 @@
...
+    LOGGER.warn("UPDATE CQL: {}", update.asCql());
+    LOGGER.warn("UPDATE VALUES: {}", positionalValues);
+
+    return queryExecutor
+        .executeWrite(dataApiRequestInfo, statement)
+        .onItem()
+        .transformToUni(UpdateTableOperation::getResultSupplier);
+  }
+
+  private static Uni<Supplier<CommandResult>> getResultSupplier(AsyncResultSet result) {
```
maheshrajamani, line 76: The modified and matched count set to `1`.  The code flow looks fine, but these needs to be revisited. As clarified in another class is the POC scope limited to 1 row operations?

maheshrajamani (approved):
I don't see any existing logic breaking because of this PR> Since it's POC only code and controlled with the table enabled flag. It's ok to be merged.

PR #1325, opened 2024-08-01
Extend error response with additional information

Hazel-Datastax:
**What this PR does**:
Extend error response with additional information
The matching family and scope for each error code are [here](https://docs.google.com/spreadsheets/d/11hk-GQSaoH7hNq--MGJmSYufb3XkB3MW8mvUsy0SNSk/edit?gid=0#gid=0)

Fixes NA

**Checklist**
- [x] Changes manually tested


Hazel-Datastax:
If the `extendError` flag is enabled, the new error response will be like:
```
{
    "errors": [
        {
            "message": "\"insertOne1\" not one of \"CollectionCommand\"s: known commands are [countDocuments, deleteMany, deleteOne, estimatedDocumentCount, find, findOne, findOneAndDelete, findOneAndReplace, findOneAndUpdate, insertMany, insertOne, updateMany, updateOne]",
            "errorCode": "COMMAND_UNKNOWN",
            "scope": "",
            "title": "Provided command unknown",
            "family": "REQUEST"
        }
    ]
}
```
By default, the flag is disabled, and the error response will not be changed:
```
{
    "errors": [
        {
            "message": "Provided command unknown: \"insertOne1\" not one of \"CollectionCommand\"s: known commands are [countDocuments, deleteMany, deleteOne, estimatedDocumentCount, find, findOne, findOneAndDelete, findOneAndReplace, findOneAndUpdate, insertMany, insertOne, updateMany, updateOne]",
            "errorCode": "COMMAND_UNKNOWN"
        }
    ]
}
```

src/main/java/io/stargate/sgv2/jsonapi/exception/JsonApiException.java
```diff
@@ -47,6 +55,10 @@ protected JsonApiException(
     super(message, cause);
     this.errorCode = errorCode;
     this.httpStatus = httpStatus;
+    this.title = errorCode.getMessage();
+    String errorFamily = getErrorFamily();
+    this.errorFamily = errorFamily;
```
jeffreyscarpenter, line 60: Can combine lines 59/60? `this.errorFamily = getErrorFamily()`

src/main/java/io/stargate/sgv2/jsonapi/exception/JsonApiException.java
```diff
@@ -83,11 +95,25 @@ public CommandResult.Error getCommandResultError(String message, Response.Status
...
     DebugModeConfig debugModeConfig = config.getConfigMapping(DebugModeConfig.class);
     final boolean debugEnabled = debugModeConfig.enabled();
-    final Map<String, Object> fields =
-        debugEnabled
-            ? Map.of(
-                "errorCode", errorCode.name(), "exceptionClass", this.getClass().getSimpleName())
-            : Map.of("errorCode", errorCode.name());
+    final boolean extendError = config.getConfigMapping(OperationsConfig.class).extendError();
+    Map<String, Object> fields = null;
+    if (debugEnabled) {
```
jeffreyscarpenter, line 100: wondering about the logic here. Where is "debugEnabled" set?

jeffreyscarpenter: Something seems strange to me about this. We're using two different booleans to pick from one of three different cases. Is it possible that debugEnabled and extendError could both be set? Which one should take precedence (right now it is debugEnabled).

src/test/offline/io/stargate/sgv2/jsonapi/service/operation/model/impl/OfflineGetStatusOperationTest.java
```diff
@@ -3,6 +3,7 @@
 import static org.junit.jupiter.api.Assertions.*;
 
 import io.micrometer.core.instrument.simple.SimpleMeterRegistry;
+import io.quarkus.test.junit.QuarkusTest;
```
jeffreyscarpenter, line 6: doesn't seem in scope of this PR. Kathir is refactoring these classes to another repo anyway.

src/main/java/io/stargate/sgv2/jsonapi/exception/JsonApiException.java
Hazel-Datastax: rethought the logic here and changed my code to 
```
  if (extendError) {
    fields =
        Map.of(
            "errorCode",
            errorCode.name(),
            "family",
            errorFamily,
            "scope",
            errorScope,
            "title",
            title);
  } else {
    fields = Map.of("errorCode", errorCode.name());
  }

  if (debugEnabled) {
    fields.put("exceptionClass", this.getClass().getSimpleName());
  }
```
It will first decide the basic error structure based on `extendError` boolean and add `exceptionClass` if debug mode is enabled. WDYT?

jeffreyscarpenter: This looks cleaner / less confusing to me, thanks @Hazel-Datastax

src/main/java/io/stargate/sgv2/jsonapi/config/OperationsConfig.java
```diff
@@ -280,6 +280,12 @@ interface ConsistencyConfig {
   @WithDefault("false")
   boolean vectorizeEnabled();
 
+  /**
+   * @return Flag to extend error response with additional information.
+   */
+  @WithDefault("false")
```
tatu-at-datastax, line 286: nit: should probably be ordered alphabetically same as other settings.
(but it sounds like we'll be grouping Feature flags together soon anyway).

src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCode.java
```diff
@@ -209,16 +225,25 @@ public String getMessage() {
   }
 
   public JsonApiException toApiException(String format, Object... args) {
+    if (extendError) {
+      return new JsonApiException(this, String.format(format, args));
```
tatu-at-datastax, line 229: Should extract `String.format()` part into helper method since it's called multiple times (to contain logic of formattion wrt `extendError`)

src/main/java/io/stargate/sgv2/jsonapi/exception/JsonApiException.java
```diff
@@ -106,4 +135,134 @@ public ErrorCode getErrorCode() {
   public Response.Status getHttpStatus() {
     return httpStatus;
   }
+
+  private String getErrorFamily() {
+    // some error codes should be classified as SERVER errors but do not have any pattern
+    Set<ErrorCode> serverFamily =
```
tatu-at-datastax, line 141: I think this is not a good way to do this -- instead of creating a `Set` here, should include this information directly in `ErrorCode`.

src/main/java/io/stargate/sgv2/jsonapi/exception/JsonApiException.java
```diff
@@ -25,6 +27,12 @@ public class JsonApiException extends RuntimeException implements Supplier<Comma
 
   private final Response.Status httpStatus;
 
+  private final String title;
+
+  private final String errorFamily;
```
tatu-at-datastax, line 32: Should probably be an enum, not String.

src/main/java/io/stargate/sgv2/jsonapi/exception/JsonApiException.java
```diff
@@ -25,6 +27,12 @@ public class JsonApiException extends RuntimeException implements Supplier<Comma
...
+
+  private final String errorScope;
```
tatu-at-datastax, line 34: Should perhaps be an enum instead of String?

src/main/java/io/stargate/sgv2/jsonapi/exception/JsonApiException.java
```diff
@@ -106,4 +135,134 @@ public ErrorCode getErrorCode() {
...
+            add(COLLECTION_CREATION_ERROR);
+            add(INVALID_QUERY);
+            add(NO_INDEX_ERROR);
+          }
+        };
+
+    if (serverFamily.contains(errorCode)
+        || errorCode.name().startsWith("SERVER")
+        || errorCode.name().startsWith("EMBEDDING")) {
+      return "SERVER";
```
tatu-at-datastax, line 165: Explicit information better than reverse-engineering from String (wrt earlier note on adding explicit family info in `ErrorCode`, not append here).

src/main/java/io/stargate/sgv2/jsonapi/exception/JsonApiException.java
```diff
@@ -106,4 +135,134 @@ public ErrorCode getErrorCode() {
...
+    }
+    return "REQUEST";
+  }
+
+  private String getErrorScope(String family) {
```
tatu-at-datastax, line 170: Same as with family, seems better to pass explicit info either by caller (eventually), or from `ErrorCode` (at least initially).

src/test/java/io/stargate/sgv2/jsonapi/api/v1/VectorSearchIntegrationTest.java
```diff
@@ -803,7 +802,7 @@ public void happyPathWithEmptyVector() {
           .body("errors", hasSize(1))
           .body("errors[0].exceptionClass", is("JsonApiException"))
           .body("errors[0].errorCode", is("SHRED_BAD_VECTOR_SIZE"))
-          .body("errors[0].message", is(ErrorCode.SHRED_BAD_VECTOR_SIZE.getMessage()));
+          .body("errors[0].message", is("$vector value can't be empty"));
```
tatu-at-datastax, line 805: Not sure why this was changed? Earlier code more robust wrt changes to error message.

src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCode.java
amorton: not sure I follow, is that to handle errors from String.format () is so +1

src/main/java/io/stargate/sgv2/jsonapi/exception/JsonApiException.java
amorton: not a terrible idea, when we started this refactoring we were not sure what all of the scopes  would be . And we were happy to keep it simple as we will be throwing it away, we now have a better idea.  @hazel maybe refactor into an enum to reduce duplicate strings

src/main/java/io/stargate/sgv2/jsonapi/exception/JsonApiException.java
amorton: we took this approach to reduce the amount of existing code we had to change, and the amount of repitition. All of the existing ErrorCode enum is going to be removed , and there is 90 odd definitions there.  So rather than change all fo them we took this approach for the lowest number of modified lines

amorton: we should move this set to be static final though so it is not created each time the function runs

src/main/java/io/stargate/sgv2/jsonapi/exception/JsonApiException.java
amorton: as above, we want to remove all the old error code in 2 to 3 weeks so this to reduce the number of changed lines. 

Hazel - this could be a Map<Set<ErrorCode> , String> (string for the scope) created as static final so you . If you do that you iterate  the entries and test the keys, it is a little less code than a Map<ErrorCode, String> and better conveys what we are doing.

amorton (changes requested):
couple of changes ot the way we are creating the family / scope - but I am still happy with the general approach rather than changing all of the existing ErrorCode enum

src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCode.java
tatu-at-datastax: Just to extract variation in creating `message` (old mode: append title + body; new mode just body) into helper method so it's not repeated in various `toApiException()`s.

src/main/java/io/stargate/sgv2/jsonapi/exception/JsonApiException.java
tatu-at-datastax: Ok. Fair enough.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/VectorSearchIntegrationTest.java
Hazel-Datastax: Because I added `extendError` - the configuration in the `ErrorCode` class. It will cause the class initialization error in this case - because it cannot find the config

tatu-at-datastax (approved):

amorton (approved):

src/test/java/io/stargate/sgv2/jsonapi/api/v1/VectorSearchIntegrationTest.java
tatu-at-datastax: Ok. We will probably change things to ensure that no loading occurs from within static initialization of `ErrorCode` (or its replacement) in future.
But is fine for now.

src/main/java/io/stargate/sgv2/jsonapi/exception/JsonApiException.java
```diff
@@ -106,4 +203,95 @@ public ErrorCode getErrorCode() {
...
+    DATABASE("DATABASE"),
+    DOCUMENT("DOCUMENT"),
+    EMBEDDING("EMBEDDING"),
+    EMPTY(""),
+    FILTER("FILTER"),
+    INDEX("INDEX"),
+    PROJECTION("PROJECTION"),
+    SCHEMA("SCHEMA"),
+    SORT("SORT"),
+    UPDATE("UPDATE");
```
vkarpov15, line 284: Should there also be an error scope for OPTIONS if you pass in an invalid option?

PR #1331, opened 2024-08-07
Publish Data API jar to maven central

kathirsvn:
**What this PR does**:
Changes to publish Data API jar to maven central

**Checklist**
- [X] Changes manually tested
- [X] Automated Tests added/updated
- [X] Documentation added/updated
- [X] CLA Signed: [DataStax CLA](https://cla.datastax.com/)

PR #1330, opened 2024-08-07
Removed offline mode code

kathirsvn:
**What this PR does**:
The offline mode code moved to internal repo

**Checklist**
- [X] Changes manually tested
- [X] Automated Tests added/updated
- [X] Documentation added/updated
- [X] CLA Signed: [DataStax CLA](https://cla.datastax.com/)

PR #1337, opened 2024-08-14
Fix jar publishing workflows

jeffreyscarpenter:
In #1330 we removed the offline code and logic to publish Data API jar to private maven repo. We had originally intended to start publishing the jar to a public repo in #1331, but there are no known use cases that would rely on this, so we decided to stick with current approach of publishing to private repo. This PR restores the `distributionManagement` section to the pom so that Maven knows where to publish the jar.

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -333,4 +333,18 @@
       </properties>
     </profile>
   </profiles>
+  <distributionManagement>
+    <repository>
+      <uniqueVersion>false</uniqueVersion>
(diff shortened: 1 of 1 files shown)


jeffreyscarpenter:
Verified Jar publish workflow is now working https://github.com/stargate/data-api/actions/runs/10391183706/job/28773483957.

PR #1348, opened 2024-08-19
Fix CreateTableOperation identifier issue, add several table ITs.

Yuqi-Du:
1. 
So in previous implementation of CreateTableOperation,
```
 public static CreateTableStart createTable(@Nullable String keyspace, @NonNull String tableName) {
        return createTable(keyspace == null ? null : CqlIdentifier.fromCql(keyspace), CqlIdentifier.fromCql(tableName));
    }
```
we are using createTable method from CQL Driver, which uses fromCql method to resolve identifier. This is a problematic, because if things are not quoted, driver will make them case-insensitive, to lower case.

e.g. 
keyspace in DB:  `Yuqi_Keyspace`
because of this string is not quoted, then CQL Driver decides to convert it to `yuqi_keyspace`, which does not exist.

So to be consistent with what we did for collection, I just added quote for keyspace and tableName in createTableCommand.

2.
Enable table feature in test resource

3. 
add AbstractTableIntegrationTestBase, add few ITs for table.

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/operation/tables/CreateTableOperation.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/operation/tables/CreateTableOperation.java
@@ -2,6 +2,7 @@ package io.stargate.sgv2.jsonapi.service.operation.tables;
 
 import static com.datastax.oss.driver.api.querybuilder.SchemaBuilder.*;
 
+import com.datastax.oss.driver.api.core.CqlIdentifier;
 import com.datastax.oss.driver.api.core.cql.AsyncResultSet;
 import com.datastax.oss.driver.api.core.cql.SimpleStatement;
 import com.datastax.oss.driver.api.core.metadata.schema.ClusteringOrder;
@@ -55,9 +56,10 @@ public class CreateTableOperation implements Operation {
   @Override
   public Uni<Supplier<CommandResult>> execute(
       DataApiRequestInfo dataApiRequestInfo, QueryExecutor queryExecutor) {
-    String keyspaceQuoted = "\"%s\"".formatted(commandContext.schemaObject().name.keyspace());
-    String tableNameQuoted = "\"%s\"".formatted(tableName);
-    CreateTableStart create = createTable(keyspaceQuoted, tableNameQuoted).ifNotExists();
+    CqlIdentifier keyspaceIdentifier =
+        CqlIdentifier.fromInternal(commandContext.schemaObject().name.keyspace());
+    CqlIdentifier tableIdentifier = CqlIdentifier.fromInternal(tableName);
+    CreateTableStart create = createTable(keyspaceIdentifier, tableIdentifier).ifNotExists();
 
     // Add all primary keys and colunms
     CreateTable createTable = addColumnsAndKeys(create);
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/operation/tables/CreateTableOperation.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/operation/tables/CreateTableOperation.java
@@ -55,9 +55,9 @@ public class CreateTableOperation implements Operation {
   @Override
   public Uni<Supplier<CommandResult>> execute(
       DataApiRequestInfo dataApiRequestInfo, QueryExecutor queryExecutor) {
-
-    CreateTableStart create =
-        createTable(commandContext.schemaObject().name.keyspace(), tableName).ifNotExists();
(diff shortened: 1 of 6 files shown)


src/main/java/io/stargate/sgv2/jsonapi/service/operation/tables/CreateTableOperation.java
```diff
@@ -55,9 +55,9 @@ public CreateTableOperation(
   @Override
   public Uni<Supplier<CommandResult>> execute(
       DataApiRequestInfo dataApiRequestInfo, QueryExecutor queryExecutor) {
-
-    CreateTableStart create =
-        createTable(commandContext.schemaObject().name.keyspace(), tableName).ifNotExists();
+    String keyspaceQuoted = "\"%s\"".formatted(commandContext.schemaObject().name.keyspace());
```
tatu-at-datastax, line 58: Do we not have helper method(s) in `CqlIdentifier` to handle quoting? Instead of manually surrounding with double quotes.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/tableIntegrationTest/AddTableIndexIntegrationTest.java
```diff
@@ -0,0 +1,106 @@
...
+                              }
+                          }
+                    """
+            .formatted(simpleTableName);
+    createTable(tableJson);
+  }
+
+  @Nested
+  @Order(1)
+  class addIndexSuccess {
```
tatu-at-datastax, line 51: nit: class names start with upper-case characters (so `AddIndexSuccess`)

src/main/java/io/stargate/sgv2/jsonapi/service/operation/tables/CreateTableOperation.java
Yuqi-Du: I did not find any method in driver that gives us CreateTableStart back and handle quote automatically.

```
    @NonNull
    public static CreateTableStart createTable(@NonNull String tableName) {
        return createTable(CqlIdentifier.fromCql(tableName));
    }

    @NonNull
    public static CreateTableStart createTable(@Nullable String keyspace, @NonNull String tableName) {
        return createTable(keyspace == null ? null : CqlIdentifier.fromCql(keyspace), CqlIdentifier.fromCql(tableName));
    }
```
I think they design this by the tableName/keyspace as cql representation. Which means if we want quoted string, we need to add by ourselves.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/tableIntegrationTest/AddTableIndexIntegrationTest.java
Yuqi-Du: thanks!

src/main/java/io/stargate/sgv2/jsonapi/service/operation/tables/CreateTableOperation.java
tatu-at-datastax: That seems very odd -- there's nothing that takes `CqlIdentifier`? We should pass

   CqlIdentifier.fromInternal(tableName)

to ensure quoting is used if (and only if) necessary. But I agree, if only `String` is allowed then we must quote.

tatu-at-datastax (approved):

tatu-at-datastax:
@Yuqi-Du let's get this merged to make it easier to add other Table ITs!

src/main/java/io/stargate/sgv2/jsonapi/service/operation/tables/CreateTableOperation.java
```diff
@@ -55,9 +56,10 @@ public CreateTableOperation(
   @Override
   public Uni<Supplier<CommandResult>> execute(
       DataApiRequestInfo dataApiRequestInfo, QueryExecutor queryExecutor) {
-
-    CreateTableStart create =
-        createTable(commandContext.schemaObject().name.keyspace(), tableName).ifNotExists();
+    CqlIdentifier keyspaceIdentifier =
+        CqlIdentifier.fromInternal(commandContext.schemaObject().name.keyspace());
+    CqlIdentifier tableIdentifier = CqlIdentifier.fromInternal(tableName);
```
tatu-at-datastax, line 61: Ok good!

vkarpov15 (commented):
LGTM :+1:

PR #1359, opened 2024-08-23
Fix `findOne`/`find` wrt `_id` column (actually all filterables)

### LINKED ISSUE #1357: Not able to filter by _id with tables
(see #1349 for background)

After fixing #1349, it will be possible to use "non-standard" column names, ones that must be quoted in CQL. But there appears to be a special case for column `_id`, likely due to special handling given to it with Collections.
So while:

1. One can create Table with `_id` column AND
2. insert rows with `_id`

`findOne()` and `find() commands seem to fail with strange failure related to filtering:

```
    "errors": [
        {
            "message": "Server internal error: Filter type not supported, unable to resolve to a filtering strategy",
            "errorCode": "SERVER_INTERNAL_ERROR",
            "exceptionClass": "JsonApiException"
```

tatu-at-datastax:
**What this PR does**:

Fixes handling of `_id` column wrt Find operations: there is something specific about this compared to all other ids (quoted or not)

Fixes #1357

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/resolver/matcher/TableFilterResolver.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/resolver/matcher/TableFilterResolver.java
@@ -11,6 +11,7 @@ import io.stargate.sgv2.jsonapi.service.operation.filters.table.NativeTypeTableF
 import io.stargate.sgv2.jsonapi.service.operation.filters.table.NumberTableFilter;
 import io.stargate.sgv2.jsonapi.service.operation.filters.table.TextTableFilter;
 import io.stargate.sgv2.jsonapi.service.operation.query.DBFilterBase;
+import io.stargate.sgv2.jsonapi.service.shredding.collections.DocumentId;
 import java.math.BigDecimal;
 import java.util.ArrayList;
 import java.util.EnumSet;
@@ -24,6 +25,7 @@ import java.util.List;
 public class TableFilterResolver<CmdT extends Command & Filterable>
     extends FilterResolver<CmdT, TableSchemaObject> {
 
+  private static final Object DYNAMIC_DOCID_GROUP = new Object();
   private static final Object DYNAMIC_TEXT_GROUP = new Object();
   private static final Object DYNAMIC_NUMBER_GROUP = new Object();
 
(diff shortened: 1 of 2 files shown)


tatu-at-datastax:
Ok, this turns out to be rather difficult to solve: basically `FilterClauseDeserializer` has quite specific handling for `_id` (DOC_ID), in:

https://github.com/stargate/data-api/blob/24af4e2cfe49689916113ba6d90719fc195c032d/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/deserializers/FilterClauseDeserializer.java#L336 

which creates value type not matched by filter resolved for Tables.

Conceptually it'd be good for deserializer to skip this special handling for Tables, but piping through that information proves difficult given that deserialization is invoked by Quarkus, via type on endpoint.

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/matcher/TableFilterResolver.java
```diff
@@ -75,20 +90,44 @@ private static List<DBFilterBase> findDynamic(CaptureExpression captureExpressio
...
       } else if (captureExpression.marker() == DYNAMIC_NUMBER_GROUP) {
         filters.add(
             new NumberTableFilter(
                 captureExpression.path(),
                 NativeTypeTableFilter.Operator.from(
                     (ValueComparisonOperator) filterOperation.operator()),
-                (BigDecimal) filterOperation.operand().value()));
+                (BigDecimal) rhsValue));
+      } else if (captureExpression.marker() == DYNAMIC_DOCID_GROUP) {
+        Object actualValue = ((DocumentId) rhsValue).value();
```
tatu-at-datastax, line 109: Should probably be refactored to avoid duplicating code to create Text/NumberTableFilter.

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/matcher/TableFilterResolver.java
```diff
@@ -75,20 +90,44 @@ private static List<DBFilterBase> findDynamic(CaptureExpression captureExpressio
...
+                  NativeTypeTableFilter.Operator.from(
+                      (ValueComparisonOperator) filterOperation.operator()),
+                  (BigDecimal) actualValue));
+        } else {
+          throw new UnsupportedOperationException(
+              "Unsupported DocumentId type: " + rhsValue.getClass().getName());
+        }
+      } else {
+        throw new UnsupportedOperationException(
+            "Unsupported dynamic filter type: " + filterOperation);
```
Yuqi-Du, line 130: I know this is just following the pattern. We can change these to Data API exception later.

Yuqi-Du (approved):
LGTM!

vkarpov15:
I confirmed yesterday that this works in https://github.com/stargate/stargate-mongoose-sample-apps/pull/397

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/matcher/TableFilterResolver.java
tatu-at-datastax: Yes, sorry I missed this one before merging.

tatu-at-datastax:
Thank you for verifying @vkarpov15 !

PR #1360, opened 2024-08-26
hide table feature in swaggerUI before it goes public

Yuqi-Du:
hide table feature in swaggerUI before it goes public
**Checklist**
- [x] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/AddIndexCommand.java
```diff
@@ -8,7 +8,8 @@
 import jakarta.validation.constraints.Size;
 import org.eclipse.microprofile.openapi.annotations.media.Schema;
 
-@Schema(description = "Command that creates an index for a column in a table.")
+// TODO, hide table feature detail before it goes public
```
tatu-at-datastax, line 11: idea: maybe link back to issue for re-enabling these annotations (may need to create one if it doesn't exist)?

tatu-at-datastax (approved):
LGTM, but I think it'd be good to add a link from comments to issue that tracks re-enabling of OpenAPI definitions.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/AddIndexCommand.java
Yuqi-Du: No problem.

PR #1362, opened 2024-08-27
Added support for optional http_response_override yaml 

amorton:
**What this PR does**:

When set it is passed through to the APIException creation and
set as a field on the exception that can be used later when we
start handling these errors and working out the response code.

Fixes #<issue number>

**Checklist**
- [*] Changes manually tested

### DIFF
--- a/src/main/resources/errors.yaml
+++ b/src/main/resources/errors.yaml
@@ -34,6 +34,10 @@
 # Each error object has:
 # - scope: UPPER_SNAKE_CASE_1
 # - code: UPPER_SNAKE_CASE_1
+# - http_response_override: (optional) The HTTP status code to return when this error is thrown. If not present, the
+#                           default status code is 200 for most things. This is not returned in the error object JSON
+#                           It is used to override the HTTP status code in the response.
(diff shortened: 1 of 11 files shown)


src/main/resources/errors.yaml
```diff
@@ -34,6 +34,10 @@
 # Each error object has:
 # - scope: UPPER_SNAKE_CASE_1
 # - code: UPPER_SNAKE_CASE_1
+# - http_response_override: (optional) The HTTP status code to return when this error is thrown. If not present, the
```
tatu-at-datastax, line 37: Any reason why we want to use "snake_case" instead of camelCase? In Java land, snake_case is rarely used except for interoperability with systems written in Python etc?
But here we control the definition.

amorton: discussed offline, I thought that was the standard but we have agreed it should be `kebab-case` will create a seperate PR to change

amorton:
merging to the feature branch without a +1 approval because feature branch will a bigger review. 

will create a PR to change to `kebab-case`

PR #1367, opened 2024-08-27
Fixes #1365: add integral number codecs, tests

### LINKED ISSUE #1365: Data API, JsonCodec testing: integral numeric values (`bigint`, `smallint`, `int`, `tinyint`, `varint`)
Add:

* Unit tests for 5 supported integer CQL types
* ITs for the same to verify creation of table with columns of these types, working InsertOne, and findOne to fetch.

tatu-at-datastax:
**What this PR does**:

Changes table-shredding to convert JSON values (filter, insert) to use 3 types -- `Long`, `BigInteger`, `BigDecimal` -- instead of just `BigDecimal`, for less wasteful conversions (BigDecimal use for integral numbers is significant overhead).

Adds more CQL codecs to convert to/from CQL integral types; tests (Unit and Integration).

Fixes #1365

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/operation/filters/table/NumberTableFilter.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/operation/filters/table/NumberTableFilter.java
@@ -1,11 +1,9 @@
 package io.stargate.sgv2.jsonapi.service.operation.filters.table;
 
-import java.math.BigDecimal;
-
 /** Filter to use any JSON number against a table column. */
-public class NumberTableFilter extends NativeTypeTableFilter<BigDecimal> {
+public class NumberTableFilter extends NativeTypeTableFilter<Number> {
 
-  public NumberTableFilter(String path, Operator operator, BigDecimal value) {
+  public NumberTableFilter(String path, Operator operator, Number value) {
     super(path, operator, value);
   }
 }
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/operation/filters/table/codecs/JSONCodec.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/operation/filters/table/codecs/JSONCodec.java
@@ -1,6 +1,7 @@
 package io.stargate.sgv2.jsonapi.service.operation.filters.table.codecs;
 
 import com.datastax.oss.driver.api.core.type.DataType;
+import com.datastax.oss.driver.api.core.type.DataTypes;
 import com.datastax.oss.driver.api.core.type.reflect.GenericType;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
@@ -166,6 +167,45 @@ public record JSONCodec<JavaT, CqlT>(
           throw new ToCQLCodecException(value, toCQLType, e);
         }
       };
+    }
+
+    static Integer safeLongToInt(Long value) {
+      long l = value.longValue();
+      if (l < Integer.MIN_VALUE) {
+        throwUnderflow(DataTypes.INT, value);
+      } else if (l > Integer.MAX_VALUE) {
+        throwOverflow(DataTypes.INT, value);
+      }
+      return (int) l;
+    }
+
+    static Short safeLongToSmallint(Long value) {
+      long l = value.longValue();
+      if (l < Short.MIN_VALUE) {
+        throwUnderflow(DataTypes.SMALLINT, value);
+      } else if (l > Short.MAX_VALUE) {
+        throwOverflow(DataTypes.SMALLINT, value);
+      }
+      return (short) l;
+    }
+
+    static Byte safeLongToTinyint(Long value) {
+      long l = value.longValue();
+      if (l < Byte.MIN_VALUE) {
+        throwUnderflow(DataTypes.TINYINT, value);
+      } else if (l > Byte.MAX_VALUE) {
+        throwOverflow(DataTypes.TINYINT, value);
+      }
+      return (byte) l;
+    }
+
+    static void throwOverflow(DataType targetCQLType, Number value) {
+      throw new ArithmeticException(String.format("Overflow, value too big for %s", targetCQLType));
+    }
+
+    static void throwUnderflow(DataType targetCQLType, Number value) {
+      throw new ArithmeticException(
+          String.format("Underflow, value too small for %s", targetCQLType));
     }
   }
 
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/operation/filters/table/codecs/JSONCodec.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/operation/filters/table/codecs/JSONCodec.java
@@ -105,7 +105,7 @@ public record JSONCodec<JavaT, CqlT>(
   }
 
   /**
-   * Function interface that is used by the codec to convert the Java value to the value CQL
+   * Functional interface that is used by the codec to convert the Java value to the value CQL
    * expects.
    *
    * <p>The interface is used so the conversation function can throw the checked {@link
@@ -170,7 +170,7 @@ public record JSONCodec<JavaT, CqlT>(
   }
 
   /**
-   * Function interface that is used by the codec to convert value returned by CQL into a {@link
+   * Functional interface that is used by the codec to convert value returned by CQL into a {@link
    * JsonNode} that can be used to construct the response document for a row.
    *
    * <p>The interface is used so the conversation function can throw the checked {@link
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/operation/filters/table/codecs/JSONCodec.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/operation/filters/table/codecs/JSONCodec.java
@@ -228,7 +228,7 @@ public record JSONCodec<JavaT, CqlT>(
    * <p>The interface is used so the conversation function can throw the checked {@link
    * ToJSONCodecException}, it is also given the CQL data type to make better exceptions.
    *
-   * <p>Use the static constructors on the interface to get instances, see it's use in the {@link
+   * <p>Use the static constructors on the interface to get instances, see its use in the {@link
    * JSONCodecRegistry}
    *
    * @param <CqlT> The type Java object the CQL driver expects
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/operation/filters/table/codecs/JSONCodec.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/operation/filters/table/codecs/JSONCodec.java
@@ -5,19 +5,21 @@ import com.datastax.oss.driver.api.core.type.DataTypes;
 import com.datastax.oss.driver.api.core.type.reflect.GenericType;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import io.stargate.sgv2.jsonapi.service.shredding.tables.RowShredder;
 import java.util.function.Function;
 
 /**
- * Handles the conversation between the in memory Java representation of a value from a JSON
+ * Handles the conversation between the in-memory Java representation of a value from a JSON
  * document and the Java type that the driver expects for the CQL type of the column.
  *
  * <p>This is codec sitting above the codec the Java C* driver uses.
  *
- * <p>The path is:
+ * <p>The path for converting values contained in incoming JSON Document into values sent to C* is:
  *
  * <ul>
- *   <li>JSON Document
- *   <li>Jackson parses and turns into Java Object (e.g. BigInteger)
+ *   <li>JSON Document parsed by Jackson into a {@link JsonNode}
+ *   <li>{@link RowShredder} converts the Jackson {@JsonNode}s values into Java Objects (from {@code
+ *       TextNode} into {@code String}, {@code BooleanNode} into {@code Boolean}, etc.)
  *   <li>JSONCodec (this class) turns Java Object into the Java type the C* driver expects (e.g.
  *       Short
  *   <li>C* driver codec turns Java type into C* type
@@ -27,6 +29,16 @@ import java.util.function.Function;
  * JSON doc from reading a row and to use it for writing a row. // TODO Mahesh, The codec looks fine
  * for primitive type. Needs a revisit when we doing complex // types where only few fields will
  * need to be returned. Will we be creating custom Codec based // on user requests?
+ *
+ * <p>Note on Jackson conversion of JSON Numbers: Jackson is used to first read JSON content as
+ * {@link JsonNode}s, and then values are converted to "natural" Java types: conversion is done in
+ * {@link RowShredder#shredValue} and results in one of the following types:
+ *
+ * <ul>
+ *   <li>{@link java.math.BigDecimal} for numbers that are not integers in JSON
+ *   <li>{@link java.math.BigInteger} for integers that are too big for {@link Long}
+ *   <li>{@link java.lang.Long} for integers that fit in 64-bit signed {@code Long}
+ * </ul>
  *
  * @param javaType {@link GenericType} of the Java object that needs to be transformed into the type
  *     CQL expects.
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/operation/filters/table/codecs/JSONCodecRegistry.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/operation/filters/table/codecs/JSONCodecRegistry.java
@@ -127,7 +127,7 @@ public class JSONCodecRegistry {
   }
 
   // Boolean
-  public static final JSONCodec<Boolean, Boolean> BOOLEAN =
+  private static final JSONCodec<Boolean, Boolean> BOOLEAN =
       new JSONCodec<>(
           GenericType.BOOLEAN,
           DataTypes.BOOLEAN,
@@ -135,21 +135,21 @@ public class JSONCodecRegistry {
           JSONCodec.ToJSON.unsafeNodeFactory(JsonNodeFactory.instance::booleanNode));
 
   // Numeric Codecs
-  public static final JSONCodec<BigDecimal, Long> BIGINT_FROM_BIG_DECIMAL =
+  private static final JSONCodec<BigDecimal, Long> BIGINT_FROM_BIG_DECIMAL =
       new JSONCodec<>(
           GenericType.BIG_DECIMAL,
           DataTypes.BIGINT,
           JSONCodec.ToCQL.safeNumber(BigDecimal::longValueExact),
           JSONCodec.ToJSON.unsafeNodeFactory(JsonNodeFactory.instance::numberNode));
 
-  public static final JSONCodec<BigInteger, Long> BIGINT_FROM_BIG_INTEGER =
+  private static final JSONCodec<BigInteger, Long> BIGINT_FROM_BIG_INTEGER =
       new JSONCodec<>(
           GenericType.BIG_INTEGER,
           DataTypes.BIGINT,
           JSONCodec.ToCQL.safeNumber(BigInteger::longValueExact),
           JSONCodec.ToJSON.unsafeNodeFactory(JsonNodeFactory.instance::numberNode));
 
-  public static final JSONCodec<Long, Long> BIGINT_FROM_LONG =
+  private static final JSONCodec<Long, Long> BIGINT_FROM_LONG =
       new JSONCodec<>(
           GenericType.LONG,
           DataTypes.BIGINT,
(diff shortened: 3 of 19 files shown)


src/main/java/io/stargate/sgv2/jsonapi/service/operation/filters/table/NumberTableFilter.java
```diff
@@ -1,11 +1,9 @@
 package io.stargate.sgv2.jsonapi.service.operation.filters.table;
 
-import java.math.BigDecimal;
-
 /** Filter to use any JSON number against a table column. */
-public class NumberTableFilter extends NativeTypeTableFilter<BigDecimal> {
+public class NumberTableFilter extends NativeTypeTableFilter<Number> {
```
tatu-at-datastax, line 4: Need to change now that we can get values in 3 distinct types (Long, BigInteger, BigDecimal).

tatu-at-datastax:
Getting there:

* Codecs added for all 5 integral CQL types, from all 3 JSON number types (`Long`, `BigInteger`, `BigDecimal`
* Range checks working EXCEPT for `Long` case (no ready-made method to use, need to implement)
* No dedicated Integration Tests yet, minor adds to "findOne" for tables but not sufficient.

tatu-at-datastax:
Bounds-/range-check added for `smallint`/`tinyint`/`int` conversion from Long; only missing `insertOne` integration tests. Ready to start review.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/util/DataApiCommandSenderBase.java
```diff
@@ -0,0 +1,73 @@
...
+import io.stargate.sgv2.jsonapi.service.embedding.operation.test.CustomITEmbeddingProvider;
+import java.util.Base64;
+import java.util.Map;
+import org.eclipse.microprofile.config.ConfigProvider;
+
+/**
+ * Helper class used for constructing and sending commands to the Data API for testing purposes
+ * (usually integration tests).
+ */
+public abstract class DataApiCommandSenderBase<T extends DataApiCommandSenderBase> {
```
tatu-at-datastax, line 17: At least 2 implementations (namespace/keyspace commands, table/collection commans); not sure if Table/Collection should be separated, probably not. Self-recursive type to allow type-safe "this" return for chaining.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/util/DataApiCommandSenderBase.java
```diff
@@ -0,0 +1,73 @@
...
+  public T expectHttpStatus(int expectedHttpStatus) {
+    this.expectedHttpStatus = expectedHttpStatus;
+    return _this();
+  }
+
+  protected T _this() {
+    return (T) this;
+  }
+
+  protected Map<String, ?> getHeaders() {
```
tatu-at-datastax, line 41: Copied from existing IT base class, no change.

Can allow overriding when needed, not added yet as no need.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/util/DataApiCommandSenders.java
```diff
@@ -0,0 +1,8 @@
+package io.stargate.sgv2.jsonapi.api.v1.util;
+
+/** Helper class used for constructing and sending commands to the Data API */
+public class DataApiCommandSenders {
+  public static DataApiTableCommandSender tableCommand(String namespace, String tableName) {
```
tatu-at-datastax, line 5: Makes little sense until adding `namespaceCommand()` (or should that be `keyspaceCommand()`), but that will be added soon.

tatu-at-datastax:
And now with more-compact-than-before abstractions for more efficient (wrt reading, writing) IT writing. Ready for constructive feedback.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/filters/table/codecs/JSONCodecRegistry.java
```diff
@@ -168,34 +177,90 @@ public static <JavaT, CqlT> JSONCodec<JavaT, CqlT> codecToJSON(DataType targetCQ
...
+          JSONCodec.ToJSON.unsafeNodeFactory(JsonNodeFactory.instance::numberNode));
+
+  public static final JSONCodec<Long, Integer> INT_FROM_LONG =
+      new JSONCodec<>(
+          GenericType.LONG,
+          DataTypes.INT,
+          JSONCodec.ToCQL.safeNumber(JSONCodec.ToCQL::safeLongToInt),
+          JSONCodec.ToJSON.unsafeNodeFactory(JsonNodeFactory.instance::numberNode));
+
+  public static final JSONCodec<BigDecimal, Short> SMALLINT_FROM_BIG_DECIMAL =
```
amorton, line 201: is this correct that we have to smallint from BigDecimal - i thought all integral would come from BigInteger ?

src/main/java/io/stargate/sgv2/jsonapi/service/operation/filters/table/codecs/JSONCodecRegistry.java
```diff
@@ -127,25 +127,34 @@ public static <JavaT, CqlT> JSONCodec<JavaT, CqlT> codecToJSON(DataType targetCQ
...
           GenericType.BIG_DECIMAL,
           DataTypes.BIGINT,
           JSONCodec.ToCQL.safeNumber(BigDecimal::longValueExact),
           JSONCodec.ToJSON.unsafeNodeFactory(JsonNodeFactory.instance::numberNode));
 
-  // TODO Tatu For performance reasons we could also consider only converting FP values into
-  // BigDecimal JsonNode -- but converting CQL integer values into long-valued JsonNode.
-  //  I think our internal handling can deal with Integer and Long valued JsonNodes and this avoids
-  // some of BigDecimal overhead (avoids conversion overhead, serialization is faster).
+  private static final JSONCodec<BigInteger, Long> BIGINT_FROM_BIG_INTEGER =
```
amorton, line 145: I get what we are doing with integral values from Jackson - question is what need to support both BigInteger and Long source values ? if it is needed can we add a comment at top to explain strategy  . 

And that the type mappings here must sync with how we pull values from Jackson.

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/matcher/TableFilterResolver.java
```diff
@@ -104,7 +103,7 @@ private static List<DBFilterBase> findDynamic(CaptureExpression captureExpressio
                 captureExpression.path(),
                 NativeTypeTableFilter.Operator.from(
                     (ValueComparisonOperator) filterOperation.operator()),
-                (BigDecimal) rhsValue));
+                (Number) rhsValue));
```
amorton, line 106: Just a note - the changes yuqi and I are doing will revert how the resolver functions here are called back to the original approach where it gets called once with everything that matched. Currently it is getting called for every match. 

Nothing to change, just a note that when we do that we should look at how to reduce the use of instanceof and casting

src/main/java/io/stargate/sgv2/jsonapi/service/shredding/tables/RowShredder.java
```diff
@@ -91,11 +91,24 @@ public WriteableTableRow shred(TableSchemaObject table, JsonNode document) {
...
     return switch (value.getNodeType()) {
-      case NUMBER -> value.decimalValue();
+      case NUMBER -> shredNumber(value);
       case STRING -> value.textValue();
       case BOOLEAN -> value.booleanValue();
       case NULL -> null;
       default -> throw new RuntimeException("Unsupported type");
     };
   }
+
```
amorton, line 101: I know this class is not finished / high quality - can you add a quick comment to say the different types for numbers need to match the codecs in JSONCodecRegistry

src/test/java/io/stargate/sgv2/jsonapi/api/v1/util/DataApiCommandSenderBase.java
```diff
@@ -0,0 +1,73 @@
...
+import org.eclipse.microprofile.config.ConfigProvider;
+
+/**
+ * Helper class used for constructing and sending commands to the Data API for testing purposes
+ * (usually integration tests).
+ */
+public abstract class DataApiCommandSenderBase<T extends DataApiCommandSenderBase> {
+  protected final String namespace;
+
+  protected int expectedHttpStatus = 200;
```
amorton, line 20: Nit: use Response.Status.OK  (payback for your ninja fix on my use of 200  :) ) 

lso, maybe change the expectedHttpStatus below to use the Response.Status enum

src/test/java/io/stargate/sgv2/jsonapi/api/v1/util/DataApiNamespaceCommandSender.java
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  /**
+   * "Untyped" method for sending a POST command to the Data API: caller is responsible for
+   * formatting the JSON body correctly.
+   *
+   * @param jsonBody JSON body to POST
+   * @return Response validator for further assertions
+   */
+  public DataApiResponseValidator postRaw(String jsonBody) {
```
amorton, line 22: this looks like it can move up to the base, the only difference between this is tables is the args passed to .post()

src/test/java/io/stargate/sgv2/jsonapi/api/v1/util/DataApiTableCommandSender.java
```diff
@@ -0,0 +1,69 @@
...
+   *
+   * @param findOneClause JSON clause to include in the "findOne" command: minimally empty JSON
+   *     Object ({@code { } })
+   * @return Response validator for further assertions
+   */
+  public DataApiResponseValidator postFindOne(String findOneClause) {
+    return postCommand("findOne", findOneClause);
+  }
+
+  public DataApiResponseValidator postInsertOne(String docAsJSON) {
```
amorton, line 58: I like how this is going, giving function names that match the commands. Consider going further and calling them insertOne etc do not need the post* prefix.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/util/DataApiTableCommandSender.java
```diff
@@ -0,0 +1,69 @@
...
+            .contentType(ContentType.JSON)
+            .body(jsonBody)
+            .when()
+            .post(CollectionResource.BASE_PATH, namespace, tableName)
+            .then()
+            .statusCode(expectedHttpStatus);
+    return new DataApiResponseValidator(response);
+  }
+
+  public DataApiResponseValidator postCommand(String command, String commandClause) {
```
amorton, line 38: Consider moving to the base, it looks like your sender classes would only define the commands we have for the type of endpoint resource.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/util/DataApiResponseValidator.java
```diff
@@ -0,0 +1,73 @@
...
+    this.response = response;
+  }
+
+  // // // Access to Hamcrest Response Matchers // // //
+
+  public ValidatableResponse response() {
+    return response;
+  }
+
+  public DataApiResponseValidator body(String path, Matcher<?> matcher, Object... args) {
```
amorton, line 26: Is this correct that it is changing the state ? Should it return a new instance ?

src/test/java/io/stargate/sgv2/jsonapi/api/v1/util/DataApiResponseValidator.java
```diff
@@ -0,0 +1,73 @@
...
+import static org.hamcrest.Matchers.nullValue;
+
+import io.restassured.response.ValidatableResponse;
+import io.stargate.sgv2.jsonapi.exception.ErrorCode;
+import org.hamcrest.Matcher;
+
+public class DataApiResponseValidator {
+  private ValidatableResponse response;
+
+  public DataApiResponseValidator(ValidatableResponse response) {
```
amorton, line 16: I like where this is going

src/test/java/io/stargate/sgv2/jsonapi/service/operation/filters/table/codecs/JSONCodecRegistryTest.java
```diff
@@ -16,7 +17,7 @@
 
 public class JSONCodecRegistryTest {
 
-  private final JSONCodecRegistryTestData TEST_DATA = new JSONCodecRegistryTestData();
+  private static final JSONCodecRegistryTestData TEST_DATA = new JSONCodecRegistryTestData();
```
amorton, line 20: as a pattern, i had been keeping the "TestData" class as instance but I guess given how tests work static makes sense.

src/test/java/io/stargate/sgv2/jsonapi/service/operation/filters/table/codecs/JSONCodecRegistryTest.java
```diff
@@ -81,9 +82,29 @@ public void codecToCQL(DataType cqlType, Object fromValue, Object expectedCqlVal
   }
 
   private static Stream<Arguments> codecToCQLTestCases() {
+    // Arguments: (CQL-type, from-caller, bound-by-driver-for-cql
+    // Note: all Numeric types accept 3 Java types: Long, BigInteger, BigDecimal
     return Stream.of(
-        Arguments.of(DataTypes.INT, BigDecimal.valueOf(100), 100) // second 100 is an int
-        );
+        // Integer types:
+        Arguments.of(DataTypes.BIGINT, -456L, -456L),
```
amorton, line 89: nit: consider rename to something like validCodecToCQL.... now you have out of range below.

src/test/java/io/stargate/sgv2/jsonapi/service/operation/filters/table/codecs/JSONCodecRegistryTest.java
```diff
@@ -81,9 +82,29 @@ public void codecToCQL(DataType cqlType, Object fromValue, Object expectedCqlVal
...
+        Arguments.of(DataTypes.SMALLINT, -3999L, (short) -3999),
+        Arguments.of(DataTypes.SMALLINT, BigInteger.valueOf(1234), (short) 1234),
+        Arguments.of(DataTypes.SMALLINT, BigDecimal.valueOf(10911.0), (short) 10911),
+        Arguments.of(DataTypes.TINYINT, -39L, (byte) -39),
+        Arguments.of(DataTypes.TINYINT, BigInteger.valueOf(123), (byte) 123),
+        Arguments.of(DataTypes.TINYINT, BigDecimal.valueOf(109.0), (byte) 109),
+        Arguments.of(DataTypes.VARINT, -39999L, BigInteger.valueOf(-39999)),
+        Arguments.of(DataTypes.VARINT, BigInteger.valueOf(1), BigInteger.valueOf(1)),
+        Arguments.of(
+            DataTypes.VARINT, BigDecimal.valueOf(123456789.0), BigInteger.valueOf(123456789)),
```
amorton, line 104: I've not checked that the list here and below is exhaustive, let me know if you want it checked.

amorton (changes requested):

amorton (changes requested):
Generally like where it is going, just small comments on nit picks and making some things clearer.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/util/DataApiCommandSenders.java
```diff
@@ -0,0 +1,12 @@
+package io.stargate.sgv2.jsonapi.api.v1.util;
+
+/** Helper class used for constructing and sending commands to the Data API */
+public class DataApiCommandSenders {
+  public static DataApiNamespaceCommandSender namespaceCommand(String namespace) {
+    return new DataApiNamespaceCommandSender(namespace);
+  }
+
+  public static DataApiTableCommandSender tableCommand(String namespace, String tableName) {
```
amorton, line 9: idea - if we name these assertTableCommand, then do static imports in tests, and make the other name change below it could look like this in a test: 

```
assertTableCommand(namespace, table)
    .findOne({})
```

src/main/java/io/stargate/sgv2/jsonapi/service/operation/filters/table/codecs/JSONCodec.java
```diff
@@ -167,10 +168,49 @@ static <JavaT extends Number, CqlT> ToCQL<JavaT, CqlT> safeNumber(
...
+      if (l < Byte.MIN_VALUE) {
+        throwUnderflow(DataTypes.TINYINT, value);
+      } else if (l > Byte.MAX_VALUE) {
+        throwOverflow(DataTypes.TINYINT, value);
+      }
+      return (byte) l;
+    }
+
+    static void throwOverflow(DataType targetCQLType, Number value) {
+      throw new ArithmeticException(String.format("Overflow, value too big for %s", targetCQLType));
```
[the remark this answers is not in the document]
tatu-at-datastax: Good point -- I was doing that, but realized that `safeNumber()` that catches `ArithmeticException` and re-wraps as `ToCQLCodecException` adds the actual value.
This is why I chose to avoid adding it in `ArithmeticException` itself.

On underflow: yeah I must admit I googled for definitions since I recall "overflow" sometimes being used to mean both, too. :)

src/main/java/io/stargate/sgv2/jsonapi/service/operation/filters/table/codecs/JSONCodecRegistry.java
tatu-at-datastax: So, with FPs, cost of processing `BigDecimal` seems to be roughly similar to `Double`.
But cost of processing `BigInteger` is significantly higher than that of boxed primitive integral types. So I figured it is a compromise to try to avoid huge number of converters by slight addition -- I think minimally we could go with `BigDecimal` and `BigInteger`.

I'll add a comment here.

100% on mentioning that handling with `RowShredder` must be kept in sync.

tatu-at-datastax: I added some class Javadoc comments in `RowShredder` and `JSONCodec`, hope that suffices.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/filters/table/codecs/JSONCodecRegistry.java
tatu-at-datastax: This is needed in cases where JSON has, say, `1.0` and we want to allow that: Jackson bases type of textual representation of JSON Number so if there's decimal point, it's FP.

I am assuming we want to allow this coercion (existing code, when only using `BigDecimal`, did allow this): if not, could drop this converter.

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/matcher/TableFilterResolver.java
tatu-at-datastax: Thank you: that sounds like a good improvement.

src/main/java/io/stargate/sgv2/jsonapi/service/shredding/tables/RowShredder.java
tatu-at-datastax: Added something in class javadoc, but will also add it right here too.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/util/DataApiCommandSenderBase.java
tatu-at-datastax: Fair, will change.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/util/DataApiCommandSenders.java
tatu-at-datastax: Great idea, will change method name.

Could change class name too, to `DataApiCommandAssertions` too?

tatu-at-datastax: The only downside is that these look odd as set up components (creating table, inserting rows, that are not part of test).
Will change, we can re-think if need be.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/util/DataApiNamespaceCommandSender.java
tatu-at-datastax: Not quite: URL construction different. But I can extract that out, I think.

tatu-at-datastax: done, created "postInternal" for sub-classes to implement.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/util/DataApiResponseValidator.java
tatu-at-datastax: Hmmh. Good question. I am not 100% sure how rest-assured (underlying test library) does it.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/util/DataApiTableCommandSender.java
tatu-at-datastax: Done.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/util/DataApiResponseValidator.java
tatu-at-datastax: Ok, changed to immutable, create new copy.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/util/DataApiTableCommandSender.java
tatu-at-datastax: Ok, I'll keep this in mind as possible change -- the only concern I have is that short prefix can group related/similar methods together, which can be nice for auto-complete, method listing etc.

But I like the compactness of dropping "post-" as well.

src/test/java/io/stargate/sgv2/jsonapi/service/operation/filters/table/codecs/JSONCodecRegistryTest.java
tatu-at-datastax: That is unfortunate, but it looks like ParameterizedTest requires source method (I tried and non-static ones do not seem to work? LMK if you know of a way) to be `static`.
I'd prefer non-static as well so if there's a way, that'd be better.

src/test/java/io/stargate/sgv2/jsonapi/service/operation/filters/table/codecs/JSONCodecRegistryTest.java
tatu-at-datastax: Excellent point, changing

src/test/java/io/stargate/sgv2/jsonapi/service/operation/filters/table/codecs/JSONCodecRegistryTest.java
tatu-at-datastax: Yeah it's not exhaustive in many ways (i.e. does not try different ranges, values), but does try to cover permutations of 3 input "plain" Java types (BigDecimal, BigInteger, Long) x CQL integer types.

So if you can sanity check, that'd be great.

amorton (approved):
approved

vkarpov15 (commented):
LGTM

PR #1371, opened 2024-08-28
Error object v2

amorton:
**What this PR does**:

Initial error object v2

This is still not in use, it is dead code. Once we get it in we will move from the "playing" package into parent "exceptions" package once we start using it. 

This  PR is to get it into the main, next setps are: 
* start using it
* start making it return the errors on the API 

Fixes #<issue number>

**Checklist**
- [*] Changes manually tested
- [ * Documentation added/updated


pom.xml
```diff
@@ -64,6 +64,21 @@
         <type>pom</type>
         <scope>import</scope>
       </dependency>
+      <dependency>
+        <groupId>com.fasterxml.jackson.dataformat</groupId>
+        <artifactId>jackson-dataformat-yaml</artifactId>
+        <version>${jackson.version}</version>
```
tatu-at-datastax, line 70: Version can/should be removed, should come from `jackson-bom` (I'll remove)

tatu-at-datastax: Oh. This is in wrong section -- `<dependencyManagement>` is for versions, not dependencies: need to move to under `<dependencies>`. Will change.

pom.xml
```diff
@@ -135,6 +135,10 @@
       <groupId>io.quarkus</groupId>
       <artifactId>quarkus-hibernate-validator</artifactId>
     </dependency>
+    <dependency>
```
tatu-at-datastax, line 138: Moved this next to other Quarkus deps.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/CreateCollectionCommand.java
```diff
@@ -137,7 +137,7 @@ public record VectorizeConfig(
           @Nullable
               @Schema(
                   description =
-                      "Optional parameters that match the template provided for the provider",
+                      "Optional parameters that match the messageTemplate provided for the provider",
```
tatu-at-datastax, line 140: @amorton Is this intentional change or bulk-replace?

src/main/java/io/stargate/sgv2/jsonapi/exception/playing/APIException.java
```diff
@@ -0,0 +1,128 @@
...
+  // All errors default to 200 HTTP status code, because we have partial failure modes.
+  // There are some overrides, e.g. a server timeout may be a 500, this is managed in the
+  // error config. See ErrorTemplate.
+  public static final int DEFAULT_HTTP_RESPONSE = 200;
+
+  /**
+   * HTTP Response code for this error. NOTE: Not using enum from quarkus because do not want
+   * references to the HTTP framework this deep into the command processing
+   */
+  public final int httpResponse;
```
tatu-at-datastax, line 54: I think it may make sense to keep as `int` to avoid conversion on deserialization from YAML.

But FWTW, it wouldn't be (or have to be) Quarkus dep: standard Java JAX-RS has `jakarta.ws.rs.core.Response.Status` that is commonly used (Quarkus is a JAX-RS / Jakarta-RS implementation) as enum for HTTP status code.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/CreateCollectionCommand.java
tatu-at-datastax: I assume bulk replace since there's no `messageTemplate` around here, will revert

tatu-at-datastax: I assume it is intentional as there's further changes in `EmbeddingProvider`.

src/main/java/io/stargate/sgv2/jsonapi/exception/playing/APIException.java
```diff
@@ -0,0 +1,129 @@
...
+  }
+
+  public APIException(
+      ErrorFamily family, ErrorScope scope, String code, String title, String body) {
+    this(new ErrorInstance(UUID.randomUUID(), family, scope, code, title, body, Optional.empty()));
+  }
+
+  @Override
+  public CommandResponseError get() {
+    return null;
```
tatu-at-datastax, line 108: @amorton Why isn't this left abstract to require sub-classes to implement?

src/main/java/io/stargate/sgv2/jsonapi/exception/playing/APIException.java
```diff
@@ -0,0 +1,129 @@
+package io.stargate.sgv2.jsonapi.exception.playing;
```
tatu-at-datastax, line 1: @amorton I assume this is temporary package name, to be changed in near future?

src/main/java/io/stargate/sgv2/jsonapi/exception/playing/ErrorCode.java
```diff
@@ -0,0 +1,75 @@
...
+ * #get()} overloads for example:
+ *
+ * <pre>
+ *   FilterException.Code.MULTIPLE_ID_FILTER.get("variable_name", "value");
+ *   RequestException.Code.FAKE_CODE.get();
+ * </pre>
+ *
+ * @param <T> Type of the {@link APIException} the error code creates.
+ */
+public interface ErrorCode<T extends APIException> {
```
tatu-at-datastax, line 26: Does this type parameter make sense? If it is used, it needs to be carried along the whole call chain. And generally all we care about is that it we create an `APIException`, as all types have same properties and the only reason (I think) we even have sub-classes is to allow possibly catching and re-throwing different types (server vs request).

It seems to me that leaving out `T` would be simpler -- will try to see if that is true, locally.

tatu-at-datastax: @amorton Removal of this type parameter seems simple enough, was able to do that in couple of minutes (minus test classes), it's all mechanical. I don't think it adds any value.

src/main/java/io/stargate/sgv2/jsonapi/exception/playing/ServerException.java
```diff
@@ -0,0 +1,46 @@
...
+ * Base for any errors that are from the {@link ErrorFamily#SERVER} family, these are server side
+ * errors not related to the structure of the request itself.
+ *
+ * <p>Scope are defined in {@link Scope} and each represents a subclass of this class.
+ *
+ * <p>See {@link APIException}
+ */
+public class ServerException extends APIException {
+
+  public static final ErrorFamily FAMILY = ErrorFamily.SERVER;
```
tatu-at-datastax, line 13: This is problematic:

1. Should not really be `public` -- why would it be needed from outside (probably `protected` for sub-classes)
2. Name is very vague, esp. when referred to from sub-classes. Maybe `SERVER_EXCEPTION_FAMILY`
3. ... or is this constant needed at all? Just use `ErrorFamily.SERVER` directly in sub-class.

src/main/java/io/stargate/sgv2/jsonapi/exception/playing/DatabaseException.java
```diff
@@ -0,0 +1,25 @@
+package io.stargate.sgv2.jsonapi.exception.playing;
+
+public class DatabaseException extends ServerException {
+
+  public static final Scope SCOPE = Scope.DATABASE;
```
tatu-at-datastax, line 5: Same as with other constants:

1. Should not be `public`?
2. Name is very generic/vague, probably `DEFAULT_SCOPE`.
3. ... but is this constant actually needed at all? Seems like its only used from sub-class.

Not 100% sure what to do with it, but just seems wrong. (1 and 2 easy enough to change, but eliminating need would be best)

src/main/java/io/stargate/sgv2/jsonapi/exception/playing/DatabaseException.java
```diff
@@ -0,0 +1,25 @@
...
+    super(errorInstance);
+  }
+
+  public enum Code implements ErrorCode<DatabaseException> {
+    FAKE;
+
+    private final ErrorTemplate<DatabaseException> template;
+
+    Code() {
+      template = ErrorTemplate.load(DatabaseException.class, FAMILY, SCOPE, name());
```
tatu-at-datastax, line 17: Loading from Enum instance constructors is extremely fragile -- I think it happens at class loading time -- and could cause really gnarly failure modes.

But then again, as long as our test coverage is good, maybe that's not a big issue.

tatu-at-datastax: Ohhhh. Much worse -- this is called once per each Enum value? Does it really try to load something, or is that misnamed method? (to me, loading means reading YAML config file).

src/main/java/io/stargate/sgv2/jsonapi/exception/playing/ErrorCode.java
tatu-at-datastax: ... but I guess maybe I get it. Caller can get instance of specific exception from factory method.
So it can be cleaner.
I'll have to think more about this. Apologies for stream-of-consciousness kind of commentary :)

src/main/java/io/stargate/sgv2/jsonapi/exception/playing/ErrorScope.java
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * <p>This is used to group errors together in a {@link ErrorFamily}.
+ *
+ * <p>This interface is used because multiple ENUM's define the scopes, the interface creates a way
+ * to treat the values from the different ENUM's in a consistent way.
+ *
+ * <p>See {@link APIException}
+ */
+@FunctionalInterface
+public interface ErrorScope {
```
tatu-at-datastax, line 14: I don't understand this -- why not just a simple Enum? What is the point in this contraption?

src/main/java/io/stargate/sgv2/jsonapi/exception/playing/EmbeddingProviderException.java
```diff
@@ -0,0 +1,28 @@
+package io.stargate.sgv2.jsonapi.exception.playing;
+
+public class EmbeddingProviderException extends ServerException {
+  public EmbeddingProviderException(ErrorInstance errorInstance) {
+    super(errorInstance);
+  }
+
+  public enum Code implements ErrorCode<EmbeddingProviderException> {
```
tatu-at-datastax, line 8: Is the idea it's ok to have same "code" (like CLIENT_ERROR) for multiple scopes? I guess that is the case.

One concern is that to know which problem you have, it is no longer possible to group failures solely by Code, as that will conflate Codes with different Scopes.

src/main/java/io/stargate/sgv2/jsonapi/exception/playing/ErrorCode.java
```diff
@@ -0,0 +1,75 @@
...
+
+  /**
+   * Gets an instance of the {@link APIException} the error code represents without providing any
+   * substitution values for the error body.
+   *
+   * @return Instance of {@link APIException} the error code represents.
+   */
+  default T get() {
```
tatu-at-datastax, line 34: This is not a good method name. Get what?

So suggesting we rename to something like "toApiException()" or "asApiException()"

src/main/java/io/stargate/sgv2/jsonapi/exception/playing/ErrorCode.java
```diff
@@ -0,0 +1,75 @@
...
+  /**
+   * Gets an instance of the {@link APIException} the error code represents, providing substitution
+   * values for the error body as a param array.
+   *
+   * @param values Substitution values for the error body. The array length must be a multiple of 2,
+   *     each pair of strings is treated as a key-value pair for example ["key-1", "value-1",
+   *     "key-2", "value-2"]
+   * @return Instance of {@link APIException} the error code represents.
+   */
+  default T get(String... values) {
```
tatu-at-datastax, line 47: Do we need this method? We already have overload with `Map`, why/where do we need this?

src/main/java/io/stargate/sgv2/jsonapi/exception/playing/ErrorConfig.java
```diff
@@ -0,0 +1,327 @@
...
+ * Java objects to hold the config from the errors config yaml file for use by {@link ErrorTemplate}
+ *
+ * <p>See the {@link #DEFAULT_ERROR_CONFIG_FILE} for description of what the yaml should look like.
+ *
+ * <p>To use this, call {@link #getInstance()} in code, this will cause the yaml file to be loaded
+ * if needed. The file will be loaded from {@link #DEFAULT_ERROR_CONFIG_FILE} resource in the JAR.
+ * Then use {@link #getErrorDetail(ErrorFamily, String, String)} if you want the template, or the
+ * use {@link #getSnippetVars()} when running templates.
+ *
+ * <p>If you need more control use the {@link #initializeFromYamlResource(String)} before any code
```
tatu-at-datastax, line 38: ... not necessarily easy/possible to do, due to class loading/initialization constraints.

src/main/java/io/stargate/sgv2/jsonapi/exception/playing/ErrorConfig.java
```diff
@@ -0,0 +1,327 @@
...
+    return snippets;
+  }
+
+  /**
+   * See {@link #getErrorDetail(ErrorFamily, String, String)} to get a template for a specific error
+   * code.
+   *
+   * @return
+   */
+  public List<ErrorDetail> requestErrors() {
```
tatu-at-datastax, line 177: Why do we have this?

src/main/java/io/stargate/sgv2/jsonapi/exception/playing/ErrorConfig.java
```diff
@@ -0,0 +1,327 @@
...
+    return requestErrors;
+  }
+
+  /**
+   * See {@link #getErrorDetail(ErrorFamily, String, String)} to get a template for a specific error
+   * code.
+   *
+   * @return
+   */
+  public List<ErrorDetail> serverErrors() {
```
tatu-at-datastax, line 187: Why do we have this?

src/main/java/io/stargate/sgv2/jsonapi/exception/playing/ErrorConfig.java
```diff
@@ -0,0 +1,327 @@
...
+   */
+  public Optional<ErrorDetail> getErrorDetail(ErrorFamily family, String scope, String code) {
+
+    var errors =
+        switch (family) {
+          case REQUEST -> requestErrors;
+          case SERVER -> serverErrors;
+        };
+    return errors.stream()
+        .filter(e -> e.scope().equals(scope) && e.code().equals(code))
```
tatu-at-datastax, line 207: Should switch order of checks as second one much more discriminating (will do).

tatu-at-datastax (approved):
As per my comment above, conditionally approving, just need to address concerns before or after merging.

src/main/java/io/stargate/sgv2/jsonapi/exception/playing/EmbeddingProviderException.java
amorton: will add a fully qualified code to the API exception, we can then use it later for metrics etc

src/main/java/io/stargate/sgv2/jsonapi/exception/playing/ErrorCode.java
amorton: will look for better name - discussed

src/main/java/io/stargate/sgv2/jsonapi/exception/playing/ErrorConfig.java
amorton: note to be - revisit the comment with how this is used in testing

src/main/java/io/stargate/sgv2/jsonapi/exception/playing/ErrorScope.java
amorton: discussed - make it clear this is here because the scope is  disjoint sets for the diff families

src/main/java/io/stargate/sgv2/jsonapi/exception/playing/APIException.java
amorton: it is removed in later changes, was a POC placeholder

src/main/java/io/stargate/sgv2/jsonapi/exception/playing/UnresolvedErrorTemplateVariable.java
```diff
@@ -0,0 +1,35 @@
...
+ * that is not resolved by the values passed when the exception is created.
+ *
+ * <p>E.G. if the template is "too many ${foo}" and the values passed are "bar" and "baz" then this
+ * exception is raised.
+ *
+ * <p>This <b>SHOULD NOT</b> happen in production code, exceptions should be triggered in testing
+ * and this exception is raised if the template and the code do not match. Note that it is OK for
+ * the code to provide more variables than the template uses, but not the other way around.
+ */
+public class UnresolvedErrorTemplateVariable extends RuntimeException {
```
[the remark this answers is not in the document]
amorton: discussed, maybe later we can look at more base exceptions

src/main/java/io/stargate/sgv2/jsonapi/exception/playing/ServerException.java
amorton: idea was for subclasses to all reference the same definition of the family during construction, e.g. in the ctor for the Code enum

src/main/java/io/stargate/sgv2/jsonapi/exception/playing/DatabaseException.java
amorton: idea was for the Code enum to be able to refence a central definition while it is constructing

src/main/java/io/stargate/sgv2/jsonapi/exception/playing/DatabaseException.java
amorton: The ErrorTemplate will cause the ErrorConfig to load, which happens once and loads all of the YAML config at once and only once. 

Then this call to load() will fail if there is an error code defined in code but not in the yaml, so we want that failure to happen.

src/main/java/io/stargate/sgv2/jsonapi/exception/playing/APIException.java
```diff
@@ -0,0 +1,130 @@
...
+
+  /** The family of the error. */
+  public final ErrorFamily family;
+
+  /**
+   * Optional scope of the error, inside the family.
+   *
+   * <p>Never {@code null}, uses "" for no scope. See {@link ErrorScope}
+   */
+  public final String scope;
```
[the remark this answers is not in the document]
amorton: the scope list is different for each family so there is no single enum with them all. i.e. we do not use the FILTER scope with a SERVER family.

src/main/java/io/stargate/sgv2/jsonapi/exception/playing/ErrorCode.java
amorton: there to make it easy when there is only one or two variables to pass. The idea was to make it as easy as possible to instantiate exceptions

amorton:
follow up tickets, and there will soon be a PR that makes this usable that will move it from the "playing" package 

https://github.com/stargate/data-api/issues/1374
https://github.com/stargate/data-api/issues/1375

merging

PR #1380, opened 2024-08-29
Fix problem with ErrorCode initialization wrt config access

tatu-at-datastax:
**What this PR does**:

Fixes a problem with `ErrorCode` initialization occurring during IT runs; allows use of ErrorCode enum values again.

N/A

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCode.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCode.java
@@ -207,17 +207,6 @@ public enum ErrorCode {
   TABLE_COLUMN_UNKNOWN("Column unknown");
 
   private final String message;
-  private final boolean extendError =
-      ApiConstants.isOffline()
-          ? new SmallRyeConfigBuilder()
-              .withMapping(OperationsConfig.class)
-              .build()
-              .getConfigMapping(OperationsConfig.class)
-              .extendError()
-          : ConfigProvider.getConfig()
-              .unwrap(SmallRyeConfig.class)
-              .getConfigMapping(OperationsConfig.class)
-              .extendError();
 
   ErrorCode(String message) {
     this.message = message;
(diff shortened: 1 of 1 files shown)


src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCode.java
```diff
@@ -207,17 +207,6 @@ public enum ErrorCode {
   TABLE_COLUMN_UNKNOWN("Column unknown");
 
   private final String message;
-  private final boolean extendError =
```
tatu-at-datastax, line 210: Access from constructor of EVERY enum constant is pretty wasteful to begin with, but more importantly caused odd initialization failure from any Integration Test that tried to refer to `ErrorCode` enums.
Hence moved to be done lazily when needed, not eagerly before everything else.

src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCode.java
```diff
@@ -241,7 +230,7 @@ public JsonApiException toApiException(Throwable cause, String format, Object...
   }
 
   private String getErrorMessage(String format, Object... args) {
-    if (extendError) {
+    if (ExtendError.enabled()) {
```
tatu-at-datastax, line 233: First access to `getErrorMessage()` of any `ErrorCode` instance will load `ExtendError` helper class, which will get configuration settings. This is properly synchronized as guaranteed by JVM classloading, and all further access is via properly initialized singleton.

src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCode.java
```diff
@@ -254,4 +243,33 @@ public JsonApiException toApiException() {
...
     return new JsonApiException(this, message, null, httpStatus);
   }
+
+  /**
+   * Helper class to cache loading of settings from the configuration. This is used to avoid having
+   * to access Config during Enum class initialization. It will also prevent repeated configuration
+   * loading calls.
+   */
+  static class ExtendError {
+    private static final ExtendError instance = new ExtendError();
```
tatu-at-datastax, line 253: Instead of private static instance and accessor, could have exposed this as `public`. But looks to me slightly cleaner from caller this way.

Yuqi-Du (approved):

PR #1384, opened 2024-08-30
Fix #1383: map most recognized SERVER_ errors to 504 or 502, leave 500 for unrecognized

### LINKED ISSUE #1383: Exceptions from Database (via CQL Driver) incorrectly mapped to HTTP 500, should be 504 (or 502)
HTTP response 500 indicates an Internal Server error (see f.ex https://en.wikipedia.org/wiki/List_of_HTTP_status_codes). There are some failure modes that should be mapped to this code, for example otherwise unhandled exceptions.

But we have many exceptions thrown by CQL driver, caught and re-processed by `ThrowableToErrorMapper`, where we use code 500 as well.
These would much better map to 504 (Gateway Timeout) or sometimes 502 (Bad Gateway).

The reason for changing these is to improve operational support: 500 means something particularly bad happened; whereas 504 generally means problem due to a system Data API depends on (DB or possibly Embedding Gateway).

tatu-at-datastax:
**What this PR does**:

Maps most former 500 http failures to 504 or 502, to indicate these are recognized failures with backend DB; 500 remains for unknown/unrecognized exceptions

Fixes #1383

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/exception/mappers/ThrowableToErrorMapper.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/exception/mappers/ThrowableToErrorMapper.java
@@ -41,6 +41,7 @@ public final class ThrowableToErrorMapper {
         // if our own exception, shortcut
         if (throwable instanceof JsonApiException jae) {
           if (jae.getErrorCode().equals(ErrorCode.SERVER_EMBEDDING_GATEWAY_NOT_AVAILABLE)) {
+            // 30-Aug-2021: [data-api#1383] Why is this special case here?
             return jae.getCommandResultError(message, Response.Status.INTERNAL_SERVER_ERROR);
           }
           return jae.getCommandResultError(message, jae.getHttpStatus());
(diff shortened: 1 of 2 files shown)


src/test/java/io/stargate/sgv2/jsonapi/service/operation/collections/FindCollectionOperationTest.java
```diff
@@ -3018,15 +3018,15 @@ public void readFailureException() throws UnknownHostException {
           ThrowableToErrorMapper.getMapperWithMessageFunction()
               .apply(failure, failure.getMessage());
       AssertionsForClassTypes.assertThat(error).isNotNull();
-      AssertionsForClassTypes.assertThat(error.message())
```
tatu-at-datastax, line 3021: Just reordered to start with more general tests (error code, exception class). And then changed testing for message according to change in `ThrowableToErrorMapper`

src/main/java/io/stargate/sgv2/jsonapi/exception/mappers/ThrowableToErrorMapper.java
```diff
@@ -88,17 +89,21 @@ private static CommandResult.Error handleDriverException(
       return handleAllNodesFailedException((AllNodesFailedException) throwable, message);
     } else if (throwable instanceof ClosedConnectionException) {
       return ErrorCode.SERVER_CLOSED_CONNECTION
-          .toApiException(message)
-          .getCommandResultError(Response.Status.INTERNAL_SERVER_ERROR);
+          .toApiException("(DriverException/ClosedConnectionException) %s", message)
+          .getCommandResultError(Response.Status.GATEWAY_TIMEOUT);
```
Hazel-Datastax, line 93: Why here is `GATEWAY_TIMEOUT` not `BAD_GATEWAY`?

tatu-at-datastax: Let me double-check; you are right seems odd.

tatu-at-datastax: @Hazel-Datastax yes, you are right: should be BAD_GATEWAY -- changing. Thank you!

maheshrajamani (approved):
LGTM

Hazel-Datastax (approved):

vkarpov15 (commented):
LGTM

PR #1355, opened 2024-08-22
Fix #1349: support "non-standard" column names

tatu-at-datastax:
**What this PR does**:

Fix issue(s) in filtering (and insertion) by column names that require quoting (have upper-case letters; non-alphanumerics, start with underscore etc etc)

**Which issue(s) this PR fixes**:
Fixes #1349

**Checklist**
- [ ] Changes manually tested
- [x] Automated Tests added/updated
- [ ] Documentation added/updated
- [x] CLA Signed: [DataStax CLA](https://cla.datastax.com/)

PR #1390, opened 2024-09-04
Fix #1355: add general-purpose Feature flag system

tatu-at-datastax:
**What this PR does**:

Adds extensible Feature-Flag system to allow both default, system prop setting, and per-request overrides.

Initially just a single feature is added, "tables":

* As YAML Config entry: `stargate.feature.flags.tables`
* As HTTP header: `Feature-Flag-tables` (case-insensitive)

That is: when starting; may specific fixed enabled/disabled via `application.yaml` or system property; latter like:

```
./mvnw quarkus:dev -Dstargate.jsonapi.operations.database-config.local-datacenter=dc1 \
  -Dquarkus.log.console.darken=2  -Dstargate.feature.flags.tables=true
```

but if no setting specified (and ONLY in that case), per-request `curl` header like:

```
curl -i --location --request POST ... --header 'Feature-Flag-tables: true'
```

can be used to enable/disable processing

Fixes #1355

**Checklist**
- [x] Changes manually tested


src/main/resources/application.yaml
```diff
@@ -25,6 +25,10 @@ stargate:
   exception-mappers:
     enabled: false
 
+  feature:
+    flags:
+      TABLES: null
```
tatu-at-datastax, line 30: `null` meaning "no default", allowing per-request override, whereas `true` would mean enabled and `false` disabled (neither allowing per-request overrides)

tatu-at-datastax: Actually, better choice is to leave it empty; either works.

tatu-at-datastax:
Ready for initial review: need to think about how this could be tested better.

CONFIGURATION.md
```diff
@@ -71,10 +71,17 @@ Other Quarkus properties that are specifically relevant for the service:
 
 
 ## Command level logging configuration
-*Configuration for command level logging, defined by [CommandLoggingConfig.java](src/main/java/io/stargate/sgv2/jsonapi/config/CommandLoggingConfig.java).*
+*Configuration for command level logging, defined by [CommandLevelLoggingConfig.java](src/main/java/io/stargate/sgv2/jsonapi/config/CommandLoggingConfig.java).*
```
tatu-at-datastax, line 74: Just fixed incorrect class name I noticed.

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/NamespaceCache.java
```diff
@@ -22,8 +22,6 @@ public class NamespaceCache {
 
   private final ObjectMapper objectMapper;
 
-  private final boolean apiTablesEnabled;
```
tatu-at-datastax, line 25: Had to remove Tables-enabled feature from here because we have to be able enable/disable it on per-request basis. So we will always load regular Table schema find, and fail request at a later point.
Seems more correct overall anyway; checking here would be too low level.

src/main/java/io/stargate/sgv2/jsonapi/config/feature/DataApiFeatureFlag.java
```diff
@@ -0,0 +1,56 @@
...
+   * commands, and support Collection commands on Tables. If disabled, those operations will fail
+   * with {@link ErrorCode#TABLE_FEATURE_NOT_ENABLED}.
+   *
+   * <p>If no configuration specified (config or request), the feature will be Disabled.
+   */
+  TABLES("tables", false);
+
+  private final String featureName;
+
+  private final boolean enabledByDefault;
```
amorton, line 25: The command above says "If no configuration specified (config or request), the feature will be Disabled."

consider: we should prob enforce this by default in the enum and remove the enabledByDefault setting. Concern would be that we end up with a mix of values and it gets confusing.

src/main/java/io/stargate/sgv2/jsonapi/config/feature/DataApiFeatureFlag.java
```diff
@@ -0,0 +1,56 @@
...
+
+  /**
+   * HTTP header name to be used to override the feature flag for a specific request: lower-case,
+   * prefixed with "x-stargate-feature-"; lookup case-insensitive.
+   */
+  private final String featureNameAsHeader;
+
+  DataApiFeatureFlag(String featureName, boolean enabledByDefault) {
+    this.featureName = featureName;
+    featureNameAsHeader = "x-stargate-feature-" + featureName;
```
amorton, line 35: can we remove "stargate" pls, happy to use "datapi" or nothing. 

also "x" prefixing has been deprecated , are we sure we want to use it ? https://datatracker.ietf.org/doc/html/rfc6648

Consider "Feature-Flag" ?

src/main/java/io/stargate/sgv2/jsonapi/config/feature/DataApiFeatureFlag.java
```diff
@@ -0,0 +1,56 @@
...
+  private final boolean enabledByDefault;
+
+  /**
+   * HTTP header name to be used to override the feature flag for a specific request: lower-case,
+   * prefixed with "x-stargate-feature-"; lookup case-insensitive.
+   */
+  private final String featureNameAsHeader;
+
+  DataApiFeatureFlag(String featureName, boolean enabledByDefault) {
+    this.featureName = featureName;
```
amorton, line 34: consider forcing lcase here, because the feature name is used raw in the header name and other places.

src/main/java/io/stargate/sgv2/jsonapi/config/feature/DataApiFeatures.java
```diff
@@ -0,0 +1,43 @@
...
+ * overrides) and possible per-request settings.
+ */
+public class DataApiFeatures {
+  private final Map<DataApiFeatureFlag, Boolean> fromConfig;
+  private final DataApiRequestInfo.HttpHeaderAccess httpHeaders;
+
+  DataApiFeatures(
+      Map<DataApiFeatureFlag, Boolean> fromConfig,
+      DataApiRequestInfo.HttpHeaderAccess httpHeaders) {
+    this.fromConfig = (fromConfig == null) ? Collections.emptyMap() : fromConfig;
```
amorton, line 18: what is the situation where the fromConfig is null ? could this lead to a subtle bug where the config is missed ?

consider enforcing non-null

nit: also, consider making ctor private if we want to construct via the static factories.

src/main/java/io/stargate/sgv2/jsonapi/config/feature/DataApiFeatures.java
```diff
@@ -0,0 +1,43 @@
...
+  public static DataApiFeatures fromConfigAndRequest(
+      DataApiFeatureConfig config, DataApiRequestInfo.HttpHeaderAccess httpHeaders) {
+    return new DataApiFeatures(config.flags(), httpHeaders);
+  }
+
+  public boolean isFeatureEnabled(DataApiFeatureFlag flag) {
+    // First check if there is definition from configuration
+    Boolean b = fromConfig.get(flag);
+
+    if (b == null) {
```
amorton, line 35: if I read this correctly, if the feature is disabled in config it is not possible to enable via request. You mention the three ways for controlling features the other day, may be best to spell this out very clearly at the top fo this class.

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/NamespaceCache.java
amorton: sounds reasonable

src/main/resources/application.yaml
```diff
@@ -25,6 +25,10 @@ stargate:
   exception-mappers:
     enabled: false
 
+  feature: # See io.stargate.sgv2.jsonapi.config.feature.DataApiFeatureConfig
```
amorton, line 28: not: can we expand the docs here, or point to expanded docs on the DataApiFeatures class to explain the logic for setting values.

src/test/java/io/stargate/sgv2/jsonapi/TestConstants.java
```diff
@@ -40,15 +41,21 @@ public final class TestConstants {
 
   // CommandContext for working on the schema objects above
 
+  public static final DataApiFeatures DEFAULT_API_FEATURES_FOR_TESTS = DataApiFeatures.empty();
```
amorton, line 44: question - how easy / desirable is it to be able to turn on features in a DataApiFeatures class so we can run unit test without needing the things the happen in the DseTestResource class ?

amorton (changes requested):

amorton (changes requested):
generally like where this is going, just comments on names and some details.

src/main/java/io/stargate/sgv2/jsonapi/api/request/DataApiRequestInfo.java
```diff
@@ -18,6 +18,7 @@ public class DataApiRequestInfo {
   private final Optional<String> tenantId;
   private final Optional<String> cassandraToken;
   private final EmbeddingCredentials embeddingCredentials;
+  private final HttpHeaderAccess httpHeaders;
```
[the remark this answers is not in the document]
tatu-at-datastax: Yes, agreed, should change as suggested. Have to think about it a bit ...

src/main/java/io/stargate/sgv2/jsonapi/api/request/DataApiRequestInfo.java
```diff
@@ -54,4 +57,39 @@ public Optional<String> getCassandraToken() {
...
+   */
+  public static class HttpHeaderAccess {
+    private final io.vertx.core.MultiMap headers;
+
+    public HttpHeaderAccess(io.vertx.core.MultiMap headers) {
+      this.headers = headers;
+    }
+
+    /**
+     * Accessor for getting value of given header (case-insensitive), as {@code Boolean} if (and
```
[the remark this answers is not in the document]
tatu-at-datastax: Yes, is due to vertx http server impl that does it for HTTP standard reason.

What I am thinking is that we only document use of lower-case header name for usage, and leave ability for case-insensitivity as impl detail: this since we are not exposing features to actual end users. If case-insensitivity wasn't provided I wouldn't bother actually adding it here.

src/main/java/io/stargate/sgv2/jsonapi/config/feature/DataApiFeatureConfig.java
```diff
@@ -0,0 +1,13 @@
...
+
+import io.smallrye.config.ConfigMapping;
+import java.util.Map;
+
+/**
+ * Configuration mapping for Data API Feature flags as read from main application configuration
+ * (with possible property / sysenv overrides).
+ */
+@ConfigMapping(prefix = "stargate.feature")
+public interface DataApiFeatureConfig {
```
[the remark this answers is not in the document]
tatu-at-datastax: Makes sense, will change.

src/main/java/io/stargate/sgv2/jsonapi/config/feature/DataApiFeatureFlag.java
tatu-at-datastax: Problem is that precedence is 3-layered:

1. Explicit value from FeaturesConfig (map keyed by feature)
2. Explicit value from request header if no (1)
3. Default if neither (1) or (2).

I originally thought of just leaving (3) to be `false` for simplicity; I can still do that and drop `enabledByDefault` as it is not strictly needed -- and can be added if/when needed (or something else that does what is needed).

src/main/java/io/stargate/sgv2/jsonapi/config/feature/DataApiFeatureFlag.java
tatu-at-datastax: I'll add validation.

src/main/java/io/stargate/sgv2/jsonapi/config/feature/DataApiFeatures.java
tatu-at-datastax: Can/will change to `private`. We get "config" from Quarkus, but I can add `null` check in caller -- but I think it probably better be there just since config mapping might indeed get `null` if our `application.yaml` has empty value (which is reasonable if we do not define any enabled features).

src/main/java/io/stargate/sgv2/jsonapi/api/request/DataApiRequestInfo.java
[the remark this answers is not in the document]
tatu-at-datastax: Hmmh. While I'd prefer all lower-case, looks like general best practice seems to be mixed case.

src/main/resources/application.yaml
tatu-at-datastax: Fixed link wrt name, but thinking of documenting only/mostly on code side.

src/test/java/io/stargate/sgv2/jsonapi/TestConstants.java
tatu-at-datastax: I am thinking that using `DseTestResource` reflects actual usage better, and as long as we do not have tons of testing for alternative setting (that is, most tests can use specific default), works for now.

Otherwise we may end up adding test scaffolding that differs from actual running.
This wrt ITs.

tatu-at-datastax:
Ok: situation at the end of week:

1. Things work wrt Config, Header overrides, with manual testing
2. Addressed doc/comment suggestions
3. Figured out how to do ITs, to verify same as (1)
4. Ran out of time to address one last major thing ("I really want to get the DataApiRequestInfo onto the CommandContext and remove the dependency injection. ") -- hoping to check that on monday.
5. Aside from (4), ready for re-review!

src/main/resources/application.yaml
amorton: works for me

src/main/java/io/stargate/sgv2/jsonapi/api/request/DataApiRequestInfo.java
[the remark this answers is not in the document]
amorton: ok, we def should not block this PR , it's a nice to have

src/main/java/io/stargate/sgv2/jsonapi/api/request/DataApiRequestInfo.java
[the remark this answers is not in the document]
amorton: I am on the edge of my seat here, seems to be ??? seems to be...

src/main/java/io/stargate/sgv2/jsonapi/config/feature/ApiFeature.java
```diff
@@ -0,0 +1,73 @@
...
+   * with {@link ErrorCodeV1#TABLE_FEATURE_NOT_ENABLED}.
+   *
+   * <p>If no configuration specified (config or request), the feature will be Disabled.
+   */
+  TABLES("tables", false);
+
+  /**
+   * Prefix for HTTP headers used to override feature flags for specific requests: prepended before
+   * {@link #featureName}, so f.ex for {@link #TABLES} flag, the header name would be {@code
+   * Feature-Flag-tables}.
```
vkarpov15, line 32: Note to self that this is the header name

vkarpov15 (commented):
LGTM :+1:

src/main/java/io/stargate/sgv2/jsonapi/config/feature/DataApiFeatureFlag.java
amorton: discussed - drop the enabledByDefault and always default to false. the mental model is these are named boolean values, that default to false.

amorton:
reviewed in person, Done for me . Will approve, and then we need to get the error object V2 flag to use features config

tatu-at-datastax:
Last changes; will merge once CI passes.

src/main/java/io/stargate/sgv2/jsonapi/api/request/DataApiRequestInfo.java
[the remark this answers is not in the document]
tatu-at-datastax: Will come back for a look later today but will merge this PR now to unblock other work.

src/main/java/io/stargate/sgv2/jsonapi/api/request/DataApiRequestInfo.java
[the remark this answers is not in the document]
tatu-at-datastax: hmmmh. Not sure why your GUI showed "seems to be" without second line ("mixed case"). Or is that secret line just shown to me? :-o
(I saw it missing when you screen shared)

PR #1394, opened 2024-09-09
Add more DriverExceptionHandler to new error code structure

Hazel-Datastax:
**What this PR does**:
Add more DriverExceptionHandler to new error code structure

Fixes [Jira C2-3681](https://datastax.jira.com/browse/C2-3681)

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/exception/DatabaseException.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/exception/DatabaseException.java
@@ -9,8 +9,8 @@ public class DatabaseException extends ServerException {
   }
 
   public enum Code implements ErrorCode<DatabaseException> {
-    TABLE_WRITE_TIMEOUT,
-    CLOSED_CONNECTION;
+    CLOSED_CONNECTION,
+    TABLE_WRITE_TIMEOUT;
 
     private final ErrorTemplate<DatabaseException> template;
 
--- a/src/main/java/io/stargate/sgv2/jsonapi/exception/DatabaseException.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/exception/DatabaseException.java
@@ -10,6 +10,7 @@ public class DatabaseException extends ServerException {
 
   public enum Code implements ErrorCode<DatabaseException> {
     CLOSED_CONNECTION,
+    DRIVER_TIMEOUT,
     TABLE_WRITE_TIMEOUT;
 
     private final ErrorTemplate<DatabaseException> template;
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/DefaultDriverExceptionHandler.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/DefaultDriverExceptionHandler.java
@@ -2,6 +2,7 @@ package io.stargate.sgv2.jsonapi.service.cqldriver.executor;
 
 import static io.stargate.sgv2.jsonapi.exception.ErrorFormatters.errVars;
 
+import com.datastax.oss.driver.api.core.DriverTimeoutException;
 import com.datastax.oss.driver.api.core.connection.ClosedConnectionException;
 import io.stargate.sgv2.jsonapi.config.constants.ErrorObjectV2Constants.TemplateVars;
 import io.stargate.sgv2.jsonapi.exception.DatabaseException;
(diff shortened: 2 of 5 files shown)


src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/DefaultDriverExceptionHandler.java
```diff
@@ -33,4 +42,136 @@ public RuntimeException handle(SchemaT schemaObject, ClosedConnectionException e
     return DatabaseException.Code.CLOSED_CONNECTION.get(
         errVars(schemaObject, map -> map.put(TemplateVars.ERROR_MESSAGE, exception.getMessage())));
   }
+
+  @Override
+  public RuntimeException handle(SchemaT schemaObject, DriverTimeoutException exception) {
+    // TODO(Hazel): Aaron said "DRIVER" is a bad code.
```
tatu-at-datastax, line 48: Agree that "DRIVER" would be too vague, but "DRIVER_TIMEOUT" might be misleading -- maybe something like "DRIVER_UKNOWN_FAILURE" or something? This is assuming that what we map here is something nothing else recognizes, and that should eventually be mapped to something specific.

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/DefaultDriverExceptionHandler.java
```diff
@@ -33,4 +42,136 @@ public RuntimeException handle(SchemaT schemaObject, ClosedConnectionException e
...
+    if (message.contains(
+            "If you want to execute this query despite the performance unpredictability, use ALLOW FILTERING")
+        || message.contains("ANN ordering by vector requires the column to be indexed")) {
+      // TODO(Hazel): Original code is NO_INDEX_ERROR, I think we need to change but am not sure
+      // what to change
+      return exception;
+    }
+    if (message.contains("vector<float,")) {
+      // It is tricky to find the actual vector dimension from the message, include as-is
+      // TODO(Hazel): the code VECTOR_SIZE_MISMATCH was added recently, should we keep using it?
```
tatu-at-datastax, line 120: I think that'd be good because it is very common failure, from ops/metrics -- we do not want to lose this information wrt metrics (leads to unnecessary investigation for common failures).

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/DefaultDriverExceptionHandler.java
```diff
@@ -33,4 +42,136 @@ public RuntimeException handle(SchemaT schemaObject, ClosedConnectionException e
...
+      return exception;
+    }
+    // Reuse the default method in the interface
+    return DriverExceptionHandler.super.handle(schemaObject, exception);
```
tatu-at-datastax, line 124: This seems somehow like abuse of default interface methods, having to specify interface of which default impl to use (but I don't know of a better way without using base class instead of interface).

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/DefaultDriverExceptionHandler.java
```diff
@@ -33,4 +42,136 @@ public RuntimeException handle(SchemaT schemaObject, ClosedConnectionException e
...
+
+  @Override
+  public RuntimeException handle(SchemaT schemaObject, UnauthorizedException exception) {
+    return AuthException.Code.INVALID_TOKEN.get(errVars(exception));
+  }
+
+  @Override
+  public RuntimeException handle(SchemaT schemaObject, ReadFailureException exception) {
+    return DatabaseException.Code.READ_FAILURE.get(
+        errVars(
```
tatu-at-datastax, line 135: Not part of this PR, but I really think it'd be cleaner to have `errVars` that takes schema object and `Map` for extra variables to add, instead of closure for adding variables.

src/main/resources/errors.yaml
```diff
@@ -152,7 +163,80 @@ server-errors:
...
+      ${SNIPPET.RETRY}
+  - scope: DATABASE
+    code: READ_TIMEOUT
+    http-status-override: 504
+    title: Database read timed out
+    body: |-
+      The Data API timed out while reading from the database.
+
+      Reading from the ${schemaType} ${keyspace}.${table} failed to complete successfully. 
+      #TODO(Hazel): ReadTimeoutException has multiple ways to format the error message. What message should we use?
```
tatu-at-datastax, line 209: Probably don't want to have comment within text block? (it will be included as text, not comment)

Hazel-Datastax: Yes, will remove it later - just want to leave the comments at the nearest place

src/main/java/io/stargate/sgv2/jsonapi/exception/DatabaseException.java
```diff
@@ -9,8 +9,14 @@ public DatabaseException(ErrorInstance errorInstance) {
   }
 
   public enum Code implements ErrorCode<DatabaseException> {
+    CLOSED_CONNECTION,
```
amorton, line 12: hold on merging this - we need to work out how to hide errors like the ones below from the users.

- CLOSED_CONNECTION,
- DRIVER_TIMEOUT,
    NO_NODE_AVAILABLE,

also, it is sloppy to have a TABLE_WRITE_TIMEOUT, and WRITE_TIMEOUT; we need to work out if these are table specific errors or not.

amorton (changes requested):
HOLD until we can discuss what errors we want to return to the users, see my one comment

Issue #1376, opened 2024-08-28
Add new keyspace commands, deprecate namespace commands


jeffreyscarpenter:
As described in #1373:

Add keyspace operations to the API to parallel the namespace operations, and mark the namespace operations as deprecated.
- FindNamespaces -> FindKeyspaces
- CreateNamespace → CreateKeyspace
- DropNamespace → DropKeyspace
- Return a warning message in the response of the keyspace operations that indicates use of a deprecated command.
- One way to do this in code: rename existing operations to their new names, add operations with the old names, marked as deprecated, which act as aliases to the new operations.
- Make sure all metrics/logging reference new operation names. We do not want to capture metrics/logging for the old operation names.
- Make sure swagger/OpenAPI marks old operation names as deprecated.

Yuqi-Du:
Note, 
What differs to the original ticket is that **we decide to keep the deprecated commands metrics and logs** . So, two different metrics and logs even these commands are the 'same'.

E.G. createNamespaceCommand, createKeyspaceCommand.
e,g.

command_processor_process_seconds_count{command="CreateKeyspaceCommand",error="false",error_class="NA",error_code="NA",module="sgv2-jsonapi",sort_type="NONE",tenant="unknown",vector_enabled="false",} 1.0

command_processor_process_seconds_count{command="CreateNamespaceCommand",error="false",error_class="NA",error_code="NA",module="sgv2-jsonapi",sort_type="NONE",tenant="unknown",vector_enabled="false",} 1.0

Yuqi-Du:
Regarding warning msg, here is the example:

{
    "status": {
        "namespaces": [
            "system_auth",
            "system_schema",
        ]
        "warnings": ["warning 1", "warning 2"]
    }
}
the change is, we add a command status key called warnings that is a list / json array of strings. Include the deprecation notice in that, and we can re-use this status field with API Tables to warn about missing indexes.

jeffreyscarpenter:
@Yuqi-Du can we close this issue now since the linked PR is merged?

PR #1397, opened 2024-09-09
deprecate namespace, convert to keyspace

Yuqi-Du:
Background: [Removal of Namespace terminology in Data API ](https://github.com/stargate/data-api/discussions/1373) , [issue](https://github.com/stargate/data-api/issues/1376)

This PR will have these related changes:

- deprecate namespace command, add keyspace command

We will still have deprecated commands for backwards compatibility, but return a warning message when used. And that will be **added in the status map of commandResult**. 

findNamespaces -> findKeyspaces
Full example
```
{
    "status": {
        "namespaces": [
            "fixs",
            "system_auth",
            "system_schema",
        ],
        "warnings": [
            "This FindNamespacesCommand has been deprecated and will be removed in future releases, use FindKeyspacesCommand instead."
        ]
    }
}
```

createNamespace -> createKeyspace
` This CreateNamespaceCommand has been deprecated and will be removed in future releases, use CreateKeyspaceCommand instead.`
dropNamespace -> dropKeyspace
`This DropNamespaceCommand has been deprecated and will be removed in future releases, use DropKeyspaceCommand instead.`

- metrics and logs

What differs to the original ticket is that we decide to **keep the deprecated commands metrics and logs** . So, two different metrics and logs even these commands are the 'same'. 
E.G. createNamespaceCommand, createKeyspaceCommand.
e,g.
```
command_processor_process_seconds_count{command="CreateKeyspaceCommand",error="false",error_class="NA",error_code="NA",module="sgv2-jsonapi",sort_type="NONE",tenant="unknown",vector_enabled="false",} 1.0

command_processor_process_seconds_count{command="CreateNamespaceCommand",error="false",error_class="NA",error_code="NA",module="sgv2-jsonapi",sort_type="NONE",tenant="unknown",vector_enabled="false",} 1.0
```

One thing to note here, I also changed micrometer binder path in application.yaml file. 
Means we have uri tag changed for
`http_server_requests_seconds{method="POST",module="sgv2-jsonapi",outcome="SUCCESS",status="200",uri="/v1/{namespace}/{collection}",quantile="0.5",} 0.067108864`
to 
`http_server_requests_seconds{method="POST",module="sgv2-jsonapi",outcome="SUCCESS",status="200",uri="/v1/{keyspace}/{collection}",quantile="0.5",} 0.035651584`
The only impacted place I think is HTTP section in the data api detailed grafana dashboard. @jeffreyscarpenter 
image
image

- OpenAPI and Swagger

Path has been changed to `v1/keyspace` and `v1/keyspace/collection` in swagger UI.
All the command examples have been changed to new commands (findKeyspaces/dropKeyspace/createKeyspace)
There are schemas for deprecated commands, but have been marked as deprecated. 
image

- Tests

Added tests for corresponding new command, keep deprecated commands tests(since they still exist for backwards support)

- Terminology

We have namespace term in the repo codebase. I did not try to refactor them all to keyspace, since it will make this PR super huge and noisy. Currently, I refactored terms for src classes, some tests files and base class. 

**Checklist**
- [ ] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/StargateJsonApi.java
```diff
@@ -21,7 +21,7 @@
     info = @Info(title = "", version = ""),
     tags = {
       @Tag(name = "General", description = "Executes general commands."),
-      @Tag(name = "Namespaces", description = "Executes namespace commands."),
+      @Tag(name = "Keyspaces", description = "Executes keyspaces commands."),
```
tatu-at-datastax, line 24: typo: "Executes keyspaces commands." -> "Executes keyspace commands."

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/CommandResult.java
```diff
@@ -47,15 +47,17 @@ public record CommandResult(
                   nullable = true)
             })
         Map<CommandStatus, Object> status,
-    @JsonInclude(JsonInclude.Include.NON_EMPTY) @Schema(nullable = true) List<Error> errors) {
+    @JsonInclude(JsonInclude.Include.NON_EMPTY) @Schema(nullable = true) List<Error> errors,
+    @Schema(description = "Deprecated command warning message.", nullable = true)
```
tatu-at-datastax, line 51: This probably needs to be discussed, whether we want another main-level property instead of a new entry within "status"

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/CreateNamespaceCommand.java
```diff
@@ -12,7 +12,12 @@
 import javax.annotation.Nullable;
 import org.eclipse.microprofile.openapi.annotations.media.Schema;
 
-@Schema(description = "Command that creates a namespace.")
+/**
+ * Command {@link CreateNamespaceCommand} is deprecated, please switch to {@link
+ * CreateKeyspaceCommand} Support it for backward-compatibility
+ */
+@Schema(
+    description = "Command that creates a namespace.(Deprecated, please switch to createKeyspace.)")
```
tatu-at-datastax, line 20: nit/formatting: should be formatted like:

description = "Command that creates a namespace (deprecated, please switch to createKeyspace).")

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/CommandResolver.java
```diff
@@ -150,10 +151,15 @@ default void addToMetrics(
       Command command,
       LogicalExpression logicalExpression,
       IndexUsage baseIndexUsage) {
-    // TODO: this functions hould not be on the CommandResolver interface, it has nothing to do with
+    // TODO: this functions should not be on the CommandResolver interface, it has nothing to do
```
tatu-at-datastax, line 154: type: "functions" -> "function"

src/main/java/io/stargate/sgv2/jsonapi/util/DeprecatedCommandUtil.java
```diff
@@ -0,0 +1,62 @@
...
+   * "deprecatedCommandWarning":"Warning: deprecated command \"FindNamespacesCommand\", please
+   * switch to \"FindKeyspacesCommand\"." "deprecatedCommandWarning":"Warning: deprecated command
+   * \"FindNamespaces\", please switch to \"FindKeyspaces\"."
+   *
+   * @param commandContext commandContext that has original command name
+   * @return a message states using of deprecated command
+   */
+  public static <U extends SchemaObject> String getDeprecatedCommandMsg(
+      CommandContext<U> commandContext) {
+    Log.error("sdad " + commandContext.commandName());
```
tatu-at-datastax, line 45: Probably left by accident, should be removed.

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/CommandResolver.java
```diff
@@ -150,10 +151,15 @@ default void addToMetrics(
...
+    // with
     // that
     // it's only here because of the use of records and interfaces, move to a base class
-    Tag commandTag = Tag.of(jsonApiMetricsConfig.command(), command.getClass().getSimpleName());
+    Tag commandTag =
+        Tag.of(
+            jsonApiMetricsConfig.command(),
+            DeprecatedCommandUtil.maybeResolveDeprecatedCommandName(
```
tatu-at-datastax, line 161: Could resolve actual command `class` instead of name?

src/test/java/io/stargate/sgv2/jsonapi/api/v1/AbstractKeyspaceIntegrationTestBase.java
```diff
@@ -22,35 +22,35 @@
...
+ * automatically creates a keyspace before all tests and drops a keyspace after all test have been
  * run.
  *
  * <p>Note that this test uses a small workaround in {@link #getTestPort()} to avoid issue that
  * Quarkus is not setting-up the rest assured target port in the @BeforeAll and @AfterAll methods
  * (see https://github.com/quarkusio/quarkus/issues/7690).
  */
 @TestInstance(TestInstance.Lifecycle.PER_CLASS)
-public abstract class AbstractNamespaceIntegrationTestBase {
+public abstract class AbstractKeyspaceIntegrationTestBase {
```
tatu-at-datastax, line 34: idea: to reduce size of this PR, could create separate PR that only renames this test base class and references to `namespaceName` (to `keyspace`) -- this is internal, non-user-visible change, and could be merged before any other changes. But would remove dozens of files where the only change is for naming.

src/main/java/io/stargate/sgv2/jsonapi/StargateJsonApi.java
Yuqi-Du: solved, thanks!

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/CreateNamespaceCommand.java
Yuqi-Du: solved, thanks

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/CommandResolver.java
Yuqi-Du: solved, thanks

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/CommandResult.java
Yuqi-Du: Thanks, that should be discussed. @amorton

src/test/java/io/stargate/sgv2/jsonapi/api/v1/AbstractKeyspaceIntegrationTestBase.java
Yuqi-Du: Thanks, let me separate from this PR.

amorton:
Discussed, pls change the response to look like below: 

```
{
    "status": {
        "namespaces": [
            "system_auth",
            "system_schema",
        ]
        "warnings": ["warning 1", "warning 2"]
    }
}
```

the change is, we add a command status key called `warnings` that is a list / json array of strings. Include the deprecation notice in that, and we can re-use this status field with API Tables to warn about missing indexes. 

 create another ticket to (after this is merged) generate the message using the new APIExceptions code so we can template the message. Use a new "WARNING" family, no scope, create the APIException and use the body / message .

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/CommandResult.java
Yuqi-Du: please refer to the comment from Aaron in this PR.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/CommandResult.java
```diff
@@ -86,6 +87,21 @@ public CommandResult(List<Error> errors) {
     this(null, null, errors);
   }
 
+  /**
+   * construct a new CommandResult with warnings added in status map
```
amorton, line 91: This is not a constructor , it is not clear that this method returns a new instance without modifying the original. e.g. add() on a list mutates the list

I could imagine that the issue could be that the status map is immutable.  

two options,: 

1. If we want to be immutable, then rename of withWarning(String) and update comments to make it clear new object is returned. I think it unlikely we will add many so ok to just pass string , not list. 
2. If we want to be mutable, stick with addWarning(string) (again no need for list here). 

In either case, enforce this with a canonical constructor for the record that makes defensive copies and of the status and errors, either immutable or mutable. And same for MultiResponseData as we are here, just List,.copyOf() we are not going to deep clone the JSON node(s) 

While I normally go with immutable objects, in this situation my feeling is make it mutable because results like this can be built up 

as a final nitpick, this may be better organised if at the bottom of the class

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/CommandStatus.java
```diff
@@ -10,9 +10,15 @@ public enum CommandStatus {
   /** The element has the count of deleted documents */
   @JsonProperty("deletedCount")
   DELETED_COUNT,
-  /** Status for reporting existing namespaces. */
+  /**
+   * Status for reporting existing namespaces. findNamespaces command is deprecated, keep this
+   * jsonProperty to support backwards-compatibility
+   */
```
amorton, line 16: nit: consider also saying we use "namespaces" when old command name used and "keyspaces" when new one

src/main/java/io/stargate/sgv2/jsonapi/service/operation/keyspaces/FindKeyspacesOperation.java
```diff
@@ -0,0 +1,54 @@
...
+import java.util.Map;
+import java.util.function.Supplier;
+
+/**
+ * Operation that list all available keyspaces into the {@link CommandStatus#EXISTING_KEYSPACES}
+ * command status.
+ *
+ * @param cqlSessionCache CQLSession cache for keyspace fetching
+ */
+public record FindKeyspacesOperation(CQLSessionCache cqlSessionCache) implements Operation {
```
amorton, line 21: -1 on having this class and FindNamespacesOperation with the same code except for 1 line. 

Recommend:
* make this a class, it was a record from peeps get too excited about records. Remove the CQLSessionCache it is not needed
* add a ctor param  useKeyspaceName and use that when creating the result. 
* Get the session case from the QueryExecutor , that is what it is there for and it already has a method on it. 
* Keep the two command resolvers, they only work with a  single command class, but they can both create the same operation,.

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/CreateKeyspaceCommandResolver.java
```diff
@@ -0,0 +1,73 @@
...
+import io.stargate.sgv2.jsonapi.service.operation.keyspaces.CreateKeyspaceOperation;
+import jakarta.enterprise.context.ApplicationScoped;
+import java.util.Map;
+
+/**
+ * Command resolver for {@link CreateKeyspaceCommand}. Responsible for creating the replication map.
+ * Resolve a {@link CreateKeyspaceCommand} to a {@link CreateKeyspaceOperation}
+ */
+@ApplicationScoped
+public class CreateKeyspaceCommandResolver implements CommandResolver<CreateKeyspaceCommand> {
```
amorton, line 16: -1 for duplicating code like this, create  a common subclass for create Keyspace and create Namespace

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/FindNamespacesCommandResolver.java
```diff
@@ -27,6 +30,9 @@ public Class<FindNamespacesCommand> getCommandClass() {
   @Override
   public Operation resolveDatabaseCommand(
       CommandContext<DatabaseSchemaObject> ctx, FindNamespacesCommand command) {
-    return new FindNamespacesOperations(cqlSessionCache);
+    // Instead of using new FindKeyspacesOperation, still using FindNamespacesOperation.
+    // This is because there is namespaces as JsonProperty returned in FindNamespacesCommand
+    // Support for backwards-compatibility
+    return new FindNamespacesOperation(cqlSessionCache);
```
amorton, line 36: see comment in FindKeyspacesOperation

amorton (changes requested):
-1 there is some code duplication we do not need to do, and there is a cleaner way to mark and detect the deprecated commands than using a Util class.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/CreateNamespaceCommand.java
```diff
@@ -12,7 +12,12 @@
 import javax.annotation.Nullable;
 import org.eclipse.microprofile.openapi.annotations.media.Schema;
 
-@Schema(description = "Command that creates a namespace.")
+/**
+ * Command {@link CreateNamespaceCommand} is deprecated, please switch to {@link
+ * CreateKeyspaceCommand} Support it for backward-compatibility
+ */
+@Schema(
+    description = "Command that creates a namespace (deprecated, please switch to createKeyspace).")
```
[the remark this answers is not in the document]
Yuqi-Du: Thanks, this looks good!

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/CommandStatus.java
Yuqi-Du: Thanks

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/CommandResult.java
Yuqi-Du: Thanks, Resolved.

src/main/java/io/stargate/sgv2/jsonapi/util/DeprecatedCommandUtil.java
```diff
@@ -0,0 +1,79 @@
+package io.stargate.sgv2.jsonapi.util;
+
+import io.stargate.sgv2.jsonapi.api.model.command.Command;
+import io.stargate.sgv2.jsonapi.api.model.command.CommandContext;
+import io.stargate.sgv2.jsonapi.api.model.command.impl.*;
+import io.stargate.sgv2.jsonapi.service.cqldriver.executor.SchemaObject;
+
+public class DeprecatedCommandUtil {
```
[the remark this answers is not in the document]
Yuqi-Du: This is a much better way, thanks

amorton (changes requested):

amorton (changes requested):
few more changes pls, see comments and we have discussed.

Yuqi-Du:
Above comments are solved. thanks!!

amorton (approved):
remove the use of this. in CommandResult.withWarning() and we are good to ship 👍

vkarpov15 (commented):
LGTM :+1:

PR #1401, opened 2024-09-10
(WIP) Fix #1396: allow use of `_id` column for tables

### LINKED ISSUE #1396: createTable command with `_id` errors out
The following createTable() command:

```
{
  "createTable": {
    "name": "bots",
    "definition": {
      "primaryKey": "_id",
      "columns": {
        "_id": {
          "type": "text"
        },
        "name": {
          "type": "text"
        }
      }
    }
  }
}

```

throws the following error:

```
[{"message":"Server failed: root cause: (java.lang.IllegalArgumentException) Invalid CQL form [_id]: needs double quotes","errorCode":"SERVER_UNHANDLED_ERROR"}
```

Looks like we also need some double quotes for createTable?

tatu-at-datastax:
**What this PR does**:

Adds test(s) for, fixes issue #1396: problem with `_id` for `createTable` command

Fixes #1396

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/src/test/java/io/stargate/sgv2/jsonapi/api/v1/tables/AbstractTableIntegrationTestBase.java
+++ b/src/test/java/io/stargate/sgv2/jsonapi/api/v1/tables/AbstractTableIntegrationTestBase.java
@@ -40,6 +40,14 @@ public class AbstractTableIntegrationTestBase extends AbstractNamespaceIntegrati
         .body("status.ok", is(1));
   }
 
+  protected DataApiResponseValidator deleteTable(String tableName) {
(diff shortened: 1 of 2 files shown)


tatu-at-datastax:
> @tatu-at-datastax As much I remember, the change for tables that was done was to support field name starting with `_` which is not limited to `_id`. If that is still the case, can the PR title be changed? It's misleading to state we do something special for _id field. Changes looks good.

Correct: fix was for quoting in general. Although... there was another fix specific to "Id" field. But I think problem reported here was quoting; will update title.

PR #1419, opened 2024-09-12
Fix #1364: add tests, fixes to text value codecs (to CQL)

### LINKED ISSUE #1364: Data API, JsonCodec testing: textual values (ascii, text, varchar)
Add

* Unit tests for 3 supported (to some degree) CQL data types: `ascii`, `text`, `varchar` (of which only `text` allowed to be created.
* Integration tests for creating tables with `text`, insert, find (unless already existing).
* Integration tests for CQL-created tables with `ascii` and `varchar` columns, with Insert/Find support; as well as failing tests to ensure creation fails.

**EDIT**: looks like there is no distinct "varchar" type and it is more of an alias.

tatu-at-datastax:
**What this PR does**:

Adds test to JSON->CQL codecs for textual values

Fixes #1364

**Checklist**
- [ ] Changes manually tested


src/test/java/io/stargate/sgv2/jsonapi/service/operation/filters/table/codecs/JSONCodecRegistryTestData.java
```diff
@@ -38,6 +38,12 @@ public class JSONCodecRegistryTestData {
 
   public final BigDecimal NOT_EXACT_AS_INTEGER = new BigDecimal("1.25");
 
+  public final String STRING_ASCII_SAFE = "ascii-safe-string";
+
+  public final String STRING_UTF8_WITH_2BYTE_CHAR = "utf8-2-byte-\u00a2"; // cent symbol
+
+  public final String STRING_UTF8_WITH_3BYTE_CHAR = "utf8-3-byte-\u20ac"; // euro symbol
```
amorton, line 45: question - Docs say java strings are UTF 16 https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/String.html

and https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Character.html#unicode

both of these examples are represented with a single UTF 16 pair 

https://codepoints.net/U+00A2?lang=en
https://codepoints.net/U+20AC?lang=en

We can keep these tests, can we also add tests that require UTF16 surrogate pairs (two examples here) https://en.wikipedia.org/wiki/UTF-16#U+D800_to_U+DFFF_(surrogates)

example https://codepoints.net/U+10437?lang=en is 4 bytes in UTF-16 

then checking in the docs for the String: A String represents a string in the UTF-16 format in which supplementary characters are represented by surrogate pairs (see the section Unicode Character Representations in the Character class for more information). Index values refer to char code units, so a supplementary character uses two positions in a String.

so if you have gotten this far :) 

* is it correct to be talking about the UTF8 encoding or should it be utf16 encoding ? 
* can we also test some chars that are beyond the Basic Multilingual Plane, some docs say there are in the  "Supplementary Ideographic Plane (SIP) or the Tertiary Ideographic Plane (TIP)"  The wiki example link above has 2 chars from the SIP 
* when we have chars that require surrogte pairs, how does the charAt() work ? I read the docs and was a little confused

src/main/java/io/stargate/sgv2/jsonapi/service/operation/filters/table/codecs/JSONCodec.java
```diff
@@ -167,6 +167,24 @@ static <JavaT> ToCQL<JavaT, JavaT> unsafeIdentity() {
       return (toCQLType, value) -> value;
     }
 
+    /**
+     * Returns given String if (and only if) it only contains 7-bit ASCII characters; otherwise it
+     * will throw an {@link ToCQLCodecException}
+     */
+    static String safeAscii(DataType targetTextType, String value) throws ToCQLCodecException {
+      for (int i = 0, len = value.length(); i < len; ++i) {
+        if (value.charAt(i) > 0x7F) {
```
amorton, line 176: see comments for JSONCodecRegistryTestData

src/test/java/io/stargate/sgv2/jsonapi/api/v1/tables/FindOneTableIntegrationTest.java
```diff
@@ -31,6 +33,16 @@ public final void createDefaultTables() {
             "name",
             Map.of("type", "text")),
         "id");
+    createTableWithColumns(
```
amorton, line 36: we should prob have a discussion at some point about what the unit tests should cover and what the integrations should, if we decide they should be independant and can cover the same ground that is cool.

amorton (changes requested):

amorton (changes requested):
Seeking clarification on the UTF 8 / 16 discusson and testing beyond the Basic Multilingual Plane with UTF16 surrogate pairs.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/filters/table/codecs/JSONCodec.java
```diff
@@ -167,6 +167,24 @@ static <JavaT> ToCQL<JavaT, JavaT> unsafeIdentity() {
...
+          Exception rootCause =
+              new IllegalArgumentException(
+                  String.format(
+                      "String contains non-ASCII character at index #%d (length=%d): \\u%04X",
+                      i, len, (int) value.charAt(i)));
+          throw new ToCQLCodecException(value, targetTextType, rootCause);
```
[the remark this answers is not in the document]
tatu-at-datastax: Agree -- was about to add alternate constructor. I think I was only stopped due to `String.format()` in base class, wasn't sure how to tackle that.
But now that I looked into that, I think I know exactly how to change that -- and will do that!

src/test/java/io/stargate/sgv2/jsonapi/service/operation/filters/table/codecs/JSONCodecRegistryTestData.java
tatu-at-datastax: Ok, yeah, I can add some comments, but more importantly will add a surrogate pair case: it is indeed 2 x UTF-16 (technically Java has "modified UCS-2" just because of details of null masking), or, 4 x UTF-8 bytes.

And as to `charAt()`: it blindly returns UTF-16 `char`, without handling surrogate pairs. But because surrogate pair characters are in `D400`-`DBFF` range, they will never be in 7-bit ASCII range, either as UTF-8 bytes UTF-16 chars. But with Java Strings we do not deal with UTF-8 bytes directly (... with JDK 9+ there is the internal optimization of using byte-based ASCII Strings... but as we saw, this details not really exposed).

Sorry, not sure above makes much sense. But:

1. Yes, I think checking of `charAt()` catches all cases we need to
2. Yes, I should still add one more check for Surrogate-pair case (4-byte)

src/test/java/io/stargate/sgv2/jsonapi/api/v1/tables/FindOneTableIntegrationTest.java
tatu-at-datastax: Right, I think some overlap can be useful, but doubling of all coverage is wasteful. Right now I am interested in end-to-end translation of exceptions,  which is partly why I add overlapping case.

Not sure how to guarantee good balance; I think we have leaned bit too much  on integration tests earlier, partly as operation unit test have been bit too mock-based (and IMO not necessarily giving confidence that Operations actually work the same way end-to-end).

src/main/java/io/stargate/sgv2/jsonapi/service/operation/filters/table/codecs/JSONCodec.java
```diff
@@ -167,6 +167,24 @@ static <JavaT> ToCQL<JavaT, JavaT> unsafeIdentity() {
       return (toCQLType, value) -> value;
     }
 
+    /**
+     * Returns given String if (and only if) it only contains 7-bit ASCII characters; otherwise it
+     * will throw an {@link ToCQLCodecException}
+     */
+    static String safeAscii(DataType targetTextType, String value) throws ToCQLCodecException {
```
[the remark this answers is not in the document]
tatu-at-datastax: `targetTextType` is indeed used: this is done in `ToCQLCodecException` constructor that builds description which comes before message we pass here. That's why it need/should not be repeated here (reference to target type would be duplicated).

src/main/java/io/stargate/sgv2/jsonapi/service/operation/filters/table/codecs/JSONCodec.java
tatu-at-datastax: Will add comment stating why this should be fine.

tatu-at-datastax: Added.

src/test/java/io/stargate/sgv2/jsonapi/service/operation/filters/table/codecs/JSONCodecRegistryTestData.java
tatu-at-datastax: Wrt ^^^: added Surrogate/non-BMP unit test so we know those are caught; and added some comments on checking to note reasons why I think test is sufficient.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/filters/table/codecs/JSONCodec.java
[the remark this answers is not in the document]
amorton: maybe question misunderstood - I was suggesting checking it was ASCII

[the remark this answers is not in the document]
amorton: yes an assertion to check , that was all

src/main/java/io/stargate/sgv2/jsonapi/exception/catchable/ToCQLCodecException.java
```diff
@@ -13,13 +13,35 @@ public class ToCQLCodecException extends CheckedApiException {
...
-            "Error trying to convert to targetCQLType `%s` from value.class `%s`, value %s. Root cause: %s",
-            targetCQLType, value.getClass().getName(), value, cause.getMessage()),
-        cause);
+        formatMessage(value, targetCQLType, (rootCause == null) ? "NULL" : rootCause.getMessage()),
+        rootCause);
     this.value = value;
     this.targetCQLType = targetCQLType;
   }
+
+  public ToCQLCodecException(Object value, DataType targetCQLType, String rootCauseMessage) {
```
amorton, line 24: +1

amorton (approved):

vkarpov15 (commented):
LGTM

PR #1430, opened 2024-09-17
add xml-format plugin

Yuqi-Du:
One step of release workflow is failing, it is the update changelog step specifically.
https://github.com/stargate/data-api/actions/runs/10911425360

**Checklist**
- [x] Changes manually tested


tatu-at-datastax:
Ok so why are we adding the xml-format plugin? Does that fix changelog update step?

Yuqi-Du:
> Ok so why are we adding the xml-format plugin? Does that fix changelog update step?

Yes, 
[release github workflow ](https://github.com/stargate/data-api/blob/fb22cf0714543d1303e2783c193dcebf1c2b8376/.github/workflows/release.yaml#L330). 

After we remove stargate dependency, this xml-format plugin is also lost, that is why v1.0.16 release workflow partially failed.

tatu-at-datastax (approved):

tatu-at-datastax:
> > Ok so why are we adding the xml-format plugin? Does that fix changelog update step?
> 
> Yes, [release github workflow ](https://github.com/stargate/data-api/blob/fb22cf0714543d1303e2783c193dcebf1c2b8376/.github/workflows/release.yaml#L330).
> 
> After we remove stargate dependency, this xml-format plugin is also lost, that is why v1.0.16 release workflow partially failed.

Ah. So we could alternatively have removed XML format invocation, but if it's called, plug-in definition is needed. This is probably safer.

PR #1433, opened 2024-09-18
Fixes #1428: support "not a number" values for 'float', 'double' columns

### LINKED ISSUE #1428: Tables number codec: support "NaN", "Infinity

tatu-at-datastax:
**What this PR does**:

Add support, tests, for NaN/FP values.

Fixes #1428

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/src/test/java/io/stargate/sgv2/jsonapi/api/v1/tables/InsertOneTableIntegrationTest.java
+++ b/src/test/java/io/stargate/sgv2/jsonapi/api/v1/tables/InsertOneTableIntegrationTest.java
@@ -19,6 +19,7 @@ import org.junit.jupiter.api.TestClassOrder;
 public class InsertOneTableIntegrationTest extends AbstractTableIntegrationTestBase {
   static final String TABLE_WITH_TEXT_COLUMNS = "findOneTextColumnsTable";
   static final String TABLE_WITH_INT_COLUMNS = "findOneIntColumnsTable";
+  static final String TABLE_WITH_FP_COLUMNS = "findOneFpColumnsTable";
 
   @BeforeAll
   public final void createDefaultTables() {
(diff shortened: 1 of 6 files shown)


src/test/java/io/stargate/sgv2/jsonapi/api/v1/tables/InsertOneTableIntegrationTest.java
```diff
@@ -48,6 +49,18 @@ public final void createDefaultTables() {
             "bigIntegerValue",
             Map.of("type", "varint")),
         "id");
+    createTableWithColumns(
```
Yuqi-Du, line 52: So for Integration test, we still need to create the table with correct columns to test against, correct?
The previous thing Aaron showed for all kinds of code combination, that is unit test?

Yuqi-Du (approved):
LGTM!

src/test/java/io/stargate/sgv2/jsonapi/api/v1/tables/InsertOneTableIntegrationTest.java
tatu-at-datastax: Correct on both: Aaron has been writing unit tests, with more scaffolding. I added some unit tests too, but those are simpler parameterized codec tests.

ITs do use some new helper classes but do require old-style table creation.
We will probably need a way to create tables with direct CQL but for now use of `createTable` works.

vkarpov15 (commented):
LGTM. I can't say the approach of sending NaN, Infinity, -Infinity as strings is my favorite, but I can't think of a better approach and it is reasonable given that we know to expect a number on the data api side.

tatu-at-datastax:
@vkarpov15 yes, it's not ideal but will do for now. And can be still revisited before GA based on feedback. In theory we could use EJSON wrapper, fwtw, but even there I think it'd be Number-or-String as wrapped value.

PR #1443, opened 2024-09-20
Fix #1417: optimize/improve codec lookup handling

### LINKED ISSUE #1417: Optimize `JSONCodecRegistry` to-CQL codec lookup: linear search won't scale
As the initial implementation, Codec lookup in `JSONCodecRegistry` for converting from JSON'ish Java value (partially converted from `JsonNode` into `String`, numbers etc) into Java type CQL driver excepts performs linear scan of all codecs, to match by source type (of value) and target CQL type. This works, functionally speaking, but with increasing number of permutations scales badly.

But it should be easy to segment codecs by CQL target type first, do a (Hash)Map lookup, and only then consider source type (for which linear match is needed).
This will change linear scan from going over dozens of choices on average into one hash lookup (to CQL type) into scan over at one or at most couple (3 numeric types).

Issue is not critical at this point but added as something that can be done when time allows.

tatu-at-datastax:
**What this PR does**:

Changes linear-scan lookup into map-based lookup (one- or two-phase depending on direction). Also improves error messages for partial misses where we support CQL type but not particular coercion (like Boolean from Integer) to indicate it's Java value mismatch and not (fully) unsupported type.

Fixes #1417

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/exception/catchable/ToCQLCodecException.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/exception/catchable/ToCQLCodecException.java
@@ -31,7 +31,7 @@ public class ToCQLCodecException extends CheckedApiException {
       Object value, DataType targetCQLType, String rootCauseMessage) {
     return String.format(
         "Error trying to convert to targetCQLType `%s` from value.class `%s`, value %s. Root cause: %s",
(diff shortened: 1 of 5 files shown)


Yuqi-Du (approved):
LGTM

PR #1449, opened 2024-09-23
Fixes #1385: add date/time codec support

### LINKED ISSUE #1385: Data API JsonCodec impl: date/time values (date, duration, time, timestamp)
Add

* Unit tests for 4 supported (to some degree) CQL data types: `date`, `duration`, `time`, `timestamp` (`timeuuid` not time-related in same way)
* Integration tests for creating tables with all types, insert, find (unless already existing).

tatu-at-datastax:
**What this PR does**:

Adds support for 4 date/time types: `date`, `duration`, `time`, `timestamp`, matching tests

Fixes #1385

**Checklist**
- [ ] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/service/operation/filters/table/codecs/JSONCodec.java
```diff
@@ -178,6 +179,24 @@ static <JavaT> ToCQL<JavaT, JavaT> unsafeIdentity() {
...
 
+    /**
+     * Method for constructing exception-handling String-to-CQL-type converter, given a function
+     * that converts a String to the target CQL type.
+     */
+    static <JavaT, CqlT> ToCQL<JavaT, CqlT> safeFromString(Function<JavaT, CqlT> function) {
+      return (toCQLType, value) -> {
+        try {
+          return function.apply(value);
+        } catch (IllegalArgumentException | DateTimeException e) {
```
tatu-at-datastax, line 190: Selecting exceptions known to be thrown for content problem, but trying to avoid catching general `Exception` to let things like NPEs flow through to be caught at higher levels (as they are not problems with input value but something different).

src/main/java/io/stargate/sgv2/jsonapi/service/operation/filters/table/codecs/JSONCodec.java
```diff
@@ -388,5 +407,16 @@ JsonNode apply(ObjectMapper objectMapper, DataType fromCQLType, CqlT value)
...
       return (objectMapper, fromCQLType, value) -> nodeFactoryMethod.apply(value);
     }
+
+    /**
+     * Returns an instance that will translate given CQL value using plain {@code toString()},
+     * constructing a {@link JsonNode} of type {@link com.fasterxml.jackson.databind.node.TextNode}.
+     *
+     * <p>See usage in the {@link JSONCodecRegistry}
+     */
+    static <CqlT> ToJSON<CqlT> toJSONUsingToString() {
```
Yuqi-Du, line 417: Feel strange to the method name at the first glance.
But understand what it does after reading more info.
Just a suggestion, probably not good, method name can be something like `toJSONFromCqlPlaneString`?

Yuqi-Du (approved):
Two minor comments, thanks!

src/main/java/io/stargate/sgv2/jsonapi/service/operation/filters/table/codecs/JSONCodec.java
tatu-at-datastax: Let me improve javadoc a bit: pertinent part to me is that it is `value.toString()` that gets called, as `String.valueOf()` is defined as:

```
    public static String valueOf(Object obj) {
        return (obj == null) ? "null" : obj.toString();
    }
```

and this is why I'd like to have "toString" in name.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/tables/InsertOneTableIntegrationTest.java
```diff
@@ -346,4 +362,99 @@ private String rawBinaryDoc(String id, String rawBinaryValue) {
...
+
+  @Nested
+  @Order(5)
+  class InsertDatetimeColumns {
+    @Test
+    void insertValidDateTimeValues() {
+      // NOTE: While `CqlDuration.from()` accepts both ISO-8601 "P"-notation (like "PT2H45M")
+      //   and Cassandra's standard compact/readable notation (like "2h45m"),
+      //   `CqlDuration.toString()` returns canonical representation so we use the latter here
+      //   to verify round-tripping
```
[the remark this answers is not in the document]
tatu-at-datastax: I also could not find a method to produce longer notation ("P....") so had to use `toString()`. I'll add a comment in `JsonCodecs`.

tatu-at-datastax:
Comments addressed; no further comments, merging.

src/test/java/io/stargate/sgv2/jsonapi/service/operation/filters/table/codecs/JSONCodecRegistryTestData.java
```diff
@@ -50,6 +50,20 @@ public class JSONCodecRegistryTestData {
...
+  public final String DATE_INVALID_STR = "not-a-date";
+
+  public final String DURATION_VALID1_STR = "1h30m";
+  public final String DURATION_VALID2_STR = "PT1H30M";
+  public final String DURATION_INVALID_STR = "not-a-duration";
+
+  public final String TIME_VALID_STR = "12:34:56.789";
+  public final String TIME_INVALID_STR = "not-a-time";
+
+  public final String TIMESTAMP_VALID_STR = "2024-09-12T10:15:30Z";
```
vkarpov15, line 64: Do timestamps support milliseconds, like `"2024-09-12T10:15:30.123Z"` ?

tatu-at-datastax: Yes, they should (and even micro-, nano-seconds I think).
It'd great if you could check on client side (and obv. should add more tests on server side).

vkarpov15: Milliseconds yes, `2024-09-30T21:30:08.212Z` seems to be stored correctly, but micro and nano doesn't seem to work, `2024-09-30T21:30:08.212123456Z` saves without error but I just get back `2024-09-30T21:30:08.212Z`

tatu-at-datastax: Ah. Now that I think, I suspect this is Cassandra limitation, unfortunately. I think it stores values as 64-bit milliseconds-since-epoch (1970-01-01/UTC).

So yeah, we can accept more granular but not retain.

vkarpov15: Might be good to throw an error in the case of microseconds? Right now the additional decimal places are silently dropped.

PR #1453, opened 2024-09-26
Add CommandResultBuilder

amorton:
Because there are several ways we create this, the builder is used in the work for OperationAttempt

**What this PR does**:

Fixes #<issue number>

**Checklist**
- [*] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/api/model/command/CommandResultBuilder.java
```diff
@@ -0,0 +1,142 @@
...
+  // If the debug mode is enabled, errors include the errorclass
+  private final boolean debugMode;
+
+  // Flagged true to include the new error object v2
+  private final boolean useErrorObjectV2;
+
+  // Created in the Ctor
+  private final Function<APIException, CommandResult.Error> apiExceptionToError;
+
+  public CommandResultBuilder(
```
tatu-at-datastax, line 39: Would it make sense to instead use per-type factory methods for improved readability?
("forSingleDoc()", "forMultiDoc()", "forStatusOnly()" or something)

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/CommandResultBuilder.java
```diff
@@ -0,0 +1,142 @@
...
+                  responseType, documents.size()));
+        }
+        break;
+    }
+
+    if (!warnings.isEmpty()) {
+      cmdStatus.put(CommandStatus.WARNINGS, warnings);
+    }
+
+    // null out values that are empty, the CommandResult serialiser will ignore them when the JSON
```
tatu-at-datastax, line 125: sidenote: this is fine, but there is also a Jackson annotation option for `@JsonInclude` to exclude both `null` and "empty" containers we can use.

but it may be better not to rely on that here since that would also drop empty Strings as values.

amorton: cool, thanks. I think there will be more work here. The initial phase is just to try to capture how it should work.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/CommandResultBuilder.java
amorton: sounds resonable, can also move the  boolean useErrorObjectV2, boolean debugMode to be builder properties 

 created https://github.com/stargate/data-api/issues/1461

PR #1454, opened 2024-09-26
Include error ID in the error response for new error format

amorton:
This was already happening for the old errors mapped to new, the new OperationAttempt work relies on this

**What this PR does**:

Fixes #<issue number>

**Checklist**
- [ ] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/exception/APIExceptionCommandErrorBuilder.java
```diff
@@ -61,6 +61,7 @@ public CommandResult.Error apply(APIException apiException) {
       errorFields.put(ErrorObjectV2Constants.Fields.FAMILY, apiException.family.name());
       errorFields.put(ErrorObjectV2Constants.Fields.SCOPE, apiException.scope);
       errorFields.put(ErrorObjectV2Constants.Fields.TITLE, apiException.title);
+      errorFields.put(ErrorObjectV2Constants.Fields.ID, apiException.errorId);
```
maheshrajamani, line 64: java doc to describe what these fields are.

maheshrajamani (approved):
LGTM

src/main/java/io/stargate/sgv2/jsonapi/exception/APIExceptionCommandErrorBuilder.java
amorton: there is a whole bunch of docs on the ApiException that explains all the fields, your comment has reminded me that it's about time we refactor the CommandResult.Error it is a mess with the map. 

if we do this, then we dont need all these fields and maps etc see https://github.com/stargate/data-api/issues/1462

PR #1464, opened 2024-09-27
Fix #1460, handle CQL nulls to  JSON gracefully

### LINKED ISSUE #1460: Handle `null` from CQL driver when converting column values to JSON (`findOne` etc)
Existing `JSONCodec` implementations "toJSON()" converters do not handle `null` value from CQL driver but throw NPE.  This should instead just map to `NullNode`: change can be done to either `JSONCode.toJSON()` method, or caller `TableRowProjection` on call site.

tatu-at-datastax:
**What this PR does**:

Ensures that CQL `null` from Driver gets safely translated into JSON responses (currently as explicit JSON null)

Fixes #1460

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/operation/filters/table/codecs/JSONCodec.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/operation/filters/table/codecs/JSONCodec.java
@@ -128,6 +128,10 @@ public record JSONCodec<JavaT, CqlT>(
    *     should convert this into the appropriate exception for the use case.
    */
   public JsonNode toJSON(ObjectMapper objectMapper, CqlT value) throws ToJSONCodecException {
+    // First: if caller passes nulls, convert to JSON Null so converter need to do null checks
+    if (value == null) {
+      return objectMapper.nullNode();
+    }
     return toJSON.apply(objectMapper, targetCQLType, value);
   }
 
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/operation/tables/TableRowProjection.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/operation/tables/TableRowProjection.java
@@ -73,7 +73,14 @@ public record TableRowProjection(
             "Column '%s' has unsupported type '%s'", columnName, column.getType().toString());
       }
       try {
-        result.put(columnName, codec.toJSON(objectMapper, row.getObject(i)));
+        final Object columnValue = row.getObject(i);
+        // We have a choice here: convert into JSON null (explicit) or drop (save space)?
+        // For now, do former: may change or make configurable later.
+        if (columnValue == null) {
+          result.putNull(columnName);
+        } else {
+          result.put(columnName, codec.toJSON(objectMapper, columnValue));
+        }
       } catch (ToJSONCodecException e) {
         throw ErrorCodeV1.UNSUPPORTED_PROJECTION_PARAM.toApiException(
             e,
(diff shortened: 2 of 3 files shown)


src/main/java/io/stargate/sgv2/jsonapi/service/operation/filters/table/codecs/JSONCodec.java
```diff
@@ -128,6 +128,10 @@ public boolean testToJSON(DataType fromCQLType) {
    *     should convert this into the appropriate exception for the use case.
    */
   public JsonNode toJSON(ObjectMapper objectMapper, CqlT value) throws ToJSONCodecException {
+    // First: if caller passes nulls, convert to JSON Null so converter need to do null checks
```
tatu-at-datastax, line 131: This just to make sure all codecs accept `null`; caller may or may not handle it instead of handing to codecs (in case it wants to suppress them, for example).

src/main/java/io/stargate/sgv2/jsonapi/service/operation/tables/TableRowProjection.java
```diff
@@ -73,7 +73,14 @@ public DocumentSource documentSource(Row row) {
             "Column '%s' has unsupported type '%s'", columnName, column.getType().toString());
       }
       try {
-        result.put(columnName, codec.toJSON(objectMapper, row.getObject(i)));
+        final Object columnValue = row.getObject(i);
+        // We have a choice here: convert into JSON null (explicit) or drop (save space)?
```
tatu-at-datastax, line 77: Added a comment since we have discussed possibility of suppressing nulls to save space: whether it'd be done here or as post-processing is an open question; at first we just want to avoid NPE.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/tables/FindOneTableIntegrationTest.java
```diff
@@ -87,6 +87,17 @@ public void findOneSingleStringKey() {
                           """;
       insertOneInTable(TABLE_WITH_STRING_ID_AGE_NAME, DOC_B_JSON);
 
+      // Third one with missing "age" and null "name"
```
tatu-at-datastax, line 90: Turns out 'text' and 'int' codecs happened to work with `null`s, although for different reasons. But was simple and cheap to add coverage here.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/tables/FindOneTableIntegrationTest.java
```diff
@@ -172,20 +209,21 @@ public void findOneDocIdKey() {
                                   "_id": 2,
                                   "desc": "b",
                                   "valueLong": 42,
-                                  "valueDouble": 0.5
+                                  "valueDouble": 0.5,
+                                  "valueBlob": null
```
tatu-at-datastax, line 213: Either missing or explicit `null` would work: this is what triggered NPE before fix.

maheshrajamani (approved):

vkarpov15 (commented):
Confirmed this works fine in Mongoose. Mongoose does treat null and undefined as interchangeable in general, so I don't foresee this having a huge impact. Only difference is that we now have more `null` fields.

tatu-at-datastax:
Null exclusion as configurable option (and maybe even as default) is something we will need to discuss in near future, for sure.

PR #1470, opened 2024-09-27
(WIP) Fix #1469: (createTable deser exposes table command) add tests, change deserialization logic

### LINKED ISSUE #1469: Do not expose table related exception if feature is not enabled.
So if the column validation failed before the feature flag check, it won’t return error like TABLE_FEATURE_NOT_ENABLED, instead it will return table related stuff. We should do feature checks before these validation happen.

For example, with Feature-Flag-tables as False
```
{
    "createTable": {
        "name": "allcolumns",
        "definition": {
            "columns": {
                "id": {
                    "type": "text"
                },
                "age": {
                    "type": "int"
                },
                "image_data": {
                    "type": "blob"
                }

            },
            "primaryKey": "id"
        }
    }
}


{
    "errors": [
        {
            "message": "Unsupported column types: Invalid column type: blob",
            "errorCode": "TABLE_COLUMN_TYPE_UNSUPPORTED"
        }
    ]
}
```

tatu-at-datastax:
**What this PR does**:

Adds reproduction of #1469 to show the issues, and changes deserialization to defer validation failures after APIFeature check.

Fixes #1469

**Checklist**
- [ ] Changes manually tested


tatu-at-datastax:
Fix here works but not completely convinced we want to take this approach -- however, if we do want to solve the issue something like this is needed.

tatu-at-datastax:
PR hard to merge due to concurrent refactorings -- leaving open in case it needs to be revived; if so, will probably need to re-create.

tatu-at-datastax:
Due to concurrent refactorings, pr no longer maintainable: will close.

PR #1471, opened 2024-09-28
Fix #1450: add support for List, Set codecs

### LINKED ISSUE #1450: Data API JsonCodec impl: unfrozen container types (Set, List)
After Scalar types we need to support some collection/container types: specifically unfrozen, non-nested `Set` and `List` types (there will be separate issue for `Map` types). Non-nested means that element/value types are scalar types, not containers (could only support nesting with frozen types).

As usual, codec implementation to be tested via unit tests (more extensive) and selected Insert-in-Table integration tests.

tatu-at-datastax:
**What this PR does**:

Adds supports for un-frozen, non-nested `Set` and `List` CQL column types

Fixes #1450

**Checklist**
- [ ] Changes manually tested


tatu-at-datastax:
First part, JSON-to-CQL added with Unit tests (for IT, need to be able to create such columns).

maheshrajamani (approved):
LGTM

tatu-at-datastax:
Great! Thank you for verifying @vkarpov15 . There still remains work for Vectors and Maps, but should be straight-forward based on work with Sets and Lists.

PR #1473, opened 2024-09-30
Create table command implementation

maheshrajamani:
**What this PR does**:
Create table api for primitive types

**Checklist**
- [x] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/api/model/command/table/definition/datatype/ColumnType.java
```diff
@@ -1,38 +1,68 @@
...
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import io.stargate.sgv2.jsonapi.api.model.command.deserializers.ColumnDefinitionDeserializer;
+import io.stargate.sgv2.jsonapi.exception.SchemaException;
+import io.stargate.sgv2.jsonapi.service.schema.tables.ApiDataType;
+import java.util.List;
+import java.util.Map;
+
+@JsonDeserialize(using = ColumnDefinitionDeserializer.class)
 
 /** Interface for column types. This is used to define the type of a column in a table. */
```
tatu-at-datastax, line 12: nit: I think Javadoc should typically come before class annotations.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/table/definition/datatype/ColumnType.java
```diff
@@ -1,38 +1,68 @@
...
+        "double",
+        "duration",
+        "float",
+        "int",
+        "smallint",
+        "text",
+        "time",
+        "timestamp",
+        "tinyint",
+        "varint");
```
tatu-at-datastax, line 34: I assume we will add types once support added (`inet` and `uuid` f.ex).

src/main/java/io/stargate/sgv2/jsonapi/exception/mappers/WebApplicationExceptionMapper.java
```diff
@@ -28,6 +37,14 @@ public RestResponse<CommandResult> webApplicationExceptionMapper(WebApplicationE
       // but leave out the root cause, as it is not useful
       toReport = ErrorCodeV1.SHRED_DOC_LIMIT_VIOLATION.toApiException(toReport.getMessage());
     }
+
+    if (toReport instanceof APIException ae) {
```
tatu-at-datastax, line 41: Could use a comment explaining why this is needed.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/tables/CreateTableOperation.java
```diff
@@ -62,6 +66,10 @@ public Uni<Supplier<CommandResult>> execute(
     CqlIdentifier tableIdentifier = CqlIdentifier.fromInternal(tableName);
     CreateTableStart create = createTable(keyspaceIdentifier, tableIdentifier).ifNotExists();
 
+    // Add if not exists flag based on request
+    if (ifNotExists) {
+      create = create.ifNotExists();
```
tatu-at-datastax, line 71: Isn't this already being set above (line 63/67)?

src/main/resources/errors.yaml
```diff
@@ -134,6 +134,95 @@ request-errors:
...
+    title: Column definition is missing for the provided primary key column.
+    body: |-
+      Column definition is missing for the provided primary key column: ${column_name}.
+
+  - scope: SCHEMA
+    code: COLUMN_TYPE_INCORRECT
+    title: Column data type not provided or format invalid in the definition
+    body: |-
+      Column data type not provided or format invalid in the definition.
+      Column definition can be defined in shorthand format as :
```
tatu-at-datastax, line 150: nit: should not have leading space but just trailing, so: `: ` not ` : `

src/main/resources/errors.yaml
```diff
@@ -134,6 +134,95 @@ request-errors:
...
+    title: Column data type used in the definition is not supported.
+    body: |-
+      Column data type ${type} used in the definition is not supported.
+      Following are the supported data types: ${supported_types}
+
+  - scope: SCHEMA
+    code: MISSING_PRIMARY_KEYS
+    title: Primary key definition needs to be provided while creating the table.
+    body: |-
+      Primary key definition needs to be provided while creating the table.
```
tatu-at-datastax, line 169: Hmmh. This looks like a very long multi-line exception message. Is this something we should be doing? I guess it should be discussed with @amorton but to me these can be noisy in logs and seems like we should have a link instead of novel-length tutorial of usage. :)

tatu-at-datastax (approved):
Added some suggestions but in general seems ok.

tatu-at-datastax:
@maheshrajamani seems like this is focused on refactoring wrt error reporting, renaming? Makes sense: I have some concerns about verbosity of exceptions (added as comments) but otherwise fine.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/tables/CreateTableOperation.java
maheshrajamani: Fixed it.

src/main/resources/errors.yaml
maheshrajamani: Had the same feel, raised it in the discussion yesterday, this is the messaging approach for error, that's planed. cc: @amorton

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/table/definition/datatype/ComplexTypes.java
```diff
@@ -0,0 +1,73 @@
+package io.stargate.sgv2.jsonapi.api.model.command.table.definition.datatype;
+
+import io.stargate.sgv2.jsonapi.service.schema.tables.ApiDataType;
+import io.stargate.sgv2.jsonapi.service.schema.tables.ComplexApiDataType;
+import io.stargate.sgv2.jsonapi.service.schema.tables.PrimitiveApiDataType;
+
+/** Interface for complex column types like collections */
+public interface ComplexTypes extends ColumnType {
+
+  /** A map type implimentation */
```
tatu-at-datastax, line 10: typo: implimenetation -> implementation

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/table/definition/datatype/ComplexTypes.java
```diff
@@ -0,0 +1,73 @@
+package io.stargate.sgv2.jsonapi.api.model.command.table.definition.datatype;
+
+import io.stargate.sgv2.jsonapi.service.schema.tables.ApiDataType;
+import io.stargate.sgv2.jsonapi.service.schema.tables.ComplexApiDataType;
+import io.stargate.sgv2.jsonapi.service.schema.tables.PrimitiveApiDataType;
+
+/** Interface for complex column types like collections */
+public interface ComplexTypes extends ColumnType {
```
tatu-at-datastax, line 8: Shouldn't this be singular `ComplexType` instead of plular?

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/table/definition/datatype/ComplexTypes.java
```diff
@@ -0,0 +1,73 @@
...
+
+  /** A map type implimentation */
+  public static class MapType implements ComplexTypes {
```
tatu-at-datastax, line 11: Instead of inner-static class, shouldn't `MapType` be main-level types.
Or is there specific reason to enclose them in `ComplexTypes`?

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/table/definition/datatype/ComplexTypes.java
```diff
@@ -0,0 +1,73 @@
...
+    private final ColumnType keyType;
+    private final ColumnType valueType;
+
+    public MapType(ColumnType keyType, ColumnType valueType) {
```
tatu-at-datastax, line 15: Why don't we have `getKeyType()` and `getValueType()` accessors?

maheshrajamani: I never needed that value anywhere outside so far, because we create ApiDataType using those. THought of adding it as needed.

tatu-at-datastax:
Ok: I think my main confusion is wrt multiple sets of type definitions introduced. I can understand that we need differences wrt:

* CQL driver already provides `DataType` which we have to use for many things
* Table create command needs to take input which differs bit from what is provided via CQL
* When binding to/from CQL we probably need another type.

but it does get bit hairy overall.
Not sure what to do about that, just noting it's quite confusing.

tatu-at-datastax (approved):
Added lots of comments, questions, but fundamentally I think this is ok; feel free to apply/ignore suggestions.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/table/definition/datatype/ComplexTypes.java
maheshrajamani: Resolved this.

PR #1496, opened 2024-10-04
Fix #1442: add codecs for UUID type(s)

### LINKED ISSUE #1442: Data API JsonCodec impl: UUID (uuid, timeuuid)
Need to add support for general `uuid` type and (in some form) also `timeuuid`.
We can probably just handle both the same way and let C* backend worry about difference, if any.

tatu-at-datastax:
**What this PR does**:

Adds codec impls, tests (unit, integration) for cql "uuid" and "timeuuid" types.

Fixes #1442

**Checklist**
- [ ] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/service/operation/filters/table/codecs/JSONCodecRegistries.java
```diff
@@ -48,6 +48,10 @@ public abstract class JSONCodecRegistries {
               JSONCodecs.DURATION_FROM_STRING,
               JSONCodecs.TIME_FROM_STRING,
               JSONCodecs.TIMESTAMP_FROM_STRING,
+              // UUID codecs
+              JSONCodecs.UUID_FROM_STRING,
+              JSONCodecs.TIMEUUID_FROM_STRING,
```
tatu-at-datastax, line 53: Although we won't allow creation of "timeuuid" type, we do allow read/write, for which codec needed

Yuqi-Du: Sorry, I may not know the context for this one.
Why do we not allow "timeuuid" type?

Yuqi-Du (approved):
LGTM, one comment.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/filters/table/codecs/JSONCodecRegistries.java
tatu-at-datastax: I am not 100% sure; I think it may be considered confusing thing -- you can store time-based UUIDs in "uuid" column anyway.
We can add support easily but right now spreadsheet includes type as unsupported.

/cc @amorton

PR #1499, opened 2024-10-05
SchemaObject changes to support multiple vector configs

maheshrajamani:
**What this PR does**:
Made the SchemaObject store and return List of VectorConfig. This is needed because tables need to support multiple vector config.
Changed the vectorize serialization to be based on VectorizeConfig object used in SchemaObject.

**Checklist**
- [x] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/TableSchemaObject.java
```diff
@@ -1,22 +1,98 @@
...
+   */
+  public static TableSchemaObject getTableSettings(
+      TableMetadata tableMetadata, ObjectMapper objectMapper) {
+    Map<String, String> extensions =
+        (Map<String, String>)
+            tableMetadata.getOptions().get(CqlIdentifier.fromInternal("extensions"));
+    String vectorize = extensions != null ? extensions.get("vectorize") : null;
+    Map<String, VectorConfig.VectorizeConfig> resultMap = new HashMap<>();
+    if (vectorize != null) {
+      String vectorizeJson = new String(ByteUtils.fromHexString(vectorize).array());
```
tatu-at-datastax, line 56: Needs to specify character set so as not to rely on platform defaults (StandardCharsets.UTF8 probably).

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/TableSchemaObject.java
```diff
@@ -1,22 +1,98 @@
...
+      // Define the TypeReference for Map<String, VectorConfig.VectorizeConfig>
+      TypeReference<Map<String, VectorConfig.VectorizeConfig>> typeRef =
+          new TypeReference<Map<String, VectorConfig.VectorizeConfig>>() {};
+
+      // Convert JSON string to Map
+      try {
+        resultMap = objectMapper.readValue(vectorizeJson, typeRef);
+      } catch (JsonProcessingException e) {
+        throw new RuntimeException(e);
```
tatu-at-datastax, line 65: This can occur due to various reasons so should probably use proper exception ("invalid vectorize configuration, root cause: ...") to indicate it is due to somehow invalid stored configuration.

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/TableSchemaObject.java
tatu-at-datastax, line 56: `ByteUtils.fromHexString()` can throw `IllegalArgumentException` for invalid hex-encoding. Should be caught, re-thrown with proper message.

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/TableSchemaObject.java
tatu-at-datastax: One possible improvement idea: first use `ObjectMapper.readTree()` to find invalid JSON (some garbage not JSON), and then checking it's `jsonNode.isObject()`; then traverse key/value pairs and use `mapper.treeToValue()` (to check structure matches). This way could more accurately indicate type of error.

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/TableSchemaObject.java
```diff
@@ -1,22 +1,98 @@
...
+        }
+        int dimension = vectorType.getDimensions();
+        VectorConfig vectorConfig =
+            new VectorConfig(
+                true,
+                column.getKey().asInternal(),
+                dimension,
+                similarityFunction,
+                resultMap.get(column.getKey().asInternal()));
+        vectorConfigs.add(vectorConfig);
```
tatu-at-datastax, line 90: So we are expecting error configs to be ordered in Map/JSON Object? Do we specify ordering when building the Map/Object?

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/VectorConfig.java
```diff
@@ -15,27 +16,56 @@
...
 
   // TODO: this is an immutable record, this can be singleton
   // TODO: Remove the use of NULL for the objects like vectorizeConfig
   public static VectorConfig notEnabledVectorConfig() {
-    return new VectorConfig(false, -1, null, null);
+    return new VectorConfig(false, null, -1, null, null);
   }
 
-  // convert a vector jsonNode from table comment to vectorConfig
+  // convert a vector jsonNode from table comment to vectorConfig, used for collection
```
tatu-at-datastax, line 30: ... not used for tables? So only collection? (if so, comment should state that)

tatu-at-datastax: I guess it means "used when reading config for Collections (not Tables)"

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/VectorConfig.java
```diff
@@ -15,27 +16,56 @@
...
     VectorizeConfig vectorizeConfig = null;
     // construct vectorizeConfig
     JsonNode vectorizeServiceNode = jsonNode.get("service");
     if (vectorizeServiceNode != null) {
+      vectorizeConfig = VectorizeConfig.fromJson(vectorizeServiceNode, objectMapper);
+    }
+    return new VectorConfig(true, fieldName, dimension, similarityFunction, vectorizeConfig);
+  }
+
+  public record VectorizeConfig(
```
tatu-at-datastax, line 61: Javadoc?

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/VectorConfig.java
```diff
@@ -51,17 +81,8 @@ public static VectorConfig fromJson(JsonNode jsonNode, ObjectMapper objectMapper
           vectorizeServiceParameterNode == null
               ? null
```
tatu-at-datastax, line 82: Could use `Map.of()` unless we want to use `null` as marker of some kind.

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/DataVectorizer.java
```diff
@@ -49,6 +51,9 @@ public DataVectorizer(
     this.nodeFactory = nodeFactory;
     this.embeddingCredentials = embeddingCredentials;
     this.schemaObject = schemaObject;
+    // This is getting element at 0 since only one vector is stored in the schema.
+    // This logic needs to be changed to handle multiple vectors columns for tables,
+    vectorConfig = schemaObject.vectorConfigs().get(0);
```
tatu-at-datastax, line 56: Should class Javadoc probably mention if it is only being used for Collections?
I assume it is because only single vector is defined as per comment here.

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/VectorConfig.java
```diff
@@ -15,27 +16,56 @@
...
+    String fieldName,
     int vectorSize,
     SimilarityFunction similarityFunction,
     VectorizeConfig vectorizeConfig) {
 
   // TODO: this is an immutable record, this can be singleton
   // TODO: Remove the use of NULL for the objects like vectorizeConfig
   public static VectorConfig notEnabledVectorConfig() {
-    return new VectorConfig(false, -1, null, null);
+    return new VectorConfig(false, null, -1, null, null);
```
tatu-at-datastax, line 27: What are the semantics of `null` vectorField? I assume we need a placeholder of some kind... ?
Put another way, how are we using this placeholder (how is it checked and so on).

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/VectorConfig.java
```diff
@@ -15,27 +16,56 @@
  */
 public record VectorConfig(
     boolean vectorEnabled,
+    String fieldName,
```
tatu-at-datastax, line 19: Javadoc comment above seems wrong, fwtw. ("incorporates vectorizeConfig into vectorConfig")

tatu-at-datastax (commented):
Ok, so, I understand that we will need to allow multiple vector definitions for API Tables (and possibly eventually for Collections). But I am bit worried about couple of aspects:

1. Use of "vector-disabled" value for `List<VectorConfig>` instead of main-level enabled/disabled -- along with special `VectorConfig` value
2. Ordering of `VectorConfig` within `List` as it's not clear how ordering would be defined/preserved for multi-vector use case (which we don't yet have). Should we use JSON Object instead of JSON Array, keyed by column name?

I think there are good reasons to do above but would be good to outline reasoning.

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/TableSchemaObject.java
maheshrajamani: We need to maintain ordering for the columns. The configuration has the field (column) name in it. FOr collection it will be set as `$vectorize`

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/VectorConfig.java
maheshrajamani: Gave null if the there is no vector field set for the table or collection

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/VectorConfig.java
maheshrajamani: The table there denotes cql table. fwiw I will just say comment to avoid confusion

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/DataVectorizer.java
maheshrajamani: This class has all field names hard coded for collections, this needs to be adapted for api tables.

maheshrajamani:
> 1. Use of "vector-disabled" value for `List<VectorConfig>` instead of main-level enabled/disabled -- along with special `VectorConfig` value

Refactored this to keep the flag outside vector definitions.

> 2\. Ordering of `VectorConfig` within `List` as it's not clear how ordering would be defined/preserved for multi-vector use case (which we don't yet have). Should we use JSON Object instead of JSON Array, keyed by column name?

We don't need to maintain ordering for the vectorize config, since they have the column name stored with it, in case of collection field name will be `$vectorize`

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/VectorConfig.java
maheshrajamani: Didn't understand this recommendation.

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/TableSchemaObject.java
tatu-at-datastax: Ok so we'll match by name tho, as necessary, not necessarily by index?

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/VectorConfig.java
```diff
@@ -2,66 +2,106 @@
...
+ * @param vectorEnabled - IF vector field is available for the table
+ * @param columnVectorDefinitions - List of columnVectorDefinitions each with respect to a
+ *     column/field
  */
 public record VectorConfig(
-    boolean vectorEnabled,
-    int vectorSize,
-    SimilarityFunction similarityFunction,
-    VectorizeConfig vectorizeConfig) {
+    boolean vectorEnabled, List<ColumnVectorDefinition> columnVectorDefinitions) {
```
tatu-at-datastax, line 18: Good! :)

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/TableSchemaObject.java
```diff
@@ -1,22 +1,107 @@
...
+        Map<String, VectorConfig.ColumnVectorDefinition.VectorizeConfig> vectorizeConfigMap =
+            new HashMap<>();
+        while (vectorizeByColumns.fields().hasNext()) {
+          Map.Entry<String, JsonNode> entry = vectorizeByColumns.fields().next();
+          VectorConfig.ColumnVectorDefinition.VectorizeConfig vectorizeConfig =
+              objectMapper.treeToValue(
+                  entry.getValue(), VectorConfig.ColumnVectorDefinition.VectorizeConfig.class);
+          vectorizeConfigMap.put(entry.getKey(), vectorizeConfig);
+        }
+      } catch (JsonProcessingException | IllegalArgumentException e) {
```
tatu-at-datastax, line 72: Ok, one last request: was hoping for separate catch blocks -- so for `readTree()` generic "Bad JSON" (because that'd be causing it) -- and then for `treeToValue()` can indicate specific vector field that was bad.

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/TableSchemaObject.java
```diff
@@ -1,22 +1,107 @@
...
+    for (Map.Entry<CqlIdentifier, ColumnMetadata> column : tableMetadata.getColumns().entrySet()) {
+      if (column.getValue().getType() instanceof VectorType vectorType) {
+        final Optional<IndexMetadata> index = tableMetadata.getIndex(column.getKey());
+        SimilarityFunction similarityFunction = SimilarityFunction.COSINE;
+        if (index.isPresent()) {
+          final IndexMetadata indexMetadata = index.get();
+          final Map<String, String> indexOptions = indexMetadata.getOptions();
+          final String similarityFunctionValue = indexOptions.get("similarity_function");
+          if (similarityFunctionValue != null) {
+            similarityFunction = SimilarityFunction.fromString(similarityFunctionValue);
```
tatu-at-datastax, line 87: Can this throw an exception for bad similarity function value? If so, should catch.

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/VectorConfig.java
```diff
@@ -2,66 +2,106 @@
...
   }
 
-  // convert a vector jsonNode from table comment to vectorConfig
-  public static VectorConfig fromJson(JsonNode jsonNode, ObjectMapper objectMapper) {
-    // dimension, similarityFunction, must exist
-    int dimension = jsonNode.get("dimension").asInt();
-    SimilarityFunction similarityFunction =
-        SimilarityFunction.fromString(jsonNode.get("metric").asText());
+  /**
+   * Configuration for a column, In case of collection this will be of size one
```
tatu-at-datastax, line 27: nit: "size one"? Does it mean there will just be one such definition or ... ?

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/VectorConfig.java
```diff
@@ -2,66 +2,106 @@
...
 import java.util.Map;
 
 /**
  * incorporates vectorizeConfig into vectorConfig
  *
- * @param vectorEnabled
- * @param vectorSize
- * @param similarityFunction
- * @param vectorizeConfig
+ * @param vectorEnabled - IF vector field is available for the table
```
tatu-at-datastax, line 13: nit: IF -> If

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/TableSchemaObject.java
maheshrajamani: Yes.

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/TableSchemaObject.java
maheshrajamani: Shouldn't have error, those are saved from the request which are already validated.

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/NamespaceCache.java
```diff
@@ -104,7 +104,7 @@ private Uni<SchemaObject> loadSchemaObject(
               }
 
               // 04-Sep-2024, tatu: Used to check that API Tables enabled; no longer checked here
```
tatu-at-datastax, line 106: We can remove my comment here.

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/TableSchemaObject.java
```diff
@@ -1,22 +1,107 @@
...
   }
+
+  /**
+   * Get table schema object from table metadata
+   *
+   * @param tableMetadata
+   * @param objectMapper
+   * @return
+   */
+  public static TableSchemaObject getTableSettings(
```
tatu-at-datastax, line 50: Ok so this is a factory method, so name should probably be something like... `from()` or `construct()`? Or whatever names we usually use. "getXxx()" suggested getter.

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/TableSchemaObject.java
tatu-at-datastax: Never say never :-)
But I guess it's unlikely enough not to matter until we actually hit that issue.

PR #1503, opened 2024-10-07
Fix #1498: enable access/request log by default

### LINKED ISSUE #1498: Enable HTTP request log (access log) in Production
Although #752 was completed months ago, it doesn't look like HTTP request logs are enabled in Production kubes. This is probably because `src/main/resources/application.yaml` has:

```
quarkus:
  http:
    # access log format, must be explicitly enabled
    access-log:
      pattern: "%h %l %t \"%r\" %s %b"

```

and does not have `enabled: true` (which as per https://quarkus.io/guides/http-reference#configuring-http-access-logs is needed, as property [quarkus.http.access-log.enabled](https://quarkus.io/guides/http-reference#quarkus-vertx-http_quarkus-http-access-log_quarkus-http-access-log-enabled).
And we do not (I think, given there are no entries in Splunk) have matching ENV override (`QUARKUS_HTTP_ACCESS_LOG_ENABLED `) either.

There are at least 2 ways we could do this:

1. Enable in `application.yaml` for all envs (although could also disable for tests, via `src/test/resources/application.yaml` to reduce noise)
2. Enable via deployment (Helm Chart) by ENV `QUARKUS_HTTP_ACCESS_LOG_ENABLED`.

Unless there are specific reasons not to, I am leaning towards (1).

tatu-at-datastax:
**What this PR does**:

Enables Quarkus HTTP access/request log by default

Fixes #1498

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/src/main/resources/application.yaml
+++ b/src/main/resources/application.yaml
@@ -76,6 +76,11 @@ quarkus:
 
   # HTTP settings
   http:
+    # access log format, must be explicitly enabled
+    access-log:
+      enabled: true
+      pattern: "%h %l %t \"%r\" %s %b"
+
     # every /v1 path is authenticated by default
     # adapt if changing the authentication mechanism
     auth:
@@ -86,10 +91,6 @@ quarkus:
         default:
           paths: /v1/*
           policy: authenticated
-
-    # access log format, must be explicitly enabled
-    access-log:
-      pattern: "%h %l %t \"%r\" %s %b"
 
     limits:
       # Let's limit low-level maximum HTTP request size to 20 megs: stricter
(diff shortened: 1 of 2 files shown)


src/main/resources/application.yaml
```diff
@@ -115,6 +116,8 @@ quarkus:
         level: ERROR
       'com.datastax.oss.driver':
         level: WARN
+      'io.quarkus.http.access-log':
```
tatu-at-datastax, line 119: This is important: with default level of WARN request/access log would not be included without.

Yuqi-Du (approved):

PR #1506, opened 2024-10-07
Support binary vectors for Collections

### LINKED ISSUE #1494: Support binary vectors for Collections
Vectors can take extra processing on the client and the server, we want to improve this by supporting a binary method for sending them....

we will support this for inserting docs, and then we will return the doc back with the same format
`{"$vector": {"$binary": "c3VyZS4="}}`

this follows the EJSON format and is what we are doing for inserting vectors into tables. 

It's base64 encoded, = padded on the server side, we will decode this into a float[] using Float.intBitsToFloat(), and then pass the array into the driver etc for handling as a vector. More efficient processing of sending the byte array straight through the driver will come later, and it does not impact client side. 

We will POC the processing this week, clients may need to be updated to provide helpers to transform $binary and efficiently create the vector, we will work with them on design and getting a test platform ready.

We think this approach prioritises client side improvements first, will give some server side, and then if/when we get to pure buffers will give the best server side perf.

when we write the API design change we will also include allowing a $binary vector for sorting

10,000 iterations running the code in the gist shows https://gist.github.com/amorton/3ed3787bbad3f2f50fe8959984de9240

```
unpack_byte_buffer:     0.006098166981 seconds   0.04x
decode_base64:  0.009328916989 seconds   0.07x
decode_and_unpack:      0.015409209009 seconds   0.11x
load_json_string:       0.135685167013 seconds   0.97x
load_csv_string:        0.140139291005 seconds   1.00x
```
(x score is the time compared to the slowest time)

Hazel-Datastax:
**What this PR does**:
Support binary vectors for Collections

Fixes #1494 

**Checklist**
- [x] Changes manually tested


src/test/java/io/stargate/sgv2/jsonapi/service/shredding/DocumentShredderWithExtendedTypesTest.java
```diff
@@ -332,6 +332,31 @@ private void _testShredUUIDAutoGeneration(CollectionIdType idType, int uuidVersi
...
 
+  @Nested
+  class OkCasesBinaryVector {
+    @Test
+    public void shredSimpleBinaryVector() throws Exception {
+      final String inputJson =
+          """
+                  {
+                    "age" : 39,
+                    "$vector": {"$binary": "PoAAAD6AAAA+gAAAPoAAAD6AAAA="}
```
Hazel-Datastax, line 343: I get this binary by using Java to encode a float array `float[] vector = {0.25f, 0.25f, 0.25f, 0.25f, 0.25f}`. Is this what we expect from the client?

src/main/java/io/stargate/sgv2/jsonapi/service/shredding/collections/WritableShreddedDocument.java
```diff
@@ -304,6 +297,28 @@ public void shredVector(JsonPath path, ArrayNode vector) {
...
+      binaryPayload = Base64Variants.MIME_NO_LINEFEEDS.decode(binaryString);
+
+      int numFloats = binaryPayload.length / 4;
+      float[] floatArray = new float[numFloats];
+      ByteBuffer byteBuffer = ByteBuffer.wrap(binaryPayload);
+
+      for (int i = 0; i < numFloats; i++) {
+        int intBits = byteBuffer.getInt(); // Get next 4 bytes as an int
+        floatArray[i] = Float.intBitsToFloat(intBits); // Convert int to float
+      }
```
Hazel-Datastax, line 317: The logic assumes the binary could be decoded correctly; how could we verify if the binary has some problem?

src/main/java/io/stargate/sgv2/jsonapi/service/shredding/collections/DocumentShredder.java
```diff
@@ -313,15 +315,29 @@ private void traverseVector(JsonPath path, JsonNode value, DocumentShredderListe
     if (value.isNull()) {
       return;
     }
-    if (!value.isArray()) {
+    // should be either array or object
+    if (!value.isArray() && !value.isObject()) {
```
tatu-at-datastax, line 319: Should probably re-arrange to have:

```
if (value.isArray()) { 
   ..
} else if (value.isObject()) {
   ...
} else {
   // throw exception
}

File: src/main/java/io/stargate/sgv2/jsonapi/service/shredding/collections/DocumentShredder.java [hunk 4]
```diff
@@ -313,15 +315,29 @@ private void traverseVector(JsonPath path, JsonNode value, DocumentShredderListe
...
+      }
+      callback.shredVector(path, arr);
+    }
+    // e.g. "$vector": {"$binary": "c3VyZS4="}
+    if (value.isObject()) {
+      ObjectNode obj = (ObjectNode) value;
+      final Map.Entry<String, JsonNode> entry = obj.fields().next();
+      JsonExtensionType keyType = JsonExtensionType.fromEncodedName(entry.getKey());
+      if (keyType != BINARY) {
+        throw ErrorCodeV1.SHRED_BAD_DOCUMENT_VECTOR_TYPE.toApiException(entry.getKey());
```
tatu-at-datastax, line 337: Need to change error here: is used with `SHRED_BAD_DOCUMENT_VECTOR_TYPE` is:

```
throw ErrorCodeV1.SHRED_BAD_DOCUMENT_VECTOR_TYPE.toApiException(
          value.getNodeType().toString());
```

where node type is like `OBJECT` or `NUMBER`. But what is given here is key in JSON Object.
Should indicate that key for JSON Object needs to be `$binary`.

src/main/java/io/stargate/sgv2/jsonapi/service/shredding/collections/DocumentShredderListener.java
```diff
@@ -27,5 +27,7 @@ public interface DocumentShredderListener {
 
   void shredVector(JsonPath path, ArrayNode vector);
 
+  void shredVector(JsonPath path, ObjectNode vector);
```
tatu-at-datastax, line 30: Instead of passing the whole `ObjectNode`, should only pass `TextNode` value, or maybe just `String` contained. Caller can and should check it is textual (if (node.isTextual()). That way called method need not try to go and find it again.

src/main/java/io/stargate/sgv2/jsonapi/service/shredding/collections/DocumentShredder.java
```diff
@@ -313,15 +315,29 @@ private void traverseVector(JsonPath path, JsonNode value, DocumentShredderListe
...
+      }
+      callback.shredVector(path, obj);
```
tatu-at-datastax, line 339: as per comment below: instead of passing `obj`, pass either `JsonNode` for key `$binary`; or, maybe better text, textual value of that node (which is just `String`).

src/main/java/io/stargate/sgv2/jsonapi/service/shredding/collections/WritableShreddedDocument.java
```diff
@@ -304,6 +297,28 @@ public void shredVector(JsonPath path, ArrayNode vector) {
...
 
+    @Override
+    public void shredVector(JsonPath path, ObjectNode vector) {
+      // vector data is added only to queryVectorValues and exists keys index
+      addKey(path);
+
+      String binaryString = vector.get(JsonExtensionType.BINARY.encodedName()).asText();
+
+      byte[] binaryPayload;
+      binaryPayload = Base64Variants.MIME_NO_LINEFEEDS.decode(binaryString);
```
tatu-at-datastax, line 308: That makes sense, but we'll probably want to catch exception from decoder in case invalid String is passed (and there needs to be test(s) for it).

src/main/java/io/stargate/sgv2/jsonapi/service/shredding/collections/WritableShreddedDocument.java
```diff
@@ -304,6 +297,28 @@ public void shredVector(JsonPath path, ArrayNode vector) {
...
+
+      int numFloats = binaryPayload.length / 4;
```
tatu-at-datastax, line 310: As per our comments, should at least verify that byte[] is divisible by 4. So something like:

```
if ((binaryPayload.length & 3) != 0) { 
   // report error
}
```

src/main/java/io/stargate/sgv2/jsonapi/service/shredding/collections/WritableShreddedDocument.java
tatu-at-datastax: As per our discussion, I think 3 main possibilities:

1. Base64 decoding fails with exception (bad character etc)
2. Length decoded is not divisible by 4 
3. Number of floats does not match vector definition.

(1) and (2) can be checked here; (3) would probably need to be validated at a later point.
Should at least test with integration tests to see what happens with (3); Database will verify for sure.

src/test/java/io/stargate/sgv2/jsonapi/service/shredding/DocumentShredderWithExtendedTypesTest.java
tatu-at-datastax: It is, although you  will probably want to create payload `Base64Variant.encode()` and helper methods from `Float` to construct it instead of hard-coding.

But I think even more important part is that we need Integration Tests to ensure end-to-end handling -- that is, what is sent with `insertOne()` is successfully read back as same vector.

tatu-at-datastax (commented):
I think this looks like a good start: I recommend some changes, but the biggest thing that is needed are Integration Tests to ensure end-to-end handling.

Although I guess based on the way Projection works, it may not verify actual decoding.
But that's good enough I think.

src/main/java/io/stargate/sgv2/jsonapi/service/shredding/collections/DocumentShredderListener.java
```diff
@@ -27,5 +27,7 @@ public interface DocumentShredderListener {
 
   void shredVector(JsonPath path, ArrayNode vector);
 
+  void shredVector(JsonPath path, String binaryVector);
```
tatu-at-datastax, line 30: nit: binaryVector -> base64Vector

src/main/java/io/stargate/sgv2/jsonapi/service/shredding/collections/WritableShreddedDocument.java
```diff
@@ -304,6 +297,38 @@ public void shredVector(JsonPath path, ArrayNode vector) {
...
+      } catch (IllegalArgumentException e) {
+        throw ErrorCodeV1.SHRED_BAD_BINARY_VECTOR_VALUE.toApiException(
+            "Invalid content in EJSON $binary wrapper: not valid Base64-encoded String; problem: %s",
+            e.getMessage());
+      }
+
+      if ((binaryPayload.length & 3) != 0) {
+        // report error
+        throw ErrorCodeV1.SHRED_BAD_BINARY_VECTOR_VALUE.toApiException(
+            "Invalid content in EJSON $binary wrapper: decoded value is not a multiple of 4 bytes long");
```
tatu-at-datastax, line 317: add actual length in error message too. Something like:

```
"Invalid content in EJSON $binary wrapper: decoded value is not a multiple of 4 bytes long (%d bytes)"
```

src/test/java/io/stargate/sgv2/jsonapi/api/v1/VectorSearchIntegrationTest.java
```diff
@@ -405,6 +403,178 @@ public void insertInvalidVectorData() {
...
+          .body("status.insertedIds[0]", is(notNullValue()))
+          .body("data", is(nullValue()))
+          .body("errors", is(nullValue()));
+
+      // get the document and verify the vector value
+      Response response =
+          given()
+              .headers(getHeaders())
+              .contentType(ContentType.JSON)
+              .body("{\"find\": { \"projection\" : {\"$vector\" : 1}}}")
```
tatu-at-datastax, line 440: Shouldn't this use "filter" to find actual document inserted? I guess it maybe happens to get right one as first but that's not guaranteed?

src/test/java/io/stargate/sgv2/jsonapi/api/v1/VectorSearchIntegrationTest.java
```diff
@@ -405,6 +403,178 @@ public void insertInvalidVectorData() {
           .body("errors[0].errorCode", is("SHRED_BAD_VECTOR_VALUE"))
           .body("errors[0].exceptionClass", is("JsonApiException"));
     }
+
+    @Test
+    public void insertSimpleBinaryVector() {
+      final String id = UUID.randomUUID().toString();
+      final float[] expectedVector = new float[] {0.25f, 0.25f, 0.25f, 0.25f, 0.25f};
```
tatu-at-datastax, line 410: I forgot to suggest but let's use vector with some variation in values, so not all same.

Like 

```
{ 0.25, -1.5, 0.0, .... }
```

or whatever.

tatu-at-datastax (commented):
I like added tests, esp. failing ones. Just couple of last questions and I can approve this.

tatu-at-datastax (approved):
LGTM!

PR #1517, opened 2024-10-08
Add "inet" codec

### LINKED ISSUE #1386: Data API JsonCodec impl: misc other type(s) (inet)
In addition to other type categories to support, there are "leftovers" to support.

This includes:

* `boolean` (straight-forward, but one-off; already supported -- but allows coercions?)
* `inet` (IP address String, ipv4/v6)

We will need to support these, and so need to add Codecs and matching Unit and Integration tests. And possible clarify expected functioning as well (which coercions allowed etc)

**EDIT**: no longer includes "binary", moved to #1403

**EDIT**: no longer includes UUIDS, moved to #1442

tatu-at-datastax:
**What this PR does**:

Adds codecs for "inet" CQL type

Fixes #1386

**Checklist**
- [ ] Changes manually tested


Yuqi-Du (approved):
LGTM.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/filters/table/codecs/JSONCodec.java
```diff
@@ -363,6 +365,14 @@ static ByteBuffer byteBufferFromEJSON(DataType targetCQLType, EJSONWrapper wrapp
       }
       return ByteBuffer.wrap(binaryPayload);
     }
+
+    static InetAddress inetAddressFromString(String value) throws IllegalArgumentException {
+      try {
+        return InetAddress.getByName(value);
```
Yuqi-Du, line 371: Oh, so it handles invalid values for us.

tatu-at-datastax: yes, ones that don't convert. Exception thrown here will be re-caught and give proper API error response.

Issue #1518, opened 2024-10-09
Enforce CommandResultBuilder 


amorton:
code in #1497 added a CommandResultBuilder that makes it easier  and more reliable to create the CommandResult

However we still need to: 
* convert the CommandResult to a class, move the builder to be inner, and protect the constructor so it can only be created using the builder. 
* make the resolvers create the CommandResultBuilder in inject into the operations (I think) or do something else so we are reliably detecting if debug mode is set. See  `CommandResult.singleDocumentBuilder()` for comments. 
* we may not need / want to though, the OperationAttempt framework uses a page builder that already uses this CommandResultBuilder so it may not be necessary to make changes only for collections right now. 

this ticket is to remember, and decide if we want to make the change

PR #1519, opened 2024-10-09
Use WarningException, expand CommandResultBuilder 

### LINKED ISSUE #1497: Change the warnings section of the status to be full error object
Currently it looks like this: 

```
{
    "data": {
        "documents": [
            {
                "partition": "partition_1",
                "value": 2,
                "key": 2
            },
            {
                "partition": "demo_1",
                "value": 2,
                "key": 2
            }
        ],
        "nextPageState": null
    },
    "status": {
        "warnings": [
            "The filter includes columns that are not indexed. \n\nThe table demo.partitioned_tinyint_key has the primary key: partition(text), key(tinyint).\nAnd has indexes on the columns: [None].\nThe request filtered on the un-indexed columns: value.\n\nThe query was executed without taking advantage of the primary key or indexes on the table, this can have performance implications on large tables.\n\nSee documentation at XXXX for best practices for filtering."
        ]
    }
}
```

The warnings we make for allow filtering in the code use the ApiException, and in writing the code it would have been easier to just see the ID for the error we created (for the warning)., 

Change the warnings list to be a list of ApiExceptions and return the fill error object V2 structure in the response so it has family, scope etc. 

change so in the warnings, it is a list of error object v2 like below: 

```
{
            "message": "Only columns defined in the table schema can be filtered on.\n\nThe table demo.partitioned_tinyint_key defines the columns: partition(text), key(tinyint), value(int).\nThe filter included the following unknown columns: valueq.\n\nResend the request using only defined columns.",
            "scope": "FILTER",
            "errorCode": "UNKNOWN_TABLE_COLUMNS",
            "id": "c0c68d62-615d-4650-8275-b87264bed1c8",
            "family": "REQUEST",
            "title": "Unknown table columns in filter"
        }
```

amorton:
Fixes #1497 

Changes the warnings associated commands to use the new ApiExcpetions,
and `warnings` in the return status now returns a list of error object V2.

To do that needed to improve the way CommandResult was built, so it always
had a status map so the CommandProcessor could append a warning. To do that
expanded the CommandResultBuilder added for the OperationAttempts, removed
the many overloads used for CommandResult, and updated all creation of the
CommandResult to use the builder.

See also https://github.com/stargate/data-api/issues/1518 to continue this

**What this PR does**:

Fixes #<issue number>

**Checklist**
- [ ] Changes manually tested


tatu-at-datastax:
I have a philosophical objection here: Exceptions should be used for Exceptional situations, failures, and not as internal communication mechanism (i.e. not for convenience). And warnings by their nature are not such things.
It seems to me usage here is due to convenience of using message construction/templating part, and not because Exceptions would make sense otherwise.

Beyond philosophical part, Exceptions are also rather expensive things to use wrt performance: they are inefficient things to use if (although only if) created fresh, with stack traces.

So I am not sure we should take this approach for generating warnings as part of response payload.

amorton:
> Beyond philosophical part, Exceptions are also rather expensive things to use wrt performance: they are inefficient things to use if (although only if) created fresh, with stack traces.


I can make the WarningException overirde `fillInStackTrace()`  so it does not create a stack trace, and I can see a way to run the template without needing to instantiate a an `ApiException` sub class, I'll spin up a ticket when this is ready. It is already almost there: running the template creates a `ErrorInstance` record, which is used to construct the `ApiException`, we can use that and just then need a way to turn them into the CommandErrorV2 for serialisation. Should be a reasonable fast follow with no external impact .  

> It seems to me usage here is due to convenience of using message construction/templating part, and not because Exceptions would make sense otherwise.

Yup I was optimising for managing the content in the yaml file, and having a standard structure for errors and warnings.

amorton:
@tatu-at-datastax created this to track the change to not use a full exception https://github.com/stargate/data-api/issues/1549

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/DeprecatedCommand.java
```diff
@@ -16,16 +17,10 @@ public interface DeprecatedCommand extends Command {
...
-   * This ${commandName} has been deprecated and will be removed in future releases, use
-   * ${useCommandName} instead.
-   *
-   * @return String
-   */
-  default String getDeprecationMessage() {
-    return String.format(
-        "This %s has been deprecated and will be removed in future releases, use %s instead.",
-        this.commandName().getApiName(), useCommandName().getApiName());
+  /** A warning message will be added to commandResult when deprecated command is used. */
```
Yuqi-Du, line 20: cool, this part is also included!

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/SchemaChangeListener.java
```diff
@@ -47,4 +51,13 @@ public void onTableCreated(@NonNull TableMetadata table) {
         table.getKeyspace().asInternal(),
         table.getName().asInternal());
   }
+
+  @Override
```
Yuqi-Du, line 55: Thought Mahesh is going to have this fix in another PR?

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/CommandResultBuilder.java
```diff
@@ -132,8 +155,8 @@ public CommandResult build() {
           case SINGLE_DOCUMENT ->
               documents.isEmpty()
                   ? null
```
Yuqi-Du, line 157: Is this the correct response when documents are empty?

So for the previous collection path, if no documents, it will be
{
    "data": {
        "document": null
    }
}
But now, it will just be
{}.

Before this PR, table path is already this way, {}.
Wonder if this is intentional or?

```
          case SINGLE_DOCUMENT ->
              documents.isEmpty()
                  ?  new CommandResult.SingleResponseData(null)
                  : new CommandResult.SingleResponseData(documents.getFirst());
```

src/main/java/io/stargate/sgv2/jsonapi/service/processor/CommandProcessor.java
```diff
@@ -102,7 +105,11 @@ public <T extends Command, U extends SchemaObject> Uni<CommandResult> processCom
 
                     // yet more mucking about with suppliers everywhere :(
                     yield (Supplier<CommandResult>)
-                        () -> new CommandResult(List.of(errorBuilder.apply(apiException)));
+                        () ->
```
Yuqi-Du, line 108: OK, so this is the cut point for collection path.
See another comment, when returned documents are empty, the response body is just
```
{}.
```
Is this intended?

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/CommandResultBuilder.java
amorton: fixed below, and does not return responseData if there is any errors. All integration tests updated to catch this bug.

Yuqi-Du (approved):

PR #1525, opened 2024-10-10
  Retry operation when fail due to missing ALLOW FILTERING

### LINKED ISSUE #1491: Add retry on ALLOW FILTERing required for ReadAttempt
following from #1448 

we will also need to handle the case where ALLOW FILTERING was not added, but should have been, because we will want to avoid adding it when not needed - as we must also give a warning about WHY we turned it on. 

if we get an error that allow filtering was needed, then we should retry the request. This will / must count towards one of the retries in the retry policy

amorton:
Fixes #1491 

Using a retry policy in the ReadAttempt to retry when allow
 filtering was not added, adds a warning to the attempt, and updates the CQL options for the query to retry.

**What this PR does**:

Fixes #<issue number>

**Checklist**
- [ ] Changes manually tested


src/test/java/io/stargate/sgv2/jsonapi/service/operation/ReadAttemptTest.java
```diff
@@ -0,0 +1,68 @@
...
+            .formatted(
+                TEST_DATA.names.KEYSPACE_NAME.asCql(true), TEST_DATA.names.TABLE_NAME.asCql(true));
+
+    fixture
+        .attempt()
+        .setStatus(OperationAttempt.OperationStatus.READY)
+        .queryExecutor()
+        .doExecuteReadThrowThenReturn(allowFilteringError, fixture.resultSet())
+        .attempt()
+        .assertCompleted()
```
Yuqi-Du, line 44: Nit, wonder if the name should indicate it actually call execute for the attempt.
Similar to doExecuteReadThrowThenReturn() above, that one is more clear.

src/test/java/io/stargate/sgv2/jsonapi/service/operation/ReadAttemptTest.java
```diff
@@ -0,0 +1,68 @@
...
+        .attempt()
+        .verifyOnSuccessCalled(1, msg)
+        .attempt()
+        .verifyOnSuccessResultSet(fixture.resultSet(), msg)
+        .attempt()
+        .verifyOneWarning(WarningException.Code.QUERY_RETRIED_DUE_TO_INDEXING, msg)
+        .attempt()
+        .verifyWarningContains("The original query used the CQL: " + cql1, msg)
+        .attempt()
+        .verifyWarningContains("The original query used the parameters: " + List.of(), msg);
```
Yuqi-Du, line 66: Cool, so verifyWarningContains called twice is for the convenience of checking msgs.

src/test/java/io/stargate/sgv2/jsonapi/service/operation/OperationAttemptAssertions.java
```diff
@@ -0,0 +1,199 @@
...
+  public FixtureT verifyOnSuccessCalled(int times, String message) {
+    verify(
+            target,
+            times(times)
+                .description("onSuccess() called %s times when: %s".formatted(times, message)))
+        .onSuccess(any());
+    return fixture;
+  }
+
+  public FixtureT verifyOnSuccessResultSet(AsyncResultSet expected, String message) {
```
Yuqi-Du, line 166: I saw you also check these times(x) method in multiple place.

Take OnSuccess as example.
When you do the fluent way of test in ReadAttemptTest.

You have this:

```
 .verifyOnSuccessCalled(1, msg)
 .attempt()
 .verifyOnSuccessResultSet(fixture.resultSet(), msg)
 .attempt()
```

verifyOnSuccessResultSet() is doing the amount of method call check again.
Wonder if these two can be combined?

src/main/java/io/stargate/sgv2/jsonapi/service/operation/ReadAttempt.java
```diff
@@ -166,4 +180,60 @@ static class ReadResult {
...
+      // clear the retry context so that we don't keep a reference to the last statement
+      var currentRetryContext = retryContext;
+      retryContext = null;
+
+      var allowFilteringMissing =
+          throwable instanceof InvalidQueryException
+              && throwable.getMessage().endsWith("use ALLOW FILTERING");
+
+      if (allowFilteringMissing) {
+        if (LOGGER.isWarnEnabled()) {
```
Yuqi-Du, line 211: I think we definitely want this log line, correct?
I could image we may need to dig in some patterns or get some 'complains' from ALLOW FILTERING.

So, do we want to just log it anyway.
I don't know we set LOGGER.isWarnEnabled() anywhere.

Yuqi-Du (approved):
questions and some nit.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/ReadAttempt.java
amorton: the next highest level to log is ERROR and dont think it is an error, happy to keep as an warn for now. Added a ticket for metrics for retry and retry when allow filtering https://github.com/stargate/data-api/issues/1583

src/test/java/io/stargate/sgv2/jsonapi/service/operation/ReadAttemptTest.java
amorton: discussing - assertExecuteCompleted() 

maybe also change the doExcecute etc to be when* - they were called do because it was originally using do() from mock library

PR #1527, opened 2024-10-10
Add one IT for binary vector support

### LINKED ISSUE #1494: Support binary vectors for Collections

Hazel-Datastax:
**What this PR does**:
In the previous PR #1506, one suggestion was verifying whether the decoded vector dimension matched the vector definition. I thought we needed to explicitly verify it somewhere in the code but was not sure where was the best place. So I merged that PR first and tried to figure it out later. Then I realized we can defer the vector dimension verification to DB - the same as the normal vector array value. 

So I don't add any code but add one Integration test to make sure it works.

Fixes #1494 

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/src/test/java/io/stargate/sgv2/jsonapi/api/v1/VectorSearchIntegrationTest.java
+++ b/src/test/java/io/stargate/sgv2/jsonapi/api/v1/VectorSearchIntegrationTest.java
@@ -459,7 +459,7 @@ public class VectorSearchIntegrationTest extends AbstractKeyspaceIntegrationTest
       // Convert the byte array to a float array
       float[] decodedVector = decodeBase64BinaryVectorToFloatArray(base64VectorFromResponse);
 
-      // Step 7: Verify that the decoded float array is equal to the expected vector
+      // Verify that the decoded float array is equal to the expected vector
       Assertions.assertArrayEquals(expectedVector, decodedVector, 0.0001f);
     }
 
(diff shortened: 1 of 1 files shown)


src/test/java/io/stargate/sgv2/jsonapi/api/v1/VectorSearchIntegrationTest.java
```diff
@@ -577,6 +577,38 @@ public void insertBinaryVectorWithInvalidDecodedValue() {
               is(
                   "Bad binary vector value to shred: Invalid content in EJSON $binary wrapper: decoded value is not a multiple of 4 bytes long (3 bytes)"));
     }
+
+    @Test
+    public void insertBinaryVectorWithUnmatchedVectorDimension() {
```
tatu-at-datastax, line 582: naming: maybe start with word "fail" (like "failToInsert....")?

I know existing tests aren't named that way but I think it is good for test method name to indicate that it is testing a failing insert.

tatu-at-datastax (approved):
LGTM: just one minor naming suggestion

PR #1535, opened 2024-10-14
List tables refactor

maheshrajamani:
**What this PR does**:
List tables refactor to use OperationAttempt

**Checklist**
- [x] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/service/operation/MetadataAttempt.java
```diff
@@ -0,0 +1,275 @@
...
+    }
+    return Uni.createFrom().item(new EmptyAsyncResultSet());
+  }
+
+  /**
+   * Convert table schema object to table response which is returned as response for `listTables`
+   *
+   * @return
+   */
+  protected TableResponse getTableSchema(TableSchemaObject tableSchemaObject) {
```
amorton, line 88: if this is only use by ListTables it should not be on the super class ? 

Also created this ticket to refactor the serialisartion later so we can test it better https://github.com/stargate/data-api/issues/1544

src/main/java/io/stargate/sgv2/jsonapi/service/operation/MetadataAttempt.java
```diff
@@ -0,0 +1,275 @@
...
+  @JsonPropertyOrder({"name", "definition"})
+  @JsonInclude(JsonInclude.Include.NON_NULL)
+  public record TableResponse(String name, TableDefinition definition) {
+
+    @JsonPropertyOrder({"columns", "primaryKey"})
+    @JsonInclude(JsonInclude.Include.NON_NULL)
+    record TableDefinition(Map<String, ColumnType> columns, PrimaryKey primaryKey) {}
+  }
+
+  protected List<TableSchemaObject> getTables() {
```
amorton, line 203: If this is only used by ListTables it should not be on the super class.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/MetadataAttempt.java
```diff
@@ -0,0 +1,275 @@
...
+    return keyspaceMetadata
+        .get()
```
amorton, line 205: nit: lint checking kicks in to say this is access to an Optional without checking if present, we should unwrap the optional in executeStatement and reference the KeyspaceMetadata rather than use Optional<KeyspaceMetadata>

src/main/java/io/stargate/sgv2/jsonapi/service/operation/MetadataAttempt.java
```diff
@@ -0,0 +1,275 @@
...
+import java.util.concurrent.CompletionStage;
+import java.util.stream.Collectors;
+
+/** An attempt to execute commands that need data from metadata */
+public abstract class MetadataAttempt<SchemaT extends SchemaObject>
+    extends OperationAttempt<MetadataAttempt<SchemaT>, SchemaT> {
+  // this will be set on executeStatement
+  private Optional<KeyspaceMetadata> keyspaceMetadata;
+
+  private static ObjectMapper objectMapper = new ObjectMapper();
```
amorton, line 49: this can be final

src/main/java/io/stargate/sgv2/jsonapi/service/operation/MetadataAttempt.java
```diff
@@ -0,0 +1,275 @@
...
+          "No next page. Use #hasMorePages before calling this method to avoid this error.");
+    }
+
+    @Override
+    public boolean wasApplied() {
+      return true;
+    }
+  }
+
+  public static class NoRetryPolicy extends RetryPolicy {
```
amorton, line 264: This is the same as OperationAttempt.RetryPolicy.NO_RETRY but changes the default from 1 ms to 10 ms , is there a reason not to use the existing ? 

also, if all it needs is to change the retry time create an instance of the OperationAttempt.RetryPolicy

src/main/java/io/stargate/sgv2/jsonapi/service/operation/MetadataAttempt.java
```diff
@@ -0,0 +1,275 @@
...
+        .map(table -> TableSchemaObject.from(table, objectMapper))
+        // get as list
+        .toList();
+  }
+
+  /**
+   * EmptyAsyncResultSet implementation to be used only for metadata attempt where no cql query is
+   * run.
+   */
+  private static class EmptyAsyncResultSet implements AsyncResultSet {
```
amorton, line 222: we can move this from an inner class, it's not related to the MetadataAttempt , the service.cqldriver package would be a good place

src/main/java/io/stargate/sgv2/jsonapi/service/operation/MetadataAttemptPage.java
```diff
@@ -0,0 +1,74 @@
...
+  public static <SchemaT extends KeyspaceSchemaObject>
+      MetadataAttemptPage.Builder<SchemaT> builder() {
+    return new MetadataAttemptPage.Builder<>();
+  }
+
+  @Override
+  protected void buildCommandResult() {
+    addAttemptWarningsToResult();
+    addAttemptErrorsToResult();
+    if (attempts.get(0).status() != OperationAttempt.OperationStatus.ERROR) {
```
amorton, line 39: attempts is a OperationAttemptContainer and has completedAttempts() to get completed attempts , it checks actually COMPLETED not just not ERROR

we should prob also check if all the size of attempts is != 1 and throw a regular InvalidStateExcpetion as a double check.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/MetadataAttemptPage.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    public Builder<SchemaT> usingCommandStatus(CommandStatus statusKey) {
+      this.statusKey = statusKey;
+      return this;
+    }
+
+    public Supplier<CommandResult> getOperationPage() {
+      var resultBuilder =
+          new CommandResultBuilder(
```
amorton, line 68: I changed the way this was build in this PR, still not merged cause of failing IT https://github.com/stargate/data-api/pull/1519

leave this for now it can change when this is done https://github.com/stargate/data-api/issues/1518

amorton (approved):
there are some things I think can be improved, such as functions on MetadatAttempt that should not be there. 

but nothing that looks blocking, and we can keep iterating  - the main goal was getting it into the OperationAttempt so we fit into the new standard way , and results /errors are handled consistently. 

thanks

PR #1536, opened 2024-10-14
Fix #1531: revert change to error messages

### LINKED ISSUE #1531: Issue with `COLLECTION_NOT_EXIST` error w/ Errors v2
When performing an operation on a nonexistent error on `v1.0.17` in dev, getting the following response from the server:
```
{
  errors: [
    {
      message: 'non_existent_collection',
      errorCode: 'COLLECTION_NOT_EXIST',
      id: '3b0dcaec-d46f-43b0-bc4c-ad9425595c44',
      family: 'REQUEST',
      title: 'Collection does not exist, collection name',
      scope: 'EMPTY'
    }
  ]
}
```

The `title` is expected to actually be of the following format, with a trailing colon & the collection name:
```
Collection does not exist, collection name: [COLLECTION NAME]
```

Beyond, due to unfortunate decisions in the TS client in its early stages, it actually attempts to parse the collection name from the `message` field, expecting an error message of the format `.*:(.*?)`, e.g. `'Collection does not exist, collection name: [COLLECTION NAME]'`, to extract the collection name.

It's not strictly necessary to have the message be of this format, but it'd be detrimental to the TS client in that the `CollectionNotFoundError.collectionName` field will be permanently `undefined` as it would no longer parse the error message correctly (since atm it's just `'non_existent_collection'`).

It shouldn't be a huge issue however, given that it will be fixed in future client versions, and it's likely not the most common error. However, it would be nice if the message would abide to that format for the time being.

tatu-at-datastax:
**What this PR does**:

Will fix #1531: unintended change to error messages wrt ErrorsV2

Fixes #1531

**Checklist**
- [x] Changes manually tested


tatu-at-datastax:
NOTE: first version shows breakage, intentionally, by enabling ErrorsV2.

tatu-at-datastax:
As can be seen from https://github.com/stargate/data-api/actions/runs/11330530094/job/31508553734?pr=1536 lots of unit tests fail:

```
Error:  Tests run: 2186, Failures: 130, Errors: 0, Skipped: 0
```

when ErrorsV2 enabled and similarly many ITs:

```
Error:  Tests run: 748, Failures: 94, Errors: 0, Skipped: 2
```

Will now revert message generation change to resolve these.

src/test/java/io/stargate/sgv2/jsonapi/exception/JsonApiExceptionTest.java
```diff
@@ -30,7 +30,6 @@ public void happyPath() {
               error -> {
                 assertThat(error.message()).isEqualTo("Provided command unknown");
                 assertThat(error.fields())
-                    .hasSize(2)
```
tatu-at-datastax, line 33: No need to check exact count: ErrorsV2 will add more entries. Same for other `hasSize()` cases.

tatu-at-datastax:
Ok: all test errors resolved -- will now remove ErrorsV2 config override.

tatu-at-datastax:
ErrorsV2 override removed, tests still pass (as they should).

src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCodeV1.java
```diff
@@ -230,9 +225,6 @@ public JsonApiException toApiException(Throwable cause, String format, Object...
   }
 
   private String getErrorMessage(String format, Object... args) {
-    if (ExtendError.enabled()) {
```
tatu-at-datastax, line 233: This is the real fix: V2 handling should not be imposed on V1 due to incompatible handling -- change to V2 should be done by upgrading to new `ErrorCode` approach, on per error-code instance (moving over all `ErrorCodeV1` uses one by one).

amorton (approved):

PR #1564, opened 2024-10-18
Alter table command implementation

### LINKED ISSUE #1563: Implement `alterTable` command
Implement `alterTable` command to add and drop columns. Add add and drop vectorize service option for existing vector columns.

add Columns
`{
    "alterTable": {
        "operation": {
            "add": {
                "columns": {
                    "new_col_1": "text",
                    "new_col_2": {
                        "type": "map",
                        "keyType": "text",
                        "valueType": "text"
                    },
                    "content": {
                        "type": "vector",
                        "dimension": 1024,
                        "service": {
                            "provider": "nvidia",
                            "modelName": "NV-Embed-QA"
                        }
                    },
                    "embedding": {
                        "type": "vector",
                        "dimension": 1024
                    }
                }
            }
        }
    }
}` 

Drop columns:
`{
    "alterTable": {
        "operation": {
            "drop": {
                "columns": ["new_col_1", "new_col_2"]
            }
        }
    }
}`

Add vectorize service config
`{
    "alterTable": {
        "operation": {
            "addVectorize": {
                "columns": {
                    "embedding": {
                        "provider": "mistral",
                        "modelName": "mistral-embed"
                    }
                }
            }
        }
    }
}`

Drop vectorize service config:
`{
    "alterTable": {
        "operation": {
            "dropVectorize": {
                "columns": ["embedding"]
            }
        }
    }
}`

maheshrajamani:
**What this PR does**:
Supports alter table command to perform

1. Add columns
2. Drop columns
3. Add vectorize service config
4. Drop vectorize service config

Fixes #1563 

**Checklist**
- [x] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/api/model/command/table/definition/datatype/ComplexTypes.java
```diff
@@ -32,6 +33,24 @@ public String keyTypeName() {
     public String valueTypeName() {
       return valueType.getApiDataType().getApiName();
     }
+
+    @Override
```
maheshrajamani, line 37: Added equals and hasCode to implement unit test.

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/TableSchemaObject.java
```diff
@@ -1,22 +1,14 @@
 package io.stargate.sgv2.jsonapi.service.cqldriver.executor;
```
maheshrajamani, line 1: Class changed to use `TableMetadataUtils` to reuse code

src/main/java/io/stargate/sgv2/jsonapi/service/operation/SchemaAttempt.java
```diff
@@ -38,6 +45,24 @@ protected Uni<AsyncResultSet> executeStatement(CommandQueryExecutor queryExecuto
 
   protected abstract SimpleStatement buildStatement();
 
+  // Add customProperties which has table properties for vectorize
```
maheshrajamani, line 48: Moved it to schema attempt since it's needed for both Create table and Alter table commands

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/AlterTableCommandResolver.java
```diff
@@ -0,0 +1,285 @@
...
+                      return new VectorConfig.ColumnVectorDefinition.VectorizeConfig(
+                          vectorizeConfig.provider(),
+                          vectorizeConfig.modelName(),
+                          vectorizeConfig.authentication(),
+                          vectorizeConfig.parameters());
+                    }));
+    final AlterTableAttempt addColumnsAttempt =
+        new AlterTableAttemptBuilder(schemaObject).addColumns(addColumns).build();
+    if (!vectorizeConfigMap.isEmpty()) {
+      // Reading exising vectorize config from the table metadata
```
tatu-at-datastax, line 117: typo: exising -> existing

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/AlterTableCommandResolver.java
```diff
@@ -0,0 +1,285 @@
...
+                      final VectorizeConfig vectorizeConfig = vectorType.getVectorConfig();
+                      validateVectorize.validateService(vectorizeConfig, vectorType.getDimension());
+                      return new VectorConfig.ColumnVectorDefinition.VectorizeConfig(
+                          vectorizeConfig.provider(),
+                          vectorizeConfig.modelName(),
+                          vectorizeConfig.authentication(),
+                          vectorizeConfig.parameters());
+                    }));
+    final AlterTableAttempt addColumnsAttempt =
+        new AlterTableAttemptBuilder(schemaObject).addColumns(addColumns).build();
```
tatu-at-datastax, line 115: It seems to me again that the Builder wouldn't be needed here; instead, 4 factory methods would suffice.
But I guess there may be some need to follow this approach? Like, incrementing `position` for later attempts?

Anyway -- what seems like code smell are all those `assert`s: we get one (and only one) alter operation, so we should have very little actual state. But yet with Builder we need to guard that unnecessary state. So there seems to be no value for Builder here.

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/AlterTableCommandResolver.java
```diff
@@ -0,0 +1,285 @@
...
+    // Validate the columns to be dropped are present
+    List<CqlIdentifier> primaryKeys =
+        tableMetadata.getPrimaryKey().stream().map(ColumnMetadata::getName).toList();
+    for (String columnName : dropColumns) {
+      CqlIdentifier column = CqlIdentifierUtil.cqlIdentifierFromUserInput(columnName);
+      if (primaryKeys.contains(column)) {
+        throw SchemaException.Code.COLUMN_CANNOT_BE_DROPPED.get(
+            Map.of(
+                "reason",
+                "Primary key column `%s` cannot be dropped".formatted(column.asCql(true))));
```
tatu-at-datastax, line 153: Instead of `column.asCql` should use method from `CqlIdentifierUtil`.

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/AlterTableCommandResolver.java
```diff
@@ -0,0 +1,285 @@
...
+                vectorizeConfig.modelName(),
+                vectorizeConfig.authentication(),
+                vectorizeConfig.parameters());
+        vectorizeConfigMap.put(entry.getKey(), dbVectorConfig);
+      } else {
+        throw SchemaException.Code.NON_VECTOR_TYPE_COLUMN.get(Map.of("column", entry.getKey()));
+      }
+    }
+
+    // Reading exising vectorize config from the table metadata
```
tatu-at-datastax, line 232: typo: exising->existing

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/AlterTableCommandResolver.java
```diff
@@ -0,0 +1,285 @@
...
+    alterTableAttempts.add(
+        new AlterTableAttemptBuilder(schemaObject).customProperties(customProperties).build());
+    return alterTableAttempts;
+  }
+
+  private List<AlterTableAttempt> handleDropVectorize(
+      AlterTableOperationImpl.DropVectorize dc, TableSchemaObject schemaObject) {
+    TableMetadata tableMetadata = schemaObject.tableMetadata();
+    List<AlterTableAttempt> alterTableAttempts = new ArrayList<>();
+    // Reading exising vectorize config from the table metadata
```
tatu-at-datastax, line 249: typo: exising->existing

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/AlterTableCommandResolver.java
```diff
@@ -0,0 +1,285 @@
...
+
+      final Optional<IndexMetadata> first =
+          tableMetadata.getIndexes().values().stream()
+              .filter(indexMetadata -> indexMetadata.getTarget().equals(column.asCql(true)))
+              .findFirst();
+      if (first.isPresent()) {
+        throw SchemaException.Code.COLUMN_CANNOT_BE_DROPPED.get(
+            Map.of(
+                "reason",
+                "Index exists on the column `%s`, drop `%s` index to drop the column"
```
tatu-at-datastax, line 168: Should there not be property "column" to bind via template (and probably "index" as well)?

src/main/resources/errors.yaml
```diff
@@ -458,8 +458,35 @@ request-errors:
...
     body: |-
-      Provided index configuration is not valid: ${reason}.
-  
+      Provided index configuration is not valid: ${reason}.V
+
+  - scope: SCHEMA
+    code: COLUMN_ALREADY_EXISTS
+    title: Column already exist in the table.
+    body: |-
+      Column `${column}` already exist in the table.
```
tatu-at-datastax, line 467: typo: exist->exists

src/main/resources/errors.yaml
```diff
@@ -458,8 +458,35 @@ request-errors:
...
+    code: COLUMN_NOT_FOUND
+    title: Column doesn't exist in the table.
+    body: |-
+      Column `${column}` doesn't exist in the table.
+
+  - scope: SCHEMA
+    code: COLUMN_CANNOT_BE_DROPPED
+    title: Column can't be dropped.
+    body: |-
+      Column can't be dropped: ${reason}.
```
tatu-at-datastax, line 479: Should have `${column}` passed for templating

tatu-at-datastax (approved):
Looks good; a few suggestions to consider.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/AlterTableOperation.java
```diff
@@ -0,0 +1,13 @@
...
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, include = JsonTypeInfo.As.WRAPPER_OBJECT)
+@JsonSubTypes({
+  @JsonSubTypes.Type(value = AlterTableOperationImpl.AddColumns.class),
+  @JsonSubTypes.Type(value = AlterTableOperationImpl.DropColumns.class),
+  @JsonSubTypes.Type(value = AlterTableOperationImpl.AddVectorize.class),
+  @JsonSubTypes.Type(value = AlterTableOperationImpl.DropVectorize.class)
+})
+public interface AlterTableOperation {}
```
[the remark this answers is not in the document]
maheshrajamani: Other places interface and implementation are kept separate (like ColumnType, Command, etc). So maintained the same code structure.

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/SchemaChangeListener.java
```diff
@@ -47,4 +47,13 @@ public void onTableCreated(@NonNull TableMetadata table) {
         table.getKeyspace().asInternal(),
         table.getName().asInternal());
   }
+
+  /** When table is updated, drop the corresponding collectionSetting cache entry if existed */
```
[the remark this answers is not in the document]
maheshrajamani: This is needed for alter table, but it exposed bug in filter clause. Fixed the issue as per @amorton suggestion

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/TableMetadataUtils.java
```diff
@@ -0,0 +1,88 @@
...
+        Iterator<Map.Entry<String, JsonNode>> it = vectorizeByColumns.fields();
+        while (it.hasNext()) {
+          Map.Entry<String, JsonNode> entry = it.next();
+          try {
+            var vectorizeConfig =
+                objectMapper.treeToValue(
+                    entry.getValue(), VectorConfig.ColumnVectorDefinition.VectorizeConfig.class);
+            vectorizeConfigMap.put(entry.getKey(), vectorizeConfig);
+          } catch (JsonProcessingException | IllegalArgumentException e) {
+            throw SchemaException.Code.INVALID_VECTORIZE_CONFIGURATION.get(
```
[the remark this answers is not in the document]
maheshrajamani: Don't think error message should include, Will add a logger though

src/main/resources/errors.yaml
maheshrajamani: Appended the column as part of the reason 
`"Primary key column `%s` cannot be dropped".formatted(column.asCql(true))`

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/AlterTableCommandResolver.java
maheshrajamani: Didn't get this, Have the column and index name binded as part of the reason text.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/tables/WhereCQLClauseAnalyzer.java
```diff
@@ -89,6 +89,11 @@ AnalysisStrategy getStrategy(WhereCQLClauseAnalyzer analyzer) {
           DataTypes.BLOB,
           DataTypes.UUID);
 
+  // Datatypes that need ALLOW FILTERING even when there is a SAI on the column when <, >, <=, >= is
```
maheshrajamani, line 92: fix for filter clause IT errors

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/AlterTableOperation.java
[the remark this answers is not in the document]
tatu-at-datastax: I am not sure that is a good reason to create separation without actual reason other than other things (for which this may make sense) do that.
Especially since this "tag" interface with no methods.

src/main/resources/errors.yaml
tatu-at-datastax: I know but why not follow the usual way with templating, same as other messages?

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/AlterTableCommandResolver.java
tatu-at-datastax: Just asking why pre-format these in code, instead of using templating mechanism ErrorsV2 offers, like other error messages do.
Using templating mechanism could also allow us to include more metadata about failure itself  later on (if we included key/value pairs of information in JSON -- not something we do yet).

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/TableMetadataUtils.java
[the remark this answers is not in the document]
tatu-at-datastax: Hmmh. Why not? If we log, clients cannot give us this information but instead will require us to try to find the logged entry.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/AlterTableOperation.java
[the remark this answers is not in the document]
tatu-at-datastax: Having said that, it's ok -- just something to consider going forward, adding things when needed and not if there is no foreseeable need.

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/SchemaChangeListener.java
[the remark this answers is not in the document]
tatu-at-datastax: In future let's try to isolate fixes to their own PRs were possible. Fine in this case.

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/AlterTableCommandResolver.java
tatu-at-datastax: Using a Builder ok; will raise the issue with Aaron.

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/AlterTableCommandResolver.java
tatu-at-datastax: One note: if it is not possible to use same templated message for all instance of Error, then fine to pre-format message to include via templating. But in general I think all variable parts that can be handled by yaml template should handled by it, and not by `String.format()`. This to keep templating in one place (ideally).

src/main/resources/errors.yaml
maheshrajamani: I am trying to balance on number of error codes that api emit vs providing meaningful error message. If I have to avoid formatting text in the code then it will end up with many types of error code.

tatu-at-datastax: Ok in that case that makes sense. I don't want to force creation of multiple one-off codes for related cases.

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/TableMetadataUtils.java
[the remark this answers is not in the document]
maheshrajamani: We have not to sent technical error message to the user. In that line of thought sending `e.getMessage()` from `JsonProcessingException | IllegalArgumentException` is avoided.

vkarpov15 (commented):
API looks good. One small question: what happens if there's a drop vectorize command and a command that uses vectorize in parallel? Would the command that uses vectorize throw an error?

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/TableMetadataUtils.java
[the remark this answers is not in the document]
tatu-at-datastax: Ok, something to discuss later on. I know Aaron is +1 on "don't confuse user", but to me hiding relevant information they can easily pass on makes more sense in some cases.

But in here it probably won't matter too much as it should never happen :)

PR #1574, opened 2024-10-21
Create table issue fixes.

### LINKED ISSUE #1565: Bug - use Cql Identifier for specifying clustering key order
https://github.com/stargate/data-api/blob/main/src/main/java/io/stargate/sgv2/jsonapi/service/operation/tables/CreateTableAttempt.java#L140

Here, when specifying clustering order, it is not using the CQL IndentifierUtil, what will happen is if you have a clustering key “clusterKey” defined as column, here it looks for “clusterkey” to specify order, and will Not Found. (edited)

### LINKED ISSUE #1570: Restrict map key types to be text or ascii for create table
Initial release only supports string, text or ascii, key types because that is a better map to a JSON object.

maheshrajamani:
**What this PR does**:
Create table bug fixes for using CqlIdentifier for Clustering column ordering
Map type only supports key to be text or ascii.

Fixes #1565 
Fixes #1570 

**Checklist**
- [ ] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/api/model/command/table/definition/datatype/ColumnType.java
```diff
@@ -59,15 +59,23 @@ static ColumnType fromString(
...
+            valueColumnType = fromString(valueType, null, null, dimension, vectorConfig);
           } catch (SchemaException se) {
-            throw SchemaException.Code.MAP_TYPE_INCORRECT_DEFINITION.get();
+            throw SchemaException.Code.MAP_TYPE_INCORRECT_DEFINITION.get(
+                Map.of("reason", "Data types used for `keyType` or `valueType` are not supported"));
+          }
+          if (!(PrimitiveTypes.TEXT.equals(keyColumnType)
+              || PrimitiveTypes.ASCII.equals(keyColumnType))) {
+            throw SchemaException.Code.MAP_TYPE_INCORRECT_DEFINITION.get(
+                Map.of("reason", "`keyType` must be `text` or `ascii`"));
```
tatu-at-datastax, line 76: Should this include actual type used as well? (" was %s")

src/main/resources/errors.yaml
```diff
@@ -393,7 +393,7 @@ request-errors:
     code: MAP_TYPE_INCORRECT_DEFINITION
     title: Map column data type definition provided in the request is incorrect.
     body: |-
-      Map column data type definition provided in the request is incorrect.
+      Map column data type definition provided in the request is incorrect: ${reason}
```
tatu-at-datastax, line 396: should it be "invalid" instead of "incorrect"? Or maybe even better "unsupported".

tatu-at-datastax (approved):
LGTM: added some suggestions (as usual :) )

PR #1579, opened 2024-10-21
Fix #1493: support Base64-encoded Vectors

### LINKED ISSUE #1493: Data API JsonCodec: Base64-binary support for vectors
(note: follow up to #1492)

After basic "JSON number array" input, also need to add support for optimized variant for vectors: Base64-encoded "packed floats" binary, in which client:

1. First converts from `float[]` (or platform equivalent) into IEEE-754 packed (4 bytes per float) byte sequence
2. Encodes resulting byte sequence into padded "MIME" Base64 String
3. Encloses String into EJSON "$binary" wrapping -- same as `blob` values

and Data API will decode, unpack things, passes to CQL driver in structure it expects.

This issue DOES NOT change handling on return path: it will use "simple"/"plain" JSON Number array representation -- this because we do not know which representation was used for insert/update (unlike with Collections where we keep a copy in `doc_json`)

For further work we'd file another issue as necessary to allow requesting optimal representation in results.

tatu-at-datastax:
**What this PR does**:

Adds support for Base64-encoded Vector values for API Tables

Fixes #1493

**Checklist**
- [ ] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/service/operation/filters/table/codecs/JSONCodec.java
```diff
@@ -345,10 +350,15 @@ static ByteBuffer byteBufferFromEJSON(DataType targetCQLType, EJSONWrapper wrapp
         if (value.isBinary()) {
```
maheshrajamani, line 350: Can we add a test case to pass the value as binary?

src/main/java/io/stargate/sgv2/jsonapi/service/operation/filters/table/codecs/VectorCodecs.java
```diff
@@ -26,38 +29,40 @@ public static <JavaT, CqlT> JSONCodec<JavaT, CqlT> arrayToCQLFloatVectorCodec(
...
-          "expected vector of length "
-              + expLen
-              + ", got one with "
-              + vectorIn.size()
-              + " elements");
-    }
-    List<Float> floats = new ArrayList<>(expLen);
+    validateVectorLength(vectorType, vectorIn, vectorIn.size());
+
+    List<Float> floats = new ArrayList<>(vectorIn.size());
```
maheshrajamani, line 65: I thought Jonathan Ellis said there is a way to pass binary encoded vector as input, if so we should use it.

src/main/java/io/stargate/sgv2/jsonapi/util/Base64Util.java
```diff
@@ -0,0 +1,13 @@
+package io.stargate.sgv2.jsonapi.util;
+
+import com.fasterxml.jackson.core.Base64Variants;
+
+public interface Base64Util {
```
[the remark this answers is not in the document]
maheshrajamani: Ah, should have seen this first.

src/main/java/io/stargate/sgv2/jsonapi/util/CqlVectorUtil.java
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * Static utility methods for working with vector value representations; specifically with Cassandra
+ * {@code float}-based vector types.
+ */
+public interface CqlVectorUtil {
+  static CqlVector<Float> bytesToCqlVector(byte[] packedBytes) {
+    return floatsToCqlVector(bytesToFloats(packedBytes));
+  }
+
+  static float[] bytesToFloats(byte[] packedBytes) throws IllegalArgumentException {
```
maheshrajamani, line 16: Can you add some comments on what is done?

maheshrajamani (approved):
LGTM., some nit comments, on things I spent time understanding.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/filters/table/codecs/JSONCodec.java
tatu-at-datastax: Hmmh. Let me think about that... it is not possible for JSON to produce `BinaryNode` (it is only possible via custom nodes or binary formats -- like CBOR or Smile).

But maybe I can hand-code it. Let me see about that.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/filters/table/codecs/VectorCodecs.java
tatu-at-datastax: I couldn't find a method to do that but I'll double-check now.

tatu-at-datastax: No, unfortunately `CqlVector<T extends Number>` is internally represented as

```
private final List<T> list;
```

so I don't think there is a way around that.

Unless we could use type other than `CqlVector` to bind `vector` column from?

src/main/java/io/stargate/sgv2/jsonapi/util/CqlVectorUtil.java
tatu-at-datastax: Will do.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/filters/table/codecs/JSONCodec.java
tatu-at-datastax: Unfortunately not quite as easy to push through plumbing. Will need to defer this (but leaving "binary" handling in there since it was not added in this pr.

vkarpov15 (commented):
LGTM :+1:

PR #1600, opened 2024-10-23
Fix #1451: add Codec support for non-nested unfrozen Maps

### LINKED ISSUE #1451: Data API JsonCodec impl: unfrozen `map` type
After Scalar types we need to support unfrozen `Map` type, with non-nested value type (that is, scalar values, not containers).

As usual, codec implementation to be tested via unit tests (more extensive) and selected Insert-in-Table integration tests.

tatu-at-datastax:
**What this PR does**:

Adds Codec support (Insert, Find) for support Map values (non-nested, unfrozen)

Fixes #1451

**Checklist**
- [ ] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/service/operation/filters/table/codecs/CollectionCodecs.java
```diff
@@ -103,14 +132,38 @@ static Set<Object> toCQLSet(
...
       result.add(elementCodec.toCQL(element));
     }
     return result;
   }
 
-  private static JSONCodec<Object, Object> findElementCodec(
+  static Map<String, Object> toCQLMap(
+      List<JSONCodec<?, ?>> valueCodecs, DataType elementType, Map<?, ?> rawMapValue)
+      throws ToCQLCodecException {
+    Map<String, JsonLiteral<?>> mapValue = (Map<String, JsonLiteral<?>>) rawMapValue;
```
maheshrajamani, line 145: Currently only support String key type, but eventually it support for other key data types by having different json structure. Should this code try to resolve key type codec and resolve key?

src/main/java/io/stargate/sgv2/jsonapi/service/operation/filters/table/codecs/CollectionCodecs.java
```diff
@@ -143,4 +211,22 @@ static JsonNode toJsonNode(
...
   }
+
+  /** Method that will convert from driver-provided CQL Map type into JSON output. */
+  static JsonNode toJsonNodeFromMap(
+      JSONCodec<?, ?> elementCodec0, ObjectMapper objectMapper, Object mapValue)
+      throws ToJSONCodecException {
+    JSONCodec<?, Object> elementCodec = (JSONCodec<?, Object>) elementCodec0;
+    final ObjectNode result = objectMapper.createObjectNode();
+    for (Map.Entry<Object, Object> entry : ((Map<Object, Object>) mapValue).entrySet()) {
+      final String key = String.valueOf(entry.getKey());
```
maheshrajamani, line 222: Are we validating anywhere to not read if map key is not text? Not all cassandra data type can be supported with `String.valueOf(entry.getKey())`

src/main/java/io/stargate/sgv2/jsonapi/service/operation/filters/table/codecs/JSONCodecRegistry.java
```diff
@@ -204,6 +222,19 @@ public <JavaT, CqlT> JSONCodec<JavaT, CqlT> codecToJSON(DataType fromCQLType) {
       return (JSONCodec<JavaT, CqlT>)
           CollectionCodecs.buildToJsonSetCodec(valueCodecCandidates.get(0));
     }
+    if (fromCQLType instanceof MapType mt) {
+      final DataType keyType = mt.getKeyType();
+      if (!isSupportedMapKeyType(keyType)) {
```
maheshrajamani, line 227: Is this to check if the column has supported key type.

src/test/java/io/stargate/sgv2/jsonapi/service/operation/filters/table/codecs/JSONCodecRegistryTest.java
```diff
@@ -337,6 +343,32 @@ private static Stream<Arguments> validCodecToCQLTestCasesCollections() {
             Set.of(-0.75, 42.5)));
   }
 
+  private static Stream<Arguments> validCodecToCQLTestCasesMaps() {
```
maheshrajamani, line 346: Can you create an issue to write tests for codec for CQL->Json? We need to mock rock and test the behavior for non String key maps?

maheshrajamani: Thinking loud this needs to be done for all Operations of api tables to check behavior of unsupported datatypes.

maheshrajamani (approved):
LGTM

src/main/java/io/stargate/sgv2/jsonapi/service/operation/filters/table/codecs/CollectionCodecs.java
tatu-at-datastax: I think that since we probably need re-designed codec for that case, adding key codec handling that we don't yet need is not probably worth doing, since it would probably need to be changed anyway. So I think I will defer that work until we add support for non-TEXT/ASCII keys.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/filters/table/codecs/JSONCodecRegistry.java
tatu-at-datastax: Yes.

src/test/java/io/stargate/sgv2/jsonapi/service/operation/filters/table/codecs/JSONCodecRegistryTest.java
tatu-at-datastax: I am not sure we really need exhaustive testing, as long as some are spot checked on per main CQL type basis (rest should follow).

But I will add a failing case for CQL->JSON for unsupported key type.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/filters/table/codecs/CollectionCodecs.java
amorton: agree not needed yet, but this is a kind of ugly looking unchecked cast based on that assumption. can we add a comment that we are assuming only string keys because that is all we support now

src/main/java/io/stargate/sgv2/jsonapi/service/operation/filters/table/codecs/CollectionCodecs.java
```diff
@@ -103,14 +132,38 @@ static Set<Object> toCQLSet(
...
+      String key = entry.getKey();
+      Object element = entry.getValue().value();
+      if (element == null) {
+        result.put(key, null);
+        continue;
+      }
+      // In most cases same codec can be used for all elements, but we still need to check
+      // since same CQL value type can have multiple codecs based on JSON value type
+      // (like multiple Number representations; or simple Text vs EJSON-wrapper)
+      if (elementCodec == null || !elementCodec.handlesJavaValue(element)) {
```
amorton, line 158: Is there any value in caching previously seen elementCodec ? Or do we think this is rare enough (needing a diff codec) it's not worth it, or the cost is low  ?

src/main/java/io/stargate/sgv2/jsonapi/service/operation/filters/table/codecs/CollectionCodecs.java
```diff
@@ -121,14 +174,29 @@ private static JSONCodec<Object, Object> findElementCodec(
     String msg =
         String.format(
             "no codec matching (list/set) declared element type `%s`, actual value type `%s`",
-            elementType, element.getClass());
+            elementType, element.getClass().getName());
+    throw new ToCQLCodecException(element, elementType, msg);
+  }
+
+  private static JSONCodec<Object, Object> findMapValueCodec(
```
amorton, line 181: checking ToCQLCodecException is the correct exception, do we still hae MissingCodecException ?

src/main/java/io/stargate/sgv2/jsonapi/service/operation/filters/table/codecs/CollectionCodecs.java
```diff
@@ -121,14 +174,29 @@ private static JSONCodec<Object, Object> findElementCodec(
...
+      List<JSONCodec<?, ?>> valueCodecs, DataType elementType, Object element)
+      throws ToCQLCodecException {
+    for (JSONCodec<?, ?> codec : valueCodecs) {
+      if (codec.handlesJavaValue(element)) {
+        return (JSONCodec<Object, Object>) codec;
+      }
+    }
+    String msg =
+        String.format(
+            "no codec matching map declared value type `%s`, actual value type `%s`",
```
amorton, line 191: nit: useful to include the list of codecs that was checked ?

src/main/java/io/stargate/sgv2/jsonapi/service/operation/filters/table/codecs/CollectionCodecs.java
```diff
@@ -143,4 +211,22 @@ static JsonNode toJsonNode(
     }
     return result;
   }
+
+  /** Method that will convert from driver-provided CQL Map type into JSON output. */
+  static JsonNode toJsonNodeFromMap(
+      JSONCodec<?, ?> elementCodec0, ObjectMapper objectMapper, Object mapValue)
```
amorton, line 217: just checking: we support multiple element codecs going toCQL and then only need one going toJson

src/main/java/io/stargate/sgv2/jsonapi/service/operation/filters/table/codecs/JSONCodecRegistry.java
```diff
@@ -110,14 +111,31 @@ public <JavaT, CqlT> JSONCodec<JavaT, CqlT> codecToCQL(
...
           }
           return (JSONCodec<JavaT, CqlT>)
               CollectionCodecs.buildToCQLSetCodec(valueCodecCandidates, st.getElementType());
         }
         // fall through
+      } else if (columnType instanceof MapType mt) {
+        List<JSONCodec<?, ?>> valueCodecCandidates = codecsByCQLType.get(mt.getValueType());
+        if (valueCodecCandidates != null) {
+          // Must check key type: only text/ascii supported
+          if (!isSupportedMapKeyType(mt.getKeyType())) {
```
amorton, line 126: we have this check for supported key types in a few places. Do you also have a check for supported value types ? 
Do you have the same for the valueTypes for lists , sets, and vectors ? 

if so can you put them public in places that make sense and I can call them from higher up when working out if we support the schema for a table 

e.g. I have this in a PR for working out if we support the map schema for a table 

```java
    public static boolean isCqlTypeSupported(MapType cqlMapType) {
      Objects.requireNonNull(cqlMapType, "cqlMapType must not be null");

      // cannot be frozen
      if (cqlMapType.isFrozen()) {
        return false;
      }
      // keys must be text or ascii, because keys in JSON are string
      if (!(cqlMapType.getKeyType() == DataTypes.TEXT
          || cqlMapType.getKeyType() == DataTypes.ASCII)) {
        return false;
      }
      // must be a primitive type value
      return cqlMapType.getValueType() instanceof PrimitiveType;
    }

```

src/main/java/io/stargate/sgv2/jsonapi/service/operation/filters/table/codecs/JSONCodecRegistry.java
amorton: see comment above about all the checks for key and value types

amorton (approved):
Nothing blocking, comments and code tidy. 

one suggestion on making the valdiation of Cql Collection types public from the Codecs  so we can use it higher up

src/main/java/io/stargate/sgv2/jsonapi/service/operation/filters/table/codecs/CollectionCodecs.java
tatu-at-datastax: Yes, I will add a comment.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/filters/table/codecs/CollectionCodecs.java
tatu-at-datastax: Added checking that we indeed get `String` key, throw exception if not.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/filters/table/codecs/CollectionCodecs.java
tatu-at-datastax: We probably should look into bit better caching once codecs are all there; for now keeping track of last `elementCodec` here should work (we will have more than one codecs for Number types, f.ex, since we have 3 from-JSON input number types to convert; similarly for Vectors we have 2 input types).

src/main/java/io/stargate/sgv2/jsonapi/service/operation/filters/table/codecs/CollectionCodecs.java
tatu-at-datastax: @amorton We do have `MissingCodecException`, but it seems to require `TableMetadata` and `ColumnMetadata` which is no longer available at this point (would need to be passed along). If necessary, could change.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/filters/table/codecs/CollectionCodecs.java
tatu-at-datastax: That could be: one problem is that of how to represent `JSONCodec`s as text -- will try something.

tatu-at-datastax: Ok cool; was able to get useful List easily -- great suggestion!

src/main/java/io/stargate/sgv2/jsonapi/service/operation/filters/table/codecs/CollectionCodecs.java
tatu-at-datastax: Exactly. It sort of suggests possible benefits of splitting `JSONCodec` into 2 parts (since one is many-to-many, another one-to-many). But for now there's some duplication and to-JSON case can basically pick any of matches.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/filters/table/codecs/JSONCodecRegistry.java
tatu-at-datastax: Ok: I'll do a follow-up ticket for this, need to think it through. One concern I have with "isXxx()" methods like this is that they need to find most of the info, then throw it away -- so, f.ex when checking if Map type supported, we need to lookup value codec (since that is how we know if type is supported), throw it away. And then in case of constructing Map codec we need to again lookup value codec.

tatu-at-datastax: Oh. One bigger concern: supported how? To read, write, create column?

src/main/java/io/stargate/sgv2/jsonapi/exception/catchable/CheckedApiException.java
```diff
@@ -24,4 +31,27 @@ public CheckedApiException(String message, Throwable cause) {
   public CheckedApiException(Throwable cause) {
     super(cause);
   }
+
+  // Add a place to slightly massage value; can be further improved
```
tatu-at-datastax, line 35: Moved these down from sub-class as they need to be shared b/w ToCQLCodecException and ToJSONCodecException

Issue #1509, opened 2024-10-07
C* Java Driver requirements tracker 


Yuqi-Du:
For Data API table feature, we decide to stick on C* Java Driver QueryBuilder, instead of using our own.
Here are 4 C* tickets that Driver is working on.

- Vector stuff [JAVA-3118 Add support for vector data type in Schema Builder, QueryBuilder](https://datastax-oss.atlassian.net/browse/JAVA-3118?atlOrigin=eyJpIjoiOGU0ZGFhODkxYzQzNDUwODg2MDkzNDg1ZmRhOWNhZjEiLCJwIjoiamlyYS1zbGFjay1pbnQifQ)
- AND/OR logical operator https://issues.apache.org/jira/browse/CASSANDRA-19931
- NOT operator https://issues.apache.org/jira/browse/CASSANDRA-19930
- access extensions in TableMetadata  https://issues.apache.org/jira/browse/CASSANDRA-19932

For OR feature in OSS driver distribution. Driver team has reached out to engineers maintaining C* OSS / CNDB and there seems to be no agreement in shipping this syntax to OSS version any time soon. Even non-vector databases do not have it currently. Driver also is not doing like a custom branch containing Astra-specific functionality or anything. So this ticket is no-go currently. Lukasz points us to hack that help us generate complex OR and ().

tatu-at-datastax:
Now that Java driver 4.19.0 is available:

https://mvnrepository.com/artifact/org.apache.cassandra/java-driver-core

we should probably revisit this issue: can we remove some of work-arounds, if not the whole thing?

amorton:
Keep open to review.

PR #1607, opened 2024-10-24
check in driver queryBuilder private build

Yuqi-Du:
All details in https://github.com/stargate/data-api/issues/1509

**Checklist**
- [x] Changes manually tested


pom.xml
```diff
@@ -192,11 +192,19 @@
       <artifactId>java-driver-metrics-micrometer</artifactId>
       <version>${driver.version}</version>
     </dependency>
+    <!--Note, using Java Driver QueryBuilder from a preview build. See https://github.com/stargate/data-api/issues/1509 -->
     <dependency>
       <groupId>com.datastax.oss</groupId>
       <artifactId>java-driver-query-builder</artifactId>
-      <version>${driver.version}</version>
-    </dependency>
+      <version>4.19.0-preview1</version>
```
tatu-at-datastax, line 199: I think base version like 4.19.0 makes sense to use as the base of version id.
I know the version we use here does not really matter a lot (since we specify exact filename), but it would be nice to include more information for our own benefit, like date it was built. Plus that it is for use by data-api. So something like this (it's bit long but shows the idea)

     4.19.0.data-api-2024-10-24

?

tatu-at-datastax: Also: probably more important, should include information on where the jar comes from. That is, which sources are used to build (link to repo, if public, branch).

tatu-at-datastax (approved):
LGTM: added suggestions for improvements as usual.

pom.xml
Yuqi-Du: thanks, Tatu, fixed

PR #1615, opened 2024-10-29
Test Java Driver QueryBuilder private build, 4.19.0-preview1

Yuqi-Du:
The [previous PR](https://github.com/stargate/data-api/pull/1607) for checking in the private java Driver 4.19.0-preview1 uses 'systemPath' to reference the jar. That is fine for Data API, however, it causes install problem to EGW which uses Data API as a jar. And it complains.
`[ERROR] 'dependencies.dependency.systemPath' for com.datastax.oss:java-driver-query-builder:jar must specify an absolute path but is ${project.basedir}/lib/java-driver-query-builder-4.19.0-preview1.jar @`

So, we still need to check in the Driver jar into Data API repo, but can not use <systemPath>. So I changed in Data API pom to search Driver queryBuilder in local repository. In this way, we can have a walk around.

Also, tested in EGW, it can be successfully built with Data API v1.0.19-SNAPSHOT now. 

**Checklist**
- [x] Changes manually tested


tatu-at-datastax:
I think this is failing due to missing `.sha1` files for checksums. So, basically if there is:

    jackson-core-2.18.0.jar

there needs to be matching

    jackson-core-2.18.0.jar.sha1 

(and same for .pom files)

pom.xml
```diff
@@ -79,6 +79,14 @@
       </dependency>
     </dependencies>
   </dependencyManagement>
+  <!-- Add local repository for private java driver queryBuilder build to avoid using systemPath in dependency-->
```
amorton, line 82: nit: can we include a link to an issue in our repo or the java driver that explains what we are waiting on, and how we can checked if the work has been done. We need the issue to stay open until we revert this change.

amorton (approved):
approved - but please add comment with link to tracking issue , see comment

pom.xml
Yuqi-Du: added tacker issue

PR #1625, opened 2024-10-30
Added flag to mark if a model is deprecated

kathirsvn:
**What this PR does**:
Added flag to mark if a model is deprecated

**Checklist**
- [X] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/service/embedding/configuration/EmbeddingProvidersConfig.java
```diff
@@ -121,6 +121,11 @@ interface ModelConfig {
 
       @JsonProperty
       Map<String, String> properties();
+
+      /** Flag to mark if support for the model is deprecated. */
+      @Nullable
+      @JsonProperty
+      Optional<Boolean> deprecated();
```
tatu-at-datastax, line 128: Do we need to use `Optional` here? I know it's "optional" in protoc but should be fine to leave this as regular `Boolean`, as it defaults to `false` (which I think is proper default for our usage).

tatu-at-datastax (approved):
LGTM: have one question on need for `Optional` but otherwise fine.

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/configuration/EmbeddingProvidersConfig.java
kathirsvn: Makes sense. Changed it to regular `Boolean` in this commit https://github.com/stargate/data-api/pull/1625/commits/62a48fc3348211dc56f545f749d2e9b38d9f2707

kathirsvn: Reverted this change, since it expects the existing configs to have this property set up
```
java.util.NoSuchElementException: SRCFG00014: The config property stargate.jsonapi.embedding.providers.cohere.models[0].deprecated is required but it could not be found in any config source
```

tatu-at-datastax: Ah. I wonder what component fails (Quarkus build?) -- but if it is needed then it is needed.

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/configuration/EmbeddingProvidersConfigImpl.java
```diff
@@ -40,7 +42,13 @@ public ModelConfigImpl(
                 ? Optional.of(grpcModelConfig.getVectorDimension())
                 : Optional.empty(),
             modelParameterList,
-            grpcModelConfig.getPropertiesMap());
+            grpcModelConfig.getPropertiesMap(),
+            grpcModelConfig.hasDeprecated()
+                ? Optional.of(grpcModelConfig.getDeprecated())
+                : Optional.empty(),
+            grpcModelConfig.hasServiceUrl()
+                ? Optional.of(grpcModelConfig.getServiceUrl())
```
tatu-at-datastax, line 50: In this case since `String` is Object, I think you can just use

   Optional.ofNullable(grpcModelConfig.getServiceUrl());

tatu-at-datastax (approved):
LGTM; just one suggestion.

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/configuration/EmbeddingProvidersConfigImpl.java
kathirsvn: - `grpcModelConfig.getServiceUrl()` is assumed to be a non-null value (though I don't see any `@NonNull` annotation or anything like that on that method, it assumes we check if the serviceUrl is present by calling `grpcModelConfig.hasServiceUrl()` before calling the `get` method).
- So, when I change the code to `Optional.ofNullable(grpcModelConfig.getServiceUrl());` the IDE suggests we change this it to `Optional.of(grpcModelConfig.getServiceUrl())` since`grpcModelConfig.getServiceUrl()` is a non-null value.

image

So, leaving this the way it is

PR #1627, opened 2024-10-31
Fix #1586: allow EJSON/$date wrapper use for CQL timestamp, for Collection compatibility

### LINKED ISSUE #1586: Allow use of  `$date` (EJSON wrapper) for API Table `timestamp` values (similar to Collections)
Currently (v1.0.18) API Table columns of type `timestamp` accept and produce JSON String values (using ISO-8601 encoding). This is different from Collections where EJSON-wrapping with numeric timestamp like:

```
{ "$date" : 123005946 }
```

is accepted and produced.

Questions: should we use or allow EJSON-wrapped values for API Tables as well? There are couple of questions to decide:

1. Should we accept both notations ("plain" String and EJsonWrapper) or just one -- and if so, which one?
2. If both are accepted, which one should be produced (returned with "findXxx" commands (if only one accepted that'd be one produced too)
3. If EJSON-wrapped value accepted, should it accept numeric Timestamp (like Collections), or ISO-8601 String (like current API Table).

These decisions should be based on what makes most sense from client perspective: since API Tables not yet externally released we can still change the notation/representation to use.

**EDIT**:

Answers: we will work to maximize interoperability with Collections.

1. Should accept BOTH, for interoperability with Collections (so yes accept ALSO EJSON-wrapped values)
2. Return "plain" ISO-8601 String representation (same as now, no change)
3. Accept Timestamp (seconds since Epoch, Java `long`), same as with Collections, for interoperability.

tatu-at-datastax:
**What this PR does**:

Alllows "$date" EJSON-wrapper notation as input for CQL `timestamp` values (in addition to "plain" String.

Fixes #1586

**Checklist**
- [ ] Changes manually tested


src/test/java/io/stargate/sgv2/jsonapi/service/operation/filters/table/codecs/JSONCodecRegistryTest.java
```diff
@@ -264,7 +264,11 @@ private static Stream<Arguments> validCodecToCQLTestCasesDatetime() {
         Arguments.of(
             DataTypes.TIMESTAMP,
             TEST_DATA.TIMESTAMP_VALID_STR,
-            Instant.parse(TEST_DATA.TIMESTAMP_VALID_STR)));
+            Instant.parse(TEST_DATA.TIMESTAMP_VALID_STR)),
```
tatu-at-datastax, line 267: Only "to CQL" test (from JSON into CQL value for driver) because reverse (from CQL to JSON) will always use "plain" notation, not EJSON wrapper.

vkarpov15:
LGTM :+1:

Yuqi-Du (approved):
LGTM

PR #1633, opened 2024-11-01
Table read sparse data

Yuqi-Du:
By default, table read will not return null column values,.

This nullOption is originally included in this PR, @amorton wants to think on naming stuff, so separate out as an issue.
[#1636]

**Checklist**
- [x] Changes manually tested


amorton (approved):
discused, will cause a merge conflict with the in memory sorting but will fix that in that PR

Issue #1611, opened 2024-10-27
(nit) Typo in the warning message for allow-filtering


hemidactylus:
It currently (1.0.18) starts with

> The Data API returned a warning: Zero filters were provided in the filer for this query. [...]

But: `filer` => `filter`

maheshrajamani:
This issue is resolved.

Issue #1602, opened 2024-10-24
TODO: WhereClauseAnalyzer should error for filtering on a complex type 


amorton:
And generally improve the code for handling mapping to the API types. 

See changes cominng into warnMissingIndexOnScalar() as part of #1590

Yuqi-Du:
fixed https://github.com/stargate/data-api/pull/1637

PR #1637, opened 2024-11-01
error typo, filter exception on complex type

Yuqi-Du:
Fixes:
https://github.com/stargate/data-api/issues/1611 (nit) Typo in the warning message for allow-filtering
https://github.com/stargate/data-api/issues/1602 WhereClauseAnalyzer should error for filtering on a complex type

**Checklist**
- [x] Changes manually tested


src/main/resources/errors.yaml
```diff
@@ -350,7 +361,7 @@ request-errors:
     code: ZERO_FILTER_OPERATIONS
     title: Zero operations provided in query filter
     body: |-
-      Zero filters were provided in the filer for this query. 
+      Zero filters were provided in the filter for this query.
```
tatu-at-datastax, line 364: Title says "Zero operations provided" so this should probably also say "Zero operations" -- or alternatively `title` should be changed to refer to "filters" instead of "operations".

Also: I know `code` probably shouldn't be changed but in general "No filters provided" (or "No operations") seems better than "Zero filters provided", so maybe change that part of message too.

amorton: Created https://github.com/stargate/data-api/issues/1662 for generally fixing error messages , will work with docs people on this

amorton (approved):
reviewed together, ready

Issue #1622, opened 2024-10-30
Return an error if a sort clause is used for delete and update with a table


amorton:
The Command objects for these are Sortable, but we cannot do sorting when used with a table.

amorton:
the work in #1620  adds a SortException it should be a code from that.

PR #1638, opened 2024-11-01
Can not sort table update delete

Yuqi-Du:
Fix for
https://github.com/stargate/data-api/issues/1622 Return an error is a sort clause is used for delete and update with a table

**Checklist**
- [x] Changes manually tested


vkarpov15:
LGTM. FWIW, MongoDB doesn't support `sort` with `updateOne()` or `deleteOne()`, so this behavior has precedent. However, `findOneAndUpdate()` and `findOneAndDelete()` do support sort, so that may make it more difficult if/when we look at `findOneAndUpdate()` for tables.

amorton (approved):
discussed, tweak one error msg all good

PR #1645, opened 2024-11-04
Fix #1592: remove duplication of Command Names

### LINKED ISSUE #1592: Code Tidy: @JsonTypeName on API Command objects duplicates command names
noticed when doing a review we have code like 

```
@JsonTypeName("dropIndex")
public record DropIndexCommand(
```

And we now have the CommandName enum, so we could do `CommandName.DROP_INDEX.getApiName()` 

this ticket is to update the `@JsonTypeName` for all commands: collections, keyspaces, tables etc .

tatu-at-datastax:
**What this PR does**:

Removes duplication of Command name Strings (like "alterTable")

Fixes #1592

**Checklist**
- [x] Changes manually tested


tatu-at-datastax:
Initial Proof-of-Concept to show how this could be done, different from the original plan, due to limitations Java Annotation type imposes.

tatu-at-datastax:
@amorton @maheshrajamani Was hoping to give feedback on whether this is worth pursuing or not? Leads to minor duplication (due to limitations of Java annotations) but can avoid more.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/AlterTableCommand.java
```diff
@@ -7,10 +7,9 @@
 import org.eclipse.microprofile.openapi.annotations.media.Schema;
 
 @Schema(description = "Command that alters the column definition in a table.")
-@JsonTypeName("alterTable")
+@JsonTypeName(CommandName.Names.ALTER_TABLE_NAME)
```
maheshrajamani, line 10: After the last merge I think we can set this as `CommandName.ALTER_TABLE.getApiName()`

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/CommandName.java
```diff
@@ -96,4 +96,8 @@ public static List<CommandName> filterByTarget(CommandTarget target) {
         .filter(commandName -> commandName.getTargets().contains(target))
         .collect(Collectors.toList());
   }
+
+  public interface Names {
```
maheshrajamani, line 100: Don't think we need this interface we can just use the apiName property of the enum.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/AlterTableCommand.java
tatu-at-datastax: Unfortunately no -- annotation values must be actual constants, like String constants; cannot use `Constant.method()` :-(
(I tried that, won't compile)

maheshrajamani (approved):

PR #1646, opened 2024-11-04
Fix #1589: extract CommandStatus name constants

### LINKED ISSUE #1589: Code Tidy: make JsonProperty for CommandStatus accessible in code
In this 
```
public enum CommandStatus {
  /** The element has the count of document */
  @JsonProperty("count")
  COUNTED_DOCUMENT,
  /** The element has the count of deleted documents */
  @JsonProperty("deletedCount")
  DELETED_COUNT,
```

Not sure how to get the value of the json property from the status, so we end up with ""sortVector" or "ok" in lots of places in tests etc. 

need a way to get to the JSONProperty of the enum in code, so tests can reference the enum and not hard code the field name. 

e.g. add the json field name as a property to the enum and have a serialiser

tatu-at-datastax:
**What this PR does**:

This is a proof-of-concept to resolve #1589, to extract out name used for `CommandStatus` enum when used as Object key in JSON: it would allow tests to reference String constant instead of hard-coding

Fixes #1589

**Checklist**
- [x] Changes manually tested


tatu-at-datastax:
Still not 100% sure we want this, but change is complete within scope (although not replacing all possible usage of name constants.

tatu-at-datastax:
@amorton @maheshrajamani Similar to PR #1645 wanted to get an opinion of if this makes sense.

maheshrajamani (approved):
LGTM

PR #1648, opened 2024-11-04
Fix #1647: support negative CqlDuration/duration values

### LINKED ISSUE #1647: Support use of negative CQL `duration` values
Apparently (see f.ex https://docs.datastax.com/en/drivers/java/4.4/com/datastax/oss/driver/api/core/data/CqlDuration.html) CQL `CqlDuration` values can be negative. This is not something ISO-8601 supports it seems (https://en.wikipedia.org/wiki/ISO_8601#Durations).

So, need to investigate how handling works, if users insert negative duration values via CQL (or Data API using C* duration syntax), what should we do on output path?

1. Throw an exception for invalid value (Codec failing)?
2. Product special value to indicate issue?

or something else?

tatu-at-datastax:
**What this PR does**:

Adds support (and tests) for accepting negative CQL `duration` values, serialization.
This because `CqlDuration` (and C* `duration` columns) allow such values... unlike ISO-8601 duration. But we can use minus-prefixed presentation since CQL side handles things ok, even if such values won't be good for ISO-8601 compliant systems.

Fixes #1647

**Checklist**
- [ ] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/service/operation/filters/table/codecs/CqlDurationConverter.java
```diff
@@ -22,7 +22,21 @@ public static String toISO8601Duration(CqlDuration value) {
     int days = value.getDays();
     long nanoSeconds = value.getNanoseconds();
 
-    final StringBuilder sb = new StringBuilder("P");
+    // Negative value? To detect see if any value negative (all must be 0 or negative
+    // if one is -- but `CqlDuration` ensures that invariant). Bit silly there's no
+    // `CqlDuration.isNegative()` (or `CqlDuration.negate()` to create opposite value)
```
Yuqi-Du, line 27: Yep, they should have these 'helper' method.

Yuqi-Du (approved):

PR #1650, opened 2024-11-05
Vectorize for insert and sort on tables 

### LINKED ISSUE #1643: Support Vectorize for Tables - Insert and Read (sort)
Initial design will allow passing a string as the value for a vector column

```
{
"my_vector" : "I like cheese", 
}
```

If the column is a vector column && it has vectorize defined we can vectorize it. 

We also need to support this for sorting, so if the command has 

```
"sort" : {"my_vector" : "i like cheese"},
```

we should also vectorise that.

amorton:
Handles inserting and sorting with vectorize

no unit or IT tests
has placeholder errors , see TODO 's

**What this PR does**:

Fixes #1643 

**Checklist**
- [ ] Changes manually tested


amorton (commented):
added 2 comments

Yuqi-Du (approved):

PR #1661, opened 2024-11-06
[WIP] Binary vector handling for sort clause

maheshrajamani:
**What this PR does**:
Binary vector handling for sort clause

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/filter/EJSONWrapper.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/filter/EJSONWrapper.java
@@ -4,6 +4,8 @@ import com.fasterxml.jackson.annotation.JsonValue;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.node.JsonNodeFactory;
 import com.fasterxml.jackson.databind.node.ObjectNode;
+import io.stargate.sgv2.jsonapi.util.Base64Util;
+import io.stargate.sgv2.jsonapi.util.CqlVectorUtil;
 import java.nio.ByteBuffer;
 import java.util.Map;
 import java.util.Objects;
@@ -117,5 +119,31 @@ public class EJSONWrapper {
     byte[] bytes = new byte[buffer.remaining()];
     buffer.get(bytes);
     return bytes;
+  }
+
+  public float[] getVectorValueForBinary() {
+    if (type != EJSONType.BINARY) {
+      throw new RuntimeException("Vector value can only be extracted from binary EJSON wrapper");
+    }
+    return toFloatArray(value.textValue());
+  }
+
+  // Utility method for converting binary wrapped value to float array
+  public static float[] toFloatArray(String base64Vector) {
+    byte[] binaryPayload;
+    try {
+      binaryPayload = Base64Util.decodeFromMimeBase64(base64Vector);
+    } catch (IllegalArgumentException e) {
+      throw new RuntimeException(
+          "Invalid content in EJSON $binary wrapper: not valid Base64-encoded String, problem: %s"
+              .formatted(e.getMessage()));
+    }
+
+    try {
+      return CqlVectorUtil.bytesToFloats(binaryPayload);
+    } catch (IllegalArgumentException e) {
+      throw new RuntimeException(
+          "Invalid content in EJSON $binary wrapper, problem: %s".formatted(e.getMessage()));
+    }
   }
 }
--- a/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/deserializers/SortClauseDeserializer.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/deserializers/SortClauseDeserializer.java
@@ -6,6 +6,7 @@ import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.deser.std.StdDeserializer;
 import com.fasterxml.jackson.databind.node.ArrayNode;
 import com.fasterxml.jackson.databind.node.ObjectNode;
+import io.stargate.sgv2.jsonapi.api.model.command.clause.filter.EJSONWrapper;
 import io.stargate.sgv2.jsonapi.api.model.command.clause.sort.SortClause;
 import io.stargate.sgv2.jsonapi.api.model.command.clause.sort.SortExpression;
 import io.stargate.sgv2.jsonapi.config.constants.DocumentConstants;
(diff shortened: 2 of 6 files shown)


src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/filter/EJSONWrapper.java
```diff
@@ -118,4 +121,46 @@ private static byte[] bytesFromByteBuffer(ByteBuffer buffer) {
     buffer.get(bytes);
     return bytes;
   }
+
+  // Returns the float array value from the binary EJSON wrapper
```
maheshrajamani, line 125: Refactored code for binary vector to here to avoid code duplication.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/filter/EJSONWrapper.java
```diff
@@ -118,4 +121,46 @@ private static byte[] bytesFromByteBuffer(ByteBuffer buffer) {
...
+  public float[] getVectorValueForBinary() {
+    if (type != EJSONType.BINARY) {
+      throw new RuntimeException("Vector value can only be extracted from binary EJSON wrapper");
```
amorton, line 128: nit: this would be a IllegalState or UnsupportedOperation

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/filter/EJSONWrapper.java
```diff
@@ -118,4 +121,46 @@ private static byte[] bytesFromByteBuffer(ByteBuffer buffer) {
...
+    }
+    return toFloatArray(value.textValue());
+  }
+
+  // Utility method for converting binary wrapped value to float array
+  public static float[] toFloatArray(String base64Vector) {
```
amorton, line 134: This looks similar / duplicate to JsonCodec.byteArrayFromEJSON() ok to merge but can you make sure to discuss with @tatu to see if there is a better way to centralise this. 

Also, it would be prob be best to use IllegalArgumentException rather than change to RuntimeException, good to give a better message with more context, and also pass the original error as the cause when building the new error

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/filter/EJSONWrapper.java
```diff
@@ -118,4 +121,46 @@ private static byte[] bytesFromByteBuffer(ByteBuffer buffer) {
...
+      throw new RuntimeException(
+          "Invalid content in EJSON $binary wrapper, problem: %s".formatted(e.getMessage()));
+    }
+  }
+
+  // Utility method for converting float array to binary vector value
+  public static String binaryFormatString(float[] vector) {
+    ByteBuffer byteBuffer = ByteBuffer.allocate(vector.length * 4); // 4 bytes per float
+
+    for (float val : vector) {
```
amorton, line 156: Fix before merge: the code in CqlVectorUtil.floatsToBytes can be used to turn the float[] into byte[] you can use that without needing to go to byte buffer.

OK to fix after merge:
This looks to duplicate some of the features in Base64Util.encodeAsMimeBase64, the last line is the same as in that function. 

 work with @tatu on how to refactor , this feels like it should be a function on Base64Util (it can be called from here but should be there)

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/sort/SortClause.java
```diff
@@ -93,6 +93,13 @@ public void validate(CollectionSchemaObject collection) {
         throw ErrorCodeV1.UNINDEXED_SORT_PATH.toApiException(
             "sort path '%s' is not indexed", sortExpression.path());
       }
+      if (!(DocumentConstants.Fields.VECTOR_EMBEDDING_FIELD.equals(sortExpression.path())
+              || DocumentConstants.Fields.VECTOR_EMBEDDING_TEXT_FIELD.equals(sortExpression.path()))
+          && sortExpression.vector() != null) {
+        throw ErrorCodeV1.INVALID_SORT_CLAUSE.toApiException(
+            "Vector sort for collection can be done only using `%s` field, field name used: `%s`  ",
```
amorton, line 100: Is this correct ?  looks like it is saying we can only sort on $vector for collections but we also allow sort on $vectorize

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/deserializers/SortClauseDeserializer.java
```diff
@@ -51,36 +52,69 @@ public SortClause deserialize(JsonParser parser, DeserializationContext ctxt) th
     while (fieldIter.hasNext()) {
       Map.Entry<String, JsonNode> inner = fieldIter.next();
       String path = inner.getKey().trim();
-
+      float[] arrayVals = null;
```
amorton, line 55: I've not fig too deeply into this , I know it was already a mess

src/main/java/io/stargate/sgv2/jsonapi/service/shredding/collections/WritableShreddedDocument.java
```diff
@@ -302,20 +301,10 @@ public void shredVector(JsonPath path, String base64Vector) {
...
-            "Invalid content in EJSON $binary wrapper: not valid Base64-encoded String, problem: %s",
-            e.getMessage());
-      }
-
-      try {
-        queryVectorValues = CqlVectorUtil.bytesToFloats(binaryPayload);
-      } catch (IllegalArgumentException e) {
-        throw ErrorCodeV1.SHRED_BAD_BINARY_VECTOR_VALUE.toApiException(
-            "Invalid content in EJSON $binary wrapper, problem: %s", e.getMessage());
+        queryVectorValues = EJSONWrapper.toFloatArray(base64Vector);
```
amorton, line 305: is this change necessary ? It is essentially calling the same functions, and it is may not be  generating less useful errors ? 

Is this used when using binary vectors for sorts ?  I would prefer this change happened working with tatu if it is not needed

amorton (changes requested):

amorton (changes requested):
some changes - looks like duplication in the way you are decoding float[] to byte[] and if we can avoid changing the WritableShreddedDocument we should or we need to get a review from @tatu-at-datastax  before that change

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/filter/EJSONWrapper.java
maheshrajamani: Make sense was just refactoring what was used in Collections. Changed to use method `CqlVectorUtil.floatsToBytes()`. @tatu-at-datastax Should we create a different util class for handling binary format, I don't think it should use `Cql`?

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/sort/SortClause.java
maheshrajamani: Want to say you can use only `$vector` field if sort is based on embedding vector as input. Yes, `$vectorize` does ANN sort but for end user it's based on text. Changed the error message, let me know.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/filter/EJSONWrapper.java
```diff
@@ -118,4 +119,18 @@ private static byte[] bytesFromByteBuffer(ByteBuffer buffer) {
     buffer.get(bytes);
     return bytes;
   }
+
+  // Returns the float array value from the binary EJSON wrapper
+  public float[] getVectorValueForBinary() {
```
tatu-at-datastax, line 124: This makes sense for immediate usage. In future may want to further refactor to expose just binary value, separate decoding of Vector value from binary to `float[]` (or `CqlVector`)

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/sort/SortClause.java
```diff
@@ -93,6 +93,13 @@ public void validate(CollectionSchemaObject collection) {
         throw ErrorCodeV1.UNINDEXED_SORT_PATH.toApiException(
             "sort path '%s' is not indexed", sortExpression.path());
       }
+      if (!(DocumentConstants.Fields.VECTOR_EMBEDDING_FIELD.equals(sortExpression.path())
```
tatu-at-datastax, line 96: Could use a comment explaining logic: I have hard time figuring out exact intent wrt needing `$vector` or `$vectorize` so others might have too.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/deserializers/SortClauseDeserializer.java
```diff
@@ -51,36 +52,69 @@ public SortClause deserialize(JsonParser parser, DeserializationContext ctxt) th
...
+      if (inner.getValue().isObject()) {
+        var ejsonWrapped = EJSONWrapper.maybeFrom((ObjectNode) inner.getValue());
+        if (ejsonWrapped == null) {
+          throw ErrorCodeV1.INVALID_SORT_CLAUSE_VALUE.toApiException(
+              "Only binary vector object values is supported for sorting. Path: %s, Value: %s.",
+              path, inner.getValue().toString());
+        }
+        try {
+          arrayVals = ejsonWrapped.getVectorValueForBinary();
+        } catch (RuntimeException e) {
```
tatu-at-datastax, line 65: Similar to the other case, would we only expect `IllegalArgumentException`? If so, catch just that.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/deserializers/SortClauseDeserializer.java
tatu-at-datastax, line 55: Maybe rename to `vectorFloats` -- `array` here might seem to relate JSON Number Array value (the other Vector representation).

tatu-at-datastax (approved):
LGTM

amorton (approved):

Issue #1688, opened 2024-11-12
Forbid the use of $in for Table updateOne and deleteOne command


Yuqi-Du:
By using $in in table filters, updateOne and deleteOne can affect multiple rows. 
Need to ban $in from these two commands.

PR #1693, opened 2024-11-12
Ban in filter for table updateOne and deleteOne command

Yuqi-Du:
This is to ban $in and $nin from table updateOne and deleteOne commands.
Fixes:
https://github.com/stargate/data-api/issues/1688

**Checklist**
- [x] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/service/operation/tables/WhereCQLClauseAnalyzer.java
```diff
@@ -237,6 +239,35 @@ private void checkFilteringOnComplexColumns(Map<CqlIdentifier, TableFilter> iden
...
+   * <p>For UpdateOne and DeleteOne table commands, the use of $in/$nin can affect multiple rows, so
+   * add a check rule to ban the usage.
+   */
+  private void checkNoInFilterUsage(Map<CqlIdentifier, TableFilter> identifierToFilter) {
+
+    var inFilterColumns =
+        identifierToFilter.entrySet().stream()
+            .filter(
+                entry -> {
+                  TableFilter tableFilter = entry.getValue();
```
amorton, line 254: nit: could this be one line ? we are not using the tableFilter var

src/main/resources/errors.yaml
```diff
@@ -240,6 +240,16 @@ request-errors:
       
       Note, DeleteMany without a filter truncates the table.
 
+  - scope: FILTER
+    code: INVALID_IN_FILTER_FOR_UPDATE_ONE_DELETE_ONE
+    title: In filter is not allowed for UpdateOne and DeleteOne commands
+    body: |-
+      An UpdateOne or DeleteOne command was issued with $in/$nin filter, filter columns are ${inFilterColumns}.
```
amorton, line 247: can we use below for the message pls, trying to keep the lines that include the variable about what went wrong kind of standard. 

The command used a `$in` or `$nin` filter to select the rows the modify, however these filters can select more than one row.

Update One and Delete One commands can only modify a single rows, and so cannot use `$in` or `$nin`.

The command used an invalid filter on the columns: ${inFilterColumns}.

Resend the request without using the `$in` or `$nin` filter.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/UpdateAttemptBuilder.java
```diff
@@ -24,11 +24,12 @@ public class UpdateAttemptBuilder<SchemaT extends TableSchemaObject> {
   private final SchemaT tableBasedSchema;
   private final WhereCQLClauseAnalyzer whereCQLClauseAnalyzer;
 
-  public UpdateAttemptBuilder(SchemaT tableBasedSchema) {
+  public UpdateAttemptBuilder(SchemaT tableBasedSchema, boolean updateOne) {
```
amorton, line 27: the updateOne param appears to never be used

src/test/java/io/stargate/sgv2/jsonapi/service/operation/tables/DeleteUpdateWhereAnalyzerTest.java
```diff
@@ -301,4 +304,27 @@ public void skip1and3of3ClusteringKey(
...
+      StatementType statementType, InTableFilter.Operator inTableFilterOperator) {
+
+    var fixture =
+        TEST_DATA
+            .whereAnalyzer()
+            .table2PK3Clustering1Index("in_filter_on_" + statementType.name(), statementType);
+    fixture
+        .expression()
+        .inOnOnePartitionKey(inTableFilterOperator)
+        .analyzeMaybeFilterError(FilterException.Code.INVALID_IN_FILTER_FOR_UPDATE_ONE_DELETE_ONE);
```
amorton, line 328: Can this also check the message ? see the comment for the error.yaml, we can check the message contains - replace the var

The command used an invalid filter on the columns: ${inFilterColumns}.

amorton (approved):
there are some changes for the error message, and I think the test can check the message. couple of other nit picks. 

Approving for you to fix those and merge.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/tables/WhereCQLClauseAnalyzer.java
Yuqi-Du: fixed

src/main/resources/errors.yaml
Yuqi-Du: fixed

src/main/java/io/stargate/sgv2/jsonapi/service/operation/UpdateAttemptBuilder.java
Yuqi-Du: yes, leave this way intentionally, cause currently we do not support updateMany for tables.
We enforce full primary key for updateOne to identify single row and update.

However, there is actually one way to do CQL 'updateMany'.
That is to use in operator. 
So we may consider adding updateMany later, more complicated surely, so leave this boolean here.

src/test/java/io/stargate/sgv2/jsonapi/service/operation/tables/DeleteUpdateWhereAnalyzerTest.java
Yuqi-Du: fixed

src/main/resources/errors.yaml
```diff
@@ -273,6 +273,18 @@ request-errors:
       
       Note, DeleteMany without a filter truncates the table.
 
+  - scope: FILTER
+    code: INVALID_IN_FILTER_FOR_UPDATE_ONE_DELETE_ONE
+    title: In filter is not allowed for UpdateOne and DeleteOne commands
+    body: |-
+      The command used a $in or $nin filter to select the rows the modify, however these filters can select more than one row.
+      
+      Update One and Delete One commands can only modify a single rows, and so cannot use $in or $nin.
```
vkarpov15, line 282: Minor typo: should be "modify a single row" not "modify a single rows"

src/main/resources/errors.yaml
```diff
@@ -273,6 +273,18 @@ request-errors:
       
       Note, DeleteMany without a filter truncates the table.
 
+  - scope: FILTER
+    code: INVALID_IN_FILTER_FOR_UPDATE_ONE_DELETE_ONE
+    title: In filter is not allowed for UpdateOne and DeleteOne commands
+    body: |-
+      The command used a $in or $nin filter to select the rows the modify, however these filters can select more than one row.
```
vkarpov15, line 280: I'm sure there's an internal reason to disallow calling `updateOne()` with `$in`/`$nin`, but this sentence doesn't seem to justify why you shouldn't use `$in` or `$nin` with `updateOne()`.

For example, a common pattern I've seen with `updateOne()` is `updateOne({ _id: 42, status: { $in: ['pending', 'active'] } }, update)`: update the document with `_id: 42` if it has status 'pending' or 'active'. This pattern can only update at most one row.

Similarly, `updateOne({ isDeleted: false })` can select many rows.

Not critical, but it would be nice to make this error message more accurately summarize the reasoning for this behavior.

PR #1706, opened 2024-11-13
fixes #1700 default limit for ANN find now 1k

### LINKED ISSUE #1700: HTTP 500 when vector search w/ no limit
When doing a `table.find({}, { sort: { vector: ... } })` on an indexed vector column w/out an explicit limit set:

DataAPIHttpError: HTTP error (500):

```
{
  "data":{"documents":[],"nextPageState":null},
  "status":{"sortVector":[1.0,1.0,1.0,1.0,1.0]},
  "errors":[{"message":"An unexpected server error occurred while processing the request. \n\nError Class: InvalidQueryException\nError Message: Use of ANN OF in an ORDER BY clause requires a LIMIT that is not greater than 1000. LIMIT was NO LIMIT","scope":"","errorCode":"UNEXPECTED_SERVER_ERROR","id":"f00d2d83-3c13-4f4f-bfe4-b0c5fb8bdeeb","family":"SERVER","title":"Unexpected server error"}]
}
```

amorton:
using the same configured AN find limit as collections, which is currently 1k

**What this PR does**:

Fixes #1700

**Checklist**
- [ ] Changes manually tested


Yuqi-Du (approved):

amorton:
> so if user specify a limit that is large than 1000 with ANN. We will do that and return back the DB error. Correct?

good call, added this so we can get the fix in quickly https://github.com/stargate/data-api/issues/1707

PR #1711, opened 2024-11-13
Fix for countDocument regression

### LINKED ISSUE #1708: (regression) countDocuments returns a count capped at 20 for collections
Hello, I detected a regression with "countDocuments" on main (at least for collection). If more than 20 documents, the API returns a count of 20. By bisection, I identified the first commit where this was introduced: 59d6b0cb Fixes #1677 - page state returned (#1689)

maheshrajamani:
**What this PR does**:
Count command fix

Fixes #1708 

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/operation/collections/CollectionReadOperation.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/operation/collections/CollectionReadOperation.java
@@ -1,5 +1,6 @@
 package io.stargate.sgv2.jsonapi.service.operation.collections;
 
+import com.datastax.oss.driver.api.core.AsyncPagingIterable;
 import com.datastax.oss.driver.api.core.cql.AsyncResultSet;
 import com.datastax.oss.driver.api.core.cql.Row;
 import com.datastax.oss.driver.api.core.cql.SimpleStatement;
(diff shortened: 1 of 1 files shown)


src/main/java/io/stargate/sgv2/jsonapi/service/operation/collections/CollectionReadOperation.java
```diff
@@ -389,21 +390,37 @@ default Uni<CountResponse> countDocumentsByKey(
...
+                  .onItem()
+                  .invoke(
+                      rs -> {
+                        counter.addAndGet(rs.remaining());
+                        stateRef.set(rs);
+                      });
+            })
+        // Documents read until pageState available, max records read is deleteLimit + 1 TODO
+        // COMMENTS
+        .whilst(AsyncPagingIterable::hasMorePages)
```
tatu-at-datastax, line 419: Quick question: where is the "max count" checked? Looks like this would read through all entries and not stop at specific count? (I am probably missing something)

amorton: this looks like the code we came up with for in memory sorting for tables

amorton:
moving to draft due to last minute situation, the simple fix is #1715

src/main/java/io/stargate/sgv2/jsonapi/service/operation/collections/CollectionReadOperation.java
tatu-at-datastax: Ah. Yes, since that's within `SimpleStatement` being passed.

PR #1739, opened 2024-11-19
Fix commands against table with unsupported datatype, add IT ability to directly execute cqlStatement

### LINKED ISSUE #1730: API commands against table with unsupported data type generate SERVER_UNHANDLED_ERROR
https://github.com/stargate/data-api/blob/main/src/main/java/io/stargate/sgv2/jsonapi/service/schema/tables/UnsupportedApiDataType.java#L11

E.G.
CountCommands against a table with frozen datatype, errors out.

```
{
    "errors": [
        {
            "message": "Server failed: root cause: (java.lang.UnsupportedOperationException) UnsupportedCqlApiDataType does not have a name",
            "id": "f03bb89c-1db5-407d-890d-e0c38290ef39",
            "title": "Server failed",
            "family": "SERVER",
            "scope": "DATABASE"
        }
    ]
}
```

Yuqi-Du:
This PR fixes #1730 first, with the fix, Data API commands can be executed against table with unsupported dataType without failing ungracefully.
E.G. Currently table does not support countCommands, so it should fail with `UNSUPPORTED_TABLE_COMMAND`

Also, this PR adds IT to cover the fix. 
Previously, we don't have the ability for Integration Tests to execute cql statement directly against C* backend. Now with this PR, we can do that, and then we can create cql tables with DataTypes that Data API does not support.
fixes: #1731, https://datastax.jira.com/browse/C2-3799

**Checklist**
- [x] Changes manually tested


src/test/java/io/stargate/sgv2/jsonapi/api/v1/AbstractKeyspaceIntegrationTestBase.java
```diff
@@ -50,6 +60,25 @@ public void createKeyspace() {
     createKeyspace(keyspaceName);
   }
 
+  @BeforeAll
+  public void createDriverSession() {
```
tatu-at-datastax, line 64: Do we want to always create session for every test? Maybe it could be created on demand instead?

Also: should this be closed after tests are run so as not to leave connections hanging?

src/test/java/io/stargate/sgv2/jsonapi/api/v1/AbstractKeyspaceIntegrationTestBase.java
```diff
@@ -274,4 +281,29 @@ protected float[] decodeBase64BinaryVectorToFloatArray(String base64Vector) {
       throw new RuntimeException("Not a valid binary vector", e);
     }
   }
+
+  protected boolean executeCqlStatement(SimpleStatement statement) {
```
tatu-at-datastax, line 285: Idea: maybe create/change to take `SimpleStatement ...` to allow multiple statements to be executed within same session? This would avoid overhead of multiple separate sessions if multiple statements are always needed to initialize something.

tatu-at-datastax (approved):
LGTM -- some suggestions

src/test/java/io/stargate/sgv2/jsonapi/api/v1/AbstractKeyspaceIntegrationTestBase.java
Yuqi-Du: thanks, fixed

src/test/java/io/stargate/sgv2/jsonapi/api/v1/ResponseAssertions.java
```diff
@@ -62,6 +62,11 @@ public class ResponseAssertions {
         "responseIsDDLSuccess", Presence.FORBIDDEN, Presence.REQUIRED, Presence.FORBIDDEN);
   }
 
+  public static TypeSafeMatcher<Map<String, ?>> responseIsCountSuccess() {
```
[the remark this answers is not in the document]
Yuqi-Du: resolved

Issue #1735, opened 2024-11-18
[Tables] Sorted `find` returns a nextPageState that cannot be subsequently used (because: no pagination with sort)


hemidactylus:
Running a `find` on a table with a regular (non-ANN) `sort` clause returns up to 20 results, sorted as expected, but also returns a non-null `nextPageState` token. Using this as the pageState of a second request, predictably, gives an error as follows:

```
Request invalid: Request invalid: field 'command.options.pageState' value 
"FindCommand[filterClause=FilterClause[logicalExpression=LogicalExpression{logicalRelation='AND', 
totalComparisonExpressionCount=2, logicalExpressions=[], comparisonExpressions=[ComparisonExpression{path='p_ascii', 
filterOperations=[ValueComparisonOperation[operator=EQ, operand=JsonLiteral{type=STRING, value(String)=A}]], 
dbFilters=null}, ComparisonExpression{path='p_bigint', filterOperations=[ValueComparisonOperation[operator=EQ, 
operand=JsonLiteral{type=NUMBER, value(BigDecimal)=100}]], dbFilters=null}]}], projectionDefinition=null, 
sortClause=SortClause[sortExpressions=[SortExpression[path=p_int, ascending=false, vector=null, vectorize=null]]], 
options=Options[limit=251, skip=null, pageState=AAgABAAAAOYBAYDn8H///+s=, includeSimilarity=false, 
includeSortVector=false]]" not valid. Problem: pageState is not supported with sort clause. (COMMAND_FIELD_INVALID)
```

In particular:

1.  this would mislead a CURL user who, seeing that token, would use it to get more stuff (we know in this case "pagination" is a  manual affair involving skip+limit, and "to be discouraged" somewhat because of resource waste, in-mem sorting, etc)
2. for clients this is a big nuisance since the `find` method for client does follow page after page on the user's behalf, so right now the clients error like seen above.

PR #1746, opened 2024-11-20
Fix sort clause and page state usage in collections and tables

### LINKED ISSUE #1736: [Tables] Passing empty `sort={}` and a pageState to a find results in "pageState is not supported with sort clause." error
When running e.g. the second page of a `find` command, thus passing a `pageState: <string>` to the options,
it is sufficient to provide an empty `sort: {}` to trigger the following error (the find being otherwise working correctly):

```python
    "errors": [
        {
            "message": "Request invalid: field 'command.options.pageState' value
 \"FindCommand[filterClause=FilterClause[logicalExpression=LogicalExpression
{logicalRelation='AND', totalComparisonExpressionCount=2, logicalExpressions=[], 
comparisonExpressions=[ComparisonExpression{path='p_ascii', filterOperations=
[ValueComparisonOperation[operator=EQ, operand=JsonLiteral{type=STRING, 
value(String)=A}]], dbFilters=null}, ComparisonExpression{path='p_bigint', 
filterOperations=[ValueComparisonOperation[operator=EQ, operand=JsonLiteral
{type=NUMBER, value(BigDecimal)=100}]], dbFilters=null}]}], projectionDefinition=null, 
sortClause=SortClause[sortExpressions=[]], options=Options[limit=null, skip=null, 
pageState=AAgABAAAABMBAPB////r8H///+s=, includeSimilarity=false, includeSortVector=false]]\" 
not valid. Problem: pageState is not supported with sort clause.",
            "errorCode": "COMMAND_FIELD_INVALID",
            "id": "f18f73b3-3fa5-4d3c-ad17-ccfe21affd9b",
            "family": "REQUEST",
            "title": "Request invalid",
            "scope": "EMPTY"
        }
    ]
```

Hazel-Datastax:
**What this PR does**:

**Collection fix**:
- Collection only supports in-memory sort, so it doesn't allow non-empty sort clause with non-empty page state

**Table fix**
- If it's CQL sort, it allows sort + pageState
- If it's in-memory, it doesn't allow non-empty sort clause with non-empty page state (same as the collection).

Fixes #1736 #1735

**Checklist**
- [x] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/api/model/command/validation/FindOptionsValidation.java
```diff
@@ -56,9 +56,12 @@ public boolean isValid(FindCommand value, ConstraintValidatorContext context) {
...
     }
 
-    if (options.pageState() != null && value.sortClause() != null) {
+    if (options.pageState() != null
+        && value.sortClause() != null
+        && !value.sortClause().isEmpty()) {
       context
-          .buildConstraintViolationWithTemplate("pageState is not supported with sort clause")
+          .buildConstraintViolationWithTemplate(
+              "pageState is not supported with non empty sort clause")
```
tatu-at-datastax, line 64: nit: "non empty" -> "non-empty"

src/test/java/io/stargate/sgv2/jsonapi/api/model/command/impl/FindCommandTest.java
```diff
@@ -183,7 +210,7 @@ public void invalidOptionsPageStateWithSort() throws Exception {
       assertThat(result)
           .isNotEmpty()
           .extracting(ConstraintViolation::getMessage)
-          .contains("pageState is not supported with sort clause");
+          .contains("pageState is not supported with non empty sort clause");
```
tatu-at-datastax, line 213: same here to match (-> "non-empty")

tatu-at-datastax (approved):
LGTM: just one typo to change

src/main/java/io/stargate/sgv2/jsonapi/exception/RequestException.java
```diff
@@ -41,6 +41,7 @@ public String scope() {
   public enum Code implements ErrorCode<RequestException> {
     UNSUPPORTED_TABLE_COMMAND,
     UNSUPPORTED_COLLECTION_COMMAND,
+    UNSUPPORTED_PAGINATION_WHEN_USING_IN_MEMORY_SORTING,
```
tatu-at-datastax, line 44: could shorten slightly as `UNSUPPORTED_PAGINATION_WITH_IN_MEMORY_SORTING`

src/main/resources/errors.yaml
```diff
@@ -157,6 +157,17 @@ request-errors:
       
       Resend using one of the supported commands. 
 
+  - scope:
+    code: UNSUPPORTED_PAGINATION_WHEN_USING_IN_MEMORY_SORTING
+    title: Pagination not supported when using in memory sorting
```
tatu-at-datastax, line 162: nit: "in memory" -> "in-memory"

src/main/resources/errors.yaml
```diff
@@ -157,6 +157,17 @@ request-errors:
...
+    body: |-
+      Pagination is not supported when the data is sorted in memory.
+
+      The table ${keyspace}.${table} has the partition sorting columns: ${partitionSorting}.
+      The command sorted on the columns: ${sortColumns}.
+
+      ${SNIPPET.INEFFICIENT_SORT}
```
tatu-at-datastax, line 169: I don't think we should include this since command actually failed; description says

"The command was executed using in memory sorting..."

which is technically not true.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/tables/FindPaginationTableIntegrationTest.java
```diff
@@ -1,23 +1,30 @@
...
 import io.quarkus.test.junit.QuarkusIntegrationTest;
-import io.stargate.sgv2.jsonapi.api.v1.util.scenarios.KeyValueTable100Scenario;
+import io.stargate.sgv2.jsonapi.api.v1.util.scenarios.PaginationTable100Scenario;
+import io.stargate.sgv2.jsonapi.api.v1.util.scenarios.ThreeClusteringKeysTableScenario;
+import io.stargate.sgv2.jsonapi.exception.RequestException;
 import io.stargate.sgv2.jsonapi.testresource.DseTestResource;
 import java.util.List;
 import java.util.Map;
 import org.junit.jupiter.api.*;
+import org.testcontainers.shaded.com.google.common.collect.ImmutableMap;
```
tatu-at-datastax, line 15: This is bad import (probably due to IDEA auto-complete), please fix to use Guava `ImmutableMap`, instead of private shaded copy by testcontainers.

tatu-at-datastax (commented):
Makes sense, needs couple of small changes.

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/FindCommandResolver.java
```diff
@@ -173,6 +174,14 @@ public Operation resolveCollectionCommand(
     }
 
     SortClause sortClause = command.sortClause();
+
+    // collection always uses in memory sorting, so we don't support page state with sort clause
+    // empty sort clause and empty page state are treated as no sort clause and no page state
+    if (sortClause != null && !sortClause.isEmpty() && pageState != null && !pageState.isEmpty()) {
```
amorton, line 180: +1 for isEmpty, any non zero length string is considered page state

src/main/java/io/stargate/sgv2/jsonapi/service/operation/tables/TableRowSorter.java
```diff
@@ -1,5 +1,7 @@
 package io.stargate.sgv2.jsonapi.service.operation.tables;
 
+import static io.stargate.sgv2.jsonapi.exception.ErrorFormatters.*;
```
amorton, line 3: nit: not needed

src/main/resources/errors.yaml
```diff
@@ -157,6 +157,15 @@ request-errors:
...
 
+  - scope:
+    code: UNSUPPORTED_PAGINATION_WITH_IN_MEMORY_SORTING
+    title: Pagination not supported when using in-memory sorting
+    body: |-
+      Pagination is not supported when the data is sorted in-memory.
+
+      The table ${keyspace}.${table} has the partition sorting columns: ${partitionSorting}.
+      The command sorted on the columns: ${sortColumns}.
+
```
amorton, line 168: add last paragraph "Resend the command without pagination. "

src/main/java/io/stargate/sgv2/jsonapi/exception/RequestException.java
```diff
@@ -41,6 +41,7 @@ public String scope() {
   public enum Code implements ErrorCode<RequestException> {
     UNSUPPORTED_TABLE_COMMAND,
     UNSUPPORTED_COLLECTION_COMMAND,
+    UNSUPPORTED_PAGINATION_WITH_IN_MEMORY_SORTING,
```
amorton, line 44: given this refers to sorting, lets move to the SORT scope

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/sort/TableMemorySortClauseResolver.java
```diff
@@ -70,6 +77,20 @@ public WithWarnings<RowSorter> resolve(
       throw new IllegalStateException("Cannot do table vector sorts in memory");
     }
 
+    // in memory sort does not support page state, error out if it is present
```
amorton, line 80: nit: tweak the comment to say because pagination currently only comes from the CQL driver and in memory sort reads all the data

src/test/java/io/stargate/sgv2/jsonapi/api/v1/FindIntegrationTest.java
```diff
@@ -150,6 +150,59 @@ public void wrongKeyspace() {
...
+          .headers(getHeaders())
+          .contentType(ContentType.JSON)
+          .body(
+              """
+                {
+                 "find": {
+                    "filter" : {"username" : "user1"},
+                    "sort" : {"username" : 1},
+                    "options" : {
+                      "pageState" : ""
```
amorton, line 194: lets add another test where `pageState: null`

src/test/java/io/stargate/sgv2/jsonapi/api/v1/util/scenarios/PaginationTable100Scenario.java
```diff
@@ -8,13 +8,15 @@
 import java.util.List;
 import java.util.Map;
 
-/** basic key-value table with 100 rows */
-public class KeyValueTable100Scenario extends TestDataScenario {
+/** basic table with 100 rows to test pagination */
+public class PaginationTable100Scenario extends TestDataScenario {
```
amorton, line 12: make this a new scenario and keep the keyvalue one , and rename this to PartitionedKeyValueTableScenario, and insert data for three partitions , and change the number of rows so the last API page of data (which is 20 rows) is not full

src/test/java/io/stargate/sgv2/jsonapi/api/v1/tables/FindPaginationTableIntegrationTest.java
```diff
@@ -31,11 +38,86 @@ public final void dropScenario() {
 
   @Test
   public void findWithPageState() {
+    // `find_all` command to get the next page state
```
amorton, line 41: also add a test that goes through all the pages in a partition, checks the rows have the clustering key values we expect, in the order we expect, and the last page is a  partial page

amorton (changes requested):

vkarpov15 (commented):
LGTM

tatu-at-datastax:
I think this looks good, but @amorton can have the final say (has check mark for requiring changes)

src/test/java/io/stargate/sgv2/jsonapi/api/v1/tables/FindPaginationTableIntegrationTest.java
```diff
@@ -30,12 +39,134 @@ public final void dropScenario() {
...
+              .body(
+                  "data.documents",
+                  jsonEquals(
+                      SCENARIO.getPageContent(0, 20, partitionedKeyValue))) // verify the content
+              .hasNextPageState()
+              .extractNextPageState();
+
+      Map<String, Object> options = ImmutableMap.of("pageState", pageState);
+
+      // The next page should have 10 rows and no page state
```
amorton, line 65: the test needs to span atleast three pages, so there is a first, middle, and last page.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/tables/FindPaginationTableIntegrationTest.java
```diff
@@ -30,12 +39,134 @@ public final void dropScenario() {
...
+              jsonEquals(
+                  SCENARIO.getPageContent(20, 20, partitionedKeyValue))) // verify the content
+          .doesNotHaveNextPageState();
+    }
+  }
+
+  @Test
+  @Order(2)
+  public void findWithEmptySortAndPageState() {
+    // This test checks the empty sort (equals no sort clause)
```
amorton, line 82: good to have comments like this, better to use javadoc standard for a function with /**

src/test/java/io/stargate/sgv2/jsonapi/api/v1/tables/FindPaginationTableIntegrationTest.java
```diff
@@ -30,12 +39,134 @@ public final void dropScenario() {
...
+        .wasSuccessful()
+        .doesNotHaveNextPageState();
+
+    // error if page state is passed with in memory sort
+    Map<String, Object> options = ImmutableMap.of("pageState", "AAgABAAAAOYBAYDn8H///+s=");
+    assertTableCommand(keyspaceName, TABLE_NAME)
+        .templated()
+        .find(FILTER_ID, List.of(), sort, options)
+        .hasSingleApiError(
+            SortException.Code.UNSUPPORTED_PAGINATION_WITH_IN_MEMORY_SORTING,
```
amorton, line 160: a better test would check that the message is specific to the request. 

We keep lines in the error message like this one: 
`The command sorted on the columns: ${sortColumns}.`

for that, so check the line has the name of the sort column(s)

amorton: prob also need a multi column sort

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/sort/TableMemorySortClauseResolver.java
```diff
@@ -70,6 +77,22 @@ public WithWarnings<RowSorter> resolve(
...
+    if (!cqlPageState.isEmpty()) {
+      var apiTableDef = commandContext.schemaObject().apiTableDef();
+
+      throw SortException.Code.UNSUPPORTED_PAGINATION_WITH_IN_MEMORY_SORTING.get(
+          errVars(
+              commandContext.schemaObject(),
+              map -> {
+                map.put(
+                    "partitionSorting", errFmtApiColumnDef(apiTableDef.clusteringKeys().values()));
+                map.put("sortColumns", errFmtCqlIdentifier(sortClause.sortColumnIdentifiers()));
```
amorton, line 92: We want the error message to be stable, so we can check it in tests. 

So we need to sort the list of `sortClause.sortColumnIdentifiers()` , use the comparator in the CqlIdentifierUtil class

amorton (approved):
approved but some more improvements pls

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/sort/TableMemorySortClauseResolver.java
amorton: code is correct, this is the sort order from the user we should keep that.

Issue #1729, opened 2024-11-15
`deleteOne` and `updateOne` w/ `$ne` can delete (/update) multiple documents


toptobes:
Example (`text` is partition key; `int` is clustering key):

```ts
await table.insertMany(Array.from({ length: 10 }, (_, i) => ({ text: key, int: i })));

await table.deleteOne({ text: key, int: { $ne: 5 } });

for (let i = 0; i < 10; i++) {
  if (i === 5) {
    assert.deepStrictEqual(await table.findOne({ text: key, int: i }, { projection: { text: 1, int: 1 } }), { text: key, int: i });
  } else {
    assert.deepStrictEqual(await table.findOne({ text: key, int: i }), null);
  }
}
```

amorton:
the fix is probably making sure that we check there is a filter on every column in the OK, which we do, and also check that the operations is $eq (probably not checking the operation)

amorton:
also for updateOne

amorton:
merged to main and cherry picked into https://github.com/stargate/data-api/tree/feature/dec-hotfix

PR #1748, opened 2024-11-21
only allow $eq for updateOne and deleteOne commands

Yuqi-Du:
Add a filterAnalyzer rule for checking non-$eq usage for table updateOne and deleteOne command

fixes: #1729 

**Checklist**
- [x] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/service/operation/tables/WhereCQLClauseAnalyzer.java
```diff
@@ -70,6 +70,8 @@ AnalysisStrategy getStrategy(WhereCQLClauseAnalyzer analyzer) {
                     analyzer::checkNoFilters,
                     analyzer::checkAllColumnsExist,
                     analyzer::checkNonPrimaryKeyFilters,
+                    analyzer::checkOnlyEqUsage,
```
tatu-at-datastax, line 73: Fine for now but in future we probably will want to pre-create these `List`s to reduce garbage generation.
(so no change proposed for this PR, just mention as future improvement)

src/main/java/io/stargate/sgv2/jsonapi/service/operation/tables/WhereCQLClauseAnalyzer.java
```diff
@@ -70,6 +70,8 @@ AnalysisStrategy getStrategy(WhereCQLClauseAnalyzer analyzer) {
...
+                    // TODO, checkOnlyEqUsage rule make checkNoInFilterUsage redundant.
```
tatu-at-datastax, line 74: Not sure I understand this comment; is it obsolete, or should it be rewritten?

src/main/java/io/stargate/sgv2/jsonapi/service/operation/tables/WhereCQLClauseAnalyzer.java
```diff
@@ -239,6 +241,39 @@ private void checkFilteringOnComplexColumns(Map<CqlIdentifier, TableFilter> iden
...
+        identifierToFilter.entrySet().stream()
+            .filter(
+                entry -> {
+                  TableFilter tableFilter = entry.getValue();
+                  if (!(tableFilter instanceof NativeTypeTableFilter<?> nativeTypeTableFilter)) {
+                    return true;
+                  }
+                  return nativeTypeTableFilter.operator != NativeTypeTableFilter.Operator.EQ;
+                })
+            .map((Map.Entry::getKey))
```
tatu-at-datastax, line 263: nit: extra parentheses (remove) :)

src/main/java/io/stargate/sgv2/jsonapi/service/operation/tables/WhereCQLClauseAnalyzer.java
```diff
@@ -239,6 +241,39 @@ private void checkFilteringOnComplexColumns(Map<CqlIdentifier, TableFilter> iden
...
+   * UpdateOne.
+   */
+  private void checkOnlyEqUsage(Map<CqlIdentifier, TableFilter> identifierToFilter) {
+
+    var nonEqFilterColumns =
+        identifierToFilter.entrySet().stream()
+            .filter(
+                entry -> {
+                  TableFilter tableFilter = entry.getValue();
+                  if (!(tableFilter instanceof NativeTypeTableFilter<?> nativeTypeTableFilter)) {
```
tatu-at-datastax, line 258: Why is this needed? And should it really return `true` (included as failing filter) instead of `false` (ignored, I think)

src/main/resources/errors.yaml
```diff
@@ -300,6 +300,16 @@ request-errors:
...
 
+  - scope: FILTER
+    code: UNSUPPORTED_FILTER_FOR_UPDATE_ONE_DELETE_ONE
+    title: updateOne and deleteOne commands do not support filter operation other than `$eq`
+    body: |-
+      The Filtering using the non `$eq` filter operations can select more than one row, and so cannot be used with the `updateOne` and `deleteOne` commands as they can only modify one row.
+      
+      The command used an invalid filter on the columns: ${unsupportedFilterColumns}.
+      
+      Resend the command only using `$eq` filter.
```
tatu-at-datastax, line 311: maybe: "filter" -> "filters" ? (since one can have more than one)
also maybe rephrase "command only using" -> "command that only has"

src/test/java/io/stargate/sgv2/jsonapi/fixtures/testdata/WhereAnalyzerTestData.java
```diff
@@ -175,6 +175,15 @@ public WhereAnalyzerFixture assertExceptionOnDurationColumns(CqlIdentifier... co
       return assertExceptionContains(warning);
     }
 
+    public WhereAnalyzerFixture assertExceptionOnNonEqFilerForUpdateOneAndDeleteOne(
+        CqlIdentifier... columns) {
+      var identifiers = Arrays.stream(columns).sorted(CQL_IDENTIFIER_COMPARATOR).toList();
+      var warning =
```
tatu-at-datastax, line 181: nit: this is error, not warning? Maybe name as such

src/main/resources/errors.yaml
amorton: Resend the command using only `$eq` filters.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/tables/WhereCQLClauseAnalyzer.java
amorton: agree, remove the $in check if not needed any more and remove the exception

src/main/java/io/stargate/sgv2/jsonapi/service/operation/tables/WhereCQLClauseAnalyzer.java
```diff
@@ -239,6 +241,39 @@ private void checkFilteringOnComplexColumns(Map<CqlIdentifier, TableFilter> iden
...
   }
 
+  /**
+   * Check if any non $eq filter is used.
+   *
+   * <p>e.g. Table UpdateOne and DeleteOne, DATA API does not know the exact modified row count in
+   * DB, so it needs to restrict on the filters. Make sure only $eq is used for DeleteOne and
+   * UpdateOne.
+   */
+  private void checkOnlyEqUsage(Map<CqlIdentifier, TableFilter> identifierToFilter) {
```
amorton, line 251: from discussions: we should not allow a sort() clause for updateOne and deleteOne for Tables because CQL does not allow sorting on these commands. Create a new ticket if we need to do the work.

amorton (changes requested):
some changes needed

src/main/java/io/stargate/sgv2/jsonapi/service/operation/tables/WhereCQLClauseAnalyzer.java
```diff
@@ -239,6 +241,39 @@ private void checkFilteringOnComplexColumns(Map<CqlIdentifier, TableFilter> iden
     }
   }
 
+  /**
+   * Check if any non $eq filter is used.
+   *
+   * <p>e.g. Table UpdateOne and DeleteOne, DATA API does not know the exact modified row count in
```
vkarpov15, line 247: Isn't it also possible for `$eq` to match multiple rows? For example, would `deleteOne({ isDeleted: false })` delete multiple documents if multiple documents have `isDeleted: false`? Potentially silly question, I just wanted to make sure. Otherwise this PR looks reasonable.

tatu-at-datastax: Yes, if we did not require filters to cover the whole primary key, but we do (for deletes and updates), I think. So only-eq-filters is a necessary but not sufficient constraints -- filtering on all primary key columns is also required.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/tables/WhereCQLClauseAnalyzer.java
Yuqi-Du: removed.

src/test/java/io/stargate/sgv2/jsonapi/fixtures/testdata/WhereAnalyzerTestData.java
Yuqi-Du: thanks, good catch.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/tables/WhereCQLClauseAnalyzer.java
Yuqi-Du: Thanks folks
Correct, for both deleteOne and updateOne, there is a rule checking (checkFullPrimaryKey) to enforce the full primary key is constraint in filter clause.
And yes, so `only-eq-filters is a necessary but not sufficient constraints`

src/main/java/io/stargate/sgv2/jsonapi/service/operation/tables/WhereCQLClauseAnalyzer.java
Yuqi-Du: only nativeTypeTableFilter has $eq, so if a tableFilter is not a nativeTypeTableFilter, then it is using operator other than $eq, so return true to keep that in the filter stream.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/tables/WhereCQLClauseAnalyzer.java
Yuqi-Du: confirmed, these sort() on deleteOne and updateOne have already been covered by

- UNSUPPORTED_SORT_FOR_TABLE_UPDATE_COMMAND
- UNSUPPORTED_SORT_FOR_TABLE_DELETE_COMMAND

amorton (approved):

Issue #1744, opened 2024-11-19
Ensure all API Codec errors are turned into nice API Exceptions 


amorton:
Split from #1573 got to all places we use the JSON Code Repository and make sure the errors are translated. This may be a good place to use the ExceptionHandler as a pattern for translating errors (it may need to change to handle throwable not runtime) 

Checklist of places the user can send values that should go through codecs: 

- [ ] inserting
- [ ] filter
- [ ] update

amorton:
removing from hotifx, we can check again, I think this is mostly fixed by #1568

Issue #1740, opened 2024-11-19
HTTP 500 on wrong filter value type


toptobes:
For example, when doing `filter: { text: 1 }`, the following error ensues:
```
DataAPIHttpError: HTTP error (500): {"errors":[{"message":"Server failed: root cause: (io.smallrye.mutiny.CompositeException) Multiple exceptions caught:\n\t[Exception 0] java.lang.IllegalStateException: retryContext must not be null\n\t[Exception 1] java.lang.RuntimeException: io.stargate.sgv2.jsonapi.exception.checked.ToCQLCodecException: Error trying to convert to targetCQLType `TEXT` from value.class `java.math.BigDecimal`, value 1. Root cause: no codec matching value type","errorCode":"SERVER_UNHANDLED_ERROR","id":"f4eddc09-8328-4195-a03a-d746edf26a20","title":"Server failed","family":"SERVER","scope":"DATABASE"}]}
```

or with `filter: { text: { $in: [0, '1'] } }`
```
DataAPIHttpError: HTTP error (500): {"errors":[{"message":"Server failed: root cause: (io.smallrye.mutiny.CompositeException) Multiple exceptions caught:\n\t[Exception 0] java.lang.IllegalStateException: retryContext must not be null\n\t[Exception 1] java.lang.RuntimeException: io.stargate.sgv2.jsonapi.exception.checked.ToCQLCodecException: Error trying to convert to targetCQLType `TEXT` from value.class `java.math.BigDecimal`, value 0. Root cause: no codec matching value type","errorCode":"SERVER_UNHANDLED_ERROR","id":"51b70c37-fe6b-4022-aea7-341e2ab35f5f","title":"Server failed","family":"SERVER","scope":"DATABASE"}]}
```

Yuqi-Du:
close this ticket. 
 https://github.com/stargate/data-api/pull/1753

PR #1753, opened 2024-11-22
Fix 500s happen in Table Filter, Update columnAssignments, when codec not found, ToCQLCodecException

### LINKED ISSUE #1749: table update, wrong dataType for set operation, get 500 error
Table

`REATE TABLE filter.animal (
    id text,
    name text,
    age int,
    height smallint,
    description text,
    PRIMARY KEY ((id, name), age, height)`
    
Update

```json
{
    "updateOne": {
        "filter": {
            "id": "1",
            "name": "dqy",
            "age": 1,
            "height": 2
        },
        "update": {
            "$set": {
                "description": 123
            }
        }
    }
}
```

response

```json
{
    "errors": [
        {
            "message": "Server failed: root cause: (java.lang.RuntimeException) io.stargate.sgv2.jsonapi.exception.checked.ToCQLCodecException: Error trying to convert to targetCQLType `TEXT` from value.class `java.lang.Long`, value 123. Root cause: no codec matching value type",
            "id": "ff27c5d5-8814-450a-bdf7-1ca5279e9d07",
            "family": "SERVER",
            "title": "Server failed",
            "scope": "DATABASE"
        }
    ]
}
```

Yuqi-Du:
This PR is to fix, the 500s server_unhandled error, which happen in Table Filter and Update Operation Path. 
We should map these 500s into corresponding API exception.

Fixes #1749 #1740 #1744

**Checklist**
- [x] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/service/operation/query/ColumnAssignment.java
```diff
@@ -82,13 +92,30 @@ protected void addPositionalValues(List<Object> positionalValues) {
               .codecToCQL(tableMetadata, column, rawValue)
               .toCQL(rawValue));
     } catch (MissingJSONCodecException e) {
-      // TODO: Better error handling
-      throw new RuntimeException(e);
+      throw DocumentException.Code.UNSUPPORTED_COLUMN_TYPES.get(
+          errVars(
+              TableSchemaObject.from(tableMetadata, new ObjectMapper()),
```
tatu-at-datastax, line 97: Pass `OBJECT_MAPPER` instead

src/main/resources/errors.yaml
```diff
@@ -249,6 +240,17 @@ request-errors:
       
       ${SNIPPET.RESEND_USING_ONLY_DEFINED_COLUMNS}
 
+  - scope: FILTER
+    code: INVALID_FILTER_COLUMN_VALUES
+    title: Column value in filter is not valid
+    body: |-
+      Only values that are supported by the column data type can be included when filtering.
```
tatu-at-datastax, line 247: Not 100% sure but this might be misleading. Not sure which of the problems occurs here:

1. Filter refers to columns of types we do not support (like Counter). If so, message is bit misleading
2. Filter refers to supported column but JSON type not supported. If so, message makes sense.

So just wanted to check which case we have here.

tatu-at-datastax: Ok looking at test case(s), I think this is (2) so we are good. Sorry for the noise.

tatu-at-datastax (approved):
LGTM: suggested one improvement (wrt `ObjectMapper` use), but approved.

tatu-at-datastax (approved):

PR #1759, opened 2024-11-25
Fix alterTable command with empty/null columns

### LINKED ISSUE #1681: alterTable with e.g. `"add": {}` gives a 500 Internal Server Error
"Server failed: root cause: (java.lang.IllegalStateException) columnsToAdd must be non null and non empty.", while a graceful "regular" error is probably the right thing to do

Yuqi-Du:
Originally, a placeHolder exception is there after 'validating' null or empty columns.
The correct logic should be when columns are null or empty.
Take add columns operation as an example. Below 3 should work as NO_OP.

- "add":{}
- "add":{"columns":null}
- "add":{"columns":{}}

Fixes #1681

**Checklist**
- [x] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/service/operation/tables/AlterTableAttempt.java
```diff
@@ -92,25 +90,5 @@ public enum AlterTableType {
     ADD_COLUMNS,
     DROP_COLUMNS,
     UPDATE_EXTENSIONS;
-
```
amorton, line 95: This can stay, it is the OperationAttempt making sure it is correctly constructed - without worrying about whatever checks the caller has done

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/AlterTableCommandResolver.java
```diff
@@ -91,6 +91,13 @@ private List<AlterTableAttempt> handleAddColumns(
       TableSchemaObject tableSchemaObject,
       AlterTableOperationImpl.AddColumns addColumnsOperation) {
 
+    // 1. "add":{}
```
amorton, line 94: +1 for checking like this, but we should then return an error if no attempts were created in the resolveTableCommand() function. 

code is probably a new MISSING_ALTER_TABLE_OPERATIONS and we can re-use for all the diff alter table types

amorton (changes requested):
need to throw a ApiException in the resolver if no attempts are made

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/AlterTableCommandResolver.java
Yuqi-Du: solved, thanks

src/main/java/io/stargate/sgv2/jsonapi/service/operation/tables/AlterTableAttempt.java
Yuqi-Du: solved, thanks

amorton (approved):
I updated the error message to match the pattern, and then updated the IT to include checking that error message has request specific information in it.

amorton:
approved, but will check with @Yuqi-Du if we actually need this in the hotfix

PR #1762, opened 2024-11-26
(DO NOT MERGE!) Fixes #1664: allow ":" and "/" in Document property names

### LINKED ISSUE #1664: Feature Request: support additional characters allowed in keys
Context: [slack discussion thread](https://datastax.slack.com/archives/C05PTSGBYL9/p1730785234810079)

In comparison with competing products, 

- MongoDB natively allows a user to write the full kubernetes manifest json into a document, including `.`, `/` and `-`. For example, a very commonly seen label looks like

```yaml
app.kubernetes.io/managed-by: kustomize
```

- AWS DocumentDB does not allow `.`.  See [AWS Documentation](https://docs.aws.amazon.com/documentdb/latest/developerguide/functional-differences.html)
> Amazon DocumentDB does not support dots “.” in a document field name, for example, db.foo.insert({‘x.1’:1}).
> Amazon DocumentDB also does not support the $ prefix in field names.

- Currently, AstraDB Data API only supports alphanumeric characters plus `_`. See [document on naming convention](https://docs.datastax.com/en/astra-db-serverless/api-reference/overview.html#naming-conventions)
> Can contain uppercase letters A-Z, lowercase letters a-z, numbers 0-9, and underscores (_).

# Minimal Request

Given the difficulty of `.`,  we at the DataStax Platform request adding support to at least `/` and `-`.  We are currently using AWS DocumentDB and have already implemented a transformation for `.` (see https://github.com/riptano/lf-controltower/pull/33).  Supporting `/` and `-` would make it possible for us to migrate to use CNDB.  Not only dogfooding  makes business sense,  but also the performance, global replication and multiple-cloud support of AstraDB makes it the ideal technical solution. 

We need this feature for the platform control-tower initiative, which is necessary to support all products (serverless, classic, langflow, gpu, open-search...).   In the same time, I am sure this feature gap with MongoDB and DocumentDB might be an issue for many customers.

tatu-at-datastax:
**What this PR does**:

Allows use of ":" and "/" in Document property names

Fixes #1664

**Checklist**
- [ ] Changes manually tested


tatu-at-datastax:
No longer relevant wrt Real fix to #1664: closing.

PR #1772, opened 2024-12-02
Fix listIndexes bug with double quoted columnName in the index target from indexMetadata

### LINKED ISSUE #1770: API Fails to recognise vector index created via the API
Something in parsing the CQL "describe index" goes wrong when the indexed column is quoted (e.g. when a column name is in SnakeCase) and the index is interpreted as `"column": "UNKNOWN"`. Therefore, ANN fails (and presumably so do other operations).

_(originally reported by Kavin; I did diagnose and came up with this tentative explanation)_

### Steps to reproduce

(adapted from Kavin's)

Table

```
{"createTable": {
  "name": "dreams",
  "definition": {
    "columns": {
      "matchId": "text",
      "round": "tinyint",
      "mVector": { "type": "vector", "dimension": 3 },
      "score": "int",
      "when": "timestamp",
      "winner": "text",
      "fighters": { "type": "set", "valueType": "uuid" }
    },
    "primaryKey": { "partitionBy": [ "matchId" ], "partitionSort": { "round": 1 } }
  },
  "options": { "ifNotExists": true }
}}
```

Index

```
{"createVectorIndex": {
  "name": "m_vector_index",
  "definition": {
    "column": "mVector",
    "options": { "sourceModel": "ada002", "metric": "dot_product" }
  }
}}
```

(unknown-returning) List indexes

```
{"listIndexes": {"options": {"explain": true}}}
```

(failing) ANN search

```
{"findOne": {
  "filter": {},
  "sort": { "mVector": [ 0.2, -0.3, -0.5 ] }
}}
```

### Sanity check

The same as above, just all-lowercase names for columns (the only change):

```
{"createTable": {
  "name": "dreams",
  "definition": {
    "columns": {
      "match_id": "text",
      "round": "tinyint",
      "m_vector": { "type": "vector", "dimension": 3 },
      "score": "int",
      "when": "timestamp",
      "winner": "text",
      "fighters": { "type": "set", "valueType": "uuid" }
    },
    "primaryKey": { "partitionBy": [ "match_id" ], "partitionSort": { "round": 1 } }
  },
  "options": { "ifNotExists": true }
}}
```

```
{"createVectorIndex": {
  "name": "m_vector_index",
  "definition": {
    "column": "m_vector",
    "options": { "sourceModel": "ada002", "metric": "dot_product" }
  }
}}
```

```
{"listIndexes": {"options": {"explain": true}}}
```

```
{"findOne": {
  "filter": {},
  "sort": { "m_vector": [ 0.2, -0.3, -0.5 ] }
}}
```

Yuqi-Du:
**What this PR does**:
Summary, we got something like this, columnName is doubleQuoted in index target.
1. The regex does not match it.
2. We need to manually remove the double quote and then create cqlIndentifier.

Reproduction:

Table
```
{"createTable": {
  "name": "dreams",
  "definition": {
    "columns": {
      "matchId": "text",
      "round": "tinyint",
      "mVector": { "type": "vector", "dimension": 3 },
      "score": "int",
      "when": "timestamp",
      "winner": "text",
      "fighters": { "type": "set", "valueType": "uuid" }
    },
    "primaryKey": { "partitionBy": [ "matchId" ], "partitionSort": { "round": 1 } }
  },
  "options": { "ifNotExists": true }
}}
```

Index

```
{"createVectorIndex": {
  "name": "m_vector_index",
  "definition": {
    "column": "mVector",
    "options": { "sourceModel": "ada002", "metric": "dot_product" }
  }
}}
```

(unknown-returning) List indexes
```
{
    "status": {
        "indexes": [
            {
                "name": "abc_AH_123_a_1121",
                "definition": {
                    "column": "UNKNOWN",
                    "apiSupport": {
                        "createIndex": false,
                        "filter": false,
                        "cqlDefinition": "CREATE CUSTOM INDEX \"abc_AH_123_a_1121\" ON filter.testindex (\"V22eector2\")\nUSING 'StorageAttachedIndex'\nWITH OPTIONS = {\n    'similarity_function' : 'DOT_PRODUCT',\n    'source_model' : 'ADA002'}"
                    }
                }
            }
       }
}
```

Fixes #1770

**Checklist**
- [x] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/service/schema/tables/CQLSAIIndex.java
```diff
@@ -4,6 +4,7 @@
 import com.datastax.oss.driver.api.core.metadata.schema.IndexKind;
 import com.datastax.oss.driver.api.core.metadata.schema.IndexMetadata;
 import com.datastax.oss.driver.internal.core.adminrequest.AdminRow;
+import com.datastax.oss.driver.internal.core.util.Strings;
```
tatu-at-datastax, line 7: It seems dangerous to use classes from driver's "internal" packages....

src/main/java/io/stargate/sgv2/jsonapi/service/schema/tables/CQLSAIIndex.java
```diff
@@ -102,11 +107,14 @@ static boolean indexClassIsSai(String className) {
...
-   *   <li>"values(monkeys)": group 1 - "values" group 2 - "monkeys"
+   *   <li>'monkeys': group 1 - 'monkeys', group 2 - null
+   *   <li>'values(monkeys)': group 1 - 'values' group 2 - 'monkeys'
+   *   <li>'"Monkeys": group 1 - '"Monkeys"', group 2 - null
+   *   <li>'values("Monkeys")': group 1 - 'values' group 2 - '"monkeys"'
    * </ul>
    */
-  private static Pattern INDEX_TARGET_PATTERN = Pattern.compile("^(\\w+)?(?:\\((\\w+)\\))?$");
+  private static Pattern INDEX_TARGET_PATTERN =
+      Pattern.compile("^(\"?\\w+\"?)?(?:\\((\"?\\w+\"?)\\))?$");
```
tatu-at-datastax, line 117: omg. This is... not easy to decipher. I don't know if anything can be done to help since regexps get pretty complicated for nested cases.

src/main/java/io/stargate/sgv2/jsonapi/service/schema/tables/CQLSAIIndex.java
```diff
@@ -144,6 +152,16 @@ static IndexTarget indexTarget(IndexMetadata indexMetadata)
...
+    // At this point, after matcher, we may get a columnName doubleQuoted string from the index
+    // target
+    // E.G. 'target': 'values("a_list_column")' -> "a_list_column", 'target':
+    // 'values("a_regular_column")' -> "a_regular_column"
+    // 1. We can NOT use fromInternal which will keep the double quote in the identifier
+    // 2. We can NOT use fromCQL, since it will lowercase the unquoted string, E.G. 'BigApple' ->
+    // bigapple
+    // So we should just stripe the doubleQuote if needed
+    columnName =
+        Strings.isDoubleQuoted(columnName) ? Strings.unDoubleQuote(columnName) : columnName;
```
tatu-at-datastax, line 164: Looks like `unDoubleQuote()` immediate checks `isDoubleQuoted()` so can simplify as:

    columnName = Strings.unDoubleQuote(columnName);

and works the same, but avoids calling `isDoubleQuoted()` twice.

tatu-at-datastax (commented):
Ok this looks quite complicated despite small size of diffs.

I am not super happy with the way things are handled but it may be necessary to fix the issue so that's ok.

But what really would be needed would be a test to show that fix works, if at all possible. And will guard against regression (avoid bug from reappearing due to changes)

src/main/java/io/stargate/sgv2/jsonapi/service/schema/tables/CQLSAIIndex.java
Yuqi-Du: thanks, solved!

src/main/java/io/stargate/sgv2/jsonapi/service/schema/tables/CQLSAIIndex.java
Yuqi-Du: We use it in other places though.
What do you recommend Tatu, writing our own unquote() method is better?

src/main/java/io/stargate/sgv2/jsonapi/service/schema/tables/CQLSAIIndex.java
Yuqi-Du: Agree, let me write some tests at least.

src/main/java/io/stargate/sgv2/jsonapi/service/schema/tables/CQLSAIIndex.java
tatu-at-datastax: No, let's use that method for now.

tatu-at-datastax (approved):

amorton:
I would not have merged this, it is a very good candidate to write a unit test , so I have re-opened the ticket https://github.com/stargate/data-api/issues/1770

src/main/java/io/stargate/sgv2/jsonapi/service/schema/tables/CQLSAIIndex.java
```diff
@@ -82,38 +83,44 @@ static boolean indexClassIsSai(String className) {
...
+   * {'class_name': 'StorageAttachedIndex', 'target': 'values("listWithCapitalLetters")'}
    *
-   * <ul>
-   *   <li>"monkeys": group 1 - "monkeys", group 2 - null
-   *   <li>"values(monkeys)": group 1 - "values" group 2 - "monkeys"
-   * </ul>
+   * </pre>
    */
-  private static Pattern INDEX_TARGET_PATTERN = Pattern.compile("^(\\w+)?(?:\\((\\w+)\\))?$");
+  private static final Pattern TARGET_REGEX =
```
amorton, line 112: This code is copied from the CQL class, we should add the name of the CQL class

src/main/java/io/stargate/sgv2/jsonapi/service/schema/tables/CQLSAIIndex.java
```diff
@@ -128,29 +135,68 @@ static IndexTarget indexTarget(IndexMetadata indexMetadata)
...
-      functionName = null;
-    } else {
+    CqlIndexType cqlIndexType;
+    if (matcher.matches()) {
+      cqlIndexType = CqlIndexType.fromString(matcher.group(1));
       columnName = matcher.group(2);
-      functionName = matcher.group(1);
+    } else {
+      columnName = target;
+      cqlIndexType = CqlIndexType.VALUES;
```
amorton, line 149: I am not sure about this, why is the default Values ? 
What about what this is an index on a text field ?

src/main/java/io/stargate/sgv2/jsonapi/service/schema/tables/CQLSAIIndex.java
```diff
@@ -128,29 +135,68 @@ static IndexTarget indexTarget(IndexMetadata indexMetadata)
...
+    // will be enclosed in quotes, which we need to unwrap.
+    // 2. It may also include quote characters internally, escaped like so:
+    // abc"def -> abc""def, then we need to un-escape as abc"def -> abc"def
+    // to get the actual column name
+    if (columnName.startsWith(QUOTE)) {
+      columnName = StringUtils.substring(StringUtils.substring(columnName, 1), 0, -1);
+      columnName = TWO_QUOTES.matcher(columnName).replaceAll(QUOTE);
+    }
+
+    return new IndexTarget(CqlIdentifier.fromInternal(columnName), cqlIndexType);
```
amorton, line 162: Why did this stop using existing ApiIndexFunction ?

src/main/java/io/stargate/sgv2/jsonapi/service/schema/tables/CQLSAIIndex.java
```diff
@@ -128,29 +135,68 @@ static IndexTarget indexTarget(IndexMetadata indexMetadata)
...
+    KEYS,
+    KEYS_AND_VALUES,
+    FULL,
+    SIMPLE;
+
+    public String toString() {
+      switch (this) {
+        case KEYS:
+          return "keys";
+        case KEYS_AND_VALUES:
```
amorton, line 179: why is this called KEY_AND_VALUES when it is the entries ?

src/main/java/io/stargate/sgv2/jsonapi/service/schema/tables/CQLSAIIndex.java
```diff
@@ -128,29 +135,68 @@ static IndexTarget indexTarget(IndexMetadata indexMetadata)
...
+        case SIMPLE:
+          return "";
+        default:
+          return "";
+      }
+    }
+
+    public static CqlIndexType fromString(String s) {
+      if ("".equals(s)) return SIMPLE;
+      else if ("values".equals(s)) return VALUES;
```
amorton, line 194: this is a case sensitive match and duplicates the logic in toString

src/main/java/io/stargate/sgv2/jsonapi/service/schema/tables/CQLSAIIndex.java
```diff
@@ -128,29 +135,68 @@ static IndexTarget indexTarget(IndexMetadata indexMetadata)
...
+      else if ("keys".equals(s)) return KEYS;
+      else if ("entries".equals(s)) return KEYS_AND_VALUES;
+      else if ("full".equals(s)) return FULL;
+
+      throw new AssertionError("Unrecognized index target type " + s);
```
amorton, line 199: This should be IllegalArgument we have not used AssertionError anywhere in the code

src/main/java/io/stargate/sgv2/jsonapi/service/schema/tables/ApiVectorIndex.java
```diff
@@ -268,13 +268,6 @@ protected ApiIndexDef create(
                 .formatted(ApiIndexType.VECTOR, apiIndexType));
       }
 
-      // also, we  must not have an index function
```
amorton, line 271: Why is this check removed ?

src/main/java/io/stargate/sgv2/jsonapi/service/schema/tables/ApiIndexType.java
```diff
@@ -53,31 +53,29 @@ public static ApiIndexType fromTypeName(String typeName) {
...
       ApiColumnDef apiColumnDef, CQLSAIIndex.IndexTarget indexTarget, IndexMetadata indexMetadata)
       throws UnsupportedCqlIndexException {
-
-    // If there is no function on the indexTarget, and the column is a scalar, then it is a regular
-    // index on
-    // an int, text, etc
-    if (indexTarget.indexFunction() == null && apiColumnDef.type().isPrimitive()) {
+    // If cqlIndexType is values, and the column is a scalar
+    // then it is a regular index on primitive types
+    if (indexTarget.cqlIndexType() == CQLSAIIndex.CqlIndexType.VALUES
```
amorton, line 58: this feels risky - we are inferring the VALUES function when it is not defined in the CQL for the index and then using that inferred value. The previous code did not infer a function when there was none defined.

src/main/java/io/stargate/sgv2/jsonapi/service/schema/tables/ApiRegularIndex.java
```diff
@@ -186,13 +186,6 @@ protected ApiIndexDef create(
                 .formatted(ApiIndexType.REGULAR, apiIndexType));
       }
 
-      // also, we should not have an index function
```
amorton, line 189: why is this check removed, there is no function for a regular index

amorton (commented):
-1 on this merging, it makes some assumptions about things in the CQL target value that are risks, removes some validation checks , and adds a new enum rather than use the existing ApiIndexFunction

PR #1785, opened 2024-12-09
[Table] Add error message when inserting data using multiple vectorize configs

### LINKED ISSUE #1752: Table vectorize, different providers, different dimension, 500 Internal Server
Split from #1726.
Current ticket is for dec-hot-fix to solve.

### Two columns with same provider, model and dimension

Inserting a row with `{v1: "blabla", v2: [...]}` (i.e. one string and one vector) works

Both vectors: works

Both strings: works (two different embedding vectors got stored for two different texts, as expected)

### Two columns with same provider and model, but different dimension

Passing both as strings _does not work_: the API mistakenly thinks both vectors are the same dim:

```
The Embedding Provider returned an unexpected response: The Embedding Provider
returned an unexpected response: Embedding provider 'openai' did not return expected
embedding length. Expect: '333'. Actual: '123'
(EMBEDDING_PROVIDER_UNEXPECTED_RESPONSE)
```

### Two columns with same provider, different models, different dimensions

Same error as above (and I suspect at this point even if the dimension did match it would either error or work in the wrong way)

### Two columns with different providers

This time, inserting both as strings leads to a 500 Internal Server Error:

```
Server failed: root cause: (java.lang.IllegalArgumentException) Must be single
embedding provider name, got [openai, jinaAI]. Server error '500 Internal Server
Error' for url 'https://[...]-us-west-2.apps.astra-dev.datastax.com/[...]/<TABLE>
```

If one of the two is passed as a vector, instead, and the other is a string, depending on which one two things happen:

1. it works
2. the dimension-mismatch error seen above is triggered (probably because the dimension is still a table-level setting somewhere and is used for the wrong model)

Hazel-Datastax:
**What this PR does**:
Currently, we allow users to create tables using multiple configs but don't support it in insert. This PR adds a proper error message to indicate that the function will be supported in the future. 

This change is intended for the `hotfix` branch as a temporary solution. Full support will be implemented in a future update and merged into the `main` branch.

Fixes: #1752 

**Checklist**
- [x] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/service/schema/tables/ApiVectorType.java
```diff
@@ -73,6 +73,23 @@ public static ApiVectorType from(int dimension, VectorizeDefinition vectorizeDef
...
+  /**
+   * This method is only used for Dec hotfix. Currently, we don't support multiple vectorize service
+   * config. 1. If one doesn't have vectorizeDefinition (use user's own vector) and the other one
+   * have (use embedding service), it's ok 2. If both have vectorizeDefinition, only support if they
+   * are the same and have the same dimension
+   *
+   * @param v the other vector type to compare
+   * @return true if the two vector types are supported together
+   */
+  public boolean isTwoVectorColumnSupported(ApiVectorType v) {
```
tatu-at-datastax, line 85: Should probably be named more like "isCompatibleWith(...)"? (since it checks for partial equality).

src/main/java/io/stargate/sgv2/jsonapi/service/schema/tables/ApiVectorType.java
```diff
@@ -73,6 +73,23 @@ public static ApiVectorType from(int dimension, VectorizeDefinition vectorizeDef
...
+    if (this.vectorizeDefinition == null || v.getVectorizeDefinition() == null) {
```
tatu-at-datastax, line 86: Maybe add a comment above; I think it means that one or both of Vector is not using Vectorization, and so there is no conflict?

amorton (changes requested):
-1 for this, the hot fix is just to present a proper error for the insert path and then make it work in main

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/DataVectorizer.java
```diff
@@ -418,4 +417,26 @@ public void setVector(float[] vector) {
           .set(i, SortExpression.tableVectorSort(sortExpression.path(), vector));
     }
   }
+
+  record VectorizeIdentity(String providerName, String modelName, int dimension) {
```
tatu-at-datastax, line 421: Brief javadoc?

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/DataVectorizer.java
```diff
@@ -263,19 +264,17 @@ public <T> Uni<T> vectorizeTasks(
...
     // goes wrong :)
-    var embeddingProviderNames =
-        tasks.stream()
-            .map(task -> task.vectorType.getVectorizeDefinition().provider())
-            .distinct()
-            .toList();
-    if (embeddingProviderNames.size() != 1) {
-      throw new IllegalArgumentException(
-          "Must be single embedding provider name, got " + embeddingProviderNames);
+    var vectorizeIdentities =
```
tatu-at-datastax, line 267: Add a comment linking issue URL and brief explanation, to indicate this is temporary limitation.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/tables/CreateTableIntegrationTest.java
```diff
@@ -923,6 +923,38 @@ private static Stream<Arguments> allTableData() {
                   true,
                   SchemaException.Code.MISSING_DIMENSION_IN_VECTOR_COLUMN.name(),
                   "The dimension is required for vector columns if the embedding service is not specified.")));
+
+      // Two vector columns with the one has vectorizeDefinition and the other one doesn't
```
tatu-at-datastax, line 927: Maybe add a note that this acceptable (only one distinct vectorize definition)? (this is implied by having `null` for failure, but just to keep things readable)

src/test/java/io/stargate/sgv2/jsonapi/api/v1/tables/InsertOneTableIntegrationTest.java
```diff
@@ -173,6 +181,107 @@ public final void createDefaultTables() {
                 Map.of("type", "vector", "valueType", "float", "dimension", 3)),
             "id")
         .wasSuccessful();
+
+    assertNamespaceCommand(keyspaceName)
```
tatu-at-datastax, line 185: Instead of adding every table in init block, tables that are only used by a single test method could (and I think should) be moved to that test method: first create table, then try insert which succeeds or fails.
And only use init method for tables shared across multiple tests.

Doing this keeps table and usage close to each other and avoids humongous methods.

tatu-at-datastax (commented):
Looks good, hoping for couple of small changes.

tatu-at-datastax (approved):
LGTM!

amorton (approved):

PR #1798, opened 2024-12-12
Fix #1698: serialize Vector key correctly for "insertedIds" status

### LINKED ISSUE #1698: `vector` primary key not returned properly in `insertedIds`
If I create a table like so:
```ts
const table = await db.createTable('t2est_table2', { 
  definition: { columns: { vec: { type: 'vector', dimension: 3 } }, primaryKey: 'vec' }, 
  ifNotExists: true 
});
```

and insert it either way:
```ts
// serializes as number[]
await table.insertOne({ vec: [0.1, 0.2, 0.3] });

// serializes as $binary
await table.insertOne({ vec: new DataAPIVector([0.1, 0.2, 0.3]) });
```

You can see in this resulting event that the `vector` was returned simply as `{ exists: true }` for some reason
```
CommandSucceededEvent {
  command: {
    insertOne: { document: { a: [ 0.1, 0.2, 0.3 ] } } // same result if serialized as $binary
  },
  resp: {
    status: {
      insertedIds: [ [ { empty: false } ] ],
      primaryKeySchema: { blob: { type: 'vector', dimension: 3 } }
    },
    ...
  },
  ...
}
```

tatu-at-datastax:
**What this PR does**:

Fixes serialization of `CqlVector` as part of primary key; other test improvements

Fixes #1698

**Checklist**
- [ ] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/service/operation/tables/TableInsertAttempt.java
```diff
@@ -43,11 +44,10 @@ public Optional<ColumnsDescContainer> schemaDescription() {
     if (row == null) {
       return Optional.empty();
     }
-
     var apiColumns = schemaObject.apiTableDef().primaryKeys();
-    if (!apiColumns.filterByUnsupported().isEmpty()) {
-      throw new IllegalStateException(
-          "Unsupported columns primary key: %s" + apiColumns.filterByUnsupported());
+    var unsupported = apiColumns.filterBySupport(ApiSupportDef::isUnsupportedInsert);
```
tatu-at-datastax, line 48: Fix required to find columns that do not support Insert and NOT columns that miss support for any of operations -- f.ex. `time_uuid` does support Insert but not Create (table).
Without fix added test fails so the fix is added via this PR.

src/test/java/io/stargate/sgv2/jsonapi/api/configuration/ObjectMapperConfigurationTest.java
```diff
@@ -55,7 +56,7 @@ class ObjectMapperConfigurationTest {
   @Inject DocumentLimitsConfig documentLimitsConfig;
 
   @Nested
-  class unmatchedOperationCommandHandlerTest {
+  class UnmatchedOperationCommandHandlerTest {
```
tatu-at-datastax, line 59: Minor naming fix; Java classes should follow "upper camel case".

src/test/java/io/stargate/sgv2/jsonapi/api/v1/tables/InsertOneTableIntegrationTest.java
```diff
@@ -195,7 +209,8 @@ void insertWithTextColumns() {
       assertTableCommand(keyspaceName, TABLE_WITH_TEXT_COLUMNS)
           .templated()
           .insertOne(DOC_JSON)
-          .wasSuccessful();
+          .wasSuccessful()
+          .hasInsertedIds(List.of("abc"));
```
tatu-at-datastax, line 213: Add verification for "insertedIds" for existing tests, was missing.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/util/DataApiResponseValidator.java
```diff
@@ -263,7 +264,7 @@ public DataApiResponseValidator hasWarning(
     return validator;
   }
 
-  public DataApiResponseValidator mayHasSingleWarning(WarningException.Code warningExceptionCode) {
+  public DataApiResponseValidator mayHaveSingleWarning(WarningException.Code warningExceptionCode) {
```
tatu-at-datastax, line 267: Minor naming fix.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/util/DataApiResponseValidator.java
```diff
@@ -288,6 +289,11 @@ public DataApiResponseValidator hasInsertedIdCount(int count) {
     return body("status.insertedIds", hasSize(count));
   }
 
+  public DataApiResponseValidator hasInsertedIds(List<?>... ids) {
```
tatu-at-datastax, line 292: Added convenience verification method.

src/main/java/io/stargate/sgv2/jsonapi/api/configuration/ObjectMapperConfiguration.java
```diff
@@ -58,15 +60,20 @@ private ObjectMapper createMapper(DocumentLimitsConfig documentLimitsConfig) {
              */
             .disable(StreamWriteFeature.WRITE_BIGDECIMAL_AS_PLAIN)
             .build();
+    // For now, just one custom serializer to register
+    SimpleModule cqlSerializersModule =
+        new SimpleModule("CqlSerializersModule").addSerializer(new CqlVectorSerializer());
```
maheshrajamani, line 65: nit: Should this be called CqlVectorSerializersModule?

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/serializer/CqlVectorSerializer.java
```diff
@@ -0,0 +1,23 @@
+package io.stargate.sgv2.jsonapi.api.model.command.serializer;
+
+import com.datastax.oss.driver.api.core.data.CqlVector;
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.databind.SerializerProvider;
+import com.fasterxml.jackson.databind.ser.std.StdSerializer;
+import java.io.IOException;
+
+public class CqlVectorSerializer extends StdSerializer<CqlVector<Float>> {
```
maheshrajamani, line 9: java doc

maheshrajamani (approved):
Change LGTM. Side question if vector field is used as partition key think we can't create index on it. Meaning we can only do equals search on vector value. Is this a real use case?

src/main/java/io/stargate/sgv2/jsonapi/api/configuration/ObjectMapperConfiguration.java
tatu-at-datastax: Hmmh. For current state, yes, `CqlVectorSerializerModule`. Can change name if more serializers (or other things) need registered.

tatu-at-datastax:
> Change LGTM. Side question if vector field is used as partition key think we can't create index on it. Meaning we can only do equals search on vector value. Is this a real use case?

That is a valid question: as per issue discussion, considered an edge case. But since Kavin found it, guessing some users might as well.

src/main/java/io/stargate/sgv2/jsonapi/service/schema/tables/ApiSupportDef.java
```diff
@@ -57,6 +57,10 @@ default boolean isUnsupportedDML() {
     return !insert() || !read() || !filter();
   }
 
+  default boolean isUnsupportedInsert() {
+    return !insert();
+  }
+
```
Hazel-Datastax, line 63: According to Aaron's design, it may not be a good idea to add an extra method here, otherwise, we need to create a method for each combination. [Here](https://github.com/stargate/data-api/blob/5c9863860960789ea00365ac6e60a074fe1d242d/src/main/java/io/stargate/sgv2/jsonapi/service/schema/tables/ApiColumnDefContainer.java#L145) is the example of how Aaron use `NO_MATCHES` for different combinations.

tatu-at-datastax: That example does not seem to support negation but only is-supported so that wouldn't work as-is I think. But I can probably change this to user `Predicate.negate()` or so to avoid adding new methods.
Thanks!

tatu-at-datastax: @Hazel-Datastax here goes: https://github.com/stargate/data-api/pull/1805 -- simpler code, can just use simple lambda; I am not sure I understand why `Matcher` is added but it's not needed in this case (and it has no `negate()` method anyway).
Thanks!

This might be something to discuss wrt Java closures/lambdas chapter of "Java for Impatienet"; there are multiple ways to achieve the result.

vkarpov15 (commented):
LGTM :+1:

PR #1805, opened 2024-12-20
Minor clean up on #1698 fix based on code review comments

tatu-at-datastax:
**What this PR does**:

Cleans up PR #1798 (for #1698) to simplify code, avoid adding new method.
Based on PR review comments added post-merge.

N/A

**Checklist**
- [x] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/service/operation/tables/TableInsertAttempt.java
```diff
@@ -45,7 +44,7 @@ public Optional<ColumnsDescContainer> schemaDescription() {
       return Optional.empty();
     }
     var apiColumns = schemaObject.apiTableDef().primaryKeys();
-    var unsupported = apiColumns.filterBySupport(ApiSupportDef::isUnsupportedInsert);
+    var unsupported = apiColumns.filterBySupport(x -> !x.insert());
```
Hazel-Datastax, line 47: Not sure I fully understand why Matcher cannot be used in this case. Aaron has the negation example in [WriteableTableRowBuilder](https://github.com/stargate/data-api/blob/35f0da45acc17e314616ca5c59dabb07f591e1eb/src/main/java/io/stargate/sgv2/jsonapi/service/operation/tables/WriteableTableRowBuilder.java#L34) (sorry, there are so many `NO_MATCHES` examples and I just randomly picked one previously.)

I tried `ApiSupportDef.Matcher.NO_MATCHES.withInsert(false)` here and it could pass the test `insertValidVectorKey` you added in the previous PR. So I think it works? I don't mean we always need to have the same way, just curious.

Hazel-Datastax (approved):

src/main/java/io/stargate/sgv2/jsonapi/service/operation/tables/TableInsertAttempt.java
tatu-at-datastax: To me this way seems bit simpler than building and using a Matcher, that's all. I am sure Matcher could be used, and would make sense for more complex cases.
So maybe just matter of preference.

Issue #1814, opened 2025-01-01
Error on keyspaces starting with a number


toptobes:
Got an issue from someone reporting that the Data API does not accept otherwise perfectly valid keyspaces which start with a number, such as `2DQJAUtSzzzKP9buDTvUvPk3`. I can confirm that such a keyspace can be created from the portal with absolutely no issues.

> I'm able to create a Keyspace both via REST API and Admin area with the value "2DQJAUtSzzzKP9buDTvUvPk3" but unable to retrieve and use it via the REST api.
> 
> ```ts
> const dbAdmin = admin.dbAdmin(dbEndpoint)
> await dbAdmin.createKeyspace('2DQJAUtSzzzKP9buDTvUvPk3')
> 
> // then i can list: 
> const keyspaces = await dbAdmin.listKeyspaces()
> console.log(keyspaces)
> 
> {
>   "existingKeyspaces": [
>     "default_keyspace",
>     "2DQJAUtSzzzKP9buDTvUvPk3"
>   ]
> }
> ```
> 
> Then when trying to use the keyspace:
> 
> ```ts
> const collectionNames = await db.listCollections({ nameOnly: true, keyspace: '2DQJAUtSzzzKP9buDTvUvPk3' })
> // or 
> await db.createCollection('collectionname', { keyspace: '2DQJAUtSzzzKP9buDTvUvPk3' })
> ```
> 
> This returns:
> 
> > DataAPIResponseError: Request invalid: field 'keyspace' value "2DQJAUtSzzzKP9buDTvUvPk3" not valid. Problem: must match "[a-zA-Z][a-zA-Z0-9_]*".

https://github.com/datastax/astra-db-ts/issues/94

Hazel-Datastax:
Similar to #1674, will solve them together

Hazel-Datastax:
Solved, see PR #1815

PR #1815, opened 2025-01-03
Update restrictions on schema names

### LINKED ISSUE #1674: createVectorIndex  has unwanted restrictions on the source model name
This is on the options for create index 

```
      @Nullable
          @Size(min = 1, max = 48)
          @Pattern(regexp = "[a-zA-Z][a-zA-Z0-9_]*")
          @Schema(description = "Model name used to generate the embeddings.")
          String sourceModel) {}
```

Why does it limit the characters and the length ? 

it looks like this has been copy - pasted into many places: 

```
        @Size(min = 1, max = 48)
        @Pattern(regexp = "[a-zA-Z][a-zA-Z0-9_]*")
```

This ticket should also check all the places this is present to see if it should be there.

Hazel-Datastax:
**What this PR does**:
Update naming restrictions on keyspace, collection, table, and index creation.
- Remove injection validation in `Command`.
- Add validation in `CommandResolver`
- New validation will support numbers and underscore as the first character
- Use the new error structure for the exception 

Remove the naming restriction when using the schema (keyspace, collection, table, and index)
- We restrict the naming length to 48 but the DB maximum allowed length is 222. If users create a table outside the DataAPI, we will still allow them to use it

Fixes #1674 #1814

**Checklist**
- [x] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/CreateCollectionCommand.java
```diff
@@ -19,6 +19,8 @@
 @JsonTypeName(CommandName.Names.CREATE_COLLECTION)
 public record CreateCollectionCommand(
     @NotNull
+        // TODO: according to issue #1814, it's ok to create a table with a name that starts with a
```
tatu-at-datastax, line 22: #1814 is about keyspaces (that already exist?), this is Collection name to create. So bit different. But we do need to decide on constraints for names of Collections we allow to be created.

Hazel-Datastax: Yes, I should remove the issue tag to reduce the confusion. They are a bit different, but they have the same restriction in the CQL. (according to the [doc](https://docs.datastax.com/en/cql-oss/3.3/cql/cql_reference/ref-lexical-valid-chars.html?hl=valid%2Ccharacters%2Cnames), `Keyspace and table names must begin with an alpha-numeric character and can only contain alpha-numeric characters and underscores.`)

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/CreateIndexCommand.java
```diff
@@ -16,6 +16,9 @@
 @JsonTypeName(CommandName.Names.CREATE_INDEX)
 public record CreateIndexCommand(
     @NotNull
+        // TODO: it seems CQL allows index name to begin with and contain any character. Should we
+        // completely remove it?
+        // https://docs.datastax.com/en/cql-oss/3.x/cql/cql_reference/ref-lexical-valid-chars.html
```
Yuqi-Du, line 21: Index is not mentioned in this uri though.

src/main/java/io/stargate/sgv2/jsonapi/util/NamingValidationUtil.java
```diff
@@ -0,0 +1,23 @@
...
+import java.util.regex.Pattern;
+
+public interface NamingValidationUtil {
+
+  String KEYSPACE_SCHEMA_NAME = "keyspace";
+  String COLLECTION_SCHEMA_NAME = "collection";
+  String TABLE_SCHEMA_NAME = "table";
+  String INDEX_SCHEMA_NAME = "index";
+
+  String NULL_SCHEMA_NAME = "null";
```
amorton, line 12: -1 on ideas like a NULL schema name, it's not a valid value for a typeof schema

src/main/java/io/stargate/sgv2/jsonapi/util/NamingValidationUtil.java
```diff
@@ -0,0 +1,23 @@
+package io.stargate.sgv2.jsonapi.util;
+
+import java.util.regex.Pattern;
+
+public interface NamingValidationUtil {
```
amorton, line 5: ok, I see where this is going, we should look at the WhereCQLClauseAnalyzer.StatementType as a pattern. so something like...

An interface / abstract class called NamingRule with: 
- name() - friendly name 
- apply() - runs the rule on a string 
- implementations for all the naming rules 

- A static class called NamingRules that has singleton instances of each of the rule classes.

amorton (changes requested):
expand the NamingRules see comments

src/main/java/io/stargate/sgv2/jsonapi/util/naming/SchemaObjectNamingRule.java
```diff
@@ -0,0 +1,20 @@
...
+  public static final int NAME_LENGTH = 48;
+  public static final Pattern PATTERN_WORD_CHARS = Pattern.compile("\\w+");
+
+  @Override
+  public abstract String name();
+
+  @Override
+  public boolean apply(String name) {
+    return name != null
+        && !name.isBlank()
```
tatu-at-datastax, line 16: Can use `!name.isEmpty()` instead, regexp can handle "all whitespace" case (blank checks for empty (0 length) as well as all white space)

src/main/java/io/stargate/sgv2/jsonapi/util/naming/SchemaObjectNamingRule.java
```diff
@@ -0,0 +1,20 @@
+package io.stargate.sgv2.jsonapi.util.naming;
+
+import java.util.regex.Pattern;
+
+public abstract class SchemaObjectNamingRule implements NamingRule {
+
+  public static final int NAME_LENGTH = 48;
```
tatu-at-datastax, line 7: Do all Schema entities have same max length limitation? Also, should be `MAX_NAME_LENGTH` (not just `NAME_LENGTH`).

src/main/java/io/stargate/sgv2/jsonapi/util/naming/SchemaObjectNamingRule.java
```diff
@@ -0,0 +1,20 @@
...
+  public static final Pattern PATTERN_WORD_CHARS = Pattern.compile("\\w+");
+
+  @Override
```
tatu-at-datastax, line 10: Could instead have `private final String name` here, require sub-classes to pass name in constructor, define `public String name() { return name; }` here. Bit more compact on sub-classes.

src/main/resources/errors.yaml
```diff
@@ -833,6 +833,18 @@ request-errors:
       
       Resend the command using a supported value type.
 
+  - scope: SCHEMA
+    code: UNSUPPORTED_SCHEMA_NAME
+    title: The used schema name is not supported
+    body: |-
+      The command attempted to create a ${schemeType} with a name that is not supported.
```
tatu-at-datastax, line 840: typo: `schemeType` -> `schemaType`

src/main/resources/errors.yaml
```diff
@@ -833,6 +833,18 @@ request-errors:
...
+      
+      The supported ${schemeType} names must not be empty, more than ${nameLength} characters long, or contain non-alphanumeric-underscore characters.
```
tatu-at-datastax, line 842: naming: `nameLength` -> `maxNameLength` ("nameLength" would sound like it refers to length of name being passed).

src/test/java/io/stargate/sgv2/jsonapi/api/v1/CollectionResourceIntegrationTest.java
```diff
@@ -116,64 +116,6 @@ public void unknownCommandField() {
               containsString("(\"filter\", \"options\", \"projection\", \"sort\")"));
     }
 
-    @Test
```
tatu-at-datastax, line 119: I understand that rule changed so leading number is fine. But shouldn't we change the test instead of removing it? Just because we have unit tests does not guarantee limits are applied end-to-end.

(I am fine removing resolver tests done earlier)

src/test/java/io/stargate/sgv2/jsonapi/api/v1/CreateKeyspaceIntegrationTest.java
```diff
@@ -159,12 +159,12 @@ public void invalidCommand() {
...
           .body("$", responseIsError())
-          .body("errors[0].errorCode", is("COMMAND_FIELD_INVALID"))
-          .body("errors[0].exceptionClass", is("JsonApiException"))
+          .body("errors[0].errorCode", is(SchemaException.Code.UNSUPPORTED_SCHEMA_NAME.name()))
+          .body("errors[0].exceptionClass", is(SchemaException.class.getSimpleName()))
           .body(
               "errors[0].message",
               is(
-                  "Request invalid: field 'command.name' value `null` not valid. Problem: must not be null."));
+                  "The command attempted to create a keyspace with a name that is not supported.\n\nThe supported keyspace names must not be empty, more than 48 characters long, or contain non-alphanumeric-underscore characters.\nThe command used the unsupported keyspace name: null.\n\nResend the command using a supported keyspace name."));
```
tatu-at-datastax, line 167: Could just check for part of message like

"The command attempted to create a keyspace with a name that is not supported."

so need not always change when we change error wording?

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/CommandResolver.java
```diff
@@ -213,4 +216,24 @@ private void getIndexUsageTags(DBLogicalExpression dbLogicalExpression, IndexUsa
...
+   *
+   * @param name The name to validate.
+   * @param namingRule The naming rule to apply.
+   * @throws SchemaException if the name is invalid.
+   */
+  default void validateSchemaName(String name, SchemaObjectNamingRule namingRule) {
+    if (!namingRule.apply(name)) {
+      throw SchemaException.Code.UNSUPPORTED_SCHEMA_NAME.get(
+          Map.of(
+              "schemeType",
```
tatu-at-datastax, line 231: typo: `schemeType` -> `schemaType`

tatu-at-datastax (commented):
Looks mostly good, some minor changes suggested.

src/main/java/io/stargate/sgv2/jsonapi/util/naming/SchemaObjectNamingRule.java
Hazel-Datastax: > Do all Schema entities have same max length limitation?

So far, yes.
Changed the name

src/test/java/io/stargate/sgv2/jsonapi/api/v1/CollectionResourceIntegrationTest.java
Hazel-Datastax: Yeah, we don't put a limit on naming usage. f.ex. the user can create a keyspace name longer than 48 characters using CQL, and we will still allow it to be used for insertion.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/CreateCollectionCommand.java
```diff
@@ -18,11 +18,7 @@
 @Schema(description = "Command that creates a collection.")
 @JsonTypeName(CommandName.Names.CREATE_COLLECTION)
 public record CreateCollectionCommand(
-    @NotNull
-        @Size(min = 1, max = 48)
-        @Pattern(regexp = "[a-zA-Z][a-zA-Z0-9_]*")
-        @Schema(description = "Name of the collection")
-        String name,
+    @Schema(description = "Name of the collection") String name,
```
tatu-at-datastax, line 21: Looks like we use `@NotEmpty` for some of these, but not all. Is this intentional?
It would perhaps be good to add it for all since I think "name" is not optional?

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/CommandResolver.java
```diff
@@ -213,4 +216,24 @@ private void getIndexUsageTags(DBLogicalExpression dbLogicalExpression, IndexUsa
...
+  default void validateSchemaName(String name, SchemaObjectNamingRule namingRule) {
+    if (!namingRule.apply(name)) {
+      throw SchemaException.Code.UNSUPPORTED_SCHEMA_NAME.get(
+          Map.of(
+              "schemaType",
+              namingRule.name(),
+              "maxNameLength",
+              String.valueOf(SchemaObjectNamingRule.MAX_NAME_LENGTH),
+              "unsupportedSchemaName",
+              name == null ? "null" : name));
```
tatu-at-datastax, line 236: Need to use something else for `null` case since "null" would be legal table/keyspace etc name? Maybe `<null>`?

src/main/java/io/stargate/sgv2/jsonapi/util/naming/CollectionNamingRule.java
```diff
@@ -0,0 +1,7 @@
+package io.stargate.sgv2.jsonapi.util.naming;
+
+public class CollectionNamingRule extends SchemaObjectNamingRule {
+  public CollectionNamingRule(String name) {
```
tatu-at-datastax, line 4: Do we need to pass name here, or can it be just defined here as:

   super("collection")

(or whatever constant we use)?

src/main/java/io/stargate/sgv2/jsonapi/util/naming/NamingRules.java
```diff
@@ -0,0 +1,10 @@
+package io.stargate.sgv2.jsonapi.util.naming;
+
+public final class NamingRules {
+  private NamingRules() {}
+
+  public static final SchemaObjectNamingRule KEYSPACE = new KeyspaceNamingRule("keyspace");
```
tatu-at-datastax, line 6: So, basically, I think constant "keyspace" can be specified in `KeyspaceNamingRule`, need not be passed from here.

src/main/java/io/stargate/sgv2/jsonapi/util/naming/SchemaObjectNamingRule.java
```diff
@@ -0,0 +1,28 @@
+package io.stargate.sgv2.jsonapi.util.naming;
+
+import java.util.regex.Pattern;
+
+public abstract class SchemaObjectNamingRule implements NamingRule {
+
+  public static final int MAX_NAME_LENGTH = 48;
```
tatu-at-datastax, line 7: Do these need to be public? If not, then either `protected` (if only sub-classes need them) or `private` (if not referenced by anything else)

src/test/java/io/stargate/sgv2/jsonapi/api/v1/tables/CreateTableIndexIntegrationTest.java
```diff
@@ -405,6 +405,89 @@ public void createVectorIndexWithCorrectIndexType() {
   @Nested
   @Order(3)
   class CreateIndexFailure {
+    @Test
+    public void createIndexWithEmptydName() {
```
tatu-at-datastax, line 409: typo: Emptyd -> Empty

src/test/java/io/stargate/sgv2/jsonapi/api/v1/tables/CreateTableIndexIntegrationTest.java
```diff
@@ -405,6 +405,89 @@ public void createVectorIndexWithCorrectIndexType() {
...
+                          "column": "vehicle_id_7"
+                        }
+                      }
+                      """)
+          .hasSingleApiError(
+              SchemaException.Code.UNSUPPORTED_SCHEMA_NAME,
+              SchemaException.class,
+              "The command attempted to create a index with a name that is not supported.",
+              "The supported index names must not be empty, more than 48 characters long, or contain non-alphanumeric-underscore characters.",
+              "The command used the unsupported index name: .");
```
tatu-at-datastax, line 426: I think we need to quote schema names: this message looks very confusing as-is.

If it read:

    The command used the unsupported index name: ''.

it would be little less odd.

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/CommandResolver.java
tatu-at-datastax: Also, I noticed (see a later comment) that error messages look quite strange if name is empty String (or all white-space).

I think names in general should be quoted so they are easier to tell from messages.
So maybe define `quote(String name)` method that does something like:

```java
public static String quote(String name) {
     if (name == null) {
        return "<null>";
    }
   return "'%s'".formatted(name);
}
```

called from here.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/CreateCollectionCommand.java
Hazel-Datastax: Yes, "name" is not optional. I intentionally removed it because `null` and `empty` name will be captured in the naming rules and could have better error message:
```
{
    "errors": [
        {
            "message": "The command attempted to create a collection with a name that is not supported.\n\nThe supported collection names must not be empty, more than 48 characters long, or contain non-alphanumeric-underscore characters.\nThe command used the unsupported collection name: .\n\nResend the command using a supported collection name.",
            "scope": "SCHEMA",
            "errorCode": "UNSUPPORTED_SCHEMA_NAME",
            "exceptionClass": "SchemaException",
            "id": "23f1c336-fa77-4f5d-a642-ad9f8a2d78d3",
            "family": "REQUEST",
            "title": "The used schema name is not supported"
        }
    ]
}
```

If we add `@NotEmpty` in here and the user uses empty name:
```
{
    "createCollection": {
        "name": ""
    }
}
```
The error message will be generated from this annotation instead of our own error message:
```
{
    "errors": [
        {
            "message": "Request invalid: field 'command.name' value \"\" not valid. Problem: must not be empty.",
            "errorCode": "COMMAND_FIELD_INVALID",
            "id": "17e28e51-169e-43d3-8bb9-3e2e0e7c5e6b",
            "title": "Request invalid",
            "family": "REQUEST",
            "scope": "EMPTY"
        }
    ]
}
```
What do you think?

src/main/java/io/stargate/sgv2/jsonapi/util/naming/SchemaObjectNamingRule.java
Hazel-Datastax: Yes, I used it in constructing error message:
```
    if (!namingRule.apply(name)) {
      throw SchemaException.Code.UNSUPPORTED_SCHEMA_NAME.get(
          Map.of(
              "schemaType",
              namingRule.name(),
              "maxNameLength",
              String.valueOf(SchemaObjectNamingRule.MAX_NAME_LENGTH),
              "unsupportedSchemaName",
              name == null ? "null" : name));
    }
```

src/test/java/io/stargate/sgv2/jsonapi/api/v1/CollectionResourceIntegrationTest.java
tatu-at-datastax: So creation outside of Data API via CQL? Because we do seem to limit name length to 48 in Data API commands?

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/CreateCollectionCommand.java
tatu-at-datastax: @Hazel-Datastax Good point. I guess it is better for us to do it all, and just remove all `@NotEmpty`s. I think there were 2 remaining.

src/main/java/io/stargate/sgv2/jsonapi/util/naming/SchemaObjectNamingRule.java
tatu-at-datastax: Ok. In future we may need to refactor this if max lengths can vary.

Actually, maybe `NamingRule` should have `getMaxLength()` method which -- for now -- just returns `SchemaObjectNamingRule.MAX_NAME_LENGTH`

tatu-at-datastax: And the reason for using method is to allow overriding; static constants cannot be overridden.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/CreateCollectionCommand.java
amorton: I prefer the first error, our V2 error, but this will impact on the Swagger. 

So how about we update the @Schema / description (which can be improved) at the same time to be "Required name of the new Collection"  and deal with the swagger problem if/when later ?

we may be able to do something with @Constraint/validatedBy but I think @NotEmpty has set that and we cannot override it,

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/DeleteCollectionCommand.java
```diff
@@ -17,11 +15,7 @@
 @Schema(description = "Command that deletes a collection if one exists.")
 @JsonTypeName(CommandName.Names.DELETE_COLLECTION)
 public record DeleteCollectionCommand(
-    @NotNull
-        @Size(min = 1, max = 48)
-        @Pattern(regexp = "[a-zA-Z][a-zA-Z0-9_]*")
-        @Schema(description = "Name of the collection")
-        String name)
+    @NotEmpty @Schema(description = "Name of the collection") String name)
```
amorton, line 18: ok , so back to what is happening in CreateCollection command.... this will use the error we dont want there . Should this be consistent ?

Also, can we change the descripiton for all the drop to be "Required name of the Collection to remove" 

Change Collection above as need, use capital for first letter

src/main/java/io/stargate/sgv2/jsonapi/api/v1/CollectionResource.java
```diff
@@ -181,16 +179,8 @@ public CollectionResource(MeteredCommandProcessor meteredCommandProcessor) {
...
-          @NotNull
-          @Pattern(regexp = "[a-zA-Z][a-zA-Z0-9_]*")
-          @Size(min = 1, max = 48)
-          String keyspace,
-      @PathParam("collection")
-          @NotNull
-          @Pattern(regexp = "[a-zA-Z][a-zA-Z0-9_]*")
-          @Size(min = 1, max = 48)
-          String collection) {
+      @PathParam("keyspace") @NotNull String keyspace,
```
amorton, line 182: we need this here, but should it be @NotEmpty ? e.g. what is the path was "\ \"

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/CommandResolver.java
```diff
@@ -213,4 +216,24 @@ private void getIndexUsageTags(DBLogicalExpression dbLogicalExpression, IndexUsa
...
+   * @throws SchemaException if the name is invalid.
+   */
+  default void validateSchemaName(String name, SchemaObjectNamingRule namingRule) {
+    if (!namingRule.apply(name)) {
+      throw SchemaException.Code.UNSUPPORTED_SCHEMA_NAME.get(
+          Map.of(
+              "schemaType",
+              namingRule.name(),
+              "maxNameLength",
+              String.valueOf(SchemaObjectNamingRule.MAX_NAME_LENGTH),
```
amorton, line 234: using SchemaObjectNamingRule.MAX_NAME_LENGTH feels wrong, why is it static and not on the object ?

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/CommandResolver.java
```diff
@@ -213,4 +216,24 @@ private void getIndexUsageTags(DBLogicalExpression dbLogicalExpression, IndexUsa
...
+              "unsupportedSchemaName",
+              name == null ? "<null>" : name));
```
amorton, line 236: nit: use ErrorTemplate.replaceIfNull and we should move that function to the ErrorFormatters I think

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/CommandResolver.java
```diff
@@ -213,4 +216,24 @@ private void getIndexUsageTags(DBLogicalExpression dbLogicalExpression, IndexUsa
...
   }
+
+  /**
+   * Validate the name of a schema object (e.g., keyspace, table, collection, or index).
+   *
+   * @param name The name to validate.
+   * @param namingRule The naming rule to apply.
+   * @throws SchemaException if the name is invalid.
+   */
+  default void validateSchemaName(String name, SchemaObjectNamingRule namingRule) {
```
amorton, line 227: make this return the name, then it is easier to use

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/CreateCollectionCommandResolver.java
```diff
@@ -58,6 +59,9 @@ public Class<CreateCollectionCommand> getCommandClass() {
   @Override
   public Operation resolveKeyspaceCommand(
       CommandContext<KeyspaceSchemaObject> ctx, CreateCollectionCommand command) {
+
+    validateSchemaName(command.name(), NamingRules.COLLECTION);
```
amorton, line 63: see comments for changeing validateSchemaName to return the name, then assign the return to a var here and re-use the validated name

src/main/java/io/stargate/sgv2/jsonapi/util/naming/SchemaObjectNamingRule.java
amorton: +1 for moving this to  NamingRule. getMaxLength() so it can be different for different rules. e.g. when we get to using this for the field and column names

src/main/java/io/stargate/sgv2/jsonapi/util/naming/SchemaObjectNamingRule.java
```diff
@@ -0,0 +1,28 @@
...
+  public static final Pattern PATTERN_WORD_CHARS = Pattern.compile("\\w+");
```
amorton, line 8: does this need to be public ?

src/main/java/io/stargate/sgv2/jsonapi/util/naming/NamingRule.java
```diff
@@ -0,0 +1,14 @@
+package io.stargate.sgv2.jsonapi.util.naming;
+
+/** Define a naming rule that can be applied to a string */
+public interface NamingRule {
```
amorton, line 4: consider making this abstract class. 

you are using the name in API error messages, and so hidding a constant string for the name of the schema types. 

Replace name with schemaType using SchemaObjectType , and then add an apiName property to the enum. Use that string in the CommandResolver to build the error

src/main/java/io/stargate/sgv2/jsonapi/util/naming/NamingRules.java
```diff
@@ -0,0 +1,10 @@
+package io.stargate.sgv2.jsonapi.util.naming;
+
+public final class NamingRules {
```
amorton, line 3: doc comments

amorton (changes requested):
some teaks / nit picks... also I we can iterate on the NamingRule design

src/test/java/io/stargate/sgv2/jsonapi/api/v1/CollectionResourceIntegrationTest.java
Hazel-Datastax: Right. In Astra/HCD, the naming length is 222. Users can still use CQL to create a name longer than 48. I want to restrict the name to 48 because when I explored the names in Astra and HCD, longer names like 222 characters are very likely to get an error in CQL, even though the keyspace could be created. Also, remember my 223-character keyspace name could crash DB and affect other tenants? So I think shorter name could be safer.

src/main/java/io/stargate/sgv2/jsonapi/util/naming/SchemaObjectNamingRule.java
```diff
@@ -0,0 +1,28 @@
+package io.stargate.sgv2.jsonapi.util.naming;
```
amorton, line 1: move to naming package to be under service.schema

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/CommandResolver.java
amorton: actually, should not need to worry about passing null. just pass the name and it will work with the function above

src/main/java/io/stargate/sgv2/jsonapi/api/v1/KeyspaceResource.java
```diff
@@ -121,12 +119,7 @@ public KeyspaceResource(
                   })))
   @POST
   public Uni<RestResponse<CommandResult>> postCommand(
-      @NotNull @Valid KeyspaceCommand command,
-      @PathParam("keyspace")
-          @NotNull
-          @Pattern(regexp = "[a-zA-Z][a-zA-Z0-9_]*")
-          @Size(min = 1, max = 48)
-          String keyspace) {
+      @NotNull @Valid KeyspaceCommand command, @PathParam("keyspace") @NotNull String keyspace) {
```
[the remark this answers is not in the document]
Hazel-Datastax: Changed to `@NotEmpty`. 
I test `http://127.0.0.1:8181/v1/\ ` and `http://127.0.0.1:8181/v1/ `, both `@NotNull` and `@NotEmpty` return:
```
{
    "errors": [
        {
            "message": "The provided keyspace does not exist: Unknown keyspace ' ', you must create it first",
            "errorCode": "KEYSPACE_DOES_NOT_EXIST",
            "id": "23166766-2714-432e-af88-3c6ba21d6195",
            "family": "REQUEST",
            "title": "The provided keyspace does not exist",
            "scope": "EMPTY"
        }
    ]
}
```

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/DeleteCollectionCommand.java
Hazel-Datastax: For the first question, as we discussed, I created an issue #1857  to handle it later

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/CreateNamespaceCommand.java
```diff
@@ -20,12 +19,9 @@
...
 @JsonTypeName(CommandName.Names.CREATE_NAMESPACE)
 public record CreateNamespaceCommand(
-    @NotNull
-        @Pattern(regexp = "[a-zA-Z][a-zA-Z0-9_]*")
-        @Size(min = 1, max = 48)
-        @Schema(
+    @Schema(
             description =
-                "Name of the namespace. This createNamespace has been deprecated and will be removed in future releases, use createKeyspace instead.",
+                "Required name of the new Namespace. This createNamespace has been deprecated and will be removed in future releases, use createKeyspace instead.",
```
tatu-at-datastax, line 24: Since we are getting rid of term "namespace", should this say "Required name of the new Keyspace" instead? Otherwise it might lead to confusion that "namespace" is something other than alias for "keyspace".

@amorton WDYT?

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/DeleteCollectionCommand.java
```diff
@@ -17,11 +15,7 @@
 @Schema(description = "Command that deletes a collection if one exists.")
 @JsonTypeName(CommandName.Names.DELETE_COLLECTION)
 public record DeleteCollectionCommand(
-    @NotNull
-        @Size(min = 1, max = 48)
-        @Pattern(regexp = "[a-zA-Z][a-zA-Z0-9_]*")
-        @Schema(description = "Name of the collection")
-        String name)
+    @NotEmpty @Schema(description = "Required name of the Collection to remove") String name)
```
tatu-at-datastax, line 18: Did we decide to lave `@NotEmpty` for "drop" commands and not "create" commands -- or why does this remain?

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/CreateCollectionCommandResolver.java
```diff
@@ -58,14 +59,17 @@ public Class<CreateCollectionCommand> getCommandClass() {
   @Override
   public Operation resolveKeyspaceCommand(
       CommandContext<KeyspaceSchemaObject> ctx, CreateCollectionCommand command) {
+
+    var name = validateSchemaName(command.name(), NamingRules.COLLECTION);
```
tatu-at-datastax, line 63: may want to change to `final var` since we won't be changing it.

src/main/java/io/stargate/sgv2/jsonapi/service/schema/naming/NamingRule.java
```diff
@@ -0,0 +1,39 @@
+package io.stargate.sgv2.jsonapi.service.schema.naming;
+
+import io.stargate.sgv2.jsonapi.service.cqldriver.executor.SchemaObject;
+
+/** Define a naming rule that can be applied to a string */
+public abstract class NamingRule {
+  private final String name;
```
tatu-at-datastax, line 7: Javadoc?

src/main/java/io/stargate/sgv2/jsonapi/service/schema/naming/NamingRule.java
```diff
@@ -0,0 +1,39 @@
...
+  private final SchemaObject.SchemaObjectType schemaType;
```
tatu-at-datastax, line 8: I think we need to change this a bit: `name` belongs here (it's the name of thing being validated), but `schemaType` should go up to `SchemaObjectNamingRule`. This because we will have other `NamingRule`s for (Collection) Fields and maybe (Table) Columns.

src/main/java/io/stargate/sgv2/jsonapi/service/schema/naming/NamingRule.java
```diff
@@ -0,0 +1,39 @@
...
+
+  public NamingRule(SchemaObject.SchemaObjectType schemaType, String name) {
+    this.schemaType = schemaType;
+    this.name = name;
+  }
+
+  /**
+   * @return the type of schema object that this rule is applied to
+   */
+  public SchemaObject.SchemaObjectType schemaType() {
```
tatu-at-datastax, line 18: Needs to go up to `SchemaObjectNamingRule` -- if we need this method at all?

src/main/java/io/stargate/sgv2/jsonapi/service/schema/naming/NamingRule.java
```diff
@@ -0,0 +1,39 @@
...
+   */
+  public String name() {
+    return name;
+  }
+  ;
+
+  /**
+   * @return the maximum allowed length for the name.
+   */
+  public abstract int getMaxLength();
```
tatu-at-datastax, line 33: Is this method needed?

src/main/java/io/stargate/sgv2/jsonapi/service/schema/naming/NamingRule.java
```diff
@@ -0,0 +1,39 @@
...
+
+  /**
+   * @return true if the input satisfies this naming rule, false otherwise
+   */
+  public abstract boolean apply(String input);
```
tatu-at-datastax, line 38: Also needs to move up to `SchemaObjectNamingRule` since other `NamingRule`s may (do) require different signature.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/DeleteCollectionCommand.java
tatu-at-datastax: Ok so this is needed since no other validation is performed. All good.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/CreateNamespaceCommand.java
Hazel-Datastax: Changed

tatu-at-datastax (approved):
LGTM

amorton (approved):
Approving, @tatu-at-datastax  is happy I will nit pick with hazel later.

PR #1816, opened 2025-01-06
add props overrides for docker-image-publish.yaml to fix Astra Data API openapi spec

Yuqi-Du:
Data API 1.0.20-DEC-HOT-FIX image surfaces a problem for Data API openapi spec and docs.
For this version, we did NOT create a release for it, instead it is just the HOT-FIX image we published from a branch.
In the release workflow, there is a [props override ](https://github.com/stargate/data-api/blob/main/.github/workflows/release.yaml#L237)which overrides ‘Stargate DATA API’ to ‘Astra DB Serverless Data API’.
What riptano astra-api-docs [github workflow ](https://github.com/riptano/astra-api-docs/blob/main/.github/workflows/generate-pr-dataapi.yml#L66) does is to download the Data API openapi spec from a prod [DB](https://flightdeck.prod.cloud-tools.datastax.com/database/b570392a-846d-4f6a-8d76-4ddda71e1fe3)created by Ronnie Miller. It is prod, so the title is not been override.

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/src/main/resources/application.yaml
+++ b/src/main/resources/application.yaml
@@ -152,7 +152,7 @@ quarkus:
     path: /api/json/openapi
     info-title: ${quarkus.application.name}
     info-version: ${quarkus.application.version:}
-    info-description:
+    info-description: 123
     info-terms-of-service:
     info-contact-name: Stargate
     info-contact-url: https://stargate.io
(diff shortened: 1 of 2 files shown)


src/main/resources/application.yaml
```diff
@@ -152,7 +152,7 @@ quarkus:
     path: /api/json/openapi
     info-title: ${quarkus.application.name}
     info-version: ${quarkus.application.version:}
-    info-description:
+    info-description: 123
     info-terms-of-service:
```
Hazel-Datastax, line 156: Mistakenly add this `123`?

tatu-at-datastax: I am guessing it was for testing?

PR #1820, opened 2025-01-09
Fix #1812: create new `ErrorCodeV1.INVALID_REQUEST_BAD_SYNTAX`, map non-parsing errors to that

### LINKED ISSUE #1812: Invalid createIndex structure causes "Request invalid, cannot parse as JSON" error even though request is valid JSON
When sending the following `createIndex` command to Data API v1.0.20:

```
{
  "createIndex": {
    "name": {
      "name": 1
    },
    "definition": {
      "column": {
        "background": true
      },
      "options": {}
    },
    "options": {}
  }
}
```

I get back the following error:

```
HTTP error (400): {"errors":[{"message":"Request invalid, cannot parse as JSON: Cannot deserialize value of type `java.lang.String` from Object value (token `JsonToken.START_OBJECT`)\n at [Source: (ByteArrayInputStream); line: 1, column: 24] (through reference chain: io.stargate.sgv2.jsonapi.api.model.command.impl.CreateIndexCommand[\"name\"])","errorCode":"INVALID_REQUEST_NOT_JSON","id":"4353e100-2914-4af1-85da-7ed6739999f6","family":"REQUEST","title":"Request invalid, cannot parse as JSON","scope":"EMPTY"}]}
```

Admittedly the error message does contain "Cannot deserialize value of type `java.lang.String` from Object value (token `JsonToken.START_OBJECT`)", which is more correct (`name` should be string not object), but the "cannot parse as JSON" error title is misleading because the request is valid JSON, just not correct command syntax.

tatu-at-datastax:
**What this PR does**:

Improves error handling wrt #1812

Fixes #1812

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/table/definition/TableDefinitionDesc.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/table/definition/TableDefinitionDesc.java
@@ -2,7 +2,6 @@ package io.stargate.sgv2.jsonapi.api.model.command.table.definition;
 
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
-import com.fasterxml.jackson.annotation.JsonPropertyOrder;
 import io.stargate.sgv2.jsonapi.config.constants.TableDescConstants;
 import jakarta.validation.Valid;
 import org.eclipse.microprofile.openapi.annotations.enums.SchemaType;
@@ -14,8 +13,6 @@ import org.eclipse.microprofile.openapi.annotations.media.Schema;
  *
  * <p>Has the columns and the primary key definition, not the name of the table.
  */
-@JsonPropertyOrder({"name", "definition"})
-@JsonInclude(JsonInclude.Include.NON_NULL)
 public record TableDefinitionDesc(
     @Valid
         @Schema(description = "API table columns definitions", type = SchemaType.OBJECT)
--- a/src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCodeV1.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCodeV1.java
@@ -47,7 +47,7 @@ public enum ErrorCodeV1 {
   /** note: Only used by EmbeddingGateway */
   INVALID_REQUEST("Request not supported by the data store"),
 
-  INVALID_REQUEST_BAD_SYNTAX("Request invalid, JSON syntax incorrect"),
+  INVALID_REQUEST_STRUCTURE_MISMATCH("Request invalid, mismatching JSON structure"),
 
   INVALID_REQUEST_NOT_JSON("Request invalid, cannot parse as JSON"),
 
--- a/src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCodeV1.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCodeV1.java
@@ -47,6 +47,8 @@ public enum ErrorCodeV1 {
   /** note: Only used by EmbeddingGateway */
   INVALID_REQUEST("Request not supported by the data store"),
 
+  INVALID_REQUEST_BAD_SYNTAX("Request invalid, JSON syntax incorrect"),
+
   INVALID_REQUEST_NOT_JSON("Request invalid, cannot parse as JSON"),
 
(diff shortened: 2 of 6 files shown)


src/main/java/io/stargate/sgv2/jsonapi/api/model/command/table/definition/TableDefinitionDesc.java
```diff
@@ -14,8 +13,6 @@
  *
  * <p>Has the columns and the primary key definition, not the name of the table.
  */
-@JsonPropertyOrder({"name", "definition"})
```
tatu-at-datastax, line 17: Not needed as we are not serializing these definitions (but if we did, had wrong column names :) ).

src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCodeV1.java
```diff
@@ -47,6 +47,8 @@ public enum ErrorCodeV1 {
   /** note: Only used by EmbeddingGateway */
   INVALID_REQUEST("Request not supported by the data store"),
 
+  INVALID_REQUEST_BAD_SYNTAX("Request invalid, JSON syntax incorrect"),
```
tatu-at-datastax, line 50: Adding as old-style error code for minimal change right now, since code is called by `ThrowableToErrorMapper` with limited access.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/tables/CreateTableIntegrationTest.java
```diff
@@ -322,10 +322,10 @@ private static Stream<Arguments> allTableData() {
                           """,
                   "invalidPartitionSortOrderingValueTable",
                   true,
-                  ErrorCodeV1.INVALID_REQUEST_NOT_JSON.name(),
+                  ErrorCodeV1.INVALID_REQUEST_BAD_SYNTAX.name(),
```
tatu-at-datastax, line 325: Gee, figuring out how to fix this test was not easy: not a big fan of huge matrix tests because IDE makes it difficult to find out where the actual test data originates in. :-(

Hazel-Datastax (approved):
LGTM.

I saw your question in the [issue](https://github.com/stargate/data-api/issues/1812) - whether to change HTTP response code from 400 to 200. Good question, both 400 and 200 would make sense to me but I personally think maybe 200 is better, since most errors are 200 and “bad json syntax” sounds more like the user doesn’t follow our definition.

Want to hear others’ thoughts

src/test/java/io/stargate/sgv2/jsonapi/api/v1/tables/CreateTableIndexIntegrationTest.java
```diff
@@ -483,6 +485,31 @@ public void createIndexWithUnknownIndexType() {
...
+                            "definition": {
+                              "column": {
+                                "background": true
+                              },
+                            }
+                        }
+                        """)
+          .hasSingleApiError(
+              ErrorCodeV1.INVALID_REQUEST_BAD_SYNTAX,
+              "Request invalid, JSON syntax incorrect: underlying problem");
```
vkarpov15, line 511: I'm not a fan of the "JSON syntax incorrect" bit because that implies that the request is invalid JSON, which isn't true. I would recommend changing to "invalid Data API request syntax" or something similar to imply that the JSON itself is valid, just the command syntax is not correct.

tatu-at-datastax: Fundamentally it is not valid message, structurally. Maybe "mismatching"?

"mismatching JSON structure"?

I am looking for better wording and open to suggestions.

src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCodeV1.java
```diff
@@ -47,6 +47,8 @@ public enum ErrorCodeV1 {
   /** note: Only used by EmbeddingGateway */
   INVALID_REQUEST("Request not supported by the data store"),
 
+  INVALID_REQUEST_STRUCTURE_MISMATCH("Request invalid, mismatching JSON structure"),
```
tatu-at-datastax, line 50: @vkarpov15 Changed error code, wording, is this any closer to being acceptable? I am happy to keep on tweaking it until we get it right.

vkarpov15: I think this message is good enough, thanks for listening :+1:

PR #1829, opened 2025-01-17
[Deprecate] CreateIndexCommand, entries/keys/values for map/set/list collection columns

Yuqi-Du:
CreateIndexCommand entries/keys/values for map/set/list
 
There are couples of rules to get familiar if reviewing

- Data API createIndex table command has defaults indexFunction for map/set/list, defaults are entries(map), values(set), values(list). In a sense, keys and values for map column are not supported,.
- Since Data API does not support frozen map/set/list table creation, so FULL index will also not be supported.
- Only text and ascii datatypes can be analyzed.

**Checklist**
- [x] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/service/schema/tables/ApiRegularIndex.java
```diff
@@ -112,59 +83,183 @@ public ApiRegularIndex create(
...
+        }
+      }
+
       // we could check if there is an existing index but that is a race condition, we will need to
       // catch it if it fails
-      // regular indexes can only be on primitive. Adding indexes on maps, sets, lists will come
-      // later.
-      if (!apiColumnDef.type().isPrimitive()) {
-        throw SchemaException.Code.UNSUPPORTED_INDEXING_FOR_DATA_TYPES.get(
+      throw SchemaException.Code.UNSUPPORTED_INDEXING_FOR_DATA_TYPES.get(
```
amorton, line 105: the error message for UNSUPPORTED_INDEXING_FOR_DATA_TYPES says the data type is frozen, but there are no checks I can see for that.

src/main/java/io/stargate/sgv2/jsonapi/service/schema/tables/ApiRegularIndex.java
```diff
@@ -112,59 +83,183 @@ public ApiRegularIndex create(
...
+          indexOptions, optionsDesc == null ? null : optionsDesc.normalize());
+    }
+
+    /**
+     * Only text and ascii datatypes can be analyzed. <br>
+     * Only text and ascii in the collection datatype can be analyzed. It works for values(list),
+     * values(set), keys(map), values(map). Note, not for entries(map). <br>
+     * The method will error out as UNSUPPORTED_TEXT_ANALYSIS_FOR_DATA_TYPES if rules are violated.
+     */
+    private void validateAnalyzableDatatypes(
```
amorton, line 210: nit: none of these indexes are analyzed indexes, that term is only used the analyzed text indexes that specify the index_analyzer property

src/main/java/io/stargate/sgv2/jsonapi/service/schema/tables/ApiRegularIndex.java
```diff
@@ -112,59 +83,183 @@ public ApiRegularIndex create(
...
+      if (optionsDesc == null) {
+        // no need to analyze if user does not specify any RegularIndexDescOptions
+        return;
+      }
+
+      // Primitive type
+      ApiTypeName analyzedColumnType = apiColumnDef.type().typeName();
+
+      // Map collection type
+      if (apiColumnDef.type() instanceof ApiMapType) {
```
amorton, line 225: check the typeName

not sure why this errors for any map type? It is called from createApiIndexForCollection() 

we need to clarify what the rules for the options are. I thought none of the three text options would be allowed for set / list / map

src/main/java/io/stargate/sgv2/jsonapi/service/schema/tables/ApiRegularIndex.java
```diff
@@ -112,59 +83,183 @@ public ApiRegularIndex create(
...
+
+      return new ApiRegularIndex(
+          indexIdentifier, ApiIndexType.COLLECTION, columnIdentifier, indexOptions);
+    }
+
+    /**
+     * Populate the indexOptions map by pulling values from optionsDesc. indexOptions will be used
+     * to append the options in createIndex cql statement.
+     */
+    private void populateIndexOptionsMap(
```
amorton, line 193: do we want to do this for indexes on collections ?

src/main/java/io/stargate/sgv2/jsonapi/service/schema/tables/ApiRegularIndex.java
```diff
@@ -112,59 +83,183 @@ public ApiRegularIndex create(
...
-        Options.CASE_SENSITIVE.putOrDefaultStringable(
-            indexOptions, optionsDesc == null ? null : optionsDesc.caseSensitive());
-        Options.NORMALIZE.putOrDefaultStringable(
-            indexOptions, optionsDesc == null ? null : optionsDesc.normalize());
+      // Rule 4
+      String indexFunction =
+          apiColumnDef.type().typeName() == ApiTypeName.MAP
+              ? ApiIndexFunction.ENTRIES.cqlFunction
+              : ApiIndexFunction.VALUES.cqlFunction;
+      indexOptions.put("indexFunction", indexFunction);
```
amorton, line 183: -1 on this, the indexOptions are the raw options for CQL is this is not one of those.

amorton (changes requested):

amorton (approved):

amorton (changes requested):
-1 we need to get this merged which removes the idea of the COLLECTION index type https://github.com/stargate/data-api/pull/1831
*  I am not sure the rules for what you can and cannot do with an index on a collection are clear 
* needs some code and error tidy

src/main/java/io/stargate/sgv2/jsonapi/service/schema/tables/ApiRegularIndex.java
Yuqi-Du: then we have a mis-communication, that is what I thought initially, but you mentioned adding it to options

PR #1831, opened 2025-01-21
Additional tidy and fixes for indexType in list indexes

### LINKED ISSUE #1696: Consider adding an "index type" field next to the index definitions in listIndexes response
Readers have to rely on the presence of specific fields within `definition` to figure out if an index is vector, "regular" (and in the future there'll be more types).
The suggestion is also to mitigate the asymmetry with the various createXYZIndex commands and add a new field in the `listIndexes` response, such as:

```
{
    "status": {
        "indexes": [
            {
                "name": "idx_table_find_p_int",
                "indexType": "justVanilla",        <===== THIS!
                "definition": {
                    "column": "p_int",
                    "options": {}
                }
            },
            {
                "name": "idx_table_find_p_vector",
                "indexType": "vector",             <===== THIS!
                "definition": {
                    "column": "p_vector",
                    "options": {
                        "metric": "cosine",
                        "sourceModel": "other"
                    }
                }
            }
        ]
    }
}
```

Note this is not inside "definition", to preserve identity between what gets in (createIndex) and what goes out (listIndexes)

amorton:
* removed the COLLECTION index type, indexType should map to the createIndex command to use. 
* code tidy , improved the indexing code from the rushed versions for preview. Should not change functionality. 
* tests for the logic to determine the IndexType from the CQL index def.

Fixes #1696

**Checklist**
- [ ] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/api/model/command/table/IndexDesc.java
```diff
@@ -12,15 +12,19 @@
  * <p>Does not include the command options, because we don't want them in the list indexes command.
  * Needs @JsonProperty annotation because it is not using bean accessor methods.
  */
-@JsonPropertyOrder({TableDescConstants.IndexDesc.NAME, TableDescConstants.IndexDesc.DEFINITION})
+@JsonPropertyOrder({
```
tatu-at-datastax, line 15: This is only needed if we serialize type -- not used on deserialization.
So could perhaps be removed?

amorton: This is the interface used to describe indexes by the `listIndexes` command

vkarpov15 (commented):
LGTM

src/main/java/io/stargate/sgv2/jsonapi/service/schema/tables/ApiIndexType.java
```diff
@@ -2,82 +2,105 @@
...
+    return apiName;
   }
 
   /**
-   * Get the ApiIndexType from the index type name. If the type name is not known, then null is
-   * returned.
+   * Get the ApiIndexType by the API name.
+   *
+   * @param apiName The name of the index type from the public API. If the name is not found, it
+   *     will return the defaultType.
```
Hazel-Datastax, line 58: `If the name is not found, it will return the defaultType.` What is the defaultType? It seems it will return null if no name is found

src/main/java/io/stargate/sgv2/jsonapi/service/schema/tables/ApiIndexType.java
```diff
@@ -2,82 +2,105 @@
...
-  private static final List<String> ALL_TYPE_NAMES =
-      Arrays.stream(values()).map(ApiIndexType::indexTypeName).toList();
-  private static final Map<String, ApiIndexType> BY_API_NAME = new HashMap<>();
+  private static final Map<String, ApiIndexType> BY_API_NAME;
 
   static {
+    var map = new HashMap<String, ApiIndexType>();
     for (ApiIndexType type : ApiIndexType.values()) {
-      BY_API_NAME.put(type.indexTypeName, type);
+      map.put(type.apiName.toLowerCase(), type);
```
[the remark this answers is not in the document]
amorton: What is someone adds another indexType enum value and does not make them lower case ? This is an easy way to make sure the string comparision is case insensitive

src/main/java/io/stargate/sgv2/jsonapi/service/schema/tables/ApiIndexType.java
amorton: There is default value for an enum like this, if the apiName is not found then the name is not known / incorrect. If we were to default it then any apiname can be used for the default value

Hazel-Datastax (approved):
LGTM

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/CreateVectorIndexCommand.java
```diff
@@ -1,43 +1,63 @@
...
     implements CollectionCommand, IndexCreationCommand {
 
-  // This is index command option irrespective of column definition.
+  /** Options for the command */
   public record CreateVectorIndexCommandOptions(
       @Nullable
           @Schema(
-              description = "Flag to ignore if index already exists",
+              description = "True to ignore if index with the same name already exists.",
               defaultValue = "false",
```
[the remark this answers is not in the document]
amorton: Good catch, fixed

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/table/definition/indexes/VectorIndexDefinitionDesc.java
```diff
@@ -1,46 +1,63 @@
...
       @Nullable
-          @Size(min = 1, max = 48)
-          @Pattern(regexp = "[a-zA-Z][a-zA-Z0-9_]*")
-          @Schema(description = "Model name used to generate the embeddings.")
+          @Schema(
+              description =
+                  "Optional name of the model used to generate the embeddings, indexes can be optimized if the model name is known. Supported values are: "
+                      + EmbeddingSourceModel.ApiConstants.ALL)
           @JsonInclude(JsonInclude.Include.NON_NULL)
+          @JsonProperty(VectorConstants.VectorColumn.METRIC)
```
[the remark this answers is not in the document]
amorton: yeah, that broke some test :) fixed now.

PR #1833, opened 2025-01-21
Fix #1578: deleteOne of row with Blob in PK

### LINKED ISSUE #1578: Cannot deleteOne when a blob is part of the PK (Tables, v1.0.18)
When trying a deleteOne on a Table with this in the primary key:

```
...
                "p_blob": {
                    "type": "blob"
                },
...
```

the deleteOne looking like this:

```
{
    "deleteOne": {
        "filter": {
            "p_ascii": "abc",
            "p_bigint": 10000,
            "p_blob": {
                "$binary": "q83vASNFZ4k="
            },
            "p_boolean": false
       }
    }
}
```

the command fails with HTTP 200 and:

```
{
    "errors": [
        {
            "message": "Server internal error: Unrecognized JsonExtensionType: BINARY",
            "errorCode": "SERVER_INTERNAL_ERROR"
        }
    ]
}
```

Note: if trying the relaxed repr. in the `filter`, i.e. `... "p_blob": "q83vASNFZ4k=", ...`, the error becomes _Server failed: root cause: (io.stargate.sgv2.jsonapi.exception.catchable.ToCQLCodecException) Error trying to convert to targetCQLType `BLOB` from value.class `java.lang.String`, value \"q83vASNFZ4k=\". Root cause: no codec matching value type_ instead.

tatu-at-datastax:
**What this PR does**:

Fixes problem with deleting rows that have a Blob in primary key

Fixes #1578

**Checklist**
- [ ] Changes manually tested


tatu-at-datastax:
Ok: fix included -- should be extensible for other EJSON types.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/filters/table/codecs/BinaryTableFilter.java
```diff
@@ -0,0 +1,11 @@
+package io.stargate.sgv2.jsonapi.service.operation.filters.table.codecs;
```
Hazel-Datastax, line 1: nit: I saw other `TableFilter`s are in `package io.stargate.sgv2.jsonapi.service.operation.filters.table`; maybe move it to there?

Hazel-Datastax (approved):
LGTM

src/main/java/io/stargate/sgv2/jsonapi/service/operation/filters/table/codecs/BinaryTableFilter.java
tatu-at-datastax: Good catch! Will move. Thank you!

tatu-at-datastax: Fixed, waiting for CI.

Issue #1532, opened 2024-10-12
"no codec matching value type" error when using `$in`


vkarpov15:
I ran into this while testing stargate-mongoose-sample-apps with #1448. I have the following users collection:

```
createTable: {
        name: 'users',
        definition: {
          primaryKey: '_id',
          columns: {
            _id: { type: 'text' },
            email: { type: 'text' },
            firstName: { type: 'text' },
            lastName: { type: 'text' }
          }
        }
      }
```

Sending the following command:

```
{"find":{"filter":{"_id":{"$in":["670ad2783a200d3ee3b16a65"]}}}}
```

returns the following error:

```
[{"message":"Server failed: root cause: (java.lang.RuntimeException) io.stargate.sgv2.jsonapi.exception.catchable.ToCQLCodecException: Error trying to convert to targetCQLType `TEXT` from value.class `io.stargate.sgv2.jsonapi.service.shredding.collections.DocumentId$StringId`, value '670ad2783a200d3ee3b16a65'. Root cause: no codec matching value type","errorCode":"SERVER_UNHANDLED_ERROR"}]
```

If I remove the `$in` and just do `{"find":{"filter":{"_id":"670ad2783a200d3ee3b16a65"}}}` then I get the expected result back, no error. Is there some sort of special handling for DocumentId when the column in `$in` is the primary key? I've tried using `$in` on other fields, and those seem to work correctly.

Yuqi-Du:
I can not replicate this on main.
@tatu-at-datastax 
Tatu, any idea why does the source type is DocumentID?
Do we have DocumentID support for table now?

tatu-at-datastax:
@Yuqi-Du  So, `DocumentId` is due to hard-coded check for field named `_id`: so if a Table Column is named that way, special type is used. This is because with Collections this is how things work: it is one of few specifically named things Collections recognize. And since filter construction code is shared, same logic is (unfortunately) applied to Tables too.

I think I had to add some work-arounds to make things work for main/regular `_id` types (text, numbers), but I think that was only for `$eq` case, probably something similar needed for `$in`.
Although it is also possible bigger changes might be needed.

tatu-at-datastax:
I think this may have been fixed since then, @vkarpov15, could you re-test? If it still occurs, we'll take another look.

vkarpov15:
@tatu-at-datastax I checked with data-api v1.0.20, I still get the same error:

```
[...]

PR #1836, opened 2025-01-24
Fix #1835: refactor `FilterClause` deserialization to give access to `CommandContext`

### LINKED ISSUE #1835: Refactor creation of `FilterClause` to be able to pass, use `CommandContext`
Currently `FilterClause` is directly decode as part of Quarkus endpoint resolution, before Data API gets control of command processing. This is problematic as parts of handling need to be specialized wrt whether filtering is for Collections or Tables.

It is desirable to split this in two parts:

1. Light-weight decoding of JSON into `JsonNode`
2. Actual construction of `FilterClause` with validation

Note that this a pre-req for solving #1532 but there are likely other issues that will be resolved (or made possible/easier to resolve).

tatu-at-datastax:
**What this PR does**:

Refactors the way `FilterClause` deserialization works to be in 2 phases:

1. Simple JSON parsing into `JsonNode` (by Quarkus) without building or validation
2. Actual POJO construction, validation, passing `CommandContext`

This is to enable fixing #1532 (and similar) where FilterClause handling differs b/w Collections and Tables.

Fixes #1835

**Checklist**
- [x] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/api/model/command/CommandContext.java
```diff
@@ -23,7 +24,8 @@ public record CommandContext<T extends SchemaObject>(
     EmbeddingProvider embeddingProvider,
     String commandName,
     JsonProcessingMetricsReporter jsonProcessingMetricsReporter,
-    ApiFeatures apiFeatures) {
+    ApiFeatures apiFeatures,
+    OperationsConfig operationsConfig) {
```
tatu-at-datastax, line 28: OperationsConfig contains limits needed by Shredder for validation (or, eventually separate validation component); not yet available efficiently from CommandContext before changes.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/Filterable.java
```diff
@@ -1,10 +1,16 @@
...
 
 import io.stargate.sgv2.jsonapi.api.model.command.clause.filter.FilterClause;
+import io.stargate.sgv2.jsonapi.api.model.command.clause.filter.FilterSpec;
 
 /*
  * All the commands that needs FilterClause will have to implement this.
  */
 public interface Filterable {
-  FilterClause filterClause();
+  FilterSpec filterSpec();
```
tatu-at-datastax, line 10: To separate out low-level JSON decoding and higher-level deserialization/marshalling, new "FilterSpec" (which only has raw JSON as `JsonNode`) added: `FilterClause` remains the fully-processed definition to use, produced from `FilterSpec`

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/Filterable.java
```diff
@@ -1,10 +1,16 @@
...
+
+  default FilterClause filterClause(CommandContext<?> ctx) {
```
tatu-at-datastax, line 12: Smallest changes to existing non-test-code is to add this default method here. Also takes care of issue of `null` (missing or explicit null) for `filter` in JSON.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/filter/FilterSpec.java
```diff
@@ -0,0 +1,46 @@
...
+  /** The JSON value that specifies the filter clause. */
+  private final JsonNode filterJson;
+
+  /** Lazily built POJO representation of the filter clause. */
+  private FilterClause filterClause;
+
+  /**
+   * To deserialize the whole JSON value, need to ensure DELEGATING mode (instead of PROPERTIES).
+   */
+  @JsonCreator(mode = JsonCreator.Mode.DELEGATING)
```
tatu-at-datastax, line 25: Doing this allows as to tell Quarkus we want a `FilterSpec`, and Q (and Jackson) to decode as `JsonNode`, construct light-weight wrapper for it, to be later processed into `FilterClause` when we have `CommandContext` with necessary configuration info (specifically, whether it's Collection or Table that we operate on).

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/deserializers/FilterClauseDeserializer.java
```diff
@@ -2,57 +2,51 @@
...
+ *
+ * <p>NOTE: is NOT a Jackson {@code JsonDeserializer}, not invoked through Jackson (or Quarkus).
  *
  * <p>TIDY: this class has a lot of string constants for filter operations that we have defined as
  * constants elsewhere
  */
-public class FilterClauseDeserializer extends StdDeserializer<FilterClause> {
-  private final OperationsConfig operationsConfig;
+public class FilterClauseDeserializer {
+  public FilterClauseDeserializer() {}
```
Hazel-Datastax, line 30: Since we don't have any other constructors, can we remove this default one?

tatu-at-datastax: We can, default one will then be generated. Sure, will remove.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/FindCommand.java
```diff
@@ -20,7 +20,7 @@
 @JsonTypeName(CommandName.Names.FIND)
 @CheckFindOption
 public record FindCommand(
-    @Valid @JsonProperty("filter") FilterClause filterClause,
+    @Valid @JsonProperty("filter") FilterSpec filterSpec,
```
amorton, line 23: From discussions, try the @Schema.implementation property so that swagger generation works ?

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/filter/FilterClause.java
```diff
@@ -17,7 +15,6 @@
 import org.eclipse.microprofile.openapi.annotations.enums.SchemaType;
 import org.eclipse.microprofile.openapi.annotations.media.Schema;
 
-@JsonDeserialize(using = FilterClauseDeserializer.class)
```
amorton, line 20: When we make this change, will the SchemaValidatable and validate* functions from this class move to the deserializer ? They were a not good idea from me.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/filter/FilterSpec.java
```diff
@@ -0,0 +1,55 @@
...
+import com.fasterxml.jackson.databind.JsonNode;
+import io.stargate.sgv2.jsonapi.api.model.command.CommandContext;
+import io.stargate.sgv2.jsonapi.api.model.command.deserializers.FilterClauseDeserializer;
+import java.util.Objects;
+import org.eclipse.microprofile.openapi.annotations.enums.SchemaType;
+import org.eclipse.microprofile.openapi.annotations.media.Schema;
+
+/**
+ * Intermediate lightly-processed container for JSON that specifies a {@link FilterClause}, and
+ * allows for lazy deserialization into a {@link FilterClause}.
```
amorton, line 13: From discussions, is there an abstraction here we can use so this class is re-usable for different Objects ?

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/CommandContext.java
```diff
@@ -23,7 +24,8 @@ public record CommandContext<T extends SchemaObject>(
     EmbeddingProvider embeddingProvider,
     String commandName,
     JsonProcessingMetricsReporter jsonProcessingMetricsReporter,
-    ApiFeatures apiFeatures) {
+    ApiFeatures apiFeatures,
```
amorton, line 27: +1 for sneaking this in here, will make it easier to remove all the injection of it

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/Filterable.java
```diff
@@ -1,10 +1,18 @@
...
 
 import io.stargate.sgv2.jsonapi.api.model.command.clause.filter.FilterClause;
+import io.stargate.sgv2.jsonapi.api.model.command.clause.filter.FilterSpec;
 
 /*
  * All the commands that needs FilterClause will have to implement this.
  */
 public interface Filterable {
-  FilterClause filterClause();
+  /** Accessor for the filter specification in its intermediate for */
```
amorton, line 10: I know I over engineer things :) but I wonder if there is a base interface here because this code looks like it will repeat

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/matcher/FilterMatchRule.java
```diff
@@ -30,7 +30,7 @@ public record FilterMatchRule<T extends Command & Filterable>(
   @Override
   public Optional<DBLogicalExpression> apply(CommandContext commandContext, T command) {
     return matcher
-        .apply(command)
+        .apply(commandContext, command)
```
amorton, line 33: Hey stupid idea time - do we need an object for CommandWithContext ?

amorton (changes requested):
Looks like a good start, lets see where this goes :)

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/FindCommand.java
tatu-at-datastax: I added `@Schema` on `FilterSpec` itself, with `implementation = FilterClause.class` so that should work the same as before:

```
@Schema(
    type = SchemaType.OBJECT,
    implementation = FilterClause.class,
    example =
        """
                      """)
```
 
although I think the only useful part is probably `example` given that `FilterClause` structure does not (and cannot) match structurally JSON (because it's mostly wrapper around `LogicalExpression` etc).

But at least examples should be included in generated Swagger.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/filter/FilterClause.java
tatu-at-datastax: Eventually yes something like that (or maybe exactly that).

Given that it looks like there is no validation for `FilterClause`, and that empty methods are called only from `FilterResolver`, one possibility would be to remove use of `SchemaValitable` at this point, and add back when we realize what to validate.
But it would follow from schema-specific deserializer split I think.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/filter/FilterSpec.java
tatu-at-datastax: It is definitely doable. The question is if there is specific benefit from common abstraction. So we could have something like

    DeferredValue<T>

with method like

    T resolve(CommandContext<?> ctx)

(or such)
but I am not sure what immediate value there'd be for handling these things (SortClause would be next immediate choice) in common way using super type.

But I'll think about this a bit, in case something comes up.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/CommandContext.java
tatu-at-datastax: That's the idea yes. I really dislike the part of it being next to impossible to create test version of `OperationsConfig` (it being interface with magically generated implementation that merges in defaults from annotations).
But at least we can isolate magic, and eventually replace it.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/Filterable.java
tatu-at-datastax: Yeah, there's definitely a pattern of some kind. If we had "Deferred value" type (discussed above), we could have uber-interface here I think.
But not sure if that would help make code more readable.

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/matcher/FilterMatchRule.java
tatu-at-datastax: Hmmh. That might be a good idea, not sure. But it'd be more... "Command and context" wrapper, I think?
Or we could make either Command refer to context (needs to assign after construction), or vice versa (context referencing command).

Main benefit being we can pass just one thing around.

Good question, not sure what the answer is. :)

vkarpov15 (commented):
Code changes look reasonable, but just confirming that this PR doesn't fix #1532, it paves the way for future work to fix #1532?

tatu-at-datastax:
> Code changes look reasonable, but just confirming that this PR doesn't fix #1532, it paves the way for future work to fix #1532?

Exactly. I am still working on this a bit, to make context-specific deserialization easier to extract but yes that is the idea.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/JsonDefinition.java
```diff
@@ -0,0 +1,25 @@
...
+
+import com.fasterxml.jackson.databind.JsonNode;
+
+/**
+ * Value type enclosing JSON-decoded representation of something to resolve into fully definition,
+ * such as filter or sort clause. Needed to let Quarkus/Jackson decode textual JSON into
+ * intermediate form ({@link JsonNode}) from which actual deserialization and validation can be
+ * deferred until we have context we need
+ */
+public abstract class JsonDefinition {
```
tatu-at-datastax, line 11: @amorton Adding this as shareable base definition for `FilterSpec` (or whatever we call it) and similar intermediate sources. The reason for base class over interface is to allow field for actual `JsonNode`. And the reason to require sub-class is to allow/require Swagger annotation(s) (`@Schema`); also helps slightly with documentation/type-safety.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/CommandContext.java
```diff
@@ -224,4 +274,19 @@ public SmallRyeConfig getConfig() {
...
+    }
+    // During tests:
+    return getConfig(OperationsConfig.class);
+  }
+
+  public FilterClause resolveFilterClause(FilterSpec filter) {
+    if (filter == null) {
+      return null;
+    }
+    return FilterClauseDeserializer.INSTANCE.buildFilterClause(this, filter.json());
```
tatu-at-datastax, line 290: @amorton Only temporarily hard-coded, will branch out to schema-specific deserializers next.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/Filterable.java
tatu-at-datastax: Added that new `JsonDefinition` type, but I think it's difficult to generalize this tag interface, since many commands have both `Filterable` and `Sortable`, so have to avoid name clash (for `JsonDefinition.json()`).
So I think this needs to remain separate from `Sortable`.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/filter/FilterClause.java
tatu-at-datastax: Correction: there IS actual validation for `FilterClause` so cannot just drop the call.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/filter/FilterSpec.java
tatu-at-datastax: Refactored to remove all processing logic from here, move to `CommandContext`. Also based on `JsonDefinition`, so this wrapper only needed for annotations, basically.

tatu-at-datastax:
Rats. Some code somewhere is relying on an odd side-effect of old code: basically `FilterClause` deserialized for `Command` is always same object and this is relied on by `FilterMatchRules` or such... need to think of how to de-couple things OR how to avoid re-deserializing `FilterClause` from `FilterSpec`.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/filter/FilterClause.java
tatu-at-datastax: Use of `SchemaValitable` removed from `FilterClause` handling: still remains for `SortClause` (to be removed).

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/FindCommand.java
amorton: cool, thinking ahead to this being a pattern it maybe useful - though most of the time we may be using this when it is as complex as the find

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/filter/FilterOperators.java
```diff
@@ -0,0 +1,31 @@
...
+      addComparisonOperator(filterOperator);
+    }
+    for (FilterOperator filterOperator : ElementComparisonOperator.values()) {
+      addComparisonOperator(filterOperator);
+    }
+    for (FilterOperator filterOperator : ArrayComparisonOperator.values()) {
+      addComparisonOperator(filterOperator);
+    }
+    // This should not be supported from outside
+    operatorMap.remove(ArrayComparisonOperator.NOTANY.getOperator());
```
amorton, line 21: nit: super pedantic, set the static map to be Immutable with Map.copy() I know we limit access.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/filter/FilterOperators.java
```diff
@@ -0,0 +1,31 @@
+package io.stargate.sgv2.jsonapi.api.model.command.clause.filter;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/** Helper class for {@link FilterOperator} lookups. */
+public class FilterOperators {
```
amorton, line 7: nit: prevent construction as abstract and private ctor

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/Filterable.java
```diff
@@ -1,10 +1,17 @@
...
+  FilterSpec filterSpec();
+
+  default FilterClause filterClause(CommandContext<?> ctx) {
+    FilterSpec spec = filterSpec();
+    return (spec == null) ? null : spec.toFilterClause(ctx);
```
amorton, line 15: also in other comments, can we make this method always return a filter ?

src/main/java/io/stargate/sgv2/jsonapi/api/v1/CollectionResource.java
```diff
@@ -79,6 +80,8 @@ public class CollectionResource {
 
   @Inject FeaturesConfig apiFeatureConfig;
 
+  @Inject OperationsConfig operationsConfig;
+
```
amorton, line 84: +1 : can this and the   apiFeatureConfig be private ? Same on the other resources I think

src/main/java/io/stargate/sgv2/jsonapi/api/v1/GeneralResource.java
```diff
@@ -36,14 +37,15 @@
 @SecurityRequirement(name = OpenApiConstants.SecuritySchemes.TOKEN)
 @Tag(ref = "General")
 public class GeneralResource {
+  public static final String BASE_PATH = "/v1";
```
amorton, line 40: nit: can we make the Keyspace and Collection resource use this constant to build it's path?

src/main/java/io/stargate/sgv2/jsonapi/api/v1/KeyspaceResource.java
```diff
@@ -55,13 +56,20 @@ public class KeyspaceResource {
...
+  @Inject private DataApiRequestInfo dataApiRequestInfo;
 
   @Inject
-  public KeyspaceResource(MeteredCommandProcessor meteredCommandProcessor) {
+  public KeyspaceResource(
+      MeteredCommandProcessor meteredCommandProcessor,
+      FeaturesConfig apiFeatureConfig,
+      OperationsConfig operationsConfig) {
     this.meteredCommandProcessor = meteredCommandProcessor;
+    this.apiFeatureConfig = apiFeatureConfig;
```
amorton, line 71: nit: question, the other resources using @Inject on the properties not the resources, is it a big change to make them all work the same ?

src/main/java/io/stargate/sgv2/jsonapi/service/operation/filters/collection/CollectionFilter.java
```diff
@@ -123,7 +123,7 @@ protected static JsonNode toJsonNode(JsonNodeFactory nodeFactory, List<Object> l
 
   protected static JsonNode toJsonNode(JsonNodeFactory nodeFactory, Map<String, Object> mapValues) {
     final ObjectNode objectNode = nodeFactory.objectNode();
-    mapValues.forEach((k, v) -> objectNode.put(k, toJsonNode(nodeFactory, v)));
+    mapValues.forEach((k, v) -> objectNode.set(k, toJsonNode(nodeFactory, v)));
```
amorton, line 126: Checking this change from objectNode.put() to objectNode.set() was deliberate

src/main/java/io/stargate/sgv2/jsonapi/service/processor/MeteredCommandProcessor.java
```diff
@@ -258,10 +259,14 @@ private <T extends SchemaObject> Tags getCustomTags(
...
 
-  private JsonApiMetricsConfig.SortType getVectorTypeTag(Command command) {
+  private JsonApiMetricsConfig.SortType getVectorTypeTag(
+      CommandContext<?> commandContext, Command command) {
     int filterCount = 0;
-    if (command instanceof Filterable fc && fc.filterClause() != null) {
-      filterCount = fc.filterClause().logicalExpression().getTotalComparisonExpressionCount();
+    if (command instanceof Filterable filterable) {
+      FilterClause fc = filterable.filterClause(commandContext);
+      if (fc != null) {
```
amorton, line 267: these are the types of situations I think it would be good to make Filterable.filterClause always return non-null

src/main/java/io/stargate/sgv2/jsonapi/service/processor/SchemaValidatable.java
```diff
@@ -6,7 +6,8 @@
 import java.util.Objects;
 
 /**
- * TODO: AARON 9 sept 2024 - this is a failed idea of mine, we should remove as we go
+ * TODO: AARON 9 sept 2024 - this is a failed idea of mine, we should remove as we go UPDATE: Tatu
```
amorton, line 9: +1

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/matcher/FilterMatcher.java
```diff
@@ -32,14 +33,13 @@ public enum MatchStrategy {
     this.strategy = strategy;
   }
 
-  public Optional<CaptureGroups<T>> apply(T command) {
-    FilterClause filter = command.filterClause();
+  public Optional<CaptureGroups<T>> apply(CommandContext<?> ctx, T command) {
```
amorton, line 36: +1

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/matcher/FilterMatchRule.java
amorton: ok, lets not add to this PR for now. I think I would lean more towards adding Context to the Command , but the command is created before the context so to be pragmatic we may need to add the Command to the CommandContext. 

It will be a small change but hit a bunch of files, can do it later after some of this settles down

amorton: another thing, how would either approach work with Filterable.getFilter(CommandContext) ?

amorton (approved):
Nothing blocking or obvious errors. Multiple cleanups and questions for things. I have not read through all the tests, can do after we sync on this review.

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/matcher/FilterMatchRule.java
tatu-at-datastax: Btw, I came to conclusion that packaging or adding reference not necessarily a good idea. Forgot to add comment to that effect.
Either way, not really needed here anyway.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/filter/FilterSpec.java
```diff
@@ -0,0 +1,49 @@
...
+ */
+@Schema(
+    type = SchemaType.OBJECT,
+    implementation = FilterClause.class,
+    example =
+        """
+                     {"name": "Aaron", "country": "US"}
```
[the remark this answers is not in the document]
tatu-at-datastax: I am just copying existing example fwtw. But indentation is unfortunately with Google auto-formatter.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/JsonDefinition.java
```diff
@@ -0,0 +1,25 @@
...
+  /** The wrapped JSON value */
+  private final JsonNode json;
+
+  protected JsonDefinition(JsonNode json) {
+    this.json = json;
```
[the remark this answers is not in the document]
tatu-at-datastax: Good point, should add null check.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/JsonDefinition.java
```diff
@@ -0,0 +1,25 @@
...
+  }
+
+  /**
+   * @return JSON value that specifies the object
+   */
+  public JsonNode json() {
```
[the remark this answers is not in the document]
tatu-at-datastax: It is needed by converter to access JSON definition, not sure there's way around it. Making defensive copy rather expensive.
`JsonNode` is mutable; there's no immutable variant.
Ideally would be immutable or copy, but in particular case I think I'd go with documenting the situation (instead of trying to change).

src/main/java/io/stargate/sgv2/jsonapi/service/processor/MeteredCommandProcessor.java
tatu-at-datastax: Yes, I was toying with the idea of adding "empty" `FilterClause` always. That's a possibility -- unless we want to know difference between missing and empty clause? (not sure we do; probably we don't).

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/JsonDefinition.java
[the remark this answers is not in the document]
amorton: from comments, can be protected or remove and use var as protected

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/filter/FilterSpec.java
```diff
@@ -0,0 +1,49 @@
...
+                      """)
+public class FilterSpec extends JsonDefinition {
+  /**
+   * Lazily deserialized {@link FilterClause} from the JSON value. We need this due to existing
+   * reliance on specific stateful instances of {@link FilterClause}.
+   */
+  private transient FilterClause filterClause;
```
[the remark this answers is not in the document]
amorton: from talking - meant to be volatile but prob not needed

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/CommandContext.java
```diff
@@ -48,30 +53,51 @@ public static <T extends SchemaObject> CommandContext<T> forSchemaObject(
...
     if (schemaObject instanceof DatabaseSchemaObject dso) {
       return (CommandContext<T>)
           forSchemaObject(
-              dso, embeddingProvider, commandName, jsonProcessingMetricsReporter, apiFeatures);
+              dso,
+              embeddingProvider,
+              commandName,
+              jsonProcessingMetricsReporter,
+              apiFeatures,
+              operationsConfig);
```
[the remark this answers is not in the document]
tatu-at-datastax: Done.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/filter/FilterSpec.java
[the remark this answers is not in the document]
tatu-at-datastax: Done.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/CommandContext.java
```diff
@@ -224,4 +274,16 @@ public SmallRyeConfig getConfig() {
...
+
+  public OperationsConfig operationsConfig() {
+    if (operationsConfig != null) {
+      return operationsConfig;
+    }
+    // During tests:
+    return getConfig(OperationsConfig.class);
+  }
+
+  public FilterClause resolveFilterClause(JsonNode json) {
```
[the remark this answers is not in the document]
tatu-at-datastax: Moved to `FilterSpec` based on discussions.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/JsonDefinition.java
```diff
@@ -0,0 +1,25 @@
+package io.stargate.sgv2.jsonapi.api.model.command;
+
+import com.fasterxml.jackson.databind.JsonNode;
+
+/**
+ * Value type enclosing JSON-decoded representation of something to resolve into fully definition,
+ * such as filter or sort clause. Needed to let Quarkus/Jackson decode textual JSON into
+ * intermediate form ({@link JsonNode}) from which actual deserialization and validation can be
+ * deferred until we have context we need
```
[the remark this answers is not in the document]
tatu-at-datastax: Added `@see`.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/builders/CollectionFilterClauseBuilder.java
```diff
@@ -0,0 +1,128 @@
...
+      if (path.equals(DocumentConstants.Fields.DOC_ID)) {
+        FilterOperator filterOperator =
+            comparisonExpression.getFilterOperations().get(0).operator();
+        // if operator is $eq or $in, _id can be used, return
+        if (filterOperator == ValueComparisonOperator.EQ
+            || filterOperator == ValueComparisonOperator.IN) {
+          return;
+        }
+        // otherwise throw _id - specific JsonApiException
+        throw ErrorCodeV1.ID_NOT_INDEXED.toApiException(
```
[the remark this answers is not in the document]
tatu-at-datastax: Yes; existing code just moved around.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/builders/FilterClauseBuilder.java
```diff
@@ -1,62 +1,96 @@
...
-  private final OperationsConfig operationsConfig;
+public abstract class FilterClauseBuilder<T extends SchemaObject> {
+  protected final T schema;
 
-  public FilterClauseDeserializer() {
-    super(FilterClause.class);
-    SmallRyeConfig config = ConfigProvider.getConfig().unwrap(SmallRyeConfig.class);
-    operationsConfig = config.getConfigMapping(OperationsConfig.class);
+  protected FilterClauseBuilder(T schema) {
+    this.schema = schema;
```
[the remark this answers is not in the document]
tatu-at-datastax: Done.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/builders/FilterClauseBuilder.java
```diff
@@ -1,62 +1,96 @@
...
   }
 
-  /**
-   * {@inheritDoc} Filter clause can follow short-cut {"field" : "value"} instead of {"field" :
-   * {"$eq" : "value"}}
-   */
-  @Override
-  public FilterClause deserialize(
-      JsonParser jsonParser, DeserializationContext deserializationContext) throws IOException {
+  public static FilterClauseBuilder builderFor(SchemaObject schema) {
```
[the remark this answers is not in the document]
tatu-at-datastax: Fixed.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/builders/FilterClauseBuilder.java
```diff
@@ -1,62 +1,96 @@
...
+      return null;
+    }
     if (!filterNode.isObject()) {
+      // JSON `null`s are ok though
+      if (filterNode.isNull()) {
+        return null;
+      }
       throw ErrorCodeV1.UNSUPPORTED_FILTER_DATA_TYPE.toApiException();
     }
+
```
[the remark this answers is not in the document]
tatu-at-datastax: Correct: logic just moved around, not changed in any way.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/filters/collection/CollectionFilter.java
tatu-at-datastax: Yes -- "put" gave warning, being Deprecated; "set" is the intended replacement (that or "replace").

PR #1837, opened 2025-01-29
Complete V2 errors for Driver errors for the Tables 

### LINKED ISSUE #1745: Updates to DriverExceptionHandler & Driver Error Object V2 definitions
- [ ] In `DriverExceptionHandler` override the `ExceptionHandler.handleUnhandled()` to use a generic database V2 error that does not return 500's 
- [ ] Update `DriverExceptionHandler` for all of the expected driver errors to return V2 errors that are nicely written.

amorton:
**What this PR does**:

Adds V2 errors for all driver errors from the driver , still only used for Tables code. 

Fixes #1745

**Checklist**
- [*] Changes manually tested
- [*] Automated Tests added/updated


src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/DefaultDriverExceptionHandler.java
```diff
@@ -1,36 +1,516 @@
...
+  //
+  // - **InvalidConfigurationInQueryException**
+  //    - Happens if we get the DDL command wrong.
+  //    - Should not happen.
+
+  /**
+   * this is the InvalidRequestException in the Cassandra code base, there are 300+ usages in it.
+   * generally means while the syntax is correct, what you are asking it to do is not possible e.g.
+   * add to a list, but the column is not a list. the Data API should prevent this from happening,
+   * but if it does, it is a bug and maybe the user can fix it.
```
tatu-at-datastax, line 201: nit: I think if it's a bug, user cannot fix it, but can possibly work around it (change something to avoid hitting the bug)?

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/DefaultDriverExceptionHandler.java
```diff
@@ -1,36 +1,516 @@
...
+            }));
+  }
+
+  // ========================================================================
+  // QueryConsistencyException and subclasses
+  // see commend for the QueryExecutionException section
+  // ========================================================================
+
+  /**
+   * Thrown when there is a non timeout error while executing the request
```
tatu-at-datastax, line 345: nit: "non timeout" -> "non-timeout"

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/RequestFailureReason.java
```diff
@@ -0,0 +1,62 @@
...
+ *
+ * <p><b>NOTE:</b> The codes here need to be in sync with the C* code, the exceptions in the drivers
+ * only use the codes. Sync with <em>Astra</em> version of the code not the DataStax because the
+ * Astra version has more codes.
+ *
+ * <p>Used with {@link com.datastax.oss.driver.api.core.servererrors.ReadFailureException} and
+ * similar errors.
+ */
+public enum RequestFailureReason {
+  // codes must be sequential !
```
tatu-at-datastax, line 15: Do we ever need to worry about 500+ codes that seem to exist for DataStax-DBs (DSE and/or HCD)?

src/main/java/io/stargate/sgv2/jsonapi/service/operation/OperationAttempt.java
```diff
@@ -634,6 +613,24 @@ public PrettyToStringBuilder toString(PrettyToStringBuilder prettyToStringBuilde
...
+   *
+   * <p>This is really just a Pair so we can give the Statement used to the {@link
+   * DefaultDriverExceptionHandler} via the {@link DefaultDriverExceptionHandler.Factory} and the
+   * statement context into the exception handling.
+   *
+   * @param statement Statement used with the attempt, may be null.
+   * @param results A supplier that fetches the results when queried, the supplier must wait until
+   *     called to avoid blocking. Must not be null.
+   */
+  protected record StatementContext(
```
tatu-at-datastax, line 627: This is pretty well-hidden here (took a while to find via GH diffs) -- while it is sort of internal to `OperationAttempt` should it be a main-level record or (alternatively) at beginning of this class next to `OperationStatus`? Our codebase is pretty heavy on inner classes/records.
(but I can see it might make sense for non-public types).

src/main/java/io/stargate/sgv2/jsonapi/service/operation/OperationAttempt.java
```diff
@@ -178,24 +186,32 @@ protected Uni<AsyncResultSet> executeIfInProgress(CommandQueryExecutor queryExec
...
+              + positionAndAttemptId());
+    }
+    return currentStatement.results.get();
   }
 
   /**
-   * Sublasses must implement this method to build the query and execute it, they should not do
-   * anything with Uni for retry etc, that is handled in the base class {@link
-   * #execute(CommandQueryExecutor, DriverExceptionHandler)}.
+   * Subclasses must implement this method to build the query and provide a supplier to get the
```
tatu-at-datastax, line 204: Maybe: "provide a supplier that executes query and returns results" (or such?)

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/DefaultDriverExceptionHandler.java
```diff
@@ -1,36 +1,516 @@
...
+   *
+   * <pre>
+   *   return new GenericOperation<>(attempts, pageBuilder, TableDriverExceptionHandler::new);
+   * </pre>
+   *
+   * See usage of the {@link
+   * io.stargate.sgv2.jsonapi.service.operation.tables.TableDriverExceptionHandler}
+   */
+  @FunctionalInterface
+  public interface Factory<T extends SchemaObject>
```
tatu-at-datastax, line 465: Naming: this was confusing from the other end -- type defines not a factory, but something to call/apply to get/construct a factory. So maybe... `FactorySupplier`? (or `Provider`? not "FactoryFactory" tho :) ).

Not a big deal but just noticed as a minor readability concern.

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/CreateIndexCommandResolver.java
```diff
@@ -86,6 +87,7 @@ public Operation resolveTableCommand(
     return new GenericOperation<>(
         new OperationAttemptContainer<>(attempt),
         pageBuilder,
-        new CreateIndexExceptionHandler(apiIndex.indexName()));
+        DefaultDriverExceptionHandler.Factory.withIdentifier(
+            CreateIndexExceptionHandler::new, apiIndex.indexName()));
```
tatu-at-datastax, line 91: Whoa. This took time to unpack. Functional style is compact but not necessarily easy to digest.
(not sure there's any way to make it more readable, just an observation).

src/main/java/io/stargate/sgv2/jsonapi/util/CqlPrintUtil.java
```diff
@@ -0,0 +1,62 @@
...
+ *
+ * <p>Trims the long values we get with vectors etc.
+ */
+public abstract class CqlPrintUtil {
+
+  public static List<Object> trimmedPositionalValues(SimpleStatement statement) {
+    return statement.getPositionalValues().stream()
+        .map(
+            value -> {
+              if (value instanceof CqlVector<?> vector) {
```
tatu-at-datastax, line 20: Wouldn't we have some challenges with `List`s, `Set`s and `Map`s too?

src/test/java/io/stargate/sgv2/jsonapi/exception/ExceptionHandlerTest.java
```diff
@@ -22,13 +21,13 @@ public class ExceptionHandlerTest {
...
 
     var handler =
-        new ExceptionHandler<TableSchemaObject, UnsupportedOperationException>() {
+        new ExceptionHandler<UnsupportedOperationException>() {
           @Override
           public Class<UnsupportedOperationException> getExceptionClass() {
             return UnsupportedOperationException.class;
           }
         };
-    var actualEx = assertDoesNotThrow(() -> handler.maybeHandle(null, null));
```
tatu-at-datastax, line 31: I've always wondered the point of "assertDoesNotThrow" -- why not just execute; if exception is thrown, test will still fail? Or does `assertDoesNotThrow()` provide some useful extra information?

tatu-at-datastax (approved):
Not sure I could review in detail, but tried adding question where I had. What I did understand makes sense wrt our earlier discussions on requiring injection of contextual info on exception handlers (statement, schema entity name where applicable).

No required changes, added some suggestions.

src/test/java/io/stargate/sgv2/jsonapi/exception/ExceptionHandlerTest.java
amorton: it's pretty useless, and the docs say as much - any thrown error causes a test to fail. I tend to find my self using it when writing tests when I want to get the the result of the function and I am trying to be very declarative about what is expected.

src/main/java/io/stargate/sgv2/jsonapi/util/CqlPrintUtil.java
```diff
@@ -0,0 +1,62 @@
...
+                int vectorSize = vector.size();
+                List<Object> trimmedList = new ArrayList<>();
+
+                // Add elements up to a maximum of 5 or the actual vector size, whichever is
+                // smaller
+                for (int i = 0; i < Math.min(5, vectorSize); i++) {
```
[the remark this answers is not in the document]
amorton: done.

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/CreateIndexCommandResolver.java
amorton: understood, it makes it simple to use... but not simple for implementation. 

(I got the idea from the Pragmatic Java book talking about  method references for a ctor :) )

src/main/java/io/stargate/sgv2/jsonapi/service/operation/OperationAttempt.java
amorton: agree it can be better, and am not happy with the name etc. 

Am working on some refactoring for this areas so we can handle the embedding in the pipeline better and will improve as part of that.

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/RequestFailureReason.java
amorton: not for this, it's internal C* codes that writers / worker nodes return to the coordinator for why something failed. Which are then passed back to the client. 

See commend at top they are copied from RequestFailureReason in the C* code. This is where we can detect that an index is not available and give correct error

src/test/java/io/stargate/sgv2/jsonapi/exception/ExceptionHandlerTest.java
tatu-at-datastax: Right, but in theory you could say that every line that does not expect exception should be surrounded by "assertDoesNotThrow".

It's fine, I just wanted to ask as I've seen it be used -- could have missed something.

PR #1846, opened 2025-02-05
CreateIndexCommand for collection types, map/set/list

Yuqi-Du:
**Checklist**
- [x] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/CreateIndexCommand.java
```diff
@@ -26,7 +26,7 @@
 })
 public record CreateIndexCommand(
     @NotNull
-        @Schema(description = "Name of the column to create the index on.")
+        @Schema(description = "Name of the index.")
```
amorton, line 29: Nit: change to "Required name of the new Collection" 

Note: that I now realise the name of the index is optional in CQL, if not supplier it will use the tabkle name + column name. Lets keep this required for now and then we can make it optional later

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/table/definition/indexes/IndexDefinitionDesc.java
```diff
@@ -4,9 +4,9 @@
  * The definition of an index used in creation and listing, this does not include the index name
  * because the create index commands have the "name" and "definition" top level fields.
  */
-public interface IndexDefinitionDesc<T> {
+public interface IndexDefinitionDesc<IndexColumn, IndexOption> {
```
amorton, line 7: nit: add @param doc comments for the generics, explain why they do not have restrictions on them (e.g. extends) .. because they can be either a basic string or a complex type. 

If this interface is serialised in a response then I like to decorate with @JsonPropertyOrder so the order is stable and defined

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/table/definition/indexes/RegularIndexDefinitionDesc.java
```diff
@@ -15,17 +30,117 @@
   TableDescConstants.IndexDefinitionDesc.OPTIONS
 })
 public record RegularIndexDefinitionDesc(
+
+    /*
```
amorton, line 34: can we move this doc comment to be on the class

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/table/definition/indexes/RegularIndexDefinitionDesc.java
```diff
@@ -15,17 +30,117 @@
...
+            required = true)
+        @JsonDeserialize(using = RegularIndexColumnDeserializer.class)
+        @JsonSerialize(using = RegularIndexColumnSerializer.class)
         @JsonProperty(TableDescConstants.IndexDefinitionDesc.COLUMN)
-        String column,
-    //
+        RegularIndexColumn column,
     @JsonInclude(JsonInclude.Include.NON_NULL)
         @Nullable
         @Schema(description = "Indexing options.", type = SchemaType.OBJECT)
```
amorton, line 51: Update to "Options for the new index, not all options are valid for all columns. Check documentation for details"

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/table/definition/indexes/RegularIndexDefinitionDesc.java
```diff
@@ -15,17 +30,117 @@
...
+        RegularIndexDefinitionDesc.RegularIndexColumn,
+        RegularIndexDefinitionDesc.RegularIndexDescOptions> {
+
+  public record RegularIndexColumn(
+      @NotNull
+          @Schema(description = "Name of the column to index.", required = true)
+          @JsonProperty(TableDescConstants.IndexDefinitionDesc.COLUMN)
+          String columnName,
+      @Nullable
+          @Schema(description = "Index function for map/set/list column.")
```
amorton, line 64: nit: lets hide the idea of a "index function" and this only applies to maps we are not going to expose the idea of functions for set and list. 

"Property of the map to index, either $keys or $values " something like that

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/table/definition/indexes/RegularIndexDefinitionDesc.java
```diff
@@ -15,17 +30,117 @@
...
+          @JsonProperty(TableDescConstants.IndexDefinitionDesc.COLUMN)
```
amorton, line 65: bug, this is not the column

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/table/definition/indexes/RegularIndexDefinitionDesc.java
```diff
@@ -15,17 +30,117 @@
...
+          ApiIndexFunction indexFunction) {}
```
amorton, line 66: does this mean we are exposing the idea of the "entries" indexing function ? We prob need an ApiIndexDesc enum with the values we want to API to know about and the apiName . We do not want to bind the name in the API to the name from CQL

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/table/definition/indexes/RegularIndexDefinitionDesc.java
```diff
@@ -15,17 +30,117 @@
...
+   *   - Default to index on map entries: {"column": "mapColumn"}
+   *   - Index on map keys: {"column": {"mapColumn" : "$keys"}}
+   *   - Index on map values: {"column": {"mapColumn" : "$values"}}
+   *   - Index on map entries: {"column": {"mapColumn" : "$entries"}}
+   *   </pre>
+   *
+   * NOTE, this is just index function from user input, validation and default values are in {@link
+   * ApiRegularIndex} since we need the ColumnMetaData
+   */
+  private static class RegularIndexColumnDeserializer extends StdDeserializer<RegularIndexColumn> {
```
amorton, line 85: lets pull the ser/de ser out into their own files to make it easier to see the pure API

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/table/definition/indexes/RegularIndexDefinitionDesc.java
```diff
@@ -15,17 +30,117 @@
...
+   * Primitive column: {"column": "age"}, indexFunction is null <br>
+   * List column: {"column": "listColumn"} <br>
+   * Set column: {"column": "setColumn"} <br>
+   * Map column:
+   *
+   * <pre>
+   *   - Default to index on map entries: {"column": "mapColumn"}
+   *   - Index on map keys: {"column": {"mapColumn" : "$keys"}}
+   *   - Index on map values: {"column": {"mapColumn" : "$values"}}
+   *   - Index on map entries: {"column": {"mapColumn" : "$entries"}}
```
amorton, line 79: I am not sure on exposing the idea of $entries, it may make sense. But we do not use $entries anywhere else in the API - but with $keys and $values we do use them in the filtering 

I was thinking the default is entries, and the default way to filter is with entries. Then you need to do something special to work with keys and values

src/main/java/io/stargate/sgv2/jsonapi/service/schema/tables/ApiIndexFunction.java
```diff
@@ -1,30 +1,59 @@
...
-    if (FUNCTION_MAP.containsKey(cqlFunction)) {
-      return FUNCTION_MAP.get(cqlFunction);
+    if (cqlFunction == null || !FUNCTION_MAP.containsKey(cqlFunction.toLowerCase())) {
+      throw new UnknownCqlIndexFunctionException(cqlFunction);
     }
-    throw new UnknownCqlIndexFunctionException(cqlFunction);
+    return FUNCTION_MAP.get(cqlFunction.toLowerCase());
+  }
+
+  public static ApiIndexFunction fromDollarCommand(String indexFunctionWithDollarSign) {
```
amorton, line 43: -1 the API names for things should be in the api packages I think not in our core schema here. 

and we should do this as case insensitive

src/main/java/io/stargate/sgv2/jsonapi/service/schema/tables/ApiIndexFunction.java
```diff
@@ -1,30 +1,59 @@
...
+    return switch (indexFunctionWithDollarSign) {
+      case TableDescConstants.CollectionTypeComponent.keys -> KEYS;
+      case TableDescConstants.CollectionTypeComponent.values -> VALUES;
+      case TableDescConstants.CollectionTypeComponent.entries -> ENTRIES;
+      default -> throw SchemaException.Code.INVALID_FORMAT_FOR_INDEX_CREATION_COLUMN.get();
+    };
+  }
+
+  public String toDollarCommand() {
```
amorton, line 52: as above, there should be a ApiMapIndexTypeDesc enum for this in the API

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/table/definition/indexes/RegularIndexDefinitionDesc.java
```diff
@@ -15,17 +30,117 @@
...
+        return new RegularIndexColumn(columnNode.textValue(), null);
+      } else if (columnNode.isObject() && columnNode.size() == 1) {
+        Map.Entry<String, JsonNode> entry = columnNode.fields().next();
+
+        if (!entry.getValue().isTextual()) {
+          // E.G. {"column": {"mapColumn" : 123}}
+          throw SchemaException.Code.INVALID_FORMAT_FOR_INDEX_CREATION_COLUMN.get();
+        }
+        return new RegularIndexColumn(
+            entry.getKey(), ApiIndexFunction.fromDollarCommand(entry.getValue().textValue()));
```
amorton, line 105: what is the $ value is incorrect ?

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/table/definition/indexes/RegularIndexDefinitionDesc.java
```diff
@@ -15,17 +30,117 @@
...
+      } else {
+        throw SchemaException.Code.INVALID_FORMAT_FOR_INDEX_CREATION_COLUMN.get();
```
amorton, line 107: duplicate code as above, refactor to only throw this once

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/table/definition/indexes/RegularIndexDefinitionDesc.java
```diff
@@ -15,17 +30,117 @@
...
+    @Override
+    public void serialize(
+        RegularIndexColumn regularIndexColumn,
+        JsonGenerator jsonGenerator,
+        SerializerProvider serializerProvider)
+        throws IOException {
+      if (regularIndexColumn.indexFunction != null) {
+        jsonGenerator.writeStartObject();
+        jsonGenerator.writeStringField(
+            regularIndexColumn.columnName, regularIndexColumn.indexFunction.toDollarCommand());
```
amorton, line 137: lets add ApiMapIndexTypeDesc and the propery name should be apiName()

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/table/definition/indexes/RegularIndexDefinitionDesc.java
```diff
@@ -15,17 +30,117 @@
...
+        jsonGenerator.writeEndObject();
+      } else {
+        jsonGenerator.writeString(regularIndexColumn.columnName);
+      }
+    }
+  }
 
   /** Options for the index, used only with text, including text in a collection */
   @JsonPropertyOrder({
```
amorton, line 146: question - maybe we should say then these are valid with map / set /list ?

src/main/java/io/stargate/sgv2/jsonapi/config/constants/TableDescConstants.java
```diff
@@ -32,6 +32,7 @@ interface IndexDesc {
     String INDEX_TYPE = "indexType";
     String NAME = "name";
     String OPTIONS = "options";
+    String INDEX_ON = "indexOn";
```
amorton, line 35: Not used

src/main/java/io/stargate/sgv2/jsonapi/config/constants/TableDescConstants.java
```diff
@@ -50,4 +51,11 @@ interface VectorIndexDefinitionDescOptions {
     String SOURCE_MODEL = "source_model";
     String SIMILARITY_FUNCTION = "similarity_function";
   }
+
+  // These strings will be used in createIndex, filtering for collection datatypes
+  interface CollectionTypeComponent {
+    String keys = "$keys";
+    String values = "$values";
+    String entries = "$entries";
```
amorton, line 59: not sure we want to use $entries and we should proba access these as apiName() on an enum like ApiMapIndexTypeDesc

src/main/java/io/stargate/sgv2/jsonapi/exception/SchemaException.java
```diff
@@ -48,6 +50,7 @@ public enum Code implements ErrorCode<SchemaException> {
 
     // older below - seperated because they need to be confirmed
     INVALID_CONFIGURATION,
+    INVALID_FORMAT_FOR_INDEX_CREATION_COLUMN,
```
amorton, line 53: nit: move this to be above - this is the list of things to remove at some point :)

src/main/java/io/stargate/sgv2/jsonapi/service/operation/tables/CreateIndexAttempt.java
```diff
@@ -62,4 +68,25 @@ protected SimpleStatement buildStatement() {
     var extendedCreateIndex = new ExtendedCreateIndex((DefaultCreateIndex) createIndex);
     return extendedCreateIndex.build();
   }
+
+  /**
+   * Index Function keys/values/entries on map/set/list collection columns needs special handling to
+   * create the driver CreateIndex. <br>
+   * If indexFunction is null, then it is a regular index on a scalar column.
+   */
+  private CreateIndex createIndexWithIndexFunction(
```
amorton, line 77: easier to pass the ApiIndexDef

src/main/java/io/stargate/sgv2/jsonapi/service/schema/tables/ApiIndexDef.java
```diff
@@ -28,6 +28,9 @@ public interface ApiIndexDef extends PrettyPrintable {
   /** The type of index from the API perspective. */
   ApiIndexType indexType();
 
+  /** The index function from the API perspective. */
+  ApiIndexFunction indexFunction();
```
amorton, line 32: add to toString()

src/main/java/io/stargate/sgv2/jsonapi/service/schema/tables/ApiIndexFunction.java
```diff
@@ -1,30 +1,59 @@
...
 
   ApiIndexFunction(String cqlFunction) {
     this.cqlFunction = cqlFunction;
   }
 
   public static ApiIndexFunction fromCql(String cqlFunction)
       throws UnknownCqlIndexFunctionException {
-    if (FUNCTION_MAP.containsKey(cqlFunction)) {
-      return FUNCTION_MAP.get(cqlFunction);
+    if (cqlFunction == null || !FUNCTION_MAP.containsKey(cqlFunction.toLowerCase())) {
```
amorton, line 37: super duper nit pick : ) do the lcase once only

src/main/java/io/stargate/sgv2/jsonapi/service/schema/tables/CollectionApiDataType.java
```diff
@@ -3,7 +3,7 @@
 import com.datastax.oss.driver.api.core.type.DataType;
 
 /** Interface defining the api data type for collection types */
-public abstract class CollectionApiDataType implements ApiDataType {
+public abstract class CollectionApiDataType<T extends DataType> implements ApiDataType {
```
amorton, line 6: doc comments to explain the type 

also , why do we need this to be generic now ? that would be something to put in the doc comments :) 

why are we using the CQL data type ? In general these classes should be our internal data model

src/main/java/io/stargate/sgv2/jsonapi/service/schema/tables/CollectionApiDataType.java
```diff
@@ -64,4 +64,6 @@ public DataType cqlType() {
   public PrimitiveApiDataTypeDef getValueType() {
     return valueType;
   }
+
+  public abstract boolean isFrozen();
```
amorton, line 68: this must have been somewhere else in the code ? 

add comments to explain why we need it here , I think I see why but I also want to check this was not somewhere else in the structur

src/main/java/io/stargate/sgv2/jsonapi/service/schema/tables/ApiMapType.java
```diff
@@ -73,8 +73,7 @@ static ApiMapType from(ApiDataType keyType, ApiDataType valueType, boolean isFro
   public static boolean isKeyTypeSupported(ApiDataType keyType) {
     Objects.requireNonNull(keyType, "keyType must not be null");
 
-    // keys must be text or ascii, because keys in JSON are string
-    return keyType == ApiDataTypeDefs.ASCII || keyType == ApiDataTypeDefs.TEXT;
+    return keyType.isPrimitive();
```
amorton, line 76: wondering if there were any errors elsehwere that need to change , need to check

src/main/java/io/stargate/sgv2/jsonapi/service/schema/tables/ApiMapType.java
```diff
@@ -155,12 +155,13 @@ public ApiMapType create(MapType cqlType, VectorizeDefinition vectorizeDefn)
...
-      if (!(cqlType.getKeyType() == DataTypes.TEXT || cqlType.getKeyType() == DataTypes.ASCII)) {
-        return false;
-      }
       // must be a primitive type value
       return cqlType.getValueType() instanceof PrimitiveType;
     }
   }
+
+  @Override
+  public boolean isFrozen() {
```
amorton, line 164: code tidy : move this to be above the innner clases so it is with the other overrides from the same class . do this for all places

amorton (changes requested):
partial review, discussed, some tweaks and changes for the index functions we expose

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/CreateIndexCommand.java
Yuqi-Du: created an issue
https://github.com/stargate/data-api/issues/1858

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/table/definition/indexes/RegularIndexDefinitionDesc.java
Yuqi-Du: solved

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/table/definition/indexes/RegularIndexDefinitionDesc.java
Yuqi-Du: solved

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/table/definition/indexes/RegularIndexDefinitionDesc.java
Yuqi-Du: removed $entries, leave it as default

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/table/definition/indexes/RegularIndexDefinitionDesc.java
Yuqi-Du: solved, remove entries, and do not expose IndexFunction in API level

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/table/definition/indexes/RegularIndexDefinitionDesc.java
```diff
@@ -15,17 +30,117 @@
...
+     * API regular column can be primitive or map/set/list collection type.
+     * CreateIndexCommand on collection column will have index function.
+     * User can specify or use the default index function.
+     * Details see {@link RegularIndexColumnDeserializer}
+     */
     @NotNull
-        @Schema(description = "Name of the column to index.", required = true)
+        @Schema(
+            description =
+                "Name of the column to index. (Optional: can also specify the index function for map column. E.G. {\"column\": {\"mapColumn\" : \"$keys\"}}",
```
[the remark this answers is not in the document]
Yuqi-Du: solved, remove entries, and do not expose IndexFunction in API level

src/main/java/io/stargate/sgv2/jsonapi/service/schema/tables/ApiIndexFunction.java
Yuqi-Du: it is constraint by how exception was wrote.
  throw new UnknownCqlIndexFunctionException(cqlFunction);
  
```
   super(
        "Unknown CQL index function name: "
            + Objects.requireNonNull(functionName, "functionName must not be null"));
```

src/main/java/io/stargate/sgv2/jsonapi/config/constants/TableDescConstants.java
Yuqi-Du: removed

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/table/definition/indexes/RegularIndexDefinitionDesc.java
Yuqi-Du: added

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/table/definition/indexes/IndexDefinitionDesc.java
Yuqi-Du: added doc.

It won't have order messed up, since it will be serialized in single kv,  "column": "mapColumn" or "column" : {"mapColumn" : "$keys"}

src/main/java/io/stargate/sgv2/jsonapi/service/schema/tables/ApiMapType.java
Yuqi-Du: I had a look, this will unlock createTable, createIndex, listIndex for "non-String" key map

src/main/java/io/stargate/sgv2/jsonapi/service/schema/tables/ApiIndexDef.java
```diff
@@ -28,6 +28,9 @@ public interface ApiIndexDef extends PrettyPrintable {
   /** The type of index from the API perspective. */
   ApiIndexType indexType();
 
+  /** The index function from the API perspective. */
```
amorton, line 31: comment if implementations can return null for this

src/main/java/io/stargate/sgv2/jsonapi/service/schema/tables/ApiVectorType.java
```diff
@@ -148,4 +148,9 @@ public boolean isSupported(VectorType cqlType) {
       return cqlType.getElementType() == DataTypes.FLOAT;
     }
   }
+
+  @Override
+  public boolean isFrozen() {
+    return false;
```
amorton, line 154: cody tidy : locate with other functions from same interface

src/main/java/io/stargate/sgv2/jsonapi/service/schema/tables/CollectionApiDataType.java
```diff
@@ -2,8 +2,12 @@
 
 import com.datastax.oss.driver.api.core.type.DataType;
 
-/** Interface defining the api data type for collection types */
-public abstract class CollectionApiDataType implements ApiDataType {
+/**
+ * Interface defining the api data type for collection types. <br>
```
amorton, line 6: use < p > not < br >

src/main/java/io/stargate/sgv2/jsonapi/service/schema/tables/CollectionApiDataType.java
```diff
@@ -2,8 +2,12 @@
...
+ * DataType T is a generic type that can be map/set/list/vector, inheritors should implement
```
amorton, line 7: use the @param doc for T - we are making this generic because the driver has some properties only available on the leaf level data types

amorton (changes requested):
there is a bug in checking what types we can use to create a map key, it is not all primitive types because that would include counters and durations. Discussed with Yuqi - other comments are some refactoring and code tidy.

src/main/java/io/stargate/sgv2/jsonapi/service/schema/tables/ApiRegularIndex.java
```diff
@@ -91,77 +114,218 @@ public ApiRegularIndex create(
...
+      // Set collection type
+      if (apiColumnDef.type() instanceof ApiSetType apiSetType) {
+        targetColumnType = apiSetType.valueType.typeName();
+      }
+      // List collection type
+      if (apiColumnDef.type() instanceof ApiListType apiListType) {
+        targetColumnType = apiListType.valueType.typeName();
+      }
+
+      if (optionsDesc != null
```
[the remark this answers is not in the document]
Yuqi-Du: The original logic for index options (before this PR), is if user don't have optionDesc in the command, we still use the default options for the index.

So here, I don't want to exit the function when optionsDesc is null, instead, it will goes to the end of the function and populate the options with default.

src/main/java/io/stargate/sgv2/jsonapi/service/schema/tables/ApiIndexFunction.java
```diff
@@ -1,30 +1,59 @@
...
+      throw new UnknownCqlIndexFunctionException(cqlFunction);
     }
-    throw new UnknownCqlIndexFunctionException(cqlFunction);
+    return FUNCTION_MAP.get(cqlFunction.toLowerCase());
+  }
+
+  public static ApiIndexFunction fromApiMapComponent(ApiMapComponent apiMapComponent) {
+    return switch (apiMapComponent) {
```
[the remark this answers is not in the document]
Yuqi-Du: Second thought, I feel weird of doing this.

This makes ApiIndexFunction coupled with toAPIComponent() and also since we can't do for fromApiMapComponent, this makes the implementation asymmetric.

I vote for keeping these two methods.

vkarpov15:
API looks good, but can you explain how `"column" : {"mapColumn" : "$keys"}` would be different from `"column" : "mapColumn"`? What sort of queries would be covered by the former but not the latter?

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/CreateIndexCommand.java
```diff
@@ -25,7 +25,8 @@
   TableDescConstants.IndexDesc.OPTIONS
 })
 public record CreateIndexCommand(
-    @Schema(description = "Required name of the new Index")
+    @NotNull
```
Hazel-Datastax, line 28: I removed @NotNull intentionally so that we could have a better error message for null and empty name. See https://github.com/stargate/data-api/pull/1815#discussion_r1941984429 for the detailed discussion

src/main/java/io/stargate/sgv2/jsonapi/service/schema/tables/ApiIndexFunction.java
```diff
@@ -1,30 +1,59 @@
...
+ * entries(map), values(set), values(list). Data API does not support frozen map/set/list, so FULL
+ * index creation on frozen column will also not be supported.
+ */
 public enum ApiIndexFunction {
   KEYS("keys"),
   VALUES("values"),
   ENTRIES("entries");
 
-  private final String cqlFunction;
+  public final String cqlFunction;
```
Hazel-Datastax, line 19: Why change to `public`?

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/CreateIndexCommand.java
amorton: +1 this is tested in the Resolver now

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/deserializers/RegularIndexColumnDeserializer.java
```diff
@@ -0,0 +1,63 @@
...
+      // E.G. {"column": "mapColumn"}, default to entries for map
+      return new RegularIndexDefinitionDesc.RegularIndexColumn(columnNode.textValue(), null);
+    } else if (columnNode.isObject() && columnNode.size() == 1) {
+      Map.Entry<String, JsonNode> entry = columnNode.fields().next();
+      if (entry.getValue().isTextual()) {
+        // E.G. {"column": {"mapColumn" : "$keys"}}
+        // E.G. {"column": {"mapColumn" : "$values"}}
+        Optional<ApiMapComponent> apiMapComponent =
+            ApiMapComponent.fromUserInput(entry.getValue().textValue());
+        if (apiMapComponent.isEmpty()) {
```
amorton, line 53: nit: the standard with an Optional is to use .orElseThrow() 

so it would be: 

```java
var apiMapComponent = ApiMapComponent
    .fromUserInput(entry.getValue().textValue())
   .orElseThrow(() -> SchemaException.Code.INVALID_FORMAT_FOR_INDEX_CREATION_COLUMN.get());
```

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/table/definition/indexes/RegularIndexDefinitionDesc.java
```diff
@@ -3,31 +3,67 @@
...
-    implements IndexDefinitionDesc<RegularIndexDefinitionDesc.RegularIndexDescOptions> {
+    implements IndexDefinitionDesc<
+        RegularIndexDefinitionDesc.RegularIndexColumn,
+        RegularIndexDefinitionDesc.RegularIndexDescOptions> {
 
-  /** Options for the index, used only with text, including text in a collection */
+  public record RegularIndexColumn(
+      @NotNull @Schema(description = "Required name of the column to index.", required = true)
+          String columnName,
+      @Nullable @Schema(description = "Optional component of a map column to index, keys or values")
```
amorton, line 54: hit: tweak "keys or values" to be "$keys or $values"

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/table/ApiMapComponent.java
```diff
@@ -0,0 +1,45 @@
...
+  private static final Map<String, ApiMapComponent> COMPONENT_MAP;
+
+  static {
+    COMPONENT_MAP = new HashMap<>();
+    for (ApiMapComponent component : ApiMapComponent.values()) {
+      COMPONENT_MAP.put(component.apiName.toLowerCase(), component);
+    }
+  }
+
+  public static Optional<ApiMapComponent> fromUserInput(String userInput) {
```
amorton, line 27: nit: call it fromApiName to make it clear it is based on the apiName()

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/table/ApiMapComponent.java
```diff
@@ -0,0 +1,45 @@
...
+    if (userInput == null) {
+      return Optional.empty();
+    }
+    var lowerCaseInput = userInput.toLowerCase();
+    if (!COMPONENT_MAP.containsKey(lowerCaseInput)) {
+      return Optional.empty();
+    }
+    return Optional.of(COMPONENT_MAP.get(lowerCaseInput));
```
amorton, line 35: just a thought, Optional.ofNullable() can make code this like easier, it can return an empty Optional or one with a value it in

src/main/java/io/stargate/sgv2/jsonapi/service/schema/tables/ApiDataTypeDefs.java
```diff
@@ -43,7 +43,15 @@ public abstract class ApiDataTypeDefs {
 
   public static final PrimitiveApiDataTypeDef DURATION =
       new PrimitiveApiDataTypeDef(
-          ApiTypeName.DURATION, DataTypes.DURATION, ApiSupportDef.Support.FULL);
+          // Does not support duration as map key. But, duration can be map/list/set value
+          ApiTypeName.DURATION,
+          DataTypes.DURATION,
```
amorton, line 48: +1

src/main/java/io/stargate/sgv2/jsonapi/service/schema/tables/ApiDataTypeDefs.java
```diff
@@ -64,7 +72,9 @@ public abstract class ApiDataTypeDefs {
       new PrimitiveApiDataTypeDef(
           ApiTypeName.COUNTER,
           DataTypes.COUNTER,
-          new ApiSupportDef.Support(false, false, true, true));
+          // Does not support counter as primitive column, list/set value or map key/value.
+          new ApiSupportDef.Support(
+              false, ApiSupportDef.CollectionSupport.NONE, false, true, true));
```
amorton, line 77: +1

src/main/java/io/stargate/sgv2/jsonapi/service/schema/tables/ApiDataTypeDefs.java
```diff
@@ -73,7 +83,9 @@ public abstract class ApiDataTypeDefs {
       new PrimitiveApiDataTypeDef(
           ApiTypeName.TIMEUUID,
           DataTypes.TIMEUUID,
-          new ApiSupportDef.Support(false, true, true, true));
+          // Does not support counter as primitive column, list/set value or map key/value.
+          // TODO dod we want to relax this guardrail?
+          new ApiSupportDef.Support(false, ApiSupportDef.CollectionSupport.NONE, true, true, true));
```
amorton, line 88: +1 for now, I am guessing this is because we are trying to reduce the ide of using timeuuid so it makes sense. May need to change later

src/main/java/io/stargate/sgv2/jsonapi/service/schema/tables/ApiListType.java
```diff
@@ -27,26 +26,30 @@ private ApiListType(
...
-    throw new IllegalArgumentException(
-        "valueType must be primitive type, valueType%s".formatted(valueType));
+    return new ApiListType(
+        (PrimitiveApiDataTypeDef) valueType, defaultApiSupport(isFrozen), isFrozen);
   }
 
   public static boolean isValueTypeSupported(ApiDataType valueType) {
     Objects.requireNonNull(valueType, "valueType must not be null");
-    return valueType.isPrimitive();
+    return valueType.apiSupport().collectionSupport().asListValue();
```
amorton, line 52: this looks much better , thanks +1

src/main/java/io/stargate/sgv2/jsonapi/service/schema/tables/ApiListType.java
```diff
@@ -110,10 +114,13 @@ public ApiListType create(ListType cqlType, VectorizeDefinition vectorizeDefn)
     @Override
     public boolean isSupported(ListType cqlType) {
       Objects.requireNonNull(cqlType, "cqlType must not be null");
-
-      ///  we accept frozen, but change the support.
-      // must be a primitive type value
-      return cqlType.getElementType() instanceof PrimitiveType;
+      //  we accept frozen, but change the support.
+      try {
+        var valueType = TypeFactoryFromCql.DEFAULT.create(cqlType.getElementType(), null);
```
amorton, line 119: +1

src/main/java/io/stargate/sgv2/jsonapi/service/schema/tables/ApiMapType.java
```diff
@@ -48,39 +47,41 @@ private ApiMapType(
...
-      return new ApiMapType(
-          (PrimitiveApiDataTypeDef) keyType,
-          (PrimitiveApiDataTypeDef) valueType,
-          defaultApiSupport(isFrozen),
-          isFrozen);
-    }
-    throw new IllegalArgumentException(
-        "keyType and valueType must be primitive types, keyType: %s valueType: %s"
-            .formatted(keyType, valueType));
+    return new ApiMapType(
```
amorton, line 70: this has removed the `    if (isKeyTypeSupported(keyType) && isValueTypeSupported(valueType)) {` check so wil now throw a cast error if the keyType cannot be cast to PrimitiveApiDataTypeDef

can we add it back ? I see the comment about checking, but IMHO this functiuon is not private so cannot gurantee who will call it

src/main/java/io/stargate/sgv2/jsonapi/service/schema/tables/ApiRegularIndex.java
```diff
@@ -79,6 +88,19 @@ public RegularIndexDefinitionDesc definition() {
   private static class UserDescFactory
       extends IndexFactoryFromIndexDesc<ApiRegularIndex, RegularIndexDefinitionDesc> {
 
+    private static final ObjectMapper objectMapper = new ObjectMapper();
```
amorton, line 91: objectMapper is not used

src/main/java/io/stargate/sgv2/jsonapi/service/schema/tables/ApiRegularIndex.java
```diff
@@ -91,77 +114,218 @@ public ApiRegularIndex create(
...
+      }
+
       // we could check if there is an existing index but that is a race condition, we will need to
-      // catch it if it fails - the resolver needs to set up a custom error mapper
-      // regular indexes can only be on primitive. Adding indexes on maps, sets, lists will come
-      // later.
-      if (!apiColumnDef.type().isPrimitive()) {
-        throw SchemaException.Code.UNSUPPORTED_INDEXING_FOR_DATA_TYPES.get(
+      // catch it if it fails
+      throw SchemaException.Code.UNSUPPORTED_INDEXING_FOR_DATA_TYPES.get(
```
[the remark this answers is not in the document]
amorton: any thoughts on how this can happen ? Just for the next person to work on this

src/main/java/io/stargate/sgv2/jsonapi/service/schema/tables/ApiRegularIndex.java
```diff
@@ -91,77 +113,231 @@ public ApiRegularIndex create(
...
+      if (optionsDesc == null) {
+        return;
+      }
 
-      if (apiColumnDef.type().typeName() != ApiTypeName.TEXT
-          && apiColumnDef.type().typeName() != ApiTypeName.ASCII) {
+      ApiTypeName targetTypeName =
+          switch (apiColumnDef.type()) {
+            case ApiMapType apiMapType -> {
+              if (indexFunction == ApiIndexFunction.KEYS) {
```
amorton, line 266: nit: this is another good place for switch rather than a sequence of if/then

src/main/java/io/stargate/sgv2/jsonapi/service/schema/tables/ApiIndexFunction.java
Yuqi-Du: good catch, will change it back

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/CreateIndexCommand.java
Yuqi-Du: got it.

src/main/java/io/stargate/sgv2/jsonapi/service/schema/tables/ApiRegularIndex.java
```diff
@@ -91,77 +113,231 @@ public ApiRegularIndex create(
...
+        CqlIdentifier targetIdentifier,
+        ApiMapComponent apiMapComponent,
+        RegularIndexDefinitionDesc indexDesc) {
+      Map<String, String> indexOptions = new HashMap<>();
+      var optionsDesc = indexDesc.options();
+
+      // do NOT support create index for frozen map/set/list columns
+      if (apiColumnDef.type() instanceof CollectionApiDataType<?> collectionApiDataType
+          && collectionApiDataType.isFrozen()) {
+        throw SchemaException.Code.UNSUPPORTED_INDEXING_FOR_FROZEN_COLUMN.get(
```
amorton, line 206: Vector is also a collection type,  this checked elsewhere ?

src/main/java/io/stargate/sgv2/jsonapi/service/schema/tables/ApiRegularIndex.java
```diff
@@ -91,77 +113,231 @@ public ApiRegularIndex create(
...
-                  map.put("unsupportedColumns", errFmt(apiColumnDef));
+                  map.put("targetColumn", errFmt(targetIdentifier));
                 }));
       }
 
-      Map<String, String> indexOptions = new HashMap<>();
-      var optionsDesc = indexDesc.options();
+      // validate user specified index function
+      ApiIndexFunction indexFunction = null;
+      if (apiColumnDef.type() instanceof ApiMapType) {
```
amorton, line 219: -1 for instanceof, check the type name you do not need to cast here

amorton: this  can be a pretty simple single line if 

```java
var indexFunction = apiColumnDef.type().typeName == Map ?
    ApiIndexFunction.fromApiMapComponent(apiMapComponent)
    :
    ApiIndexFunction.VALUES;
```

src/main/java/io/stargate/sgv2/jsonapi/service/schema/tables/ApiRegularIndex.java
```diff
@@ -91,77 +113,231 @@ public ApiRegularIndex create(
...
+    }
+
+    /**
+     * Method to populate the index options map.
+     *
+     * <p>Note, this method should be called after the options are being validated.
+     */
+    private void populateIndexOptions(
+        RegularIndexDefinitionDesc.RegularIndexDescOptions optionsDesc,
+        Map<String, String> indexOptions) {
```
amorton, line 321: nit: there is no need for indexOptions map to be passed in, it can be the return

src/main/java/io/stargate/sgv2/jsonapi/service/schema/tables/ApiSetType.java
```diff
@@ -27,25 +26,30 @@ private ApiSetType(
...
   static ApiSetType from(ApiDataType valueType, boolean isFrozen) {
     Objects.requireNonNull(valueType, "valueType must not be null");
-
-    if (isValueTypeSupported(valueType)) {
-      return new ApiSetType(
-          (PrimitiveApiDataTypeDef) valueType, defaultApiSupport(isFrozen), isFrozen);
-    }
-    throw new IllegalArgumentException(
-        "valueType must be primitive type, valueType%s".formatted(valueType));
+    return new ApiSetType(
```
amorton, line 46: -1 for a non private function assuming validation has been done, please put the isValueTypeSupported() check back

src/main/java/io/stargate/sgv2/jsonapi/service/schema/tables/ApiListType.java
```diff
@@ -27,26 +26,30 @@ private ApiListType(
...
     Objects.requireNonNull(valueType, "valueType must not be null");
-
-    if (isValueTypeSupported(valueType)) {
-      return new ApiListType(
-          (PrimitiveApiDataTypeDef) valueType, defaultApiSupport(isFrozen), isFrozen);
-    }
-
-    throw new IllegalArgumentException(
-        "valueType must be primitive type, valueType%s".formatted(valueType));
+    return new ApiListType(
```
amorton, line 46: -1 for a non private function assuming validation has been done, please put the isValueTypeSupported() check back

src/main/java/io/stargate/sgv2/jsonapi/service/schema/tables/ApiSupportDef.java
```diff
@@ -24,6 +24,15 @@ public interface ApiSupportDef {
...
   boolean createTable();
 
+  /**
+   * The type can be used as component for collection
+   *
+   * <p>E.G. list values, set values, map keys, map values.
+   *
+   * @return <code>true</code> if the type can be used in a createTable statement.
+   */
+  CollectionSupport collectionSupport();
```
amorton, line 34: this whole structure is exposed as support() normally, so lets just rename this property to "collections"  - we know we are talking about support

src/main/java/io/stargate/sgv2/jsonapi/service/schema/tables/ApiSupportDef.java
```diff
@@ -86,47 +118,57 @@ record Support(boolean createTable, boolean insert, boolean read, boolean filter
    * @param read If non-null, will match objects where the read value is the same.
    * @param filter If non-null, will match objects where the filter value is the same.
    */
-  record Matcher(Boolean createTable, Boolean insert, Boolean read, Boolean filter)
+  record Matcher(
```
amorton, line 121: yeah, this was not my best idea maybe. Thanks for updating we may want to kill this later.

src/main/java/io/stargate/sgv2/jsonapi/service/schema/tables/ApiSupportDef.java
```diff
@@ -50,22 +59,45 @@ default boolean isUnsupportedAny() {
...
+   * Record to represent if a dataType is supported as component for map/set/list
+   *
+   * <p>
+   *
+   * @param asListValue If the type can be created as value on a List.
+   * @param asSetValue If the type can be created as value on a Set.
+   * @param asMapKey If the type can be created as map key.
+   * @param asMapValue If the type can be created as map value.
+   */
+  record CollectionSupport(
```
amorton, line 80: Just a thought for when you are doing the update work for collections, we should add a top level APISupportDef propoerty for update()

src/main/resources/errors.yaml
```diff
@@ -693,6 +693,16 @@ request-errors:
       
       Resend the command using only new columns.
 
+  - scope: SCHEMA
+    code: CANNOT_ANALYZE_ENTRIES_ON_MAP_COLUMNS
+    title: Can not create index on map column with index function entries and specify analyze options
+    body: |-
+      Index function `entries` can not apply to map column when analyze options are specified.
+      
+      The table ${keyspace}.${table} defines the columns: ${allColumns}.
```
amorton, line 702: is it possible to say what analyzer options were set ? so the user knows what they got wrong

src/main/resources/errors.yaml
```diff
@@ -805,9 +815,11 @@ request-errors:
     body: |-
       The command attempted to create a map column that used unsupported types for either the key or value.
       
-      Maps can only use `text` or `ascii` types for the key type, and any primitive type for the value.
+      Maps can only use primitive type for the key and value with following exceptions:
+      1. The key can't be duration, counter, timeuuid.
```
amorton, line 819: ok for now, if we open this up again it would be better to have the code supply the lists of supported types , but no need to change now.

src/main/resources/errors.yaml
```diff
@@ -903,6 +915,17 @@ request-errors:
...
       Resend the command using columns of the supported data types.
 
+  - scope: SCHEMA
+    code: UNSUPPORTED_INDEXING_FOR_FROZEN_COLUMN
+    title: Index creation on a frozen column is not supported
+    body: |-
+      The command attempted to create an index on a column that is frozen.
+      
+      The table ${keyspace}.${table} defines the columns: ${allColumns}.
+      '${targetColumn}` is frozen on the table.
```
amorton, line 925: nit:   '${targetColumn}` is frozen on the table.  -> to " The table has the frozen column:  ${targetColumn}"

amorton (approved):
ok to ship some comments and things about code layout and tidy that I will leave to you, great work. 

Am also going to push some changes for javadoc tidy up

src/main/java/io/stargate/sgv2/jsonapi/service/schema/tables/ApiSupportDef.java
Yuqi-Du: got it

src/main/resources/errors.yaml
Yuqi-Du: added

Yuqi-Du:
https://github.com/stargate/data-api/pull/1846#discussion_r1953643722
      
 This won't happen for now, since above checking cover primitive and collection types.
However, any primitive type that is not allow to create index should fall into this error.
Ticket. https://github.com/stargate/data-api/issues/1872

src/main/java/io/stargate/sgv2/jsonapi/service/schema/tables/ApiMapType.java
Yuqi-Du: added back

PR #1859, opened 2025-02-07
Multi Vector Refactor

amorton:
**What this PR does**:

This PR refactors the way Table commands are processed, it builds on the OperationAttempt idea to generalise into a Task that is generic enough to support EmbeddingTask later that can be used to run embedding operations that feed vectors into the next task (such as an insert task). 

This PR itself does not introduce new functionality, that will come in   #1884  and later when we have hybrid search. 

NOTE: there will be things in this PR that are not optimal, if possible we should fix them in #1884 which builds on this or future work. 

Fixes #<issue number>

**Checklist**
- [*] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/service/operation/tasks/DBTask.java
```diff
@@ -0,0 +1,188 @@
...
+  // =================================================================================================
+
+  /**
+   * A {@link UniSupplier} that for fetching a {@link AsyncResultSet}.
+   *
+   * <p>Includes the {@link SimpleStatement} so error mapping can include the CQL if needed. Note
+   * that the statement can change from request to request, e.g. if the first request missed an
+   * ALLOW FILTERING the next may have it.
+   */
+  public static class AsyncResultSetSupplier implements BaseTask.UniSupplier<AsyncResultSet> {
```
tatu-at-datastax, line 67: Would this make sense as main-level class instead of `public` inner one? Or is it only ever used with `DBTask`?

src/main/java/io/stargate/sgv2/jsonapi/service/operation/tasks/DBTaskPage.java
```diff
@@ -0,0 +1,93 @@
...
+ * Base for a page of {@link DBTask}s that have been run, and we want to get a page of results from.
+ *
+ * <p>Implements the {@link Supplier} interface to provide a {@link CommandResult}, which is what is
+ * returned by an {@link Operation}. Subclasses should normally override the {@link
+ * #buildCommandResult()} method, calling the base normally to add the errors and warnings to the
+ * result.
+ *
+ * <p>Implements some re-usable logic for building the {@link CommandResult} .
+ */
+public abstract class DBTaskPage<TaskT extends DBTask<SchemaT>, SchemaT extends SchemaObject>
```
tatu-at-datastax, line 23: First thought it's odd that a (result) Page has refs to Tasks. But maybe some other term than "page" would make sense, assuming they are sets of results (or, result builders?).
Not sure what, just thinking that as a term it's not super clear.
("page of what?")

src/main/java/io/stargate/sgv2/jsonapi/service/operation/tasks/Task.java
```diff
@@ -0,0 +1,155 @@
...
+ *
+ * <p>We have an interface to make it easier for other code to be generic when using a task, the
+ * {@link BaseTask} has the logic we need to retry and manage state etc. so subclass that.
+ *
+ * @param <SchemaT> The type of the schema object that the task works on
+ */
+public interface Task<SchemaT extends SchemaObject>
+    extends Comparable<Task<SchemaT>>, WithWarnings.WarningsSink {
+
+  enum TaskStatus {
```
tatu-at-datastax, line 23: Would it make sense to make this main-level enum, instead of inner one?

src/main/java/io/stargate/sgv2/jsonapi/service/operation/tasks/TaskBuilder.java
```diff
@@ -0,0 +1,75 @@
...
+  protected TaskBuilder(SchemaT schemaObject) {
+    this.schemaObject = Objects.requireNonNull(schemaObject, "schemaObject must not be null");
+  }
+
+  protected int nextPosition() {
+    return taskPosition += 1;
+  }
+
+  @SuppressWarnings("unchecked")
+  public <T extends TaskBuilder<TaskT, SchemaT>> T withExceptionHandlerFactory(
```
tatu-at-datastax, line 31: Similar to Accumulator, these are ugly casts: would it be possible to define SELF type (parameter) for builder case class instead?

src/main/java/io/stargate/sgv2/jsonapi/service/operation/tasks/TaskGroup.java
```diff
@@ -0,0 +1,113 @@
...
+import java.util.*;
+
+/**
+ * Ordered container of {@link Task} to be processed as a group.
+ *
+ * <p>The container can be configured to process the tasks sequentially or in parallel (default),
+ * and other config that is needed about how to process the tasks as a group should live here (e.g.
+ * if there is a delay between tasks, or if the container should fail fast).
+ *
+ * <p>TODO: aaron feb 4 - stop inheriting array list, just expose methods needed
```
tatu-at-datastax, line 13: +1

src/main/java/io/stargate/sgv2/jsonapi/service/operation/tasks/TaskGroup.java
```diff
@@ -0,0 +1,113 @@
...
+ *
+ * @param <SchemaT> Schema object type
+ */
+public class TaskGroup<TaskT extends Task<SchemaT>, SchemaT extends SchemaObject>
```
tatu-at-datastax, line 17: Are these type parameters needed? As in, do TaskGroups need to be typed by `Task` type (and thereby `SchemaObject` type?)

src/main/java/io/stargate/sgv2/jsonapi/service/operation/tasks/TaskGroup.java
```diff
@@ -0,0 +1,113 @@
...
+
+  /**
+   * @return Returns <code>true</code> if the container is configured for sequential processing.
+   */
+  public boolean getSequentialProcessing() {
+    return sequentialProcessing;
+  }
+
+  public List<TaskT> errorTasks() {
+    return stream().filter(task -> task.status() == BaseTask.TaskStatus.ERROR).toList();
```
tatu-at-datastax, line 56: nit: I think `Task.TaskStatus` is simpler reference?

src/main/java/io/stargate/sgv2/jsonapi/service/operation/tasks/TaskRetryPolicy.java
```diff
@@ -0,0 +1,78 @@
...
+ *
+ * <p>To implement a custom retry policy, subclass this class and override {@link
+ * #shouldRetry(Throwable)}.
+ */
+public class TaskRetryPolicy implements PrettyPrintable {
+
+  /** Re-usable policy to not retry. */
+  public static final TaskRetryPolicy NO_RETRY = new TaskRetryPolicy();
+
+  private final int maxRetries;
```
tatu-at-datastax, line 22: Based on usage, isn't this more like "maxTries" than retries? Esp. when `NO_RETRY` is built with `maxRetries` of `1`?

src/main/java/io/stargate/sgv2/jsonapi/service/processor/MeteredCommandProcessor.java
```diff
@@ -29,6 +28,7 @@
 import org.slf4j.LoggerFactory;
 import org.slf4j.MDC;
 
+/** */
```
tatu-at-datastax, line 31: nit: remove (or fill :) )

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/sort/TableCqlSortClauseResolver.java
```diff
@@ -348,7 +352,7 @@ private WithWarnings<OrderByCqlClause> resolveVectorSort(
         new TableOrderByANNCqlClause(
             vectorSortColumn,
             cqlVector,
-            commandContext.getConfig(OperationsConfig.class).maxVectorSearchLimit()),
+            commandContext.config().get(OperationsConfig.class).maxVectorSearchLimit()),
```
tatu-at-datastax, line 355: idea/question: why not have "CommandContext.getConfig()" as before, as convenience method?

tatu-at-datastax (approved):
LGTM. Left a few comments, can be handled post-merge.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/CommandConfig.java
```diff
@@ -0,0 +1,126 @@
...
+  /**
+   * Gets the config service to use, depends on the offline mode.
+   *
+   * <p>TODO: Copied from JsonAPIException , not sure why we need to do this
+   */
+  private static SmallRyeConfig getConfigProvider() {
+    // aaron - copied from JsonAPIException , not sure why we need to do this
+    // TODO - cleanup how we get config, this seem unnecessary complicated
+
+    if (_config_provider == null) {
```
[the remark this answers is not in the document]
amorton: thinking about this, not sure why it is static . I am guessing it was static because of where I copied it from, and being static now is a bug. Making instance and will put a double check sync on it. Rather than sync the function, becuase this will be called from multiple threads / request processing

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/CommandContext.java
```diff
@@ -1,67 +1,113 @@
...
+    // using a sync block here because the context can be accessed by multiple tasks concurrently
+    if (apiFeatures == null) {
+      synchronized (this) {
+        if (apiFeatures == null) {
+          // Merging the config for features with the request headers to get the final feature set
+          apiFeatures =
+              ApiFeatures.fromConfigAndRequest(
+                  commandConfig.get(FeaturesConfig.class), requestContext.getHttpHeaders());
+        }
+        ;
```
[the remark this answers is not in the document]
amorton: fixed

src/main/java/io/stargate/sgv2/jsonapi/service/operation/Operation.java
```diff
@@ -17,7 +20,27 @@
...
-public interface Operation {
-  Uni<Supplier<CommandResult>> execute(
-      DataApiRequestInfo dataApiRequestInfo, QueryExecutor queryExecutor);
+public interface Operation<SchemaT extends SchemaObject> {
+
+  /** Use {@link #execute(CommandContext)} this is for original legacy operations */
+  @Deprecated
+  default Uni<Supplier<CommandResult>> execute(
+      RequestContext dataApiRequestInfo, QueryExecutor queryExecutor) {
+    throw new UnsupportedOperationException(
```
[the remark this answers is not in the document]
amorton: unsure if you mean removing the default or throwing the error. 

need the default so that new tasks do not need to implement the deprecated method. 

and put the error in there incase I got something wrong and this ended up getting called on a new operation. 

can we keep for safety and remove once we stop having old code around ?

src/main/java/io/stargate/sgv2/jsonapi/service/operation/tasks/DBTask.java
amorton: It is only used with the  DBTask subclasses. 

In the next "Big PR (tm)" the EmbeddingTask does not use this result set supplier, nor does the RerankingTask

src/main/java/io/stargate/sgv2/jsonapi/service/operation/Operation.java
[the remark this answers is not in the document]
tatu-at-datastax: I meant removing the method altogether, even declaration, if it is not called from anywhere. Then there's no need for default impl either.

But it sounds like there may be code path(s) that theoretically could call it. If so, yeah, let's keep it until it can be removed.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/tasks/DBTaskPage.java
amorton: good questions, this has been there from the beginning and can be rethought. 

The idea was:The Operation runs the Tasks, when they complete (success or failure) they are sent to the Accumulator, when all  tasks are complete we ask the Accumulator for the result - which is called the Page. 

We probably had that name because to avoid using "Result" when we have CommandResult. And that for things like a read it makes sense, there is a page of results. 

Something to keep iterating on

src/main/java/io/stargate/sgv2/jsonapi/service/operation/tasks/TaskBuilder.java
amorton: yes, I got the fluent pattern wrong all over the place :(  Started improving it in the next PR

src/main/java/io/stargate/sgv2/jsonapi/service/operation/tasks/TaskGroup.java
amorton: 1) the generics got out of control, there is plenty of room for making this simple. 
2) I *think* it needs the generics because we need to know the Task type, so when it comes times to build the results we have the TaskT to do things like get the results for an insert or read task.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/tasks/TaskRetryPolicy.java
amorton: yes, leaving open to change in the next PR

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/sort/TableCqlSortClauseResolver.java
amorton: because I like making things harder, leaving open to change in the next PR

PR #1864, opened 2025-02-11
(WIP) Fix #1848: validate against document part overlap wrt dotted paths

### LINKED ISSUE #1848: Dots in Field names: Shredding, error on overlapping/conflicting Fields
(note: part of #1664; requires #1847 implementation first)

Now that Dots will be allowed in Field names, existing index creation (which is not to be changed for compatibility reasons) could produce conflicts with Documents like this:

```
{
  "price": {
    "usd": 5.0
  },
  "price.usd": 8.50
}
```

(in which index key for two leaf values both would be `"price.usd"`).

This could be highly confusing because different paths CAN be expressed for Projections (with the use of new escaping) but CANNOT be distinguished for Filtering or Sorting (due to way of index keys are constructed and used -- something we cannot change at this point), so let's add basic validation in which a `JsonApiException` is thrown if a conflict is detected.

Note that implementing this has #1847 as a pre-requisite.

tatu-at-datastax:
**What this PR does**:

Adds validation to ensure escaped dotted paths cannot conflict with non-dotted paths (see #1848 for explanation)

Fixes #1848

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/shredding/collections/DocumentShredder.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/shredding/collections/DocumentShredder.java
@@ -184,7 +184,7 @@ public class DocumentShredder {
     new IndexableValueValidator(documentLimits).validate(indexableDocument);
 
     // And finally let's traverse the document to actually "shred" (build index properties)
-    traverse(indexableDocument, b, JsonPath.rootBuilder());
+    new ShreddingTraverser(b).traverse(indexableDocument);
     return b.build();
   }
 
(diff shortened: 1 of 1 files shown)


src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCodeV1.java
```diff
@@ -65,16 +65,18 @@ public enum ErrorCodeV1 {
 
   SHRED_BAD_BINARY_VECTOR_VALUE("Bad binary vector value to shred"),
 
-  SHRED_BAD_DOCUMENT_TYPE("Bad document type to shred"),
```
tatu-at-datastax, line 68: Just moved to be more alphabetically ordered.

src/main/java/io/stargate/sgv2/jsonapi/service/shredding/collections/DocumentShredder.java
```diff
@@ -241,126 +246,6 @@ private static JsonNode wrapExtensionType(
     return jnf.objectNode().put(etype.encodedName(), value.toString());
   }
 
-  /**
```
tatu-at-datastax, line 244: Refactoring: moved inlined traverse methods to a new helper class without logic changes.

src/main/java/io/stargate/sgv2/jsonapi/service/shredding/collections/WritableShreddedDocument.java
```diff
@@ -338,7 +343,12 @@ public void shredVectorize(JsonPath path) {
      * <p>Method will add path to {@link #existKeys}.
      */
     private void addKey(JsonPath key) {
-      existKeys.add(key);
```
tatu-at-datastax, line 341: We basically use and extend existing functionality to avoid traversing multiple times, or storing duplicate Set of paths for common case: we only construct Set if there are duplicates, and then collect all duplicate paths for caller to access.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/InsertIntegrationTest.java
```diff
@@ -1309,6 +1309,42 @@ private void _verifyInsert(String docId, JsonNode doc) {
 
   @Nested
   @Order(4)
+  class InsertOneFails {
```
tatu-at-datastax, line 1312: Separate IT beyond unit tests to ensure we actually have integration working end-to-end.

src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCodeV1.java
```diff
@@ -65,16 +65,18 @@ public enum ErrorCodeV1 {
...
-
   SHRED_BAD_DOCID_TYPE("Bad type for '_id' property"),
 
+  SHRED_BAD_DOCID_EMPTY_STRING("Bad value for '_id' property: empty String not allowed"),
+
+  SHRED_BAD_DOCUMENT_PATH_CONFLICTS("Bad document to shred, contains conflicting Field path(s)"),
```
tatu-at-datastax, line 72: Using V1 errors just because everything in Collection does, easier to convert to v2 as separate step.

Hazel-Datastax (approved):
LGTM

Issue #1854, opened 2025-02-07
Dots in Field names: Projection; use new Path expression codec


tatu-at-datastax:
(note: part of #1664)

(note: directly depends on #1853 )

Once we have Path expression Decoder and ability to insert test documents with "dotted Field names", we need to change Projection code to use decoder and test it works as expected.

So, a simplified example doc like:

```
{
    "id": 123,
    "price.usd": 5.0
}
```

with Projection of:

```
{
  "projection" : {
      "price&.usd": 1
   }
}
```

should only include:

```
{
    "price.usd": 5.0
}
```

Hazel-Datastax:
Done with PR https://github.com/stargate/data-api/pull/1866

PR #1866, opened 2025-02-11
Support dots in projection with ampersand-escapes

### LINKED ISSUE #1853: Dots in Field names: Projection; Implement Path expression handler with ampersand-escapes
(note: part of #1664; will be a pre-req for many child tasks)

As per #1664 we will be allowing (and requiring) use of simple escaping mechanism, for Projection references as well as for Update operations.

Escaping rules are simple: when encoding path for Document like so:

```
{
  "pricing": {
    "price.usd": 5.0
  }
}
```

we have logical path of `["pricing", "price.usd"]` to encode. This is done by:

1. Encoding each segment so that all characters except for `.` and `&` are included as they area; `.` include as `&.` (2-character escape) and `&` as `&&` (2-character escape)
2. Encoded segments are concatenated using `.` as segment separator

and we end up with path like: `"pricing.price&.usd"` which is sent by Data API client in relevant clauses:

```
{
  "projection" : {
      "pricing.price&.usd": 1
   }
}
```

Before use, Data API will need to decode path expression, to get back to "List of Strings". This is done by:

1. Scanning encoded path String character-by-character, building segments
2. If scanned character is `.`, the current segment is complete and is added to the List; new empty segment is started
3. If scanned character is `&`, check following character if any (if not: error) and if it is either `.` or `&`, append that following character into current segment; if it is some other character, error out (only allowed to escape `&` and `.`)
4. Otherwise (scanned character neither `&` nor `.`), append character into current segment.
5. Once end-of-expression reached, the last segment is complete

This algorithm, given String `"pricing.price&.usd"` produces expected output of:

```
    ["pricing", "price.usd"]
```

which is used to construct an Object that can be used for traversing JSON Document as needed for Projection. This replaces existing implementation -- "split by dot" -- which is something like (see `PathMatchLocator.java`)

```
String[] segments = Pattern.compile(Pattern.quote(".")).split(expressionString);
```

This issue is about implementing encoder/decoder, testing, but not yet integrating.

Hazel-Datastax:
**What this PR does**:
Support dots in projection with ampersand-escapes

Fixes #1853 #1854

**Checklist**
- [x] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/service/projection/ProjectionLayer.java
```diff
@@ -108,6 +101,77 @@ private static ProjectionLayer buildLayers(
...
+   * <p>For example, the input {@code "pricing.price&.usd"} results in: {@code ["pricing",
+   * "price.usd"]}.
+   *
+   * @param path the encoded path string.
+   * @return an array of decoded segments.
+   */
+  private static String[] decodePath(String path) {
+    List<String> segments = new ArrayList<>();
+    StringBuilder segment = new StringBuilder();
+
```
Hazel-Datastax, line 143: There are some similar codes in `PathMatchLocator` that need to be replaced by this method. That one is more related to the update operation, so I didn't change. I am thinking of extracting this method to some util class. @tatu-at-datastax Do you think it's good to move it to `FieldNamingRule`?

src/test/java/io/stargate/sgv2/jsonapi/api/v1/FindIntegrationTest.java
```diff
@@ -329,98 +330,6 @@ public void byId() {
                       """));
     }
 
-    // For [json-api#634]: empty Object as Projection should work same as missing one,
-    // that is, include everything
-    @Test
```
Hazel-Datastax, line 334: I didn't move, just refactor and group projection tests to another nested class

src/main/java/io/stargate/sgv2/jsonapi/service/projection/ProjectionLayer.java
```diff
@@ -108,6 +101,77 @@ private static ProjectionLayer buildLayers(
...
+    for (int i = 0; i < path.length(); i++) {
+      char ch = path.charAt(i);
+      if (ch == '.') {
```
vkarpov15, line 146: Also, what about handling multiple consecutive dots? `foo..bar`?

tatu-at-datastax: Those infer empty Segments. Not sure if we should error on those, given that they are not legal (no empty Path segments, cannot Insert Documents with empty Object property names). If we accept them, path can never match.

src/test/java/io/stargate/sgv2/jsonapi/service/projection/DocumentProjectorTest.java
```diff
@@ -170,6 +171,87 @@ public void verifyNoDollarSimilarity() throws Exception {
...
+                              "a b": "456"
+                          }
+                      }
+                      """;
+      final String projectionString =
+          """
+                      {
+                          "pricing.price&.usd": 1,
+                          "pricing.price&&jpy": 1,
+                          "pricing.price&&&.aud": 1,
```
tatu-at-datastax, line 196: Might be better to have a test case where only some of paths are included -- otherwise test could work simply because the whole document was included?

Would also be good to have a nested entry without escaped characters.

src/main/resources/errors.yaml
```diff
@@ -612,6 +612,18 @@ request-errors:
       
       Resend the command using only supported column types.
 
+  - scope: PROJECTION
+    code: UNSUPPORTED_AMPERSAND_ESCAPE_USAGE
+    title: The usage of ampersand escape is not supported
+    body: |-
+      The usage of ampersand escape is not supported.
+      
+      Ampersand `&` escape is only allowed to escape and followed by ampersand `&` and dot `.`.
```
tatu-at-datastax, line 621: -> "Ampersand `&` can only be used as an escape character and must be followed by a `&` or `.` character"

src/main/java/io/stargate/sgv2/jsonapi/service/projection/ProjectionLayer.java
```diff
@@ -108,6 +101,77 @@ private static ProjectionLayer buildLayers(
...
+   *       <p>When the end of the expression is reached, consider the last segment complete.
+   * </ol>
+   *
+   * <p>For example, the input {@code "pricing.price&.usd"} results in: {@code ["pricing",
+   * "price.usd"]}.
+   *
+   * @param path the encoded path string.
+   * @return an array of decoded segments.
+   */
+  private static String[] decodePath(String path) {
```
tatu-at-datastax, line 140: Since we will be reusing the logic, we need a separate object to represent decoded path. That could be... `ProjectionPath`? Decoding could be a static factory method in that class.

src/main/java/io/stargate/sgv2/jsonapi/service/projection/ProjectionLayer.java
tatu-at-datastax: Ah. Ok added my comment above first before responding to this.

No, I don't think this is related to `NamingRule`s. Something like `ProjectionPath` would make sense -- even if used from other places.

We also need to allow encoding of Path for Filter and Sort use cases, where we basically need to first decode path Object, but then go ahead and encode String -- but one without escapes (since Filter paths, keys of Maps like `query_text_values`, do not support escaping).
So, for example for filtering we could have:

    "pricing.price&.usd" -> Path object -> "pricing.price.usd"

we discussed this with Aaron and decided that although we can't really benefit from escaping there, it is better for uniformness, escaping always expected with paths.

tatu-at-datastax (commented):
Looks like a good start but need to extract some kind of Path object to reuse functionality from multiple places.

src/main/java/io/stargate/sgv2/jsonapi/service/projection/ProjectionLayer.java
Hazel-Datastax: Error out if it's an empty segment.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/FindIntegrationTest.java
```diff
@@ -2152,6 +2032,269 @@ public void notWithComparisonExpressionAndLogicalExpression() {
...
+          .post(CollectionResource.BASE_PATH, keyspaceName, collectionName)
+          .then()
+          .statusCode(200)
+          .body("$", responseIsFindSuccess())
+          .body("data.documents", hasSize(1))
+          .body("data.documents[0]", jsonEquals("{}"));
+    }
+
+    @Test
+    public void withAmpersandEscape() {
```
Hazel-Datastax, line 2177: New added ITs (`withAmpersandEscape`, `failWithAmpersandEscapeAtTheEnd`, `failWithAmpersandEscapeNotFollowedByAmpersandOrDot`, `failWithEmptySegment`) start from here

Hazel-Datastax:
## Projection cases summary:

### Include All Cases:
```
"projection": ""
"projection": " "
"projection": "abc"
"projection": []
"projection": {}
"projection": {"*": 1}
```

### Invalid Projection Path Cases (with error message):
```
// Unsupported projection parameter: empty paths (and path segments) not allowed
"projection": {"": 1} 

// Ampersand & can only be used as an escape character and must be followed by a & or . character.
"projection": {"price&abc": 1} 
"projection": {"price&": 1} 

// The segments from the path in the projection cannot be empty.
"projection": {".path": 1} 
"projection": {"foo..bar": 1} 
"projection": {"path.": 1} 
```

### Maybe need to handle case
```
"projection": {" ": 1} 
```
It's valid to have a blank string now, but this will only project the field " " (black string)

### Need to handle cases
If insert: `"document": {"*": "123"}`, and use projection `"projection": {"*": 1}`.
This will project all, not only the field name matches `*`
We could either forbid users from using `*` as the field name or use `&` to escape it

tatu-at-datastax:
On these:

> Maybe need to handle case
>.   "projection": {" ": 1} 
> It's valid to have a blank string now, but this will only project the field " " (blank string)

This is fine, I think, given that blank String is valid (if strange) key. Otherwise there would be no way to refer to it.

> Need to handle case
If insert: "document": {"*": "123"}, and use projection "projection": {"*": 1}.
> This will project all, not only the field name matches *
> We could either forbid users from using * as the field name (also need the change in the insertion) or use & to escape it

I think we have some ideas, but for now behavior is fine. Users would be well-advised just not to use main level `"*"` key.

But the idea Aaron suggested was to support special path of `$*`, for projections. It would be used instead of current "plain" `*` to match-all/exclude-all so projections would be:

```
{"$*": 1} // include all
{"$*": 0} // include nothing (exclude all)
```

If we add that, it should be via separate PR.

src/main/java/io/stargate/sgv2/jsonapi/service/projection/ProjectionLayer.java
```diff
@@ -89,49 +81,51 @@ private static ProjectionLayer buildLayers(
...
-          new String[] {DocumentConstants.Fields.DOC_ID});
+          ProjectionPath.from(DocumentConstants.Fields.DOC_ID));
     }
     if (add$vector) {
       buildPath(
           failOnOverlap,
           DocumentConstants.Fields.VECTOR_EMBEDDING_FIELD,
           root,
-          new String[] {DocumentConstants.Fields.VECTOR_EMBEDDING_FIELD});
+          ProjectionPath.from(DocumentConstants.Fields.VECTOR_EMBEDDING_FIELD));
```
tatu-at-datastax, line 91: Same here, static singleton `ProjectionPath` so we do not need to always decode it.

src/main/java/io/stargate/sgv2/jsonapi/service/projection/ProjectionLayer.java
```diff
@@ -89,49 +81,51 @@ private static ProjectionLayer buildLayers(
...
     }
     if (add$vectorize) {
       buildPath(
           failOnOverlap,
           DocumentConstants.Fields.VECTOR_EMBEDDING_TEXT_FIELD,
           root,
-          new String[] {DocumentConstants.Fields.VECTOR_EMBEDDING_TEXT_FIELD});
+          ProjectionPath.from(DocumentConstants.Fields.VECTOR_EMBEDDING_TEXT_FIELD));
```
tatu-at-datastax, line 98: and here.

src/main/java/io/stargate/sgv2/jsonapi/service/projection/ProjectionLayer.java
```diff
@@ -89,49 +81,51 @@ private static ProjectionLayer buildLayers(
...
     }
     return root;
   }
 
   static void buildPath(
-      boolean failOnOverlap, String fullPath, ProjectionLayer layer, String[] segments) {
+      boolean failOnOverlap, String fullPath, ProjectionLayer layer, ProjectionPath path) {
+    List<String> segments = path.getSegments();
```
tatu-at-datastax, line 105: Instead of accessor to `segments` (`List<String>`), should we instead have `String segment(int i)` accessor in `ProjectionPath` itself?

src/main/java/io/stargate/sgv2/jsonapi/service/projection/ProjectionLayer.java
```diff
@@ -195,21 +189,22 @@ void addSlicer(boolean failOnOverlap, String fullPath, String segment, Slicer sl
...
    */
   public boolean isPathIncluded(String path) {
-    final String[] segments = DOT.split(path);
+    ProjectionPath p = ProjectionPath.from(path);
+    final List<String> segments = p.getSegments();
     return isPathIncluded(segments, 0);
   }
 
-  private boolean isPathIncluded(String[] segments, int index) {
+  private boolean isPathIncluded(List<String> segments, int index) {
```
tatu-at-datastax, line 197: ... and if we have `ProjectionPath.segment()` can (and should) just pass `ProjectionPath` here.

src/main/java/io/stargate/sgv2/jsonapi/service/projection/ProjectionPath.java
```diff
@@ -0,0 +1,152 @@
...
+ * ampersands within segments.
+ *
+ * <p>Use {@link #from(String)} to create an instance, {@link #getSegments()} to access the
+ * segments, and {@link #encode()} to get the dot-separated path string.
+ */
+public final class ProjectionPath {
+  private final List<String> segments;
+
+  private ProjectionPath(List<String> segments) {
+    // Make a defensive copy to preserve immutability.
```
tatu-at-datastax, line 23: One benefit of removing accessor to segments is that we can avoid creating defensive copy.

src/main/java/io/stargate/sgv2/jsonapi/service/projection/ProjectionPath.java
```diff
@@ -0,0 +1,152 @@
...
+    this.segments = Collections.unmodifiableList(new ArrayList<>(segments));
```
tatu-at-datastax, line 24: If we make a defensive copy, should only use one of `Collections.unmodifiableList()` OR `new ArrayList<>(oldList))` -- not both.

src/main/java/io/stargate/sgv2/jsonapi/service/projection/ProjectionPath.java
```diff
@@ -0,0 +1,152 @@
...
+    Objects.requireNonNull(path, "path cannot be null");
+    return new ProjectionPath(decode(path));
+  }
+
+  /**
+   * Returns the list of decoded segments.
+   *
+   * @return an unmodifiable list of segments
+   */
+  public List<String> getSegments() {
```
tatu-at-datastax, line 45: As per earlier notes, replace with two methods:

* `public int segment(ind index)`
* `public int segmentCount()` (or, if you prefer, `int size()`

src/main/java/io/stargate/sgv2/jsonapi/service/projection/ProjectionPath.java
```diff
@@ -0,0 +1,152 @@
...
+    return segments;
+  }
+
+  /**
+   * Encodes the ProjectionPath back into a string by joining segments with a dot. This encoding is
+   * used for filter and sort use cases, where escapes are not allowed.
+   *
+   * @return the encoded path string (e.g., "pricing.price.usd")
+   */
+  public String encode() {
```
tatu-at-datastax, line 55: Makes sense, but should the name to indicate no escapes? Like `encodePlain()` or `encodeNoEscaping()`?

src/test/java/io/stargate/sgv2/jsonapi/api/v1/FindIntegrationTest.java
```diff
@@ -2152,6 +2032,269 @@ public void notWithComparisonExpressionAndLogicalExpression() {
...
+                              "projection": { "_id":0, "username":1 }
+                            }
+                          }
+                          """;
+
+      String expected =
+          """
+                  {"username":"user1"}
+                  """;
+      given()
```
tatu-at-datastax, line 2071: Forgot to mention earlier, but for new tests (which this may or not be), we can use helper methods from `AbstractCollectionIntegrationTestBase`: for passing (non-error) calls

```
    givenHeadersPostJsonThenOkNoErrors(json)
       .body("$", ...)
```

and error cases

`givenHeadersPostJsonThenOk(json)`

that makes tests bit less verbose and easier to read (can focus on actual request and response).

tatu-at-datastax (commented):
Good, much better. Still some work to do to encapsulate `List<String> segments` within `ProjectionPath` (and not exposed outside as `List` but only segment-by-segment).
And can define 3 constant paths to refer to doc id (`_id`) and `$vector` / `$vectorize` paths, to be re-used (minor performance thing but easy to do).

src/main/java/io/stargate/sgv2/jsonapi/service/projection/ProjectionLayer.java
```diff
@@ -89,49 +81,51 @@ private static ProjectionLayer buildLayers(
           failOnOverlap,
           DocumentConstants.Fields.DOC_ID,
           root,
-          new String[] {DocumentConstants.Fields.DOC_ID});
+          ProjectionPath.from(DocumentConstants.Fields.DOC_ID));
```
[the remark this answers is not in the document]
tatu-at-datastax: Actually, constants could be created alternatively just by `ProjectionLayer` since this is where they are needed?
I'll let you decide what is the best/logical place.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/FindIntegrationTest.java
```diff
@@ -2152,6 +2032,269 @@ public void notWithComparisonExpressionAndLogicalExpression() {
...
+    public void failWithAmpersandEscapeNotFollowedByAmpersandOrDot() {
+      given()
+          .headers(getHeaders())
+          .contentType(ContentType.JSON)
+          .body(
+              """
+                        {
+                            "find": {
+                                "projection" : {
+                                    "price&abc": 1
```
vkarpov15, line 2248: I'd also like to see a test case with `price&&abc`

Hazel-Datastax: I do have one in the unit test (ProjectionPathTest)
"pricing.price&&jpy" -> ["pricing", "price&jpy"]
I could add one in the integration test

tatu-at-datastax: Quick note: ITs do not (and generally, should not) duplicate every unit test scenario. In this case it's good to have wider unit test coverage, and then an IT or two to ensure validation is triggered, but does not have verify all of them (if one is triggered, all are).

So: for unit tests, cover all different types of failures, for ITs, one or two.

Obviously judgment call how much overlap is ideal, but I wanted to share the idea. Test code is code to maintain so we do not want maximize test code but optimize (balance between value of tests vs cost of maintaining code)

src/main/java/io/stargate/sgv2/jsonapi/service/projection/ProjectionLayer.java
```diff
@@ -89,49 +81,51 @@ private static ProjectionLayer buildLayers(
...
   static void buildSlicer(boolean failOnOverlap, SliceDef slice, ProjectionLayer layer) {
     final String fullPath = slice.path;
-    String[] segments = DOT.split(fullPath);
-    final int last = segments.length - 1;
+    ProjectionPath path = ProjectionPath.from(fullPath);
+    List<String> segments = path.getSegments();
+    final int last = segments.size() - 1;
     for (int i = 0; i < last; ++i) {
-      layer = layer.findOrCreateBranch(failOnOverlap, fullPath, segments[i]);
+      layer = layer.findOrCreateBranch(failOnOverlap, fullPath, segments.get(i));
```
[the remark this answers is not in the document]
Hazel-Datastax: Not sure about the logic in `findOrCreateBranch()`, @tatu-at-datastax what do you think?

tatu-at-datastax (approved):

PR #1870, opened 2025-02-12
Fix #1849: allow filtering with "dotted names", test (but before escaping added)

### LINKED ISSUE #1849: Dots in Field names: Filtering, verify functioning with new "dotted" names, old
(note: part of #1664; requires implementation of #1847 first)

When Dots will be allowed in Field names, we need to ensure that references from `FilterClause` to both old "simple" Fields and new "dotted" names work. That is, that we can find documents like this:

```
{
  "price.usd": 20,
  "metadata": {
    "shape": "square"
  }
}
```

where references to fields would be:

* `price.usd` for the first field
* `metadata.shape` for the second field

NOTE: no escaping mechanism used (not required but not yet even supported) for Filter clauses.

This will require changes to `FilterClause` validation for the first case; and requires implementation of #1847 (relaxed Shredding validation) first.

tatu-at-datastax:
**What this PR does**:

Adds tests to ensure new "dotted paths" can be filtered as expected (before escaping)

Fixes #1849

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/src/test/java/io/stargate/sgv2/jsonapi/api/v1/FindOneIntegrationTest.java
+++ b/src/test/java/io/stargate/sgv2/jsonapi/api/v1/FindOneIntegrationTest.java
@@ -1,13 +1,11 @@
 package io.stargate.sgv2.jsonapi.api.v1;
 
-import static io.restassured.RestAssured.given;
 import static io.stargate.sgv2.jsonapi.api.v1.ResponseAssertions.*;
 import static net.javacrumbs.jsonunit.JsonMatchers.jsonEquals;
 import static org.hamcrest.Matchers.*;
 
 import io.quarkus.test.common.WithTestResource;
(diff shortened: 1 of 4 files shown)


src/test/java/io/stargate/sgv2/jsonapi/api/v1/FindOneIntegrationTest.java
```diff
@@ -925,13 +642,77 @@ public void inConditionForUUIDField() {
   @Nested
   @Order(3)
   @TestMethodOrder(MethodOrderer.OrderAnnotation.class)
+  class FindOneWithDottedPaths {
```
tatu-at-datastax, line 645: And here are actual tests: for now just 2 tests. When add support for escaping will add more, covering ampersands -- right now just matches against now legal dotted paths

Hazel-Datastax (approved):
LGTM

PR #1871, opened 2025-02-13
[Table] Add error message when inserting data using multiple vectorize configs

### LINKED ISSUE #1752: Table vectorize, different providers, different dimension, 500 Internal Server

Hazel-Datastax:
**What this PR does**:
Add error message when inserting data using multiple vectorize configs
Previous [PR 1785](https://github.com/stargate/data-api/pull/1785) was against hotfix branch because we thought we could refactor vectorize and support multi providers in main. We still cannot support it, and we need this error message in the next release

Fixes #1752 

**Checklist**
- [x] Changes manually tested


Yuqi-Du (approved):
LGTM, change from hot-fix

PR #1888, opened 2025-02-21
fix $in filter result document count mismatch limit issue for collection

Yuqi-Du:
Issue, when use $in for _id with a limit, the result document count is not derterministic.

Example
```
>>> [doc['_id'] for doc in vstore.astra_env.collection.find()]
['10', '8', '1', '9', '5', '6', '7', '2', '4', '3']
>>> 
>>> 
>>> len(list(vstore.astra_env.collection.find({'_id': {'$in': ['5', '6', '7', '8']}},limit=1)))
1
>>> len(list(vstore.astra_env.collection.find({'_id': {'$in': ['5', '6', '7', '8']}},limit=2)))
4
>>> len(list(vstore.astra_env.collection.find({'_id': {'$in': ['5', '6', '7', '8']}},limit=3)))
4
>>> len(list(vstore.astra_env.collection.find({'_id': {'$in': ['5', '6', '7', '8']}},limit=4)))
4
>>> len(list(vstore.astra_env.collection.find({'_id': {'$in': ['5', '6', '7', '8']}},limit=5)))
4
```

Fix, since Data API makes multiple queries for this filter, then the limit must apply to the documents returned by all the queries.

**Checklist**
- [x] Changes manually tested


src/test/java/io/stargate/sgv2/jsonapi/api/v1/InAndNinIntegrationTest.java
```diff
@@ -205,6 +209,40 @@ public void inCondition() {
...
+                              """
+              .formatted(filter, limit);
+      given()
+          .headers(getHeaders())
+          .contentType(ContentType.JSON)
+          .body(json)
+          .when()
+          .post(CollectionResource.BASE_PATH, keyspaceName, collectionName)
+          .then()
+          .statusCode(200)
```
Hazel-Datastax, line 241: Replace this with `givenHeadersPostJsonThenOkNoErrors` to reduce duplicated code

src/main/java/io/stargate/sgv2/jsonapi/service/operation/collections/CollectionReadOperation.java
```diff
@@ -165,6 +165,11 @@ default Uni<FindResponse> findDocument(
                   // picking the last page state
                   tempPageState = response.pageState();
                 }
+                // when $in apply to _id, the limit need to restricted to all the documents returned
+                // by multiple queries
+                if (documents.size() > limit) {
+                  documents = documents.subList(0, limit);
+                }
```
Hazel-Datastax, line 172: In terms of fixing the bug, ok, adding this block of code could solve the problem.
But I don't understand the whole `transform` logic here. Why do we want to specially handle the limit=1 case? And when limit=1, we don't return the pageState?
Maybe consider refactoring the whole block?

src/main/java/io/stargate/sgv2/jsonapi/service/operation/collections/CollectionReadOperation.java
```diff
@@ -165,6 +165,11 @@ default Uni<FindResponse> findDocument(
                   // picking the last page state
                   tempPageState = response.pageState();
                 }
+                // when $in apply to _id, the limit need to restricted to all the documents returned
+                // by multiple queries
```
Hazel-Datastax, line 169: I don't understand the comment here. `when $in apply to _id` - Why does it only apply to `_id` not others?

src/main/java/io/stargate/sgv2/jsonapi/service/operation/collections/CollectionReadOperation.java
tatu-at-datastax: I agree that refactoring would be nice. Fix as-is makes sense aside from that.

This big method is also only called from just one place so while big, it is in some ways also unnecessary. But maybe could be split into separate ones, if that would allow less if... else... style conditionals.
But refactoring may be bigger effort than what makes sense for this particular fix.

tatu-at-datastax (approved):
Looks good, although I thin @Hazel-Datastax is right wrt this area being ripe for refactoring (if and only if it's not too big a task for this fix)

src/main/java/io/stargate/sgv2/jsonapi/service/operation/collections/CollectionReadOperation.java
Yuqi-Du: I simplify and abstract out a separate method, which will be more clear to explain the limit and pageState.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/collections/CollectionReadOperation.java
```diff
@@ -165,6 +165,11 @@ default Uni<FindResponse> findDocument(
                   // picking the last page state
```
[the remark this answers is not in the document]
Yuqi-Du: handled in separate method.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/collections/CollectionReadOperation.java
Yuqi-Du: fixed

tatu-at-datastax (approved):
LGTM

PR #1890, opened 2025-02-22
Support Escape in Filter

### LINKED ISSUE #1850: Dots in Field names: Filtering, REQUIRE use of new Path expression (with escapes)
(note: part of #1664; requires implementation of #1849 first. And finally, requires handler for path expressions (issue to be filed))

So far filtering does not make use of new Path expression, but with #1892 it could and should.

So, for document like:

```
{
  "price.usd": 20
}
```

user should ise either following reference for filtering:

```
"filter" : {"price&.usd" : {"$eq": 20}}
```

to match Document, whereas attempting to use:

```
"filter" : {"price.usd" : {"$eq": 20}}
```

would NOT match the document.

Hazel-Datastax:
**What this PR does**:
Support Escape in Filter

Fixes #1850 

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/builders/FilterClauseBuilder.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/builders/FilterClauseBuilder.java
@@ -550,7 +550,7 @@ public abstract class FilterClauseBuilder<T extends SchemaObject> {
             "filter clause path cannot be empty String");
       }
       throw ErrorCodeV1.INVALID_FILTER_EXPRESSION.toApiException(
-          "filter clause path ('%s') cannot start with `$`", normalizePath);
+          "filter clause path ('%s') cannot start with `$`", path);
     }
     return normalizePath;
   }
--- a/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/builders/FilterClauseBuilder.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/builders/FilterClauseBuilder.java
@@ -10,7 +10,6 @@ import io.stargate.sgv2.jsonapi.config.constants.DocumentConstants;
 import io.stargate.sgv2.jsonapi.exception.ErrorCodeV1;
 import io.stargate.sgv2.jsonapi.service.cqldriver.executor.SchemaObject;
 import io.stargate.sgv2.jsonapi.service.cqldriver.executor.TableSchemaObject;
-import io.stargate.sgv2.jsonapi.service.projection.ProjectionPath;
 import io.stargate.sgv2.jsonapi.service.schema.collections.CollectionSchemaObject;
 import io.stargate.sgv2.jsonapi.service.schema.naming.NamingRules;
 import io.stargate.sgv2.jsonapi.service.shredding.collections.DocumentId;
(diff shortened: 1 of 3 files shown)


src/main/java/io/stargate/sgv2/jsonapi/service/resolver/matcher/FilterResolver.java
```diff
@@ -62,7 +61,6 @@ public WithWarnings<DBLogicalExpression> resolve(
       CommandContext<SchemaT> commandContext, CmdT command) {
     Preconditions.checkNotNull(commandContext, "commandContext is required");
     Preconditions.checkNotNull(command, "command is required");
-    final FilterClause filterClause = command.filterClause(commandContext);
```
Hazel-Datastax, line 65: remove as `filterClause` is not used

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/builders/FilterClauseBuilder.java
```diff
@@ -542,14 +543,15 @@ private void flip(LogicalExpression logicalExpression) {
   }
 
   private String validateFilterClausePath(String path) {
-    if (!NamingRules.FIELD.apply(path)) {
-      if (path.isEmpty()) {
+    String normalizePath = ProjectionPath.from(path).encodeNoEscaping();
```
Hazel-Datastax, line 546: We need to reuse the decode method in ProjectionPath. 
But if the path is not valid, f.ex. foo..bar, the exception will be thrown in the decode method.
The first problem is, the error message will say "projection path", not "filter path", which is confusing.
The second problem is, I used V2 error in the decode method, so the `ProjectionException` will be mapped to `SERVER_UNHANDLED_ERROR`
Here is the example of the returned error:
```
{
    "errors": [
        {
            "message": "Server failed: root cause: (io.stargate.sgv2.jsonapi.exception.ProjectionException) The path in the projection is not supported.\n\nThe segments from the path in the projection cannot be empty.\n\nThe command used the unsupported projection path: 'foo..bar'.\n\nResend the command using supported projection path.",
            "id": "349104b4-68fb-47b1-82be-21aedb0be65a",
            "family": "SERVER",
            "title": "Server failed",
            "scope": "DATABASE"
        }
    ]
}
```

tatu-at-datastax, line 546: Also: based on our discussions, I think we need to add something like

    String encoded;

which stores the original, encoded path, and then we probably would use either:

     String normalizedPath = ProjectionPath.from(path).encode();

or if separating validation just

    ProjectionPath.validate(path);

and use `path` as is.

Hazel-Datastax: We no longer need the encode-decode process after PR #1895, so I remove the added code. `ProjectionException` will still be a potential issue, will fix it in a new PR

Hazel-Datastax: Same as the above, no longer a problem here

Hazel-Datastax:
Since we don't need the extra thing to support escape in filter after PR #1895, I remove most of the part in this PR and just left the test. I also added some tests in that PR (see test [FindOneFilterWithDottedPaths](https://github.com/stargate/data-api/blob/main/src/test/java/io/stargate/sgv2/jsonapi/api/v1/FindOneIntegrationTest.java#L645C9-L645C37)). So the tests are comprehensive

tatu-at-datastax (approved):
LGTM

PR #1891, opened 2025-02-23
Support Escape in Sort

### LINKED ISSUE #1852: Dots in Field names: Sorting, REQUIRE use of new Path expression (with escapes)
(note: part of #1664; requires implementation of #1851 first. And finally, requires handler for path expressions (issue to be filed))
(note similar to #1850 but for Sorting)

So far sorting does not make use of new Path expression, but with https://github.com/stargate/data-api/issues/1892 it could and should.

So, for document like:

```
{
  "price.usd": 20
}
```

user needs to use following reference for filtering:

```
"sort" : {"price&.usd": 1}
```

to match given document, whereas following would NOT match:

```
"sort" : {"price.usd": 1}
```

since it would only match Document with nesting, that is:

```
{
  "price": {
     "usd": 1
  }
}
```

So: use of escape-by-ampersand is required. 

NOTE: this issue depends on #1892 being completed first.

Hazel-Datastax:
**What this PR does**:
Support Escape in Sort

Fixes #1852

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/deserializers/SortClauseDeserializer.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/deserializers/SortClauseDeserializer.java
@@ -11,7 +11,6 @@ import io.stargate.sgv2.jsonapi.api.model.command.clause.sort.SortClause;
 import io.stargate.sgv2.jsonapi.api.model.command.clause.sort.SortExpression;
 import io.stargate.sgv2.jsonapi.config.constants.DocumentConstants;
 import io.stargate.sgv2.jsonapi.exception.ErrorCodeV1;
-import io.stargate.sgv2.jsonapi.service.projection.ProjectionPath;
 import io.stargate.sgv2.jsonapi.service.schema.naming.NamingRules;
 import java.io.IOException;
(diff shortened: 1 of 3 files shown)


src/main/java/io/stargate/sgv2/jsonapi/api/model/command/deserializers/SortClauseDeserializer.java
```diff
@@ -191,14 +192,16 @@ private static float[] arrayNodeToVector(ArrayNode arrayNode) {
     return arrayVals;
   }
 
-  private void validateSortClausePath(String path) {
-    if (!NamingRules.FIELD.apply(path)) {
-      if (path.isEmpty()) {
+  private String validateSortClausePath(String path) {
+    String normalizePath = ProjectionPath.from(path).encodeNoEscaping();
```
Hazel-Datastax, line 196: same problem as the filter

Hazel-Datastax:
Similar to the previous filter PR #1890. We don't need the extra thing to support escape in sort after PR https://github.com/stargate/data-api/pull/1895, I removed most of the part in this PR and just left the tests

tatu-at-datastax (approved):
LGTM

PR #1894, opened 2025-02-26
fix NPE when the vectorize model is not supported

Yuqi-Du:
This is the fix for bug regarding `service-url-override` for nvidia.

If a DB has a collection with old Nvidia model setting in it, that will cause any CollectionCommand failure against this collection.
NPE in this line : 
`return modelUrlOverrides != null ? modelUrlOverrides.get(modelName).orElse(baseUrl) : baseUrl;`
```

"message": "Server failed: root cause: (java.lang.NullPointerException) Cannot invoke \"java.util.Optional.orElse(Object)\" because the return value of \"java.util.Map.get(Object)\" is null",
```

This will cause Astra Portal break down, since they send countCommand for every collection.

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/embedding/configuration/EmbeddingProviderConfigStore.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/embedding/configuration/EmbeddingProviderConfigStore.java
@@ -1,5 +1,6 @@
 package io.stargate.sgv2.jsonapi.service.embedding.configuration;
 
+import io.stargate.sgv2.jsonapi.exception.ErrorCodeV1;
 import java.util.Map;
 import java.util.Optional;
 
@@ -35,6 +36,11 @@ public interface EmbeddingProviderConfigStore {
     }
 
     public String getBaseUrl(String modelName) {
+      if (modelUrlOverrides != null && modelUrlOverrides.get(modelName) == null) {
+        throw ErrorCodeV1.VECTORIZE_MODEL_DEPRECATED.toApiException(
+            "Mode %s is deprecated, supported models for provider %s are %s",
+            modelName, serviceName, modelUrlOverrides.keySet());
+      }
       return modelUrlOverrides != null ? modelUrlOverrides.get(modelName).orElse(baseUrl) : baseUrl;
     }
   }
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/embedding/configuration/EmbeddingProviderConfigStore.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/embedding/configuration/EmbeddingProviderConfigStore.java
@@ -38,7 +38,7 @@ public interface EmbeddingProviderConfigStore {
     public String getBaseUrl(String modelName) {
       if (modelUrlOverrides != null && modelUrlOverrides.get(modelName) == null) {
         throw ErrorCodeV1.VECTORIZE_MODEL_DEPRECATED.toApiException(
-            "Mode %s is deprecated, supported models for provider %s are %s",
+            "Model %s is deprecated, supported models for provider '%s' are %s",
             modelName, serviceName, modelUrlOverrides.keySet());
       }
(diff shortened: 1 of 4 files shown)


src/main/java/io/stargate/sgv2/jsonapi/service/embedding/configuration/EmbeddingProviderConfigStore.java
```diff
@@ -35,6 +36,11 @@ public static ServiceConfig custom(Optional<Class<?>> implementationClass) {
     }
 
     public String getBaseUrl(String modelName) {
+      if (modelUrlOverrides != null && modelUrlOverrides.get(modelName) == null) {
+        throw ErrorCodeV1.VECTORIZE_MODEL_DEPRECATED.toApiException(
+            "Mode %s is deprecated, supported models for provider %s are %s",
```
tatu-at-datastax, line 41: typo: Mode -> Model.

Also, should surround second %s's with quotes, so: 

    "Model %s is deprecated, supported models for provider '%s' are: %s"

tatu-at-datastax (commented):
Makes sense. Is there an easy way to write a test to verify?

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/configuration/EmbeddingProviderConfigStore.java
```diff
@@ -35,6 +36,11 @@ public static ServiceConfig custom(Optional<Class<?>> implementationClass) {
     }
 
     public String getBaseUrl(String modelName) {
+      if (modelUrlOverrides != null && modelUrlOverrides.get(modelName) == null) {
+        throw ErrorCodeV1.VECTORIZE_MODEL_DEPRECATED.toApiException(
```
tatu-at-datastax, line 40: Code was added to `SchemaException`, but here reference made via `ErrorCodeV1`, does not compile.

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/configuration/EmbeddingProviderConfigStore.java
Yuqi-Du: thanks, fixed

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/configuration/EmbeddingProviderConfigStore.java
Yuqi-Du: sorry, commit the wrong file, first attempt was to try to make the v2 error, but since it is not going through the commandProcessor, so switch to v1 error.

Yuqi-Du:
> Makes sense. Is there an easy way to write a test to verify?

added an integration test, thanks

tatu-at-datastax (approved):
LGTM!

src/test/java/io/stargate/sgv2/jsonapi/api/v1/VectorizeSearchIntegrationTest.java
```diff
@@ -1331,4 +1332,56 @@ public void createDropDifferentVectorDimension() {
...
+                        { "findOne": {} }
+                    """;
+      given()
+          .headers(getHeaders())
+          .contentType(ContentType.JSON)
+          .body(json)
+          .when()
+          .post(CollectionResource.BASE_PATH, keyspaceName, collection)
+          .then()
+          .statusCode(200)
```
Hazel-Datastax, line 1376: Replace this with `givenHeadersPostJsonThenOk` method? It could reduce some duplicate code

Yuqi-Du: thanks, will merge for now, this class did not use givenHeadersPostJsonThenOk before, it feels weird not do the conversion for the entire class.

PR #1895, opened 2025-02-26
Support Escape in Shredding

### LINKED ISSUE #1892: Dots in Field names: Shredding, escape field names (for `.` and `&`) in query index columns
(note: part of #1664 -- but continuation of #1847) 

Currently (super)shredding uses simple concatenation of field paths, for creating fields needed to filter and sort document. So, document like:

```
{
   "price": {
      "usd": 3.0
   },
   "item.weight": 0.25
}
```

would essentially result in paths for `query_dbl_values` like:

```
"price.usd": 3.0,
"item.weight": 0.25
```

which is indistiguishable from alternate structure

```
{
  "price.usd": 3.0,
  "item": {
     "weight": 0.25
  }
}
```

(and other 2 variations).

But we can fix that! By escaping segments, the first case should lead to shredded paths of:

```
"price.usd": 3.0,
"item&.weight": 0.25
```

whereas second would produce:

```
"price&.usd": 3.0,
"item.weight": 0.25
```

Making this change is safe wrt other changes as well as existing shredded documents because:

1. Prior to this on-going work, neither dots nor ampersands have been valid in field names (path segments) AND
2. Filtering and sorting only need to be updated to validate path segments, but need not change them -- as long as they are correctly escaped, simple DB query/lookup will work as expected

NOTE: escaping will NOT be added to document stored as `doc_json`: it will be left as-is.

Hazel-Datastax:
**What this PR does**:
Support Escape in Shredding

Fixes #1892 

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/projection/ProjectionPath.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/projection/ProjectionPath.java
@@ -17,7 +17,10 @@ import java.util.Objects;
  * segments, and {@link #encodeSegment(String)} to get the dot-separated path string.
  */
 public final class ProjectionPath implements Comparable<ProjectionPath> {
-  private final String path;
+  /** The full encoded path string - with escape character. */
+  private final String encodedPath;
+
+  /** The segments from the encoded path - without escape character. */
   private final List<String> segments;
 
   /** projection path for document id */
@@ -32,7 +35,7 @@ public final class ProjectionPath implements Comparable<ProjectionPath> {
       ProjectionPath.from(DocumentConstants.Fields.VECTOR_EMBEDDING_TEXT_FIELD);
 
   private ProjectionPath(String originalPath, List<String> segments) {
-    this.path = originalPath;
+    this.encodedPath = originalPath;
     this.segments = segments;
   }
 
@@ -75,13 +78,13 @@ public final class ProjectionPath implements Comparable<ProjectionPath> {
   }
 
   /**
-   * Encodes a path string by escaping special characters.
+   * Encodes a path segment by escaping special characters.
    *
    * <p>The encoding process follows these steps:
    *
    * <ol>
    *   <li>
-   *       <p>Scan the path string character-by-character.
+   *       <p>Scan the path segment character-by-character.
    *   <li>
    *       <p>If the character is {@code '.'}, escape it as {@code '&.'}.
    *   <li>
@@ -92,27 +95,22 @@ public final class ProjectionPath implements Comparable<ProjectionPath> {
    *
    * <p>For example, the input {@code "item.weight"} results in: {@code "item&.weight"}.
    *
-   * @param path the raw path string to encode.
+   * @param segment the raw path segment to encode.
    * @return the encoded path string.
    */
-  public static String encodeSegment(String path) {
-    Objects.requireNonNull(path, "Path cannot be null");
+  public static String encodeSegment(String segment) {
+    Objects.requireNonNull(segment, "Segment cannot be null");
 
     StringBuilder result = new StringBuilder();
 
-    for (int i = 0; i < path.length(); i++) {
-      char ch = path.charAt(i);
+    for (int i = 0; i < segment.length(); i++) {
+      char ch = segment.charAt(i);
 
-      if (ch == '&') {
-        // Escape '&' as '&&'
-        result.append("&&");
-      } else if (ch == '.') {
-        // Escape '.' as '&.'
-        result.append("&.");
-      } else {
-        // Regular character
-        result.append(ch);
+      // Escape '&' or '.' as '&&' or '&.'
+      if (ch == '&' || ch == '.') {
+        result.append('&');
       }
+      result.append(ch);
     }
 
     return result.toString();
@@ -209,7 +207,7 @@ public final class ProjectionPath implements Comparable<ProjectionPath> {
    */
   @Override
   public String toString() {
-    return path;
+    return encodedPath;
   }
 
   /**
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/projection/ProjectionPath.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/projection/ProjectionPath.java
@@ -5,6 +5,7 @@ import io.stargate.sgv2.jsonapi.exception.ProjectionException;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
+import java.util.stream.Collectors;
 
 /**
  * An immutable representation of a projection path composed of segments.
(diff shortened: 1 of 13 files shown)


tatu-at-datastax:
Looks like there are couple of failing tests, related to changes.

src/main/java/io/stargate/sgv2/jsonapi/service/projection/ProjectionPath.java
```diff
@@ -174,7 +209,7 @@ private static List<String> decode(String path) {
    */
   @Override
   public String toString() {
-    return encodeNoEscaping();
+    return segments.stream().map(ProjectionPath::encode).collect(Collectors.joining("."));
```
tatu-at-datastax, line 212: I think it would be more efficient to do this differently: when constructing instance, keep hold of original, escaped "path" -- and on `toString()` just return that. Re-constructing String is quite a bit of overhead, and unnecessary.

tatu-at-datastax (commented):
Makes sense, but I think approach for encoding should be changed to avoid extra work.

Some tests fail due to changes, need to be fixed (which you may already be doing).

src/main/java/io/stargate/sgv2/jsonapi/service/shredding/collections/WritableShreddedDocument.java
```diff
@@ -343,12 +339,7 @@ public void shredVectorize(JsonPath path) {
      * <p>Method will add path to {@link #existKeys}.
      */
     private void addKey(JsonPath key) {
-      if (!existKeys.add(key)) {
-        if (duplicateExistKeys == null) {
-          duplicateExistKeys = new LinkedHashSet<>();
-        }
-        duplicateExistKeys.add(key);
-      }
```
Hazel-Datastax, line 351: Since conflicting is no longer a problem, I remove the core changes in PR #1848

src/main/java/io/stargate/sgv2/jsonapi/service/shredding/collections/DocumentShredder.java
```diff
@@ -478,7 +472,7 @@ private void traverseObject(ObjectNode obj, JsonPath.Builder pathBuilder) {
       Iterator<Map.Entry<String, JsonNode>> it = obj.fields();
       while (it.hasNext()) {
         Map.Entry<String, JsonNode> entry = it.next();
-        pathBuilder.property(entry.getKey());
+        pathBuilder.property(ProjectionPath.encodeSegment(entry.getKey()));
```
Hazel-Datastax, line 475: The name `ProjectionPath` sounds like it doesn't fit in the shredder process. I am thinking about changing the name. @tatu-at-datastax What do you think?

src/main/java/io/stargate/sgv2/jsonapi/service/projection/ProjectionPath.java
```diff
@@ -14,9 +14,10 @@
  * ampersands within segments.
  *
  * <p>Use {@link #from(String)} to create an instance, {@link #getSegment(int)} to access the
- * segments, and {@link #encodeNoEscaping()} to get the dot-separated path string.
+ * segments, and {@link #encodeSegment(String)} to get the dot-separated path string.
  */
 public final class ProjectionPath implements Comparable<ProjectionPath> {
+  private final String path;
```
tatu-at-datastax, line 20: Let's rename as `encodedPath` -- or add a comment mentioning it is full encoded path.

Would actually make sense to also add comment next to `segment` to mention that segment Strings are not escaped.

src/main/java/io/stargate/sgv2/jsonapi/service/projection/ProjectionPath.java
```diff
@@ -74,13 +75,47 @@ public String getSegment(int index) {
   }
 
   /**
-   * Encodes the ProjectionPath back into a string by joining segments with a dot. This encoding is
-   * used for filter and sort use cases, where escapes are not allowed.
+   * Encodes a path string by escaping special characters.
```
tatu-at-datastax, line 78: "path string" -> "path segment" (as opposed to full path)

src/main/java/io/stargate/sgv2/jsonapi/service/projection/ProjectionPath.java
```diff
@@ -74,13 +75,47 @@ public String getSegment(int index) {
...
    *
-   * @return the encoded path string (e.g., "pricing.price.usd")
+   * <p>The encoding process follows these steps:
+   *
+   * <ol>
+   *   <li>
+   *       <p>Scan the path string character-by-character.
```
tatu-at-datastax, line 84: string -> segment

src/main/java/io/stargate/sgv2/jsonapi/service/projection/ProjectionPath.java
```diff
@@ -74,13 +75,47 @@ public String getSegment(int index) {
...
+   * </ol>
+   *
+   * <p>For example, the input {@code "item.weight"} results in: {@code "item&.weight"}.
+   *
+   * @param path the raw path string to encode.
+   * @return the encoded path string.
    */
-  public String encodeNoEscaping() {
-    return String.join(".", segments);
+  public static String encodeSegment(String path) {
```
tatu-at-datastax, line 98: path -> segment

src/main/java/io/stargate/sgv2/jsonapi/service/shredding/collections/WritableShreddedDocument.java
```diff
@@ -151,10 +151,6 @@ private <T> Set<T> _nonNull(Set<T> set) {
       return (set == null) ? Collections.emptySet() : set;
     }
 
-    public Set<JsonPath> duplicateExistKeys() {
-      return duplicateExistKeys == null ? Collections.emptySet() : duplicateExistKeys;
```
tatu-at-datastax, line 155: Shoud also remove `duplicateExistKeys` itself (from line 101 or so)

src/test/java/io/stargate/sgv2/jsonapi/api/v1/InsertIntegrationTest.java
```diff
@@ -1309,9 +1309,9 @@ private void _verifyInsert(String docId, JsonNode doc) {
 
   @Nested
   @Order(4)
-  class InsertOneFails {
+  class InsertOneWithPathConflict {
```
tatu-at-datastax, line 1312: Maybe add a comment explaining why there is no real conflict for seemingly overlapping paths (or change name to "WithOverlappingPaths" to remove "Conflict" part)

tatu-at-datastax (approved):
Looks good: added some improvement suggestions; may be merged without another review.

PR #1905, opened 2025-03-04
Implement #1900: initial CreateCollection support for `$lexical`

### LINKED ISSUE #1900: Lexical config, column+index creation via "createCollection"
The first step for both BM25 feature and findAndRerank is to be able to create logical "$lexical" field for Collections. This is similar to "$vector" in that:

1. Needs to be defined as part of "createCollection" command
2. Gets passed in JSON doc to create: stored in `doc_json` but also in separately created `query_lexical_value` column with BM25 (text analyzed) SAI.
3. Information stored along with other Collection configuration (in table comments); accessible via Schema Object in code

Note that this issue is just for creating column, index, via `createCollection` and not support for inserts or find/sort (separate issues for those).

tatu-at-datastax:
**What this PR does**:

Adds support for `$lexical` when creating Collections

Fixes #1900

**Checklist**
- [x] Changes manually tested


tatu-at-datastax:
May not be 100% complete, but probably ready for initial review(s).

Hazel-Datastax:
No sure where is the error message, when I tested `{
    "createCollection": {
        "name": "test",
        "options": {
            "xx":{}
        }
    }
}`, I got an error message 
```
{
    "errors": [
        {
            "message": "The provided options are invalid: No option \"xx\" exists for `createCollection.options` (valid options: \"defaultId\", \"indexing\", \"vector\")",
            "errorCode": "INVALID_CREATE_COLLECTION_OPTIONS",
            "id": "3e77f70a-11da-4b29-8943-d825e50808db",
            "family": "REQUEST",
            "title": "The provided options are invalid",
            "scope": "SCHEMA"
        }
    ]
}
```
The valid options can add `lexical` now :)

Hazel-Datastax (approved):
LGTM

src/main/java/io/stargate/sgv2/jsonapi/StargateJsonApi.java
```diff
@@ -379,6 +379,23 @@
...
+                  name = "createCollectionLexical",
+                  summary = "`CreateCollection` command with lexical indexing ($lexical) enabled",
+                  value =
+                      """
+                                {
+                                  "createCollection": {
+                                      "name": "events",
+                                      "options": {
+                                          "lexical": {
+                                              "enabled": true,
```
vkarpov15, line 392: I'm curious, why is `enabled: true` necessary, vs assuming that presence of `lexical` means `enabled: true` vs absence of `lexical` means lexical is not enabled? Aaron mentioned this on our call yesterday but I didn't get any details as to why.

tatu-at-datastax: I think it's due to multiple factors. First: the default without "lexical" should be to enable, so to disable, one needs some value. Use of empty Object (`{ }`) seems bit confusing (difference b/w empty and missing being sort of semantically small).
Hence to disable, minimally:

```
{ "lexical": { "enabled": false  }}
```

and then it would be least confusing to always require specifying `enabled` if any value added.

Alternative would be to infer existence from combination of things.

So I think this is mostly due to defaulting to Lexical being enabled unless explicitly declined.

vkarpov15: Ok I think I understand, the issue is that it is perfectly valid to enable `lexical` with no options, but `lexical: {}` would break the pattern we have of "empty options same as omitting options". If that's the case, makes sense :+1:

tatu-at-datastax: Right, it's a tough call. I think we are breaking the pattern, due to starting to use `enabled` flag as a basic mandatory thing.

PR #1929, opened 2025-03-17
Change default C* container to HCD

tatu-at-datastax:
**What this PR does**:

Last (?) part of container setting changes to default to HCD as the default C* backend: this because it contains full BM25 functionality unlike DSE (6.9)

N/A

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -24,13 +24,13 @@
     <quarkus.container-image.tag>v${project.version}</quarkus.container-image.tag>
     <quarkus.container-image.additional-tags>v1</quarkus.container-image.additional-tags>
     <!-- When updating please change defaults in the DseTestResource class -->
-    <!-- 14-Mar-2025, tatu: default to HCD (was "dse-next", then DSE-6.9) -->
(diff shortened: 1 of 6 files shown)


pom.xml
```diff
@@ -416,6 +416,7 @@
         <stargate.int-test.cluster.name>dse-${stargate.int-test.cassandra.image-tag}-cluster</stargate.int-test.cluster.name>
         <stargate.int-test.use-coordinator>false</stargate.int-test.use-coordinator>
         <stargate.int-test.cluster.dse>true</stargate.int-test.cluster.dse>
+        <stargate.int-test.cluster.hcd>false</stargate.int-test.cluster.hcd>
```
tatu-at-datastax, line 419: To ensure only one is `true` (since defaults changed)

jeffreyscarpenter (approved):
Looks good

tatu-at-datastax:
Quick note: had to change to ONLY default IDE runs, but not Maven defaults -- this because there are "Unit" tests that run with `DseTestResource` that we do not/can not default to HCD (but DSE). But works for our needs.

PR #1931, opened 2025-03-17
Fix, insert empty list/set against table 

### LINKED ISSUE #1906: Empty lists cannot be inserted anymore on Tables (being parsed as empty maps)
Empty lists in tables, provided as `[]` in the JSON payload to insertions, cause the insertion command to fail.

## Repro

```
CREATE TABLE default_keyspace.bla (
    a text PRIMARY KEY,
    li list<int>
)
```

```
{
    "insertOne": {
        "document": {
            "a": "A",
            "li": [
                
            ]
        }
    }
}
```

Response _starting with commit `ea0ea498d08a44338bc59be79f15ad1028915e1a` on main_:

```
{
    "errors": [
        {
            "message": "Only values that are supported by the column data type can be included when inserting a document into a table.\n\nThe table default_keyspace.bla defines the columns: a(text), li(list<int>).\nThe command included the following columns that had invalid values: li: Error trying to convert to targetCQLType `List(INT, not frozen)` from value.class `java.util.HashMap`, value {}. Root cause: no codec matching value type.\n\nResend the command using only supported column values.",
            "scope": "DOCUMENT",
            "errorCode": "INVALID_COLUMN_VALUES",
            "id": "3ae1fe5f-c938-4e38-a507-ec7821360c75",
            "family": "REQUEST",
            "title": "Column values in documents are not valid"
        }
    ]
}
```

## Remarks

I can confirm that previous commits on main work correctly. It seems that the introduction of `ea0ea498d08a44338bc59be79f15ad1028915e1a - add non-string key map support for read and write path (#1876)` (Feb 19) induces empty _lists_ as being parsed as empty hash maps no matter what the column type.

Yuqi-Du:
**What this PR does**:
When inserting empty lists in tables, provided as [] in the JSON payload to insertions, cause the insertion command to fail.
This is because [] was been recognized as map tuple format incorrectly.

As a quick fix, need a validation for empty arrayNode for rowShredder. 
Since we don't have table context for rowShredder, we basically need to infer from the json payload. Probably need a bigger refactor.

Fixes #1906

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/shredding/tables/RowShredder.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/shredding/tables/RowShredder.java
@@ -145,7 +145,7 @@ public class RowShredder {
 
   /**
    * For non-string key maps, public API uses tuple format to represent map entries. This method
-   * will detect if the given array looks like tuple map entries, then then shred to object format
+   * will detect if the given array looks like tuple map entries, then shred to object format
    * JsonLiteral.
    *
    * <p>Tuple map entries are represented as an array of arrays where each inner array has two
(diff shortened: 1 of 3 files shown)


tatu-at-datastax (approved):
LGTM

tatu-at-datastax:
Looks like a minor typo / copy-paste issue wrt one IT, causing failure

vkarpov15 (commented):
LGTM

PR #1935, opened 2025-03-18
Ajm/fix 1916 find and rerank

amorton:
**What this PR does**:

Fixes #<issue number>

**Checklist**
- [ ] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/util/JsonFieldMatcher.java
```diff
@@ -0,0 +1,227 @@
...
+        }
+        unexpectedFields.put(entry.getKey(), entry.getValue());
+      } else {
+        if (clazz.isInstance(entry.getValue())) {
+          if (matched == null) {
+            // only setting the size for matching, as we assume that is the most likely.
+            matched = new HashMap<>(node.properties().size());
+          }
+          // clazz checked above
+          matched.put(entry.getKey(), unchecked(entry.getValue()));
```
Hazel-Datastax, line 120: Suggestion: Consider initializing the `matched` map upfront rather than using lazy initialization. This would simplify the code by:

1. Removing the null check before map initialization
2. Eliminating the null check in the `MatchResult` constructor 

While lazy initialization saves memory when there are no matches, the current approach adds complexity. Given that matching fields is the expected case (as your comment indicates), eager initialization might be clearer.

src/main/java/io/stargate/sgv2/jsonapi/util/JsonFieldMatcher.java
```diff
@@ -0,0 +1,227 @@
...
+    return new MatchResult<>(
+        clazz,
+        matched == null ? Map.of() : matched,
+        wrongType == null ? Map.of() : wrongType,
+        unexpectedFields == null ? Map.of() : unexpectedFields,
+        List.copyOf(unseenRequired),
+        expectedFields);
+  }
+
+  public static <T extends JsonNode> MatchResult<T> matchAndThrow(
```
Hazel-Datastax, line 142: No usages, what is this method used for?

src/main/java/io/stargate/sgv2/jsonapi/service/operation/reranking/DeferredCommandResult.java
```diff
@@ -0,0 +1,52 @@
+package io.stargate.sgv2.jsonapi.service.operation.reranking;
+
+import io.stargate.sgv2.jsonapi.api.model.command.CommandResult;
+import io.stargate.sgv2.jsonapi.service.shredding.Deferrable;
+import io.stargate.sgv2.jsonapi.service.shredding.Deferred;
+import io.stargate.sgv2.jsonapi.service.shredding.DeferredAction;
+import java.util.List;
+
+public class DeferredCommandResult implements Deferrable, Deferred {
```
Hazel-Datastax, line 9: Java doc? Not clear why this class is in the reranking (`io.stargate.sgv2.jsonapi.service.operation.reranking`) package. If it's only used for reranking, maybe add reranking in the name?

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/sort/FindAndRerankSortClauseDeserializer.java
```diff
@@ -0,0 +1,195 @@
...
+import io.stargate.sgv2.jsonapi.util.JsonFieldMatcher;
+import java.io.IOException;
+import java.util.*;
+
+/**
+ * {@link StdDeserializer} for the {@link FindAndRerankSort}.
+ *
+ * <p>Note: There is no validation, this is handled when resolving the command to an operation.
+ *
+ * <p>See {@link #deserialise(ObjectNode)} for examples of the JSON we support.
```
Hazel-Datastax, line 24: nit: deserialize

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/sort/FindAndRerankSortClauseDeserializer.java
```diff
@@ -0,0 +1,195 @@
...
+  }
+
+  /**
+   * Deserialise the sort clause from an object node.
+   *
+   * @param sort The sort clause as an object node, no check for nulls.
+   * @return The {@link FindAndRerankSort} that reflects the request without validation, e.g.
+   *     checking that a vectorize or vector sort is provided but not both.
+   */
+  private static FindAndRerankSort deserialise(JsonParser jsonParser, ObjectNode sort)
```
Hazel-Datastax, line 75: nit: deserialize

Hazel-Datastax (approved):
LTGM, my only concern is the use of v2 error in the collection now. It will be mapped to `SERVER_UNHANDLED_ERROR` and return 500.

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/FindAndRerankOperationBuilder.java
```diff
@@ -0,0 +1,347 @@
...
+
+    // TODO: Check vector if that is used
+
+    // TODO: Check vectorize if that is used
+
+    // TODO: check lexical if that is used
+
+    if (!commandContext.schemaObject().rerankingConfig().enabled()) {
+      // TODO: more info in the error
+      throw RequestException.Code.UNSUPPORTED_RERANKING_COMMAND.get();
```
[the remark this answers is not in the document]
amorton: fix it in this PR

amorton:
Merging because we need the build, will address comments and fix post merge

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/FindAndRerankOperationBuilder.java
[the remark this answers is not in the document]
amorton: Updated the ticket with reference to this.

src/main/java/io/stargate/sgv2/jsonapi/api/v1/GeneralResource.java
```diff
@@ -51,7 +52,8 @@ public GeneralResource(
       MeteredCommandProcessor meteredCommandProcessor,
       JsonProcessingMetricsReporter jsonProcessingMetricsReporter,
       CQLSessionCache cqlSessionCache,
-      EmbeddingProviderFactory embeddingProviderFactory) {
+      EmbeddingProviderFactory embeddingProviderFactory,
+      RerankingProviderFactory rerankingProviderFactory) {
```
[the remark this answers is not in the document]
amorton: > But why require not null for them?

If we allow null, then we either test it everywhere we use it or risk a NullPointer. 

More importantly, we have designed the CommandContext that s a well behaved class all objects it has should be non-null so that users of the class do not need to check and understand why they are not present. That does not mean that the object needs to "to the thing" it can be a NO_OP implementation. 

Given in this case the provider factories are application managed singletons, it's not a big deal

PR #1936, opened 2025-03-18
Reranking credential logic refactor

Yuqi-Du:
RerankingCredentials record that holds the API key for the reranking. It can be be
resolved from the request header 'reranking-api-key', if it is not present, then we will use the
cassandra token as the reranking api key. 

So, for Data API in Astra, if request does not have 'reranking-api-key' header, then it will use AstraCS token.
For on-prem user, need to specify the 'reranking-api-key' header, since the cassandra token is usually needed.

**Checklist**
- [x] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/api/request/HeaderBasedRerankingKeyResolver.java
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * We need to store both Data API token and Reranking Token in the RerankingCredentials record. E.G.
+ * For Astra self-hosted Nvidia reranking in the GPU plane, it requires the AstraCS token to access,
+ * so Data API in Astra will resolve the AstraCS token from the request header 'reranking-api-key'.
+ * If it is not present, then cassandra token will be used as the reranking api key. see {@link
+ * RequestContext}
+ *
+ * <p>For Data API in non-astra environment, since the token is also used for backend
+ * authentication, so the user needs to pass the reranking API key in the request header
```
tatu-at-datastax, line 17: nit: remove "so the"

src/main/java/io/stargate/sgv2/jsonapi/api/request/RequestContext.java
```diff
@@ -52,15 +52,25 @@ public RequestContext(
...
     this.embeddingCredentials =
         embeddingCredentialsResolver.get().resolveEmbeddingCredentials(routingContext);
-    this.rerankingCredentials =
-        rerankingCredentialsResolver.get().resolveRerankingCredentials(routingContext);
     this.tenantId = (tenantResolver.get()).resolve(routingContext, securityContext);
     this.cassandraToken = (tokenResolver.get()).resolve(routingContext, securityContext);
     httpHeaders = new HttpHeaderAccess(routingContext.request().headers());
     requestId = generateRequestId();
+
+    // rerankingCredentials will be resolved from the request header 'reranking-api-key'
```
tatu-at-datastax, line 63: This detail (name of Header) probably should remain in `HeaderBasedRerankingResolver` where it is defined?

src/main/java/io/stargate/sgv2/jsonapi/service/reranking/operation/NvidiaRerankingProvider.java
```diff
@@ -137,12 +137,14 @@ public Uni<RerankingBatchResponse> rerank(
             passages.stream().map(RerankingRequest.TextWrapper::new).toList(),
             TRUNCATE_PASSAGE);
 
-    // Note, Nvidia self-host reranker service use Astra token to authenticate the request.
-    // So we use token in rerankingCredentials, not apiKey in rerankingCredentials.
+    if (rerankingCredentials.apiKey().isEmpty()) {
+      throw ErrorCodeV1.RERANKING_PROVIDER_AUTHENTICATION_KEYS_NOT_PROVIDED.toApiException(
+          "In order to reranking, please provide the reranking API key.");
```
tatu-at-datastax, line 142: typo: "to reranking" -> "to rerank"

tatu-at-datastax (commented):
Looks ok, although injection of simple, non-configurable/non-overridable logic seems unnecessary.
Maybe instead have a static helper class to encapsulate logic, but not `@Inject` it?

tatu-at-datastax (approved):
Approved despite concerns, in case merging this functionality is blocking other work.

Yuqi-Du:
> Looks ok, although injection of simple, non-configurable/non-overridable logic seems unnecessary. Maybe instead have a static helper class to encapsulate logic, but not `@Inject` it?

Agree, just follow the pattern here. 
Let me have a static class to handle this instead of injecting it.

src/main/java/io/stargate/sgv2/jsonapi/api/request/RequestContext.java
Yuqi-Du: fixed

PR #1953, opened 2025-03-25
Fix #1904: support `$hybrid` for Lexical INSERT

### LINKED ISSUE #1904: `$hybrid` keyword on Collection inserts
(note: depends on #1901) 

To support convenient insert of both `$vectorize` and `$lexical` -- for common case of both using same text fragment -- let's add virtual field `$hybrid`, insertion of which will:

1. Work as if instead both `$vectorize` and `$lexical` were set to same String value
2. Not include itself in `doc_json`

Another way to think of it is as a macro. Its use is optional; clients may instead just pass `$vectorize` and `$lexical` separately.

tatu-at-datastax:
**What this PR does**:

Adds support for virtual `$hybrid` keyword on Insert'ing `$lexical` and `$vectorize`.

Fixes #1904

**Checklist**
- [ ] Changes manually tested


src/test/java/io/stargate/sgv2/jsonapi/api/v1/InsertInCollectionIntegrationTest.java
```diff
@@ -1005,138 +1004,6 @@ void successForPathOverlapping() {
     }
   }
 
-  @Nested
```
tatu-at-datastax, line 1008: Refactoring: moving to new test class ("InsertLexicalInCollectionIntegrationTest" below)

src/main/java/io/stargate/sgv2/jsonapi/api/v1/CollectionResource.java
```diff
@@ -220,17 +220,16 @@ public Uni<RestResponse<CommandResult>> postCommand(
...
-
                 // TODO No need for the else clause here, simplify
-                var apiFeatures =
-                    ApiFeatures.fromConfigAndRequest(
-                        apiFeatureConfig, requestContext.getHttpHeaders());
-                if ((schemaObject.type() == SchemaObject.SchemaObjectType.TABLE)
-                    && !apiFeatures.isFeatureEnabled(ApiFeature.TABLES)) {
-                  return Uni.createFrom()
-                      .failure(ErrorCodeV1.TABLE_FEATURE_NOT_ENABLED.toApiException());
+                if (schemaObject.type() == SchemaObject.SchemaObjectType.TABLE) {
```
tatu-at-datastax, line 225: Minor refactoring to only access `ApiFeatures` when we actually need them (for Table command).

src/test/java/io/stargate/sgv2/jsonapi/api/v1/AbstractCollectionIntegrationTestBase.java
```diff
@@ -34,15 +34,19 @@ protected AbstractCollectionIntegrationTestBase(String collectionNamePrefix) {
   }
 
   @BeforeAll
-  public void createSimpleCollection() {
+  public void createDefaultCollection() {
```
tatu-at-datastax, line 37: (not part of core changes)
Minor renaming to make it more clear this is the "default" collection that is normally created before IT.
May be overridden to change handling.

src/main/java/io/stargate/sgv2/jsonapi/service/processor/CommandProcessor.java
```diff
@@ -73,6 +73,10 @@ public <CommandT extends Command, SchemaT extends SchemaObject> Uni<CommandResul
                             command.commandName().getApiName(),
                             PrettyPrintable.print(commandContext.schemaObject().name())),
                     commandContext.schemaObject()));
+
+    // First bit of pre-processing: possible expansion of "$hybrid" alias
+    HybridFieldExpander.expandHybridField(commandContext, command);
+
```
Hazel-Datastax, line 79: Not sure, can we put this logic inside the `Task` or `Deferrable` structure?

src/main/java/io/stargate/sgv2/jsonapi/service/shredding/collections/DocumentShredder.java
```diff
@@ -73,12 +73,13 @@ public WritableShreddedDocument shred(JsonNode document) {
...
   public WritableShreddedDocument shred(JsonNode doc, UUID txId) {
     // TODO - why does this have "testCommand" for a command name ? Should it be a constant ?
+    // We do have "TestConstants.TEST_COMMAND_NAME"... but it's test dep cannot access here
     return shred(
         doc,
         txId,
         IndexingProjector.identityProjector(),
         "testCommand",
-        CollectionSchemaObject.MISSING,
+        CollectionSchemaObject.MISSING_WITH_LEXICAL,
```
Hazel-Datastax, line 82: I think it may not be a good idea to mix core methods with test-only methods. But at least for now, can we have a clearer Java doc here to indicate that this is only for tests and state why we need it?

src/test/java/io/stargate/sgv2/jsonapi/api/v1/VectorizeSearchIntegrationTest.java
```diff
@@ -1384,4 +1349,77 @@ public void findOneAndUpdate_sortClause() {
                   "Model random is deprecated, supported models for provider 'nvidia' are"));
     }
   }
+
+  private String getAllMetrics() {
+    return given().when().get("/metrics").then().statusCode(200).extract().asString();
+  }
```
Hazel-Datastax, line 1355: +1 for extracting the duplicated code

src/main/java/io/stargate/sgv2/jsonapi/service/processor/CommandProcessor.java
tatu-at-datastax: This must happen before `DataVectorizer` handles call for `$vectorize`, to make it transparent. So maybe if calling of `DataVectorizer` is moved to be use that structure, but not before that I think.

src/main/java/io/stargate/sgv2/jsonapi/service/shredding/collections/DocumentShredder.java
tatu-at-datastax: Right, it's not a good thing to have to rely on. Will add a Javadoc here.

src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCodeV1.java
```diff
@@ -172,6 +172,11 @@ public enum ErrorCodeV1 {
   LEXICAL_NOT_AVAILABLE_FOR_DATABASE("Lexical search is not available on this database"),
   LEXICAL_NOT_ENABLED_FOR_COLLECTION("Lexical search is not enabled for the collection"),
 
+  HYBRID_FIELD_UNSUPPORTED_VALUE_TYPE("Unsupported JSON value type for '$hybrid' field"),
+  HYBRID_FIELD_UNKNOWN_SUBFIELDS("Unrecognized sub-field(s) for '$hybrid' Object"),
+  HYBRID_FIELD_UNSUPPORTED_SUBFIELD_VALUE_TYPE(
+      "Unsupported JSON value type for '$hybrid' sub-field"),
```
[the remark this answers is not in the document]
tatu-at-datastax: I think it is a good idea to convert, but right now I am worried about mixing of error types which is why I wanted to complete this work with V1, and then switch to V2 as a follow-up.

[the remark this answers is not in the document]
amorton: not ideal but OK to get them in because this is a feature we are waiting for, and then we can remove all the V1 we added for hybrid in one step.

src/main/java/io/stargate/sgv2/jsonapi/service/processor/CommandProcessor.java
amorton: This is how I imagined the macro expansions would probably work, a pre-processing that is this essentially transparent later. 

> Not sure, can we put this logic inside the Task or Deferrable structure? 
So long as we are not making any remote calls, or doing significant work I think this is probably  OK. 

When we get to move all the dataVectorizer code out, I think we could but this macro expansion inside the commandResolver but that would only be a tidy up

src/main/java/io/stargate/sgv2/jsonapi/service/processor/HybridFieldExpander.java
```diff
@@ -0,0 +1,103 @@
...
+    }
+  }
+
+  // protected for testing purposes
+  protected static void expandHybridField(int docIndex, int docCount, JsonNode docNode) {
+    final JsonNode hybridField;
+
+    if ((docNode instanceof ObjectNode doc)
+        && (hybridField = doc.remove(DocumentConstants.Fields.HYBRID_FIELD)) != null) {
+      switch (hybridField) {
```
amorton, line 50: nit: you can do `switch (doc.remove(DocumentConstants.Fields.HYBRID_FIELD))` and have `case null`

src/main/java/io/stargate/sgv2/jsonapi/service/processor/HybridFieldExpander.java
```diff
@@ -0,0 +1,103 @@
...
+        case NullNode ignored -> addLexicalAndVectorize(doc, hybridField, hybridField);
+        case TextNode ignored -> addLexicalAndVectorize(doc, hybridField, hybridField);
+        case ObjectNode ob -> addFromObject(doc, ob, docIndex, docCount);
+        default ->
+            throw ErrorCodeV1.HYBRID_FIELD_UNSUPPORTED_VALUE_TYPE.toApiException(
```
amorton, line 55: in the FindAndRerankSortClauseDeserializer I was throwing the JsonMappingException when there was schema problem like this 

also , have a look at the JsonFieldMatcher use at the top of the class, I was trying to make this sort of deserialisation easier

src/main/java/io/stargate/sgv2/jsonapi/service/processor/HybridFieldExpander.java
```diff
@@ -0,0 +1,103 @@
...
+import io.stargate.sgv2.jsonapi.api.model.command.CommandContext;
+import io.stargate.sgv2.jsonapi.api.model.command.impl.InsertManyCommand;
+import io.stargate.sgv2.jsonapi.api.model.command.impl.InsertOneCommand;
+import io.stargate.sgv2.jsonapi.config.constants.DocumentConstants;
+import io.stargate.sgv2.jsonapi.exception.ErrorCodeV1;
+import io.stargate.sgv2.jsonapi.util.JsonUtil;
+import java.util.Iterator;
+
+/**
+ * Helper class that encapsulates the logic for expanding "$hybrid" fields in Collection Documents
```
amorton, line 17: nit: make it clear that this mutates the document in place.

src/main/java/io/stargate/sgv2/jsonapi/service/schema/collections/CollectionSchemaObject.java
```diff
@@ -48,6 +48,18 @@ public final class CollectionSchemaObject extends TableBasedSchemaObject {
           CollectionLexicalConfig.configForDisabled(),
           CollectionRerankDef.DISABLED);
 
+  // Collection Schema to use for testing: most settings missing, but $lexical
```
amorton, line 51: -1 can we move this to tests ? the MISSING one above was a bad idea that was hard to remove but I would like to.

src/main/java/io/stargate/sgv2/jsonapi/service/shredding/collections/DocumentShredder.java
```diff
@@ -62,26 +62,27 @@ public DocumentShredder(
...
+   * @param doc Document to shred.
+   * @return Shredded document
    */
-  public WritableShreddedDocument shred(JsonNode document) {
-    return shred(document, null);
-  }
-
-  public WritableShreddedDocument shred(JsonNode doc, UUID txId) {
-    // TODO - why does this have "testCommand" for a command name ? Should it be a constant ?
+  public WritableShreddedDocument testShred(JsonNode doc, UUID txId) {
```
amorton, line 73: this feels pretty janky, can we atleast get a @VisibleForTests . Can we merge and then try to refactor this out ? It also uses the  CollectionSchemaObject.MISSING_WITH_LEXICAL which i commented on, the original MISSING constant was not a great idea -> if the collection is not found then we should not have an object for it.

amorton (approved):
some things we can improve after merge, good to get this in now.

src/main/java/io/stargate/sgv2/jsonapi/service/processor/HybridFieldExpander.java
tatu-at-datastax: Ok I tried and it ALMOST works -- but not quite, because I can't get `default` case value needed for error message.

src/main/java/io/stargate/sgv2/jsonapi/service/processor/HybridFieldExpander.java
tatu-at-datastax: While possible, something would then need to map that exception to error code? Seems simpler to produce actual error code we want right where we know it.
Will have a look at `FindAndRerankSortClauseDeserializer`

src/main/java/io/stargate/sgv2/jsonapi/service/processor/HybridFieldExpander.java
tatu-at-datastax: Will add if that is missing.

tatu-at-datastax: Yes, line 21 has:

```Note, too, that input document(s) are modified in place ...
```

src/main/java/io/stargate/sgv2/jsonapi/service/schema/collections/CollectionSchemaObject.java
tatu-at-datastax: I'll first move (in this PR) it into `DocumentShredder`, if possible. And then figure out further action.

src/main/java/io/stargate/sgv2/jsonapi/service/shredding/collections/DocumentShredder.java
tatu-at-datastax: Yes, very gnarly. But I wanted to outline this instead of existing code effectively hiding it -- was/am worried regular code could have inadvertently called it.

We have about 80 tests calling the method but I'll do a follow-up issue to weed this out.

src/main/java/io/stargate/sgv2/jsonapi/service/processor/HybridFieldExpander.java
tatu-at-datastax: Since this is not a Jackson `JsonDeserializer`, it's easy to throw our own API errors, so it seems better to throw specific API error code -- otherwise we need another mapper for doing that, adding to complexity? (since we don't expose `JsonMappingException`s?

I can change this if we want to go the other way, just not sure of advantages of going through transformation route.

src/main/java/io/stargate/sgv2/jsonapi/service/shredding/collections/DocumentShredder.java
tatu-at-datastax: On `CollectionSchemaObject`: that is used to check collection settings like indexing config, vector/lexical enabling. So tests unfortunately do need to pass it (or bigger refactoring needed to change how collection config is passed). But I created #1971 for clean up.

PR #1960, opened 2025-03-27
Error out when lexical or rerank is disabled but configuration is provided

### LINKED ISSUE #1958: Error out when lexical or rerank is disabled but configuration is provided
When a user disables lexical search or rerank features but still provides configuration values for `lexical.analyzer` or `rerank.service`, we currently ignore these configurations instead of reporting an error.

Based on feedback from [PR #1933](https://github.com/stargate/data-api/pull/1933#discussion_r2015057563), we should validate these configurations and error out in these inconsistent scenarios.

Hazel-Datastax:
**What this PR does**:
Error out when lexical or rerank is disabled but configuration is provided

Fixes #1958 

**Checklist**
- [x] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/service/schema/collections/CollectionLexicalConfig.java
```diff
@@ -19,12 +19,38 @@ public record CollectionLexicalConfig(
...
+      if (analyzerDefinition != null) {
+        if (analyzerDefinition.isTextual()) {
+          throw new IllegalArgumentException(
+              "Analyzer definition should not have string if lexical is disabled");
+        }
+        if (analyzerDefinition.isObject() && !analyzerDefinition.isEmpty()) {
+          throw new IllegalArgumentException(
+              "Analyzer definition should be null or empty JSON if lexical is disabled");
+        }
+      }
```
Hazel-Datastax, line 53: All these things should be checked before calling the constructor; this is just a sanity check

src/main/java/io/stargate/sgv2/jsonapi/service/schema/collections/CollectionLexicalConfig.java
```diff
@@ -39,21 +65,40 @@ public static CollectionLexicalConfig validateAndConstruct(
...
+              "'lexical' is disabled, but 'lexical.analyzer' property is provided with value: '%s'",
+              lexicalConfig.analyzerDef().asText());
+        }
+        if (lexicalConfig.analyzerDef().isObject() && !lexicalConfig.analyzerDef().isEmpty()) {
+          throw ErrorCodeV1.INVALID_CREATE_COLLECTION_OPTIONS.toApiException(
+              "'lexical' is disabled, but 'lexical.analyzer' property is provided with non-empty JSON Object");
+        }
+      }
+      return configForDisabled();
+    }
```
Hazel-Datastax, line 94: I know it looks complex. This block ensures when `enabled: false`:
✅ `{"enabled": false, "analyzer": {}}` (empty JSON object) is valid
❌ `{"enabled": false, "analyzer": ""}` (empty string) is invalid
First check `isTextual()` because `isEmpty()` would treat any TextNode as empty, so cannot just use 
```
if (lexicalConfig.analyzerDef() != null && !lexicalConfig.analyzerDef().isEmpty()) {
// Throw error
}
```

src/main/java/io/stargate/sgv2/jsonapi/service/schema/collections/CollectionRerankDef.java
```diff
@@ -52,32 +52,39 @@ public class CollectionRerankDef {
...
+          Objects.requireNonNull(
+              rerankServiceDef,
+              "Rerank service configuration must not be null when reranking is enabled");
+    } else {
+      if (rerankServiceDef != null) {
+        throw new IllegalArgumentException(
+            "Rerank service configuration must be null when reranking is disabled");
+      }
+      this.rerankServiceDef = null;
+    }
```
Hazel-Datastax, line 87: Same as the lexical, all these things should be checked before calling the constructor; this is just a sanity check

tatu-at-datastax (changes requested):
Ok, I think definition of valid is not quite right wrt lexical config: I think that when "disabled", only valid values are:

* explicit `null` (node.isNull())
* missing (no "analyzer" property)

but we should not accept empty JSON Object. We can discuss this, since I think Aaron may have requested empty case?

src/main/java/io/stargate/sgv2/jsonapi/service/schema/collections/CollectionLexicalConfig.java
```diff
@@ -19,12 +19,38 @@ public record CollectionLexicalConfig(
...
 
+  /**
+   * Constructs a lexical configuration with the specified enabled state and analyzer definition.
+   *
+   * <p>Validation behavior:
+   *
+   * <ul>
+   *   <li>If lexical search is enabled (enabled = true), the analyzer definition must not be null
+   *   <li>If lexical search is disabled (enabled = false), the analyzer definition must be null or
+   *       empty
```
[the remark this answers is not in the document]
tatu-at-datastax: Clarified with discussion: "empty" also accepted, it's all good.

src/main/java/io/stargate/sgv2/jsonapi/service/schema/collections/CollectionLexicalConfig.java
```diff
@@ -39,21 +65,40 @@ public static CollectionLexicalConfig validateAndConstruct(
...
+        if (lexicalConfig.analyzerDef().isTextual()) {
+          throw ErrorCodeV1.INVALID_CREATE_COLLECTION_OPTIONS.toApiException(
+              "'lexical' is disabled, but 'lexical.analyzer' property is provided with value: '%s'",
+              lexicalConfig.analyzerDef().asText());
+        }
+        if (lexicalConfig.analyzerDef().isObject() && !lexicalConfig.analyzerDef().isEmpty()) {
+          throw ErrorCodeV1.INVALID_CREATE_COLLECTION_OPTIONS.toApiException(
+              "'lexical' is disabled, but 'lexical.analyzer' property is provided with non-empty JSON Object");
+        }
+      }
```
tatu-at-datastax, line 92: This is missing handling for other JSON types like Arrays or numbers. I think checks should be changed to test valid cases:

1. Missing (`null`)
2. JSON null (`def.isNull()`)
3. empty JSON Object

and unless it's one of these, fail indicating node type (`JsonUtil.nodeTypeAsString(node.getNodeType())`)

tatu-at-datastax (commented):
Looks quite good but I think logic for legal analyzer config (none/null/empty-ob for disabled) should be simplified and made to cover all cases: not just String and Object.

src/main/java/io/stargate/sgv2/jsonapi/service/schema/collections/CollectionLexicalConfig.java
Hazel-Datastax: Update the verification and yes, if it is one of types, I will call `configForDisabled()` which sets the definition to null.

src/main/java/io/stargate/sgv2/jsonapi/service/schema/collections/CollectionLexicalConfig.java
```diff
@@ -19,12 +20,38 @@ public record CollectionLexicalConfig(
...
-    // Clear out any analyzer settings if not enabled (but don't fail)
     if (enabled) {
       this.analyzerDefinition = Objects.requireNonNull(analyzerDefinition);
     } else {
+      if (analyzerDefinition != null) {
+        if (analyzerDefinition.isTextual()) {
+          throw new IllegalArgumentException(
+              "Analyzer definition should not have string if lexical is disabled");
+        }
+        if (analyzerDefinition.isObject() && !analyzerDefinition.isEmpty()) {
```
tatu-at-datastax, line 50: Not sure if any validation is needed here (since that is already done by caller), but if it is, should use similar logic -- if not enabled, must be `null` or empty Object.

src/main/java/io/stargate/sgv2/jsonapi/service/schema/collections/CollectionRerankDef.java
```diff
@@ -54,32 +54,39 @@ public class CollectionRerankDef {
    *
    * <p>This constructor is annotated with {@link JsonCreator} to enable Jackson deserialization.
    * {@link JsonProperty} annotations on parameters allow Jackson to map JSON properties to
-   * constructor parameters during deserialization.
+   * constructor parameters during deserialization. This constructor is public primarily to support
```
tatu-at-datastax, line 57: Note: constructors do NOT have to be `public` for Jackson -- any visibility is fine, even `private`, as long as `@JsonCreator` annotation is used to mark it for use.

tatu-at-datastax (approved):
Looks good -- I think there's one place to change wrt validation,  but otherwise good.
Approving so that with last changes can be merged without re-review as necessary.

PR #1963, opened 2025-03-28
Data API model deprecation ability for embedding and rerank models

Yuqi-Du:
By default, all configured models are at SUPPORTING status.
Need to explicitly mark model as DEPRECATED or END_OF_LIFE if needed.
E.G.
```
  models:
    - name: NV-Embed-QA
      model-support:
        status: DEPRECATED
        message: This model has been deprecated, it will be removed in a future release. It is not supported for new collection or table.
      vector-dimension: 1024
      properties:
        max-tokens: 512
    - name: nvidia/llama-3.2-nv-embedqa-1b-v2
      vector-dimension: 2048
      service-url-override: http://llama-3-2-nv-embedqa-1b-v2.nemo:8000/v1/embeddings
      properties:
        max-tokens: 8192
```

-       For DEPRECATED model, it can not be used for collection or table creation, but can be used for existing collection or table until the model is END_OF_LIFE.
-       For END_OF_LIFE model, it can not be used for collection or table creation, and can not be used for existing collection or table.

**Checklist**
- [x] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/service/resolver/VectorizeConfigValidator.java
```diff
@@ -344,6 +356,25 @@ private Integer validateModelAndDimension(
...
 
+    // validate model support
+    if (model.modelSupport().status()
+        != EmbeddingProvidersConfig.EmbeddingProviderConfig.ModelSupport.SupportStatus.SUPPORTING) {
+      //      throw ErrorCodeV1.INVALID_CREATE_COLLECTION_OPTIONS.toApiException(
+      //          "V2 TODO model is deprecated ",
+      //          userConfig.modelName(),
+      //          userConfig.provider(),
+      //          model.modelSupport().status(),
+      //          model.modelSupport().message().orElse(""));
```
Hazel-Datastax, line 367: remove?

src/test/java/io/stargate/sgv2/jsonapi/api/v1/FindRerankingProvidersIntegrationTest.java
```diff
@@ -0,0 +1,65 @@
...
+
+      given()
+          .port(getTestPort())
+          .headers(getHeaders())
+          .contentType(ContentType.JSON)
+          .body(json)
+          .when()
+          .post(GeneralResource.BASE_PATH)
+          .then()
+          .statusCode(200)
```
Hazel-Datastax, line 41: Use `givenHeadersPostJsonThenOk` to remove duplicated code

src/test/java/io/stargate/sgv2/jsonapi/api/v1/FindRerankingProvidersIntegrationTest.java
```diff
@@ -0,0 +1,65 @@
...
+          .body("$", responseIsStatusOnly())
+          .body("status.embeddingProviders", notNullValue())
+          .body(
+              "status.embeddingProviders.nvidia.models[0].name",
+              equalTo("nvidia/llama-3.2-nv-rerankqa-1b-v2"))
+          .body(
+              "status.embeddingProviders.nvidia.models[0].modelSupport.status",
```
Hazel-Datastax, line 48: Will it return the `DEPRECATED` and `END_OF_LIFE` models? From your changes in FindEmbeddingProvidersOperation, I guess the answer is yes?
If the model is `END_OF_LIFE`, I think we don't need to return it?

src/test/java/io/stargate/sgv2/jsonapi/api/v1/FindRerankingProvidersIntegrationTest.java
Yuqi-Du: fixed, thanks

src/test/java/io/stargate/sgv2/jsonapi/api/v1/FindRerankingProvidersIntegrationTest.java
Yuqi-Du: I will discuss with team. 
For now, we don't have EOL model, but I think for clarity, it may be good to show user that certain model is EOL, in that sense, they will know what is going on on the server side.

tatu-at-datastax: If not too complicated, option to include `END_OF_LIFE` (not set by default) would make sense to me -- so they are only returned if specifically requested.

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/configuration/EmbeddingConfigSourceProvider.java
```diff
@@ -14,6 +14,16 @@
 /**
  * Loading the YAML configuration file from the resource folder or env variable and making the
  * config available to the application.
+ *
+ * <ul>
+ *   <li>With env variables set, Data API loads provider config from specified resource location.
```
tatu-at-datastax, line 19: nit: "variables" -> "variable" (I think there is just one)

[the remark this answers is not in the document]
Yuqi-Du: thanks, added

Yuqi-Du: thanks, added

src/test/java/io/stargate/sgv2/jsonapi/api/v1/FindRerankingProvidersIntegrationTest.java
Yuqi-Du: added.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/FindEmbeddingProvidersCommand.java
```diff
@@ -3,12 +3,30 @@
...
+        Options options)
+    implements GeneralCommand {
+
+  /**
+   * By default, if includeModelStatus is not provided, only model in supported status will be
+   * returned.
+   */
+  public record Options(
+      @Schema(
+              description = "Use the option to include models as in target support status.",
```
tatu-at-datastax, line 26: Should probably mention what happens if not included ("only model(s) in supported status will be included")

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/configuration/EmbeddingProviderConfigStore.java
```diff
@@ -37,9 +37,13 @@ public static ServiceConfig custom(Optional<Class<?>> implementationClass) {
 
     public String getBaseUrl(String modelName) {
       if (modelUrlOverrides != null && modelUrlOverrides.get(modelName) == null) {
-        throw ErrorCodeV1.VECTORIZE_MODEL_DEPRECATED.toApiException(
-            "Model %s is deprecated, supported models for provider '%s' are %s",
-            modelName, serviceName, modelUrlOverrides.keySet());
+        // modelUrlOverride is a walk-around for self-hosted nvidia models with different url.
```
tatu-at-datastax, line 40: nit: walk-around -> work-around

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/configuration/EmbeddingProviderConfigStore.java
```diff
@@ -37,9 +37,13 @@ public static ServiceConfig custom(Optional<Class<?>> implementationClass) {
...
+        // This is bad, initial design should have url in model level instead of provider level.
+        // As best practice, when deprecate or EOL a model:
+        // we should mark the status in the configuration,
```
tatu-at-datastax, line 43: nit: should -> MUST

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/MeteredEmbeddingProvider.java
```diff
@@ -54,6 +57,21 @@ public Uni<Response> vectorize(
...
+      throw SchemaException.Code.UNSUPPORTED_PROVIDER_MODEL.get(
+          Map.of(
+              "model", embeddingProvider.model.name(),
+              "modelStatus", embeddingProvider.model.modelSupport().status().name(),
+              "message",
+                  embeddingProvider
+                      .model
+                      .modelSupport()
+                      .message()
+                      .orElse("The model is not supported.")));
```
tatu-at-datastax, line 72: Maybe change default slightly to something like:

"The model is no longer supported (status: END_OF_LIFE)"

src/main/java/io/stargate/sgv2/jsonapi/service/operation/embeddings/FindEmbeddingProvidersOperation.java
```diff
@@ -30,11 +32,23 @@ public Uni<Supplier<CommandResult>> execute(
...
+                                  EmbeddingProviderResponse.provider(
+                                      entry.getValue(), getSupportStatuses())));
               return new Result(embeddingProviders);
             });
   }
 
+  // By default, if includeModelStatus is not provided in command option, only model in supported
+  // status will be listed.
+  private Set<ModelSupport.SupportStatus> getSupportStatuses() {
+    var includeModelStatus = EnumSet.of(ModelSupport.SupportStatus.SUPPORTED);
```
tatu-at-datastax, line 45: nit: can move default value setting into `else` branch instead of here.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/embeddings/FindEmbeddingProvidersOperation.java
```diff
@@ -96,9 +116,13 @@ private static EmbeddingProviderResponse provider(
    * @param parameters Parameters for customizing the model.
    */
   private record ModelConfigResponse(
-      String name, Optional<Integer> vectorDimension, List<ParameterConfigResponse> parameters) {
+      String name,
+      ModelSupport modelSupport,
+      Optional<Integer> vectorDimension,
+      List<ParameterConfigResponse> parameters) {
```
tatu-at-datastax, line 122: and here they are called... "parameters"? Probably not a good name either.

src/main/java/io/stargate/sgv2/jsonapi/service/provider/EmbeddingAndRerankingConfigSourceProvider.java
```diff
@@ -44,8 +46,13 @@ public class EmbeddingAndRerankingConfigSourceProvider implements ConfigSourcePr
   private static final String DEFAULT_EMBEDDING_CONFIG_RESOURCE = "embedding-providers-config.yaml";
   // Default reranking config resource.
   private static final String DEFAULT_RERANKING_CONFIG_RESOURCE = "reranking-providers-config.yaml";
-  // System property name to override reranking config resource. Could be set by integration test
-  // resource.
+
+  // System property name to override embedding config resource.
+  // Could be set by integration test resource.
```
tatu-at-datastax, line 51: "Could be" -> "Is usually"?

src/main/java/io/stargate/sgv2/jsonapi/service/provider/EmbeddingAndRerankingConfigSourceProvider.java
```diff
@@ -74,9 +81,14 @@ public Iterable<ConfigSource> getConfigSources(ClassLoader forClassLoader) {
    */
   private ConfigSource getEmbeddingConfigSources(ClassLoader forClassLoader) throws IOException {
     String filePathFromEnv = System.getenv(EMBEDDING_CONFIG_FILE_PATH);
+    String resourceOverride = System.getProperty(DEFAULT_EMBEDDING_CONFIG_RESOURCE_OVERRIDE);
+
     if (filePathFromEnv != null) {
       LOGGER.info("Loading embedding config from file path: {}", filePathFromEnv);
       return loadConfigSourceFromFile(filePathFromEnv);
+    } else if (resourceOverride != null) {
```
tatu-at-datastax, line 89: maybe also ` && !resourceOverride.isBlank()` ?

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/VectorizeConfigValidator.java
```diff
@@ -344,6 +356,18 @@ private Integer validateModelAndDimension(
...
+    // validate model support
+    if (model.modelSupport().status() != ModelSupport.SupportStatus.SUPPORTED) {
+      throw SchemaException.Code.UNSUPPORTED_PROVIDER_MODEL.get(
+          Map.of(
+              "model",
+              userConfig.modelName(),
+              "modelStatus",
+              model.modelSupport().status().name(),
+              "message",
+              model.modelSupport().message().orElse("The model is not supported.")));
```
tatu-at-datastax, line 368: Same as earlier:  maybe

"The model is no longer supported (status: END_OF_LIFE)"

src/main/proto/embedding_gateway.proto
```diff
@@ -159,13 +159,19 @@ message GetSupportedProvidersResponse {
...
 
-    // ModelConfig message represents configuration for a specific model.
+    // ModelConfig message represents configuration for a specific embedding model.
     message ModelConfig {
       string name = 1;
-      optional int32 vector_dimension = 2;
-      repeated ParameterConfig parameters = 3;
-      map<string, string> properties = 4;
-      optional string service_url_override = 5;
+      ModelSupport modelSupport = 2;
```
tatu-at-datastax, line 165: For backwards-compatibility we should never change index used and always:

1. Deprecating existing properties (never removing), leaving index as-is
2. Only appending new properties with next available index.

Otherwise "protobuf" decoder will get corrupt data, as values and names must match.

So: add new `modelSupport` -- or maybe name as `model`? "Support" is part of type but not really relevant for usage -- with index `6` and leave old properties as-is.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/CreateCollectionIntegrationTest.java
```diff
@@ -873,6 +873,40 @@ public void failUnsupportedModel() {
     }
   }
 
+  @Test
+  public void failDeprecatedEmbedModel() {
```
tatu-at-datastax, line 877: Should we also test END_OF_LIFE one?

src/test/java/io/stargate/sgv2/jsonapi/api/v1/VectorizeSearchIntegrationTest.java
```diff
@@ -1301,7 +1301,13 @@ public void createDropDifferentVectorDimension() {
   @Nested
   @Order(8)
   @TestMethodOrder(MethodOrderer.OrderAnnotation.class)
-  class DeprecatedModel {
+  class UnknownExistingModel {
+
+    // As best practice, when deprecate or EOL a model, we should mark them in the configuration,
```
tatu-at-datastax, line 1306: nit: "deprecate" -> "we deprecate"

tatu-at-datastax (commented):
Looks pretty good: the only required change is protobuf definition: should not change indexes for properties but add new as the last one, with new (last + 1) index.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/embeddings/FindEmbeddingProvidersOperation.java
```diff
@@ -16,8 +18,8 @@
  * Operation that list all available and enabled vector providers into the {@link
  * CommandStatus#EXISTING_EMBEDDING_PROVIDERS} command status.
  */
-public record FindEmbeddingProvidersOperation(EmbeddingProvidersConfig config)
-    implements Operation {
+public record FindEmbeddingProvidersOperation(
```
amorton, line 21: from discussions .... it is reasonable to refactor and re-format this whole file

src/main/java/io/stargate/sgv2/jsonapi/service/provider/EmbeddingAndRerankingConfigSourceProvider.java
```diff
@@ -44,8 +46,13 @@ public class EmbeddingAndRerankingConfigSourceProvider implements ConfigSourcePr
   private static final String DEFAULT_EMBEDDING_CONFIG_RESOURCE = "embedding-providers-config.yaml";
   // Default reranking config resource.
   private static final String DEFAULT_RERANKING_CONFIG_RESOURCE = "reranking-providers-config.yaml";
-  // System property name to override reranking config resource. Could be set by integration test
-  // resource.
+
+  // System property name to override embedding config resource.
+  // Could be set by integration test resource.
+  private static final String DEFAULT_EMBEDDING_CONFIG_RESOURCE_OVERRIDE =
```
amorton, line 52: nit: we can have a better variable name like SYS_PROPERTY_EMBEDDING_CONFIG_RESOURCE

and the string can be "EMBEDDING_CONFIG_RESOURCE"

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/VectorizeConfigValidator.java
```diff
@@ -344,6 +356,18 @@ private Integer validateModelAndDimension(
                         "Model name '%s' for provider '%s' is not supported",
                         userConfig.modelName(), userConfig.provider()));
 
+    // validate model support
```
amorton, line 359: -1 this should not be hidden in this class. I think this should be in the commandResolver for createCollection where we clearly know the context is creating a new collection.

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/VectorizeConfigValidator.java
```diff
@@ -344,6 +356,18 @@ private Integer validateModelAndDimension(
...
+    if (model.modelSupport().status() != ModelSupport.SupportStatus.SUPPORTED) {
+      throw SchemaException.Code.UNSUPPORTED_PROVIDER_MODEL.get(
```
amorton, line 361: can we have a DEPRECATED_PROVIDER_MODEL specific for this situation, then we can say that deprecated models can still be used for existing collections

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/EmbeddingProvider.java
```diff
@@ -71,6 +74,30 @@ protected <T> Uni<T> applyRetry(Uni<T> uni) {
...
+   * END_OF_LIFE model will throw an exception.
+   *
+   * <p>Note, SUPPORTED and DEPRECATED models are still allowed to be used in read and write.
+   *
+   * <p>This method should be called before any vectorization operation.
+   */
+  public void checkEOLModelUsage() {
+    // Validate if the model is END_OF_LIFE
+    if (model.modelSupport().status() == ModelSupport.SupportStatus.END_OF_LIFE) {
+      throw SchemaException.Code.UNSUPPORTED_PROVIDER_MODEL.get(
```
amorton, line 90: can we have a END_OF_LIFE_PROVIDER_MODEL specific for this situation

amorton (changes requested):
some changes as discussed

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/configuration/EmbeddingProviderConfigStore.java
Yuqi-Du: fixed

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/configuration/EmbeddingProviderConfigStore.java
Yuqi-Du: fixed

src/test/java/io/stargate/sgv2/jsonapi/api/v1/VectorizeSearchIntegrationTest.java
```diff
@@ -1301,7 +1301,13 @@ public void createDropDifferentVectorDimension() {
...
+    // instead
+    // of removing the whole configuration entry as bad practice!
+    // The bad practice should only happen in dev before, add this validation to capture, and
+    // confirm it is as least not return 500,
```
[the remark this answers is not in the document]
Yuqi-Du: fixed

src/test/java/io/stargate/sgv2/jsonapi/api/v1/VectorizeSearchIntegrationTest.java
Yuqi-Du: fixed

src/main/java/io/stargate/sgv2/jsonapi/service/provider/EmbeddingAndRerankingConfigSourceProvider.java
Yuqi-Du: fixed

src/test/java/io/stargate/sgv2/jsonapi/api/v1/CreateCollectionIntegrationTest.java
Yuqi-Du: added, thanks

src/test/java/io/stargate/sgv2/jsonapi/api/v1/VectorizeSearchIntegrationTest.java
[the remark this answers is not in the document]
Yuqi-Du: fixed

src/test/java/io/stargate/sgv2/jsonapi/api/v1/VectorizeSearchIntegrationTest.java
Yuqi-Du: fixed

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/configuration/EmbeddingProvidersConfig.java
```diff
@@ -105,11 +104,16 @@ interface ModelConfig {
       @JsonProperty
       String name();
 
+      /**
+       * modelSupport marks the support status of the model and optional message for the
+       * deprecation, EOL etc.
+       */
+      @JsonProperty
+      ModelSupport modelSupport();
```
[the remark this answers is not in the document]
Yuqi-Du: added

src/main/java/io/stargate/sgv2/jsonapi/service/provider/EmbeddingAndRerankingConfigSourceProvider.java
Yuqi-Du: changed

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/configuration/EmbeddingProvidersConfig.java
```diff
@@ -105,11 +104,16 @@ interface ModelConfig {
       @JsonProperty
       String name();
 
+      /**
+       * modelSupport marks the support status of the model and optional message for the
+       * deprecation, EOL etc.
+       */
+      @JsonProperty
```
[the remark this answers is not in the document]
Yuqi-Du: changed

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/FindEmbeddingProvidersCommand.java
```diff
@@ -3,12 +3,30 @@
...
+              type = SchemaType.OBJECT,
+              implementation = ModelSupport.SupportStatus.class)
+          EnumSet<ModelSupport.SupportStatus> includeModelStatus) {}
```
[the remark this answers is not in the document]
Yuqi-Du: refactored

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/configuration/EmbeddingProvidersConfig.java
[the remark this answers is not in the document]
Yuqi-Du: added

src/main/java/io/stargate/sgv2/jsonapi/service/operation/embeddings/FindEmbeddingProvidersOperation.java
Yuqi-Du: refactored

src/main/java/io/stargate/sgv2/jsonapi/service/provider/EmbeddingAndRerankingConfigSourceProvider.java
Yuqi-Du: fixed

src/main/java/io/stargate/sgv2/jsonapi/service/operation/embeddings/FindEmbeddingProvidersOperation.java
Yuqi-Du: refactored as the rules are changed per Aaron's ask

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/VectorizeConfigValidator.java
Yuqi-Du: fixed

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/EmbeddingProvider.java
Yuqi-Du: fixed

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/VectorizeConfigValidator.java
Yuqi-Du: Per your ask of move validation into resolver, I could not find a way to do that,
the VectorizeConfigValidator is always used for create/alter schema, E.G. CreateTable/AlterTable you pass down the validator when creating the ApiTableDef, which is finally used with ColumnDef.
All the validation did not happen in resolver before, in other words, can not simply move, unless other folks know there is a way to do that.

src/main/proto/embedding_gateway.proto
Yuqi-Du: changed.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/FindEmbeddingProvidersCommand.java
Yuqi-Du: the options has changed according to Aaron's ask

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/EmbeddingProvider.java
```diff
@@ -71,6 +74,30 @@ protected <T> Uni<T> applyRetry(Uni<T> uni) {
...
+   * Checks if the vectorization will use an END_OF_LIFE model and throws an exception if it is.
+   *
+   * <p>As part of embedding model deprecation ability, any read and write with vectorization in an
+   * END_OF_LIFE model will throw an exception.
+   *
+   * <p>Note, SUPPORTED and DEPRECATED models are still allowed to be used in read and write.
+   *
+   * <p>This method should be called before any vectorization operation.
+   */
+  public void checkEOLModelUsage() {
```
amorton, line 87: does this need to be public  ?

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/AwsBedrockEmbeddingProvider.java
```diff
@@ -53,6 +54,8 @@ public Uni<Response> vectorize(
       List<String> texts,
       EmbeddingCredentials embeddingCredentials,
       EmbeddingRequestType embeddingRequestType) {
+    // Check if using an EOF model
```
amorton, line 57: The concern here is every embedding provider must check this, but they all need to remember to do it. Other than the code duplication, it is risky that we forget it sometime. 

Thinking about this as an Object Oriented programming problem, vectorize() is a behavour of the super class, that has some common behaviour (this an any other checks or metrics or metering or logging) and there is a "doVectorize" (but a better name) to actually do the non common behaviour. 

If we wanted a more robust solution we could: 

* make vectorize() non abstract on the base, make it do the check, and any other logging, metering etc that is common to all. 
* create a new abstract vectorizeInternal() or similar that the sub classes actually implement

We could either do it here, or in the next work for the metering.

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/EmbeddingProviderFactory.java
```diff
@@ -117,10 +118,26 @@ private synchronized EmbeddingProvider addService(
...
+      throw ErrorCodeV1.VECTORIZE_SERVICE_TYPE_UNAVAILABLE.toApiException(
+          "unknown service provider '%s'", configuration.serviceProvider());
+    }
+    EmbeddingProvidersConfig.EmbeddingProviderConfig.ModelConfig model =
+        embeddingProvidersConfig.providers().get(configuration.serviceProvider()).models().stream()
+            .filter(m -> m.name().equals(modelName))
+            .findFirst()
+            .orElseThrow(
+                () ->
+                    ErrorCodeV1.VECTORIZE_SERVICE_TYPE_UNAVAILABLE.toApiException(
```
amorton, line 134: checking this is an older error code we are reusing because it was used before (it's not a good match for what is happening)

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/EmbeddingProviderFactory.java
```diff
@@ -117,10 +118,26 @@ private synchronized EmbeddingProvider addService(
...
     }
+
+    // Get the provider, then get the model.
+    var providerConfig = embeddingProvidersConfig.providers().get(configuration.serviceProvider());
+    if (providerConfig == null) {
+      throw ErrorCodeV1.VECTORIZE_SERVICE_TYPE_UNAVAILABLE.toApiException(
+          "unknown service provider '%s'", configuration.serviceProvider());
+    }
+    EmbeddingProvidersConfig.EmbeddingProviderConfig.ModelConfig model =
+        embeddingProvidersConfig.providers().get(configuration.serviceProvider()).models().stream()
```
amorton, line 129: ah, this is ugly. The ProviderConfig does not have the name on it, I'll fix this later.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/embeddings/FindEmbeddingProvidersOperation.java
```diff
@@ -67,56 +71,91 @@ private record EmbeddingProviderResponse(
...
+     * @param sourceEmbeddingProviderConfig, the original provider config with all properties.
+     * @param includeModelStatus the modelSupport status that user asked for.
+     */
+    private static EmbeddingProviderResponse toResponse(
+        EmbeddingProvidersConfig.EmbeddingProviderConfig sourceEmbeddingProviderConfig,
+        Set<ModelSupport.SupportStatus> includeModelStatus) {
+
+      // if the providerConfig.models is null or empty, return an EmbeddingProviderResponse with
+      // empty models.
+      if (sourceEmbeddingProviderConfig.models() == null
```
amorton, line 88: the Objects.requireNonNull() is not needed you are checking null 

this would be clearer if you switch on the sourceEmbeddingProviderConfig.models() to build the list of models, and then have a single return

src/main/java/io/stargate/sgv2/jsonapi/service/operation/embeddings/FindEmbeddingProvidersOperation.java
```diff
@@ -67,56 +71,91 @@ private record EmbeddingProviderResponse(
...
+      ModelSupport modelSupport,
+      Optional<Integer> vectorDimension,
+      List<ParameterConfigResponse> parameters) {
+
+    private static ModelConfigResponse toResponse(
+        EmbeddingProvidersConfig.EmbeddingProviderConfig.ModelConfig sourceModelConfig) {
+
+      // if the sourceModelConfig.parameters is null or empty, return a ModelConfigResponse with
+      // empty parameters.
+      if (sourceModelConfig.parameters() == null
```
amorton, line 139: why the Objects.requireNonNull() after checking null ?

src/main/java/io/stargate/sgv2/jsonapi/service/operation/embeddings/FindEmbeddingProvidersOperation.java
```diff
@@ -67,56 +71,91 @@ private record EmbeddingProviderResponse(
...
+            sourceModelConfig.name(),
+            sourceModelConfig.modelSupport(),
+            sourceModelConfig.vectorDimension(),
+            Collections.emptyList());
       }
 
-      return new ModelConfigResponse(name, vectorDimension, parametersResponse);
+      // reconstruct each parameter for lowercase parameter type
+      List<ParameterConfigResponse> parametersResponse =
+          Objects.requireNonNull(sourceModelConfig.parameters()).stream()
```
amorton, line 150: why checking sourceModelConfig.parameters() for null again ?

src/main/java/io/stargate/sgv2/jsonapi/service/operation/embeddings/FindEmbeddingProvidersOperation.java
```diff
@@ -140,20 +179,61 @@ private record ParameterConfigResponse(
...
+   * <ul>
+   *   <li>If not provided, only SUPPORTED models will be returned.
+   *   <li>If provided with "" empty string, all SUPPORTED, DEPRECATED, END_OF_LIFE model will be
+   *       returned.
+   *   <li>If provided with specified SUPPORTED or DEPRECATED or END_OF_LIFE, only models matched
+   *       the status will be returned.
+   * </ul>
+   */
+  private Set<ModelSupport.SupportStatus> getSupportStatuses() {
+    if (command.options() == null || command.options().includeModelStatus() == null) {
```
amorton, line 212: is this a bug ? We said if null is passed it will return all statuses ?

src/main/java/io/stargate/sgv2/jsonapi/service/operation/embeddings/FindEmbeddingProvidersOperation.java
```diff
@@ -140,20 +179,61 @@ private record ParameterConfigResponse(
...
+      return EnumSet.of(ModelSupport.SupportStatus.SUPPORTED);
+    }
+
+    if (command.options().includeModelStatus().isEmpty()) {
```
amorton, line 216: use isBlank ?

src/main/java/io/stargate/sgv2/jsonapi/service/operation/embeddings/FindEmbeddingProvidersOperation.java
```diff
@@ -140,20 +179,61 @@ private record ParameterConfigResponse(
...
+      return EnumSet.of(
```
amorton, line 217: can this be simplified to a predicate that checks ? there is a code update problem here, if we add to the ENUM we need to know that this line needs to be updated. It would be safer to have no check 

This would be simplier if it returned a Predicate that matched SupportStatus, it can return true if the filter was not set, else match the single value

src/main/java/io/stargate/sgv2/jsonapi/service/operation/embeddings/FindEmbeddingProvidersOperation.java
```diff
@@ -140,20 +179,61 @@ private record ParameterConfigResponse(
...
+
+    var specifiedModelStatus = command.options().includeModelStatus().toUpperCase(Locale.ROOT);
+    return EnumSet.of(ModelSupport.SupportStatus.valueOf(specifiedModelStatus));
+  }
+
+  /**
+   * Filter models based on the provided support statuses. Only handle models wit modelStatuses that
+   * user asked for.
+   */
+  private static List<EmbeddingProvidersConfig.EmbeddingProviderConfig.ModelConfig>
```
amorton, line 231: if above we make a predicate , this is not needed

src/main/java/io/stargate/sgv2/jsonapi/service/provider/ModelSupport.java
```diff
@@ -22,6 +22,18 @@ public interface ModelSupport {
   @JsonInclude(JsonInclude.Include.NON_EMPTY)
   Optional<String> message();
 
+  /**
+   * Enumeration of support status for an embedding or reranking model.
+   *
+   * <ul>
```
amorton, line 28: nit: if you add these as doc comments on the enum values the IDE will show them , and the coments are associated with the enum values 

/** The model is supported and can be used . */
SUPPORTED,

also, add "can be used when creating new Collections and Tables"

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/VectorizeConfigValidator.java
```diff
@@ -365,6 +361,24 @@ private Integer validateModelAndDimension(
...
+    if (userConfig.modelName() == null) {
+      throw ErrorCodeV1.INVALID_CREATE_COLLECTION_OPTIONS.toApiException(
+          "'modelName' is needed for provider %s", userConfig.provider());
+    }
+    return providerConfig.models().stream()
+        .filter(m -> m.name().equals(userConfig.modelName()))
+        .findFirst()
+        .orElseThrow(
+            () ->
+                ErrorCodeV1.INVALID_CREATE_COLLECTION_OPTIONS.toApiException(
```
amorton, line 377: this PR also adds the error UNSUPPORTED_PROVIDER_MODEL

can we use that ? and I was thinking it should be UNKNOWN_

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/VectorizeConfigValidator.java
```diff
@@ -411,4 +425,36 @@ private Integer validateRangeDimension(
     }
     return userVectorDimension;
   }
+
+  /**
+   * Validates the model support for the vectorization service. This method checks if the model is
+   * SUPPORTED and throw corresponding error if the model's DEPRECATED or END_OF_LIFE.
+   */
+  private void validateModelSupportForSchemaCreation(VectorizeConfig service) {
```
amorton, line 433: we normally use check* for checking a value and throwing

src/main/proto/embedding_gateway.proto
```diff
@@ -159,13 +159,19 @@ message GetSupportedProvidersResponse {
...
+    // ModelConfig message represents configuration for a specific embedding model.
     message ModelConfig {
       string name = 1;
       optional int32 vector_dimension = 2;
       repeated ParameterConfig parameters = 3;
       map<string, string> properties = 4;
       optional string service_url_override = 5;
+      ModelSupport modelSupport = 6;
+
+      message ModelSupport{
```
amorton, line 171: per comments in other places, thinking this should be ApiModelSupport because it's not something the model supports

src/main/resources/errors.yaml
```diff
@@ -870,14 +870,36 @@ request-errors:
       
       Resend the command using supported model.
 
+  - scope: SCHEMA
+    code: DEPRECATED_PROVIDER_MODEL
+    title: The model is DEPRECATED.
```
amorton, line 875: Do not need all caps in the title - change to "Cannot use deprecated model"

src/main/resources/errors.yaml
```diff
@@ -870,14 +870,36 @@ request-errors:
...
+    body: |-
```
amorton, line 876: the approach is to use the first sentance to explain the error, and the list of the values normally has the value at the end. 

Can we change to below and include the list of supported models

The command attempted to create or alter a Collection or Table to use a AI Model that has been marked as deprecated. Deprecated models are only supported by the API for existing use and will later be  removed.

The model is: ${model}
The supported models for the provider are: ${supportedModels}
${message}

Resend the command using supported model.

src/main/resources/errors.yaml
```diff
@@ -870,14 +870,36 @@ request-errors:
...
+      The model ${model} is at ${modelStatus} status.
+      ${message}
+      
+      DEPRECATED model cannot be used for creation of new Collections or Tables, but can be used for existing Collections or Tables until model status is changed to END_OF_LIFE.
+      
+      Resend the command using supported model.
+
+  - scope: SCHEMA
+    code: END_OF_LIFE_PROVIDER_MODEL
+    title: The model is ENF_OF_LIFE.
```
amorton, line 886: change to - Cannot use end of life model

src/main/resources/errors.yaml
```diff
@@ -870,14 +870,36 @@ request-errors:
...
+    body: |-
+      The model ${model} is at ${modelStatus} status.
```
amorton, line 888: The command attempted to use an AI Model that has been marked as end of life. End of life models cannot be used in any way. Collections or Tables that use the model must be recreated as data such as embeddings is not transferrable.

The model is: ${model}
${message}

Resend the command using supported model.

src/main/resources/errors.yaml
Yuqi-Du: thanks! let me fix

Yuqi-Du: Let me do `The supported models for the provider are: ${supportedModels}`
as a followup, in the embeddingProvider, it does not have the context of all the providerInfo(config), I know we don't want any injections anymore, let me think of a better way

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/FindEmbeddingProvidersCommand.java
```diff
@@ -3,12 +3,31 @@
...
+      @Nullable
+          @Schema(
+              description =
+                  "If not provided, only SUPPORTED models will be returned."
+                      + " If provided with \"\" empty string, all SUPPORTED, DEPRECATED, END_OF_LIFE model will be returned."
+                      + " If provided with specified SUPPORTED or DEPRECATED or END_OF_LIFE, only models matched the status will be returned.",
+              type = SchemaType.STRING,
+              implementation = String.class)
+          @Pattern(
+              regexp = "^(SUPPORTED|DEPRECATED|END_OF_LIFE|supported|deprecated|end_of_life|)$")
```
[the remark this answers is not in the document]
Yuqi-Du: thanks

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/FindEmbeddingProvidersCommand.java
```diff
@@ -3,12 +3,31 @@
...
 @JsonTypeName(CommandName.Names.FIND_EMBEDDING_PROVIDERS)
-public record FindEmbeddingProvidersCommand() implements GeneralCommand, NoOptionsCommand {
+public record FindEmbeddingProvidersCommand(
+    @Valid @Nullable @Schema(type = SchemaType.OBJECT, implementation = Options.class)
+        Options options)
+    implements GeneralCommand {
+
+  public record Options(
+      @Nullable
+          @Schema(
```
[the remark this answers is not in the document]
Yuqi-Du: added

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/FindEmbeddingProvidersCommand.java
```diff
@@ -3,12 +3,31 @@
...
+              description =
+                  "If not provided, only SUPPORTED models will be returned."
```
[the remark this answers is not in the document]
Yuqi-Du: fixed

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/FindEmbeddingProvidersCommand.java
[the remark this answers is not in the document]
Yuqi-Du: added.

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/EmbeddingProvider.java
Yuqi-Du: changed to protected, as it will be used by specific provider.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/embeddings/FindEmbeddingProvidersOperation.java
Yuqi-Du: this could be IDEA being mul-funtional, keeps giving me alert on this.
I will remove.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/embeddings/FindEmbeddingProvidersOperation.java
Yuqi-Du: changed

src/main/java/io/stargate/sgv2/jsonapi/service/operation/embeddings/FindEmbeddingProvidersOperation.java
Yuqi-Du: fixed

src/main/java/io/stargate/sgv2/jsonapi/service/operation/embeddings/FindEmbeddingProvidersOperation.java
Yuqi-Du: fixed, with String default, then we can say NULL or empty string will return all statuses

src/main/java/io/stargate/sgv2/jsonapi/service/provider/ModelSupport.java
Yuqi-Du: fixed

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/VectorizeConfigValidator.java
Yuqi-Du: this is refactoring old code.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/embeddings/FindEmbeddingProvidersOperation.java
Yuqi-Du: fixed, thanks!

src/main/java/io/stargate/sgv2/jsonapi/service/operation/embeddings/FindEmbeddingProvidersOperation.java
Yuqi-Du: thanks, agree, this is better.

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/VectorizeConfigValidator.java
Yuqi-Du: renamed

src/main/proto/embedding_gateway.proto
Yuqi-Du: changed

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/EmbeddingProviderFactory.java
Yuqi-Du: this should not happen, I added this error for better coding practice.
because the modelName pass down here already been validated

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/AwsBedrockEmbeddingProvider.java
Yuqi-Du: Agree, I notice this when I added this one.
But it needs another level of abstraction, definitely desires another ticket.

Yuqi-Du: https://github.com/stargate/data-api/issues/2028

PR #2004, opened 2025-04-09
fixes #2000 support $includeScores=false

### LINKED ISSUE #2000: HTTP 500 error with "$similarity document field is not a number or is missing type" includeScores = false for findAndRerank
The issue occurs both on dev and latest main.

Examples using both vectorize and BYO (I'm using upstage; stefano reproduced w/ mistral):

Collection config : 

```json
{
  "vector": {
    "dimension": 4096,
    "metric": "cosine",
    "sourceModel": "other",
    "service": {
      "provider": "upstageAI",
      "modelName": "solar-embedding-1-large"
    }
  },
  "lexical": {
    "enabled": true,
    "analyzer": "standard"
  },
  "rerank": {
    "enabled": true,
    "service": {
      "provider": "nvidia",
      "modelName": "nvidia/llama-3.2-nv-rerankqa-1b-v2"
    }
  }
}
```

has document 
```json
{
  "_id": "603768dd-6b46-4771-b768-dd6b46d7712b",
  "$vectorize": "hello world",
  "$lexical": "goodbye world",
  "$vector": "DataAPIVector<4096>(typeof raw=number[], preview=[0.024719238, 0.0062713623, ...])"
}
```

Read with
```json
{
  "findAndRerank": {
    "filter": {},
    "sort": {
      "$hybrid": "abc"
    },
    "options": {}
  }
}
```

# ERROR

Returns error : 

```json
{
  "errors": [
    {
      "message": "Server failed: root cause: (java.lang.IllegalArgumentException) $similarity document field is not a number or is missing type=<null> _id=603768dd-6b46-4771-b768-dd6b46d7712b",
      "id": "587f861c-a969-42c4-a318-1e51bcdedf37",
      "family": "SERVER",
      "title": "Server failed",
      "scope": "EMPTY"
    }
  ]
}
```

Also fails when using BYO vector

```json
{
  "findAndRerank": {
    "filter": {},
    "sort": {
      "$hybrid": {
        "$vector": {
          "$binary": "PczMzT5MzM0+mZmaPszMzT8AAAA="
        },
        "$lexical": "abc"
      }
    },
    "options": {
      "rerankOn": "$lexical",
      "rerankQuery": "cba"
    }
  }
}
```

returns

```json
{
  "errors": [
    {
      "message": "Server failed: root cause: (java.lang.IllegalArgumentException) $similarity document field is not a number or is missing type=<null> _id=bccdc5be-7a01-4dd3-8dc5-be7a012dd3f7",
      "id": "33e11233-69aa-4494-9c72-f24d6c94a324",
      "title": "Server failed",
      "family": "SERVER",
      "scope": "EMPTY"
    }
  ]
}
```

amorton:
bug in findAndRerank would error when includeScores = false

added support in the ScoredDocument so that if we are doing the ANN read, and we do not have the similarity score, we still track the rank of the ANN read because we need this for the RRF score to tie break the reranking score

Fixes #2000

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/operation/reranking/ScoredDocument.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/operation/reranking/ScoredDocument.java
@@ -92,7 +92,11 @@ public class ScoredDocument implements Comparable<ScoredDocument> {
     DocumentScores vectorScores;
     var similarityField = document.get(VECTOR_FUNCTION_SIMILARITY_FIELD);
     if (rankSource == Rank.RankSource.VECTOR) {
-      if (similarityField instanceof NumericNode numberNode) {
+      if (similarityField == null) {
+        // this can happen when the scores are not going to be returned, OK we only need the rank
+        // for RRF
+        vectorScores = DocumentScores.fromVectorRead(rank);
+      } else if (similarityField instanceof NumericNode numberNode) {
         vectorScores = DocumentScores.fromVectorRead(numberNode.floatValue(), rank);
       } else {
         throw new IllegalArgumentException(
@@ -105,7 +109,7 @@ public class ScoredDocument implements Comparable<ScoredDocument> {
     } else {
       if (similarityField != null) {
         throw new IllegalArgumentException(
-            "rankSource is %s but the document with id '%s' has %s field=%s"
+            "rankSource is %s but the document has similarity score id=%s has %s=%s"
                 .formatted(
                     rankSource, documentId, VECTOR_FUNCTION_SIMILARITY_FIELD, similarityField));
       }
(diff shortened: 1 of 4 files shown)


src/main/java/io/stargate/sgv2/jsonapi/service/operation/reranking/ScoredDocument.java
```diff
@@ -105,7 +109,7 @@ static ScoredDocument create(
     } else {
       if (similarityField != null) {
```
Yuqi-Du, line 110: This won't happen at all, correct?
RerankSource is NOT VECTOR, but document response has similarityField.

I know this is just some assertions, but wonder if it is unnecessary, feel free to ignore though.

Yuqi-Du (approved):
LGTM

src/main/java/io/stargate/sgv2/jsonapi/service/operation/reranking/ScoredDocument.java
amorton: yeah, just sanity checking what we expect to happen

PR #2014, opened 2025-04-17
Fixes #2013: avoid fail for createCollection wrt legacy collections (pre-lexical/pre-rerank)

### LINKED ISSUE #2013: Reranking and Lexical feature enabled in production caused backwards compatibility issue for createCollection.
Data API rollout the hybrid search in certain Astra production kube, which will enable lexical and reranking features by default.
So 'legacy' collection created before the feature rollout won't have lexical and reranking setting. 
But now since Data API will include the lexical and reranking to the command by default, the createCollection command won't be idempotent in this situation, and will error out as EXISTING_COLLECTION_DIFFERENT_SETTINGS.

tatu-at-datastax:
**What this PR does**:

Fixes issue wrt "createCollection" failing for legacy Collections created before lexical/rerank functionality, due to changing defaults.
Will prevent throwing exception if:

1. Existing collection that has no lexical or reranking config
2. New creation that relies on default settings for lexical and reranking config

This should no longer fail as it considers "defaults are defaults" (despite defaults effectively changing for new Collections) 

Fixes #2013

**Checklist**
- [x] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/service/operation/collections/CreateCollectionOperation.java
```diff
@@ -162,11 +163,38 @@ public Uni<Supplier<CommandResult>> execute(
...
+              == CollectionLexicalConfig.configForPreLexical()
+          && existingCollectionSettings.rerankingConfig()
+              == CollectionRerankDef.configForPreRerankingCollection()) {
+        newCollectionSettings =
+            newCollectionSettings.withLexicalAndRerankOverrides(
+                existingCollectionSettings.lexicalConfig(),
+                existingCollectionSettings.rerankingConfig());
+        // and now re-check if settings are the same
+        sameSettings = existingCollectionSettings.equals(newCollectionSettings);
+        logger.info(
```
tatu-at-datastax, line 187: Adding this so we get some visibility into how often override needed, works.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/collections/CreateCollectionOperation.java
```diff
@@ -402,9 +440,15 @@ TableMetadata findTableAndValidateLimits(
     return null;
   }
 
-  public SimpleStatement getCreateTable(String keyspace, String table) {
+  public static SimpleStatement getCreateTable(
```
tatu-at-datastax, line 443: Changing this to `static` since we should be using that from some of Integration tests (unfortunately we have copy-pasted versions). Also ensures all arguments are explicitly passed since call paths differ a bit wrt overrides, specifically "lexicalConfig".

src/main/java/io/stargate/sgv2/jsonapi/service/operation/collections/CreateCollectionOperation.java
```diff
@@ -428,27 +472,26 @@ public SimpleStatement getCreateTable(String keyspace, String table) {
         createTableWithVector = createTableWithVector + " WITH comment = '" + comment + "'";
       }
       return SimpleStatement.newInstance(String.format(createTableWithVector, keyspace, table));
-    } else {
```
tatu-at-datastax, line 431: No functional change, just remove unnecessary "else" scope.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/collections/CreateCollectionOperation.java
```diff
@@ -162,11 +163,38 @@ public Uni<Supplier<CommandResult>> execute(
             embeddingSourceModel,
             comment,
             objectMapper);
-    // if table exists we have two choices:
+    // if table exists we have a choice:
     // (1) trying to create with same options -> ok, proceed
     // (2) trying to create with different options -> error out
-    if (existedCollectionSettings.equals(newCollectionSettings)) {
-      return executeCollectionCreation(dataApiRequestInfo, queryExecutor, true);
+    // but note that for case (1) we need to consider backwards-compatibility
```
Yuqi-Du, line 169: this line may need some tweak?

case (1) is saying oldSetting and newSetting are the same, but our special handing happens when they are **not** the same and then override the newSetting, then re-compare.

maybe changed to (Feel free to not change or adapt)
```
but note that for case (2), we need to consider backwards-compatibility, as we have collections created before some of the settings were added(namely, lexical and reranking settings), and now added in newCollectionSettings.
```

src/main/java/io/stargate/sgv2/jsonapi/service/operation/collections/CreateCollectionOperation.java
```diff
@@ -162,11 +163,38 @@ public Uni<Supplier<CommandResult>> execute(
...
+          && existingCollectionSettings.rerankingConfig()
+              == CollectionRerankDef.configForPreRerankingCollection()) {
+        newCollectionSettings =
+            newCollectionSettings.withLexicalAndRerankOverrides(
+                existingCollectionSettings.lexicalConfig(),
+                existingCollectionSettings.rerankingConfig());
+        // and now re-check if settings are the same
+        sameSettings = existingCollectionSettings.equals(newCollectionSettings);
+        logger.info(
+            "CreateCollectionOperation for {}.{} with legacy lexical/reranking settings: unification successful? {}",
```
Yuqi-Du, line 188: `unification successful?` do we need the question mark.

also, would it be redundant to also log the old and new collectionSetting

src/main/java/io/stargate/sgv2/jsonapi/service/schema/collections/CollectionLexicalConfig.java
```diff
@@ -125,18 +131,26 @@ public static CollectionLexicalConfig validateAndConstruct(
...
   /**
-   * Accessor for an instance to use for "lexical disabled" cases: either for existing collections
-   * without lexical config, or envi
+   * Accessor for a singleton instance used to represent case of default lexical configuration for
+   * newly created Collections that do not specify lexical configuration.
    */
-  public static CollectionLexicalConfig configForDisabled() {
-    return new CollectionLexicalConfig(false, null);
+  public static CollectionLexicalConfig configForDefault() {
+    return DEFAULT_CONFIG;
```
Yuqi-Du, line 146: This is nice to uniform both DEFAULT_CONFIG and MISSING_CONFIG.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/collections/CreateCollectionOperation.java
tatu-at-datastax: Good point, comment does not explain what we do... will tweak.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/collections/CreateCollectionOperation.java
tatu-at-datastax: Good idea: `CollectionSchemaObject.toString()` has `toString()` so I can do that.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/collections/CreateCollectionOperation.java
tatu-at-datastax: Changed, I hope it reads better now.

Yuqi-Du (approved):
approved

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/VectorConfig.java
```diff
@@ -167,6 +167,16 @@ public int hashCode() {
     return Objects.hash(columnVectorDefinitions, vectorEnabled);
   }
 
+  @Override
```
tatu-at-datastax, line 170: Was missing, noticed from new log lines I added.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/collections/CreateCollectionOperation.java
```diff
@@ -162,11 +163,58 @@ public Uni<Supplier<CommandResult>> execute(
...
+
+      // So: for backwards compatibility reasons we may need to override settings if
+      // (and only if) the collection was created before lexical and reranking.
+      // In addition, we need to check that new lexical settings are for defaults
+      // (difficult to check the same for reranking; for now assume that if lexical
+      // is default, reranking is also default).
+      if (oldLexical == CollectionLexicalConfig.configForPreLexical()
+          && newLexical == CollectionLexicalConfig.configForDefault()
+          && existingCollectionSettings.rerankingConfig()
+              == CollectionRerankDef.configForPreRerankingCollection()) {
```
Yuqi-Du, line 189: Miss one item here for rerank?

tatu-at-datastax: unfortunately no; we do not have "default" entry for reranking, so cannot use identity comparison. :-/
So that's a limitation -- comment above tries to explain it.

src/main/java/io/stargate/sgv2/jsonapi/service/reranking/configuration/RerankingProviderConfigProducer.java
```diff
@@ -71,6 +72,11 @@ RerankingProvidersConfig produce(
     }
 
     validateRerankingProvidersConfig(rerankingProvidersConfig);
+
+    // Initialize the default reranking provider and model in DefaultRerankingProviderDef as
+    // Singleton.
+    CollectionRerankDef.initializeDefaultRerankDef(rerankingProvidersConfig);
```
tatu-at-datastax, line 78: I wish we did not need to do this -- relying on state of Quarkus initialization. But I understand why it is necessary for what we need.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/collections/CreateCollectionOperation.java
tatu-at-datastax: update: Yuqi committed a fix to this omission so we are all good!

PR #2058, opened 2025-04-30
Add metrics for usage of hybrid search features

### LINKED ISSUE #1987: Add metrics for usage of hybrid search features
Metrics on the use of: 
- inserting with $hybrid  and $lexical 
- calling findAndRerank and sorting on $hybrid, $lexical, etc

Hazel-Datastax:
**What this PR does**:
Add metrics for usage of hybrid search features

Fixes #1987 

**Checklist**
- [x] Changes manually tested


Hazel-Datastax:
Send `findAndRerank` command
```
{
    "findAndRerank": {
        "sort": {
            "$hybrid": {
                "$vectorize": "i like cheese", 
                "$lexical" : "monkeys"
            }
        },
        "options": {
            "hybridLimits" : 10
        }
    }
}
```
The metrics will be like 
```
command_processor_process_seconds_max{command="FindAndRerankCommand",error="false",error_class="NA",error_code="NA",feature_hybrid_lexical="true",feature_hybrid_limits_number="true",feature_hybrid_vectorize="true",module="sgv2-jsonapi",sort_type="NONE",tenant="unknown",vector_enabled="true",} 4.177825209
```

We can see `feature_hybrid_lexical`, `feature_hybrid_vectorize` and `feature_hybrid_limits_number` in the tag

Hazel-Datastax:
Send insertOne Command:
```
{
    "insertOne": {
        "document": {
            "$lexical": "I like cheese",
            "$vectorize": "I like cheese"
        }
    }
}
```
metrics:
```
command_processor_process_seconds_sum{command="InsertOneCommand",error="false",error_class="NA",error_code="NA",feature_hybrid_lexical="false",feature_hybrid_limits_lexical="false",feature_hybrid_limits_number="false",feature_hybrid_limits_vector="false",feature_hybrid_string="false",feature_hybrid_vector="false",feature_hybrid_vectorize="false",feature_lexical="true",feature_vector="false",feature_vectorize="true",module="sgv2-jsonapi",sort_type="NONE",tenant="unknown",vector_enabled="true",} 4.148228625
```
We can see `feature_lexical` and `feature_vectorize` in the tag are `true`, other features are `false`

Send insertOne command:
```
{
    "insertOne": {
        "document": {
            "$hybrid": {
                "$lexical": "I like cheese",
                "$vectorize": "I like cheese"
            }
        }
    }
}
```
metrics:
```
command_processor_process_seconds_max{command="InsertOneCommand",error="false",error_class="NA",error_code="NA",feature_hybrid_lexical="true",feature_hybrid_limits_lexical="false",feature_hybrid_limits_number="false",feature_hybrid_limits_vector="false",feature_hybrid_string="false",feature_hybrid_vector="false",feature_hybrid_vectorize="true",feature_lexical="false",feature_vector="false",feature_vectorize="false",module="sgv2-jsonapi",sort_type="NONE",tenant="unknown",vector_enabled="true",} 7.453378
```

We can see `feature_hybrid_lexical` and `feature_hybrid_vectorize` in the tag are `true`, other features are `false`

src/main/java/io/stargate/sgv2/jsonapi/service/processor/HybridFieldExpander.java
```diff
@@ -42,24 +43,45 @@ public static void expandHybridField(CommandContext context, Command command) {
...
+          context.addFeature(Feature.VECTOR);
+        }
+        // `$vectorize` and `$vector` can't be used together - the check will be done later (in
+        // DataVectorizer)
+        if (doc.get(DocumentConstants.Fields.VECTOR_EMBEDDING_TEXT_FIELD) != null) {
+          context.addFeature(Feature.VECTORIZE);
+        }
+        if (doc.get(DocumentConstants.Fields.LEXICAL_CONTENT_FIELD) != null) {
+          context.addFeature(Feature.LEXICAL);
+        }
```
Hazel-Datastax, line 78: Maybe it's not good to add here, but I cannot think of any since after this point, the $hybrid will be replaced to `$lexical` or/and `$vectorize`

src/main/java/io/stargate/sgv2/jsonapi/service/operation/reranking/Feature.java
```diff
@@ -0,0 +1,43 @@
+package io.stargate.sgv2.jsonapi.service.operation.reranking;
```
tatu-at-datastax, line 1: Is this specific to reranking? If not, should probably be in some other package. But if it is, maybe name should reflect it (see another comment below)

src/main/java/io/stargate/sgv2/jsonapi/service/operation/reranking/Feature.java
```diff
@@ -0,0 +1,43 @@
...
+
+import java.util.Objects;
+
+/**
+ * Represents distinct features that can be used in the command. Each feature has an associated tag
+ * name, which can be used for metrics.
+ */
+public enum Feature {
```
tatu-at-datastax, line 9: Name "Feature" feels very generic... should probably be something more specific as we also have `ApiFeature`? If this is specific to reranking, perhaps `RerankingFeature`?
If more general, `CommandFeature`?

src/main/java/io/stargate/sgv2/jsonapi/service/operation/reranking/Feature.java
```diff
@@ -0,0 +1,43 @@
+package io.stargate.sgv2.jsonapi.service.operation.reranking;
+
+import java.util.Objects;
+
+/**
+ * Represents distinct features that can be used in the command. Each feature has an associated tag
```
tatu-at-datastax, line 6: Which command? Any command or just Reranking command?

src/main/java/io/stargate/sgv2/jsonapi/service/operation/reranking/FeatureSource.java
```diff
@@ -0,0 +1,13 @@
+package io.stargate.sgv2.jsonapi.service.operation.reranking;
+
+import io.stargate.sgv2.jsonapi.api.model.command.impl.FindAndRerankCommand;
+
+/**
+ * An interface for objects that can provide information about the {@link Feature}s they utilize.
+ * This is typically implemented by commands to declare which features they depend on. See {@link
+ * FindAndRerankCommand} as an examples
```
tatu-at-datastax, line 8: typo: "as an example" -> "as an example"

src/main/java/io/stargate/sgv2/jsonapi/service/operation/reranking/FeatureUsage.java
```diff
@@ -0,0 +1,125 @@
...
+import java.util.Objects;
+import java.util.stream.Collectors;
+
+/**
+ * Represents a collection of {@link Feature}s in a command. This class is immutable and can be used
+ * to track which features are utilized in a command. This class is immutable; methods that modify
+ * the set of features return a new {@code FeatureUsage} instance. It uses an {@link EnumSet}
+ * internally for efficient storage and operations on features.
+ */
+public final class FeatureUsage {
```
tatu-at-datastax, line 14: If we renamed `Feature` as `CommandFeature`, then this should probably become just `CommandFeatures` (without "usage").

src/main/java/io/stargate/sgv2/jsonapi/service/operation/reranking/FeatureUsage.java
```diff
@@ -0,0 +1,125 @@
...
+   * Creates a {@code FeatureUsage} instance from an array of {@link Feature}s.
+   *
+   * @param features The features to include. If null or empty, {@link #EMPTY} is returned.
+   * @return A new {@code FeatureUsage} instance containing the specified features.
+   */
+  public static FeatureUsage of(Feature... features) {
+    if (features == null || features.length == 0) {
+      return EMPTY;
+    }
+    EnumSet<Feature> set = EnumSet.noneOf(Feature.class);
```
tatu-at-datastax, line 34: Can simplify as:

    return new FeatureUsage(EnumSet.of(features[0], features));

(thanks to Claude)

src/main/java/io/stargate/sgv2/jsonapi/service/operation/reranking/FeatureUsage.java
```diff
@@ -0,0 +1,125 @@
...
+ * internally for efficient storage and operations on features.
+ */
+public final class FeatureUsage {
+  private final EnumSet<Feature> features;
+
+  /** A instance representing no features in use. */
+  public static final FeatureUsage EMPTY = new FeatureUsage(EnumSet.noneOf(Feature.class));
+
+  private FeatureUsage(EnumSet<Feature> features) {
+    this.features = EnumSet.copyOf(features);
```
tatu-at-datastax, line 21: Should perhaps let caller make a copy where needed, to avoid double-copying.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/reranking/FeatureUsage.java
```diff
@@ -0,0 +1,125 @@
...
+   * @throws NullPointerException if the feature is null.
+   */
+  public FeatureUsage withFeature(Feature feature) {
+    Objects.requireNonNull(feature, "Feature cannot be null");
+    if (this.features.contains(feature)) {
+      return this;
+    }
+    EnumSet<Feature> newSet = EnumSet.copyOf(this.features);
+    newSet.add(feature);
+    return new FeatureUsage(newSet);
```
tatu-at-datastax, line 68: Wrt copying: this `EnumSet` does not need to be copied in constructor as it is not referenced by anything else.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/reranking/FeatureUsage.java
```diff
@@ -0,0 +1,125 @@
...
+   * Returns a new {@code FeatureUsage} instance that is the union of this instance's features and
+   * the features of another {@code FeatureUsage} instance.
+   *
+   * @param other The other {@code FeatureUsage} instance to combine with. If null, empty, or the
+   *     same as this instance, this instance is returned. If this instance is empty, the other
+   *     instance is returned.
+   * @return A new {@code FeatureUsage} instance representing the combined set of features.
+   */
+  public FeatureUsage unionWith(FeatureUsage other) {
+    if (other == null || other.features.isEmpty() || other == this) {
```
tatu-at-datastax, line 81: Insted of accessing `other.features` directly, add `isEmpty()` method, and check `other.isEmpty()`.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/reranking/FeatureUsage.java
```diff
@@ -0,0 +1,125 @@
...
+    if (obj == null || getClass() != obj.getClass()) {
+      return false;
+    }
+    FeatureUsage that = (FeatureUsage) obj;
+    return Objects.equals(features, that.features);
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(features);
```
tatu-at-datastax, line 116: since `features` cannot be `null`, can just use

    return features.hashCode();

src/main/java/io/stargate/sgv2/jsonapi/service/processor/MeteredCommandProcessor.java
```diff
@@ -306,9 +310,42 @@ private <T extends SchemaObject> Tags getCustomTags(
     JsonApiMetricsConfig.SortType sortType = getVectorTypeTag(commandContext, command);
     Tag sortTypeTag = Tag.of(jsonApiMetricsConfig.sortType(), sortType.name());
 
+    // --- Command Feature Usage Tags ---
+    if (command instanceof FeatureSource fs) {
```
tatu-at-datastax, line 314: This looks odd: why not just call `addFeatureUsage` no matter what, let context decide if processing needed?

src/main/java/io/stargate/sgv2/jsonapi/service/operation/reranking/FeatureUsage.java
```diff
@@ -0,0 +1,125 @@
...
+    return new FeatureUsage(combined);
+  }
+
+  /**
+   * Gets a copy of the set of features contained in this {@code FeatureUsage}. The returned set is
+   * a defensive copy and modifications to it will not affect this {@code FeatureUsage} instance.
+   *
+   * @return An {@link EnumSet} containing the features.
+   */
+  public EnumSet<Feature> getFeatures() {
```
tatu-at-datastax, line 98: If the only purpose for this accessor is to build `Tags`, would be better to add method like `getAsTags()` or such -- handle construction here, without exposing `EnumSet` (and without having to create all these defensive copies).

Or maybe more like `Tags addTags(Tags tags)` if caller already has tags to add to.

src/main/java/io/stargate/sgv2/jsonapi/service/processor/MeteredCommandProcessor.java
```diff
@@ -306,9 +310,42 @@ private <T extends SchemaObject> Tags getCustomTags(
...
+
+  /**
+   * Adds tags for all defined features. If a feature is present in FeatureUsage, its tag value is
+   * "true", otherwise "false".
+   */
+  private Tags getCommandFeatureTags(FeatureUsage featureUsage) {
+    Tags tags = Tags.empty();
+    Set<Feature> usedFeatures = Collections.emptySet();
+
+    if (featureUsage != null && featureUsage.getFeatures() != null) {
```
tatu-at-datastax, line 339: Should this be ever called with `featureUsage == null`? If not, remove check.

But overall I think all of this logic of building `Tags` should be moved in `FeatureUsage` (whatever it will be named) -- as per my comments for that class.

tatu-at-datastax (commented):
Ok: functionality itself makes sense, but I think I have some concerns about naming ("Feature" is a bit generic term), choice of package (I think features are not rerank-specific so probably should not be in "...operation.reranking"), and placement of methods (`FeatureUsage` could build `Tags`).

src/main/java/io/stargate/sgv2/jsonapi/service/operation/reranking/Feature.java
amorton: +1 for it being general, could be in the metrics package

src/main/java/io/stargate/sgv2/jsonapi/service/operation/reranking/Feature.java
amorton: the idea was to be general, maybe CommandFeature is a good idea - most of what we do is command based

src/main/java/io/stargate/sgv2/jsonapi/service/operation/reranking/FeatureUsage.java
amorton: +1 there is a double copy happening here

src/main/java/io/stargate/sgv2/jsonapi/service/operation/reranking/FeatureUsage.java
amorton: i like the idea of tracking the features as features, that can then be turned into tags. It may be that not all features turn into tags, e.g. if we want to put something into a log line

src/main/java/io/stargate/sgv2/jsonapi/service/operation/reranking/FeatureUsage.java
```diff
@@ -0,0 +1,125 @@
...
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(features);
+  }
+
+  @Override
+  public String toString() {
+    return "FeatureUsage{"
+        + features.stream().map(Enum::name).collect(Collectors.joining(", "))
```
amorton, line 122: EumSet roughly does this

src/main/java/io/stargate/sgv2/jsonapi/service/operation/reranking/FeatureUsage.java
amorton: But +1 for this class having a getTags() function as well that will return a Tags object

src/main/java/io/stargate/sgv2/jsonapi/service/operation/reranking/FeatureUsage.java
amorton, line 14: I think we can change this from Immutable (which is normally a good idea) to mutable given the way it is used in the CommandContext. We are collecting the feature usage as we go, so it is designed to be mutated. 

Also, because this class is wrapping the EnumSet we are controlling the mutations. i.e. it is not possible to accidently remove or clear the set. 

Often we choose Immutable structures because (IMHO) most bugs come from unexpected state mutation. This class is designed to control the state mutation.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/CommandContext.java
```diff
@@ -160,6 +167,18 @@ public ApiFeatures apiFeatures() {
...
 
+  public FeatureUsage featureUsage() {
+    return featureUsage;
+  }
+
+  public void addFeature(Feature feature) {
+    this.featureUsage = this.featureUsage.withFeature(feature);
+  }
+
+  public void addFeatureUsage(FeatureUsage featureUsage) {
```
amorton, line 178: see comments in the FeatureUsage class to make this mutable. 

As personal preference we the CommandContext should just probide access to the FeatureUsage, not provide helpers to call functions that are on it. That breaks encapsulation.

src/main/java/io/stargate/sgv2/jsonapi/service/processor/MeteredCommandProcessor.java
amorton: I think we can move this to the CommandProcessor proper - we want to eventually kill this class.

src/main/java/io/stargate/sgv2/jsonapi/service/processor/MeteredCommandProcessor.java
```diff
@@ -306,9 +310,42 @@ private <T extends SchemaObject> Tags getCustomTags(
...
+            vectorEnabled,
+            sortTypeTag)
+        .and(commandFeatureTags);
+  }
+
+  /**
+   * Adds tags for all defined features. If a feature is present in FeatureUsage, its tag value is
+   * "true", otherwise "false".
+   */
+  private Tags getCommandFeatureTags(FeatureUsage featureUsage) {
```
amorton, line 335: This shoudl be on the FeatureUsage class, that is the holder of all of the features

src/main/java/io/stargate/sgv2/jsonapi/service/processor/MeteredCommandProcessor.java
```diff
@@ -306,9 +310,42 @@ private <T extends SchemaObject> Tags getCustomTags(
...
+    Tags tags = Tags.empty();
+    Set<Feature> usedFeatures = Collections.emptySet();
+
+    if (featureUsage != null && featureUsage.getFeatures() != null) {
+      usedFeatures = featureUsage.getFeatures();
+    }
+
+    for (Feature feature : Feature.values()) {
+      boolean isFeaturePresent = usedFeatures.contains(feature);
```
amorton, line 344: We only want to emit tags when a feature is used. We can safely assume for any given request it will only use a minority of the features. But this code will be tracking all the things it did not use as well, so when scalled up to many requests we are tracking a lot. 

So when we move this into the FeatureUsage , we can iterate the EnumSet.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/sort/FindAndRerankSort.java
```diff
@@ -26,17 +28,25 @@
               {"$sort" : {"$hybrid" : {"$vectorize" : "vectorize sort query" , "$lexical": "lexical sort" }}}
               {"$sort" : {"$hybrid" : {"$vector" : [1,2,3] , "$lexical": "lexical sort" }}}
       """)
-public record FindAndRerankSort(String vectorizeSort, String lexicalSort, float[] vectorSort)
-    implements Recordable {
+public record FindAndRerankSort(
+    String vectorizeSort, String lexicalSort, float[] vectorSort, FeatureUsage featureUsage)
```
amorton, line 32: -1  for adding this as a property for the sort.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/sort/FindAndRerankSortClauseDeserializer.java
```diff
@@ -97,6 +105,7 @@ private static FindAndRerankSort deserializeHybridObject(
       JsonParser jsonParser, ObjectNode hybridObject) throws JsonMappingException {
 
     var sortMatch = MATCH_SORT_FIELDS.matchAndThrow(hybridObject, jsonParser, ERROR_CONTEXT);
+    EnumSet<Feature> features = EnumSet.noneOf(Feature.class);
```
amorton, line 108: why not use the FeatureUsage class ?

amorton (changes requested):
some discussions in here, and lets change FeatureUsage to be mutable and attached

src/main/java/io/stargate/sgv2/jsonapi/service/operation/reranking/Feature.java
Hazel-Datastax: General command, renamed to `CommandFeature`

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/sort/FindAndRerankSort.java
```diff
@@ -49,12 +59,13 @@ public DataRecorder recordTo(DataRecorder dataRecorder) {
   public boolean equals(Object obj) {
     return Objects.equals(vectorizeSort, ((FindAndRerankSort) obj).vectorizeSort)
```
tatu-at-datastax, line 60: nit: should probably check that `obj` is a `FindAndRerankSort` first, to avoid `ClassCastException` and having to cast on every line. So like:

```
    return (obj instanceof FindAndRerankSort other)
        && Objects.equals(vectorizeSort, other.vectorizeSort)
        && Objects.equals(lexicalSort, other.lexicalSort)
        ...
```

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/sort/FindAndRerankSort.java
Hazel-Datastax: If we don't want to add this property, how can we know the feature? In `FindAndRerankSortClauseDeserializer`, we've replaced `$hybrid` string or object to `$vectorize` and `$lexical`

src/main/java/io/stargate/sgv2/jsonapi/service/processor/MeteredCommandProcessor.java
Hazel-Datastax: I know this looks odd. We try to get the `CommandFeature` in two different ways:
1. Let the command implement `FeatureSource` and the `CommandFeature` by calling `getCommandFeatures`
2. Gather the `CommandFeature`s  from the command processing process and store them in the `CommandContext`

The reason why I have to do this is we expand `$hybrid` in two different ways:
1. In `FindAndRerankCommand`, we handle `$hybrid` in the deserializers (see `FindAndRerankSortClauseDeserializer` and `HybridLimitsDeserializer`) and we lose info on how the user uses `$hybrid` after the deserializers but before we touch the `CommandContext`
2. For insert commands (`InsertOneCommand` and `InsertManyCommand`), we expand `$hybrid` in `HybridFieldExpander` - after the deserialization, and it's in the command processing process. I thought there are two ways to handle this:
   - Add deserializers for insert commands - similar to `FindAndRerankCommand`
   - Use `CommandContext` to store the information for me
   - I used the second one. 


> I think we can move this to the CommandProcessor proper - we want to eventually kill this class.

Yes, we can. But we still need two different ways to gather, unless we want to refactor some structures.

src/main/java/io/stargate/sgv2/jsonapi/service/processor/MeteredCommandProcessor.java
Hazel-Datastax: I actually tried that before, but that will cause a problem. For example, if I first send
```
{
    "insertOne": {
        "document": {
            "$hybrid": {
                "$lexical": "I like cheese",
                "$vectorize": "I like cheese"
            }
        }
    }
}
```
I can see the tags and metrics in the `/metrics` endpoint: `{name="...", command="InsertOneCommand", ..., feature_hybrid_lexical="true", feature_hybrid_vectorize="true", ...}`

but if then I send the second request
```
{
    "insertOne": {
        "document": {
            "$lexical": "I like cheese",
            "$vectorize": "I like cheese"
        }
    }
}
```

I CANNOT find the tags and metrics in the `/metrics` endpoint: `{name="...", command="InsertOneCommand", ..., feature_lexical="true", feature_vectorize="true", ...}`

The reason is, 
- Micrometer identifies these as two distinct meters because their `Meter.Id` (name + tags) is different. When Request 2 comes in with a new combination of feature tags, Micrometer registers a new `Timer` instance.
- The "disappearing" second metric is likely a race condition with Prometheus's scrape interval. The new meter for Request 2 is registered upon its first use. If the Prometheus scrape happens before this newly registered meter is included in the /metrics endpoint's next snapshot (or before the application has had a chance to fully prepare it for exposure), Prometheus won't see it on that scrape. It typically appears after one or two subsequent scrape cycles once the registration is fully processed and the meter has data. (AI told me, but I still cannot see it no matter how I refresh the endpoint, not sure if this is related to my local environment)

This change could also cause some problems:

1. Increased Cardinality: This is the most significant implication. While we are reducing the number of tags per metric instance (no more feature_x="false"), we are potentially increasing the total number of unique time series in Prometheus if there are many different combinations of features that can be used.
2. Metric "Discovery" Latency: As observed, new combinations of features (and thus new time series) might take a scrape cycle or two to appear in monitoring dashboards. This isn't a flaw in the code but a characteristic of how new series are discovered by scraping systems.
3. Querying Complexity (Minor): When querying in Prometheus/Grafana, users will need to be aware that the feature tags are only present if the feature was used. This is different from before where they could query for feature_x="true" or feature_x="false". Queries might look more like metric{feature_x="true"} or metric_count unless ON(instance) metric{feature_x="true"} to find instances without a feature if the tag is simply absent. This is a standard pattern when dealing with presence-based tags.

src/main/java/io/stargate/sgv2/jsonapi/service/processor/HybridFieldExpander.java
```diff
@@ -42,24 +43,45 @@ public static void expandHybridField(CommandContext context, Command command) {
...
+          }
+          case ObjectNode ob -> addFromObject(context, doc, ob, docIndex, docCount);
+          default ->
+              throw ErrorCodeV1.HYBRID_FIELD_UNSUPPORTED_VALUE_TYPE.toApiException(
+                  "expected String, Object or `null` but received %s (Document %d of %d)",
+                  JsonUtil.nodeTypeAsString(hybridField), docIndex + 1, docCount);
+        }
+      } else {
+        // No $hybrid field, check other fields and add feature usage to CommandContext
+        if (doc.get(DocumentConstants.Fields.VECTOR_EMBEDDING_FIELD) != null) {
```
amorton, line 68: From discussions - move this to the Insert commands so they are tracking what they see. This class can just track if it sees the $hybrid feature

src/main/java/io/stargate/sgv2/jsonapi/service/processor/MeteredCommandProcessor.java
amorton: Form discussions: we dont want to emit FALSE tags because most features wont be used on most commands, e.g. when we have this for insertOne it wont be using the hybrid sort features. 

Then the next thing is for the two commands above, should they have the same tags ? The answer is one should be a subset of the other: 
* first - should be tagged with hybrid=true, lexical=true, vectorize=true
* second - should be tagged with lexical=true, vectorize=true

So if we was looking for insert commands that used leixcal or vectorize they would both be included. Or if I wanted insert that used the hybrid keyword and only vectorize I could do that. (using promql `metric_name{label_name!~".*"}` ) 

So the answer may be that we want to remove these features from the enum 

```java
  /** The usage of $hybrid Object with $lexical in the command */
  HYBRID_LEXICAL("feature.hybrid.lexical"),
  /** The usage of $hybrid Object with $vector in the command */
  HYBRID_VECTOR("feature.hybrid.vector"),
  /** The usage of $hybrid Object with $vectorize in the command */
  HYBRID_VECTORIZE("feature.hybrid.vectorize"),
```

we can use the ones below AND set the $hybrid feature what that is used 

```java
  /** The usage of $lexical in the command */
  LEXICAL("feature.lexical"),
  /** The usage of $vector in the command */
  VECTOR("feature.vector"),
  /** The usage of $vectorize in the command */
  VECTORIZE("feature.vectorize"),
```

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/sort/FindAndRerankSort.java
amorton: good point, ok for now and we need to learn how to use FilterSpec :) from @tatu-at-datastax

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/FindAndRerankCommand.java
```diff
@@ -48,6 +51,19 @@ public CommandName commandName() {
     return CommandName.FIND_AND_RERANK;
   }
 
+  @Override
+  public void addCommandFeaturesToCommandContext(CommandContext<?> context) {
+    context
+        .commandFeatures()
+        .addAll((sortClause != null) ? sortClause.getCommandFeatures() : CommandFeatures.EMPTY);
```
tatu-at-datastax, line 58: nit: instead of chained calls maybe simpler to do:

```
if (sortClause != null) {
  context
    .commandFeatures()
    .addAll(sortClause.getCommandFeatures());
}
```

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/FindAndRerankCommand.java
```diff
@@ -48,6 +51,19 @@ public CommandName commandName() {
...
+    context
+        .commandFeatures()
+        .addAll(
```
tatu-at-datastax, line 61: and same here; move `(options != null && options.hybridLimits() != null)` check into if-statement.

src/main/java/io/stargate/sgv2/jsonapi/service/processor/HybridFieldExpander.java
```diff
@@ -42,24 +43,51 @@ public static void expandHybridField(CommandContext context, Command command) {
...
+            addFromObject(context, doc, ob, docIndex, docCount);
+          }
+          default ->
+              throw ErrorCodeV1.HYBRID_FIELD_UNSUPPORTED_VALUE_TYPE.toApiException(
+                  "expected String, Object or `null` but received %s (Document %d of %d)",
+                  JsonUtil.nodeTypeAsString(hybridField), docIndex + 1, docCount);
+        }
+      } else {
+        // No $hybrid field, check other fields and add feature usage to CommandContext
+        if (doc.get(DocumentConstants.Fields.VECTOR_EMBEDDING_FIELD) != null
```
tatu-at-datastax, line 71: suggestion: could use `if (doc.has(DocumentConstants.Fields.VECTOR_EMBEDDING_FIELD) ...` instead of `if (doc.get(...) != null) { ...`

tatu-at-datastax (approved):
LGTM: added very minor suggestions.

src/main/java/io/stargate/sgv2/jsonapi/service/processor/HybridFieldExpander.java
```diff
@@ -42,24 +43,51 @@ public static void expandHybridField(CommandContext context, Command command) {
...
+            addFromObject(context, doc, ob, docIndex, docCount);
+          }
+          default ->
+              throw ErrorCodeV1.HYBRID_FIELD_UNSUPPORTED_VALUE_TYPE.toApiException(
+                  "expected String, Object or `null` but received %s (Document %d of %d)",
+                  JsonUtil.nodeTypeAsString(hybridField), docIndex + 1, docCount);
+        }
+      } else {
+        // No $hybrid field, check other fields and add feature usage to CommandContext
+        if (doc.has(DocumentConstants.Fields.VECTOR_EMBEDDING_FIELD)
```
amorton, line 71: this should move to be part of the insertOne and insertMany commands, we can do that as a fast follow if we want. 

That is to handle the case where the insert is used without $hybrid

src/main/java/io/stargate/sgv2/jsonapi/service/processor/HybridFieldExpander.java
```diff
@@ -42,24 +43,51 @@ public static void expandHybridField(CommandContext context, Command command) {
...
-                "expected String, Object or `null` but received %s (Document %d of %d)",
-                JsonUtil.nodeTypeAsString(hybridField), docIndex + 1, docCount);
+    if (docNode instanceof ObjectNode doc) {
+      // Check for $hybrid field
+      if ((hybridField = doc.remove(DocumentConstants.Fields.HYBRID_FIELD)) != null) {
+        switch (hybridField) {
+            // this is {"$hybrid" : null}
+          case NullNode ignored -> addLexicalAndVectorize(doc, hybridField, hybridField);
+          case TextNode ignored -> {
+            context.commandFeatures().addFeature(CommandFeature.HYBRID);
```
amorton, line 57: we can add the feature if the if statement is true, which means the user include $hybrid in the command.

src/main/java/io/stargate/sgv2/jsonapi/service/processor/HybridFieldExpander.java
```diff
@@ -76,9 +104,15 @@ private static void addFromObject(ObjectNode doc, ObjectNode hybrid, int docInde
     lexical =
         validateSubFieldType(
             lexical, DocumentConstants.Fields.LEXICAL_CONTENT_FIELD, docIndex, docCount);
+    if (!lexical.isNull()) {
```
amorton, line 107: this and the vectorize below should be moved to be tracked in the insertOne and insertMany commands. 

the only thing this needs to do is track if the user include $hybrid in the command.

src/main/java/io/stargate/sgv2/jsonapi/service/processor/MeteredCommandProcessor.java
```diff
@@ -301,14 +301,24 @@ private <T extends SchemaObject> Tags getCustomTags(
...
     // --- Sort Type Tag ---
     // Determine the type of sorting used (if any), primarily for FindCommand.
-    // NOTE: This logic might need refinement or replacement when FeatureUsage is fully integrated,
-    // especially for FindAndRerankCommand.
+    // NOTE: This logic might need refinement or replacement when CommandFeatures is fully
+    // integrated, especially for FindAndRerankCommand.
     JsonApiMetricsConfig.SortType sortType = getVectorTypeTag(commandContext, command);
     Tag sortTypeTag = Tag.of(jsonApiMetricsConfig.sortType(), sortType.name());
 
+    // --- Command CommandFeature Usage Tags ---
```
amorton, line 309: nit: comment is not adding much here :)

src/main/java/io/stargate/sgv2/jsonapi/service/processor/MeteredCommandProcessor.java
```diff
@@ -301,14 +301,24 @@ private <T extends SchemaObject> Tags getCustomTags(
...
     return Tags.of(
-        commandTag, tenantTag, errorTag, errorClassTag, errorCodeTag, vectorEnabled, sortTypeTag);
+            commandTag,
+            tenantTag,
+            errorTag,
+            errorClassTag,
+            errorCodeTag,
+            vectorEnabled,
+            sortTypeTag)
+        .and(commandFeatureTags);
```
amorton, line 321: nit: you could turn this around the other way `commandFeatureTags.and(....)`

amorton (approved):
ok to merge, I did some nit picking. The bigges thing is that insertOne and insertMany need to be tracking command feature usage . That can be a fast follow

src/main/java/io/stargate/sgv2/jsonapi/metrics/CommandFeatures.java
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  @Override
+  public int hashCode() {
+    return commandFeatures.hashCode();
+  }
+
+  @Override
+  public String toString() {
+    return "CommandFeatures" + commandFeatures.toString();
```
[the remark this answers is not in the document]
tatu-at-datastax: I think our formatter does whatever is required (for better or worse).

src/main/java/io/stargate/sgv2/jsonapi/service/processor/CommandProcessor.java
```diff
@@ -64,18 +64,21 @@ public <CommandT extends Command, SchemaT extends SchemaObject> Uni<CommandResul
...
                     commandContext.schemaObject()));
 
     return Uni.createFrom()
-        .item(initialCommand)
+        .item(command)
         .onItem()
 
-        // Step 1: Expand any hybrid fields in the command (synchronous, in-place modification)
+        // Step 1: Expand any hybrid fields in the command (synchronous) and record the command
+        // features
```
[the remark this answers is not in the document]
Hazel-Datastax: I cannot use `command` in the first `invoke()` and get an error message `Variable 'command' is already defined in the scope`. I changed all the command name to `cmd`

PR #2060, opened 2025-04-30
simplify how we add tenant to the cql session

### LINKED ISSUE #2059: Simplify tenant id management for cql driver
Related to #2048 - will do this first 

We currently put the tenant ID into the session name, and the on the CQLSessionCache 

AND via the TenantAwareCqlSessionBuilder into the "startup Optons" 

```java
    protected Map<String, String> buildStartupOptions() {
      Map<String, String> existing = super.buildStartupOptions();
      return NullAllowingImmutableMap.<String, String>builder(existing.size() + 1)
          .putAll(existing)
          .put(TENANT_ID_PROPERTY_KEY, tenantId)
          .build();
    }
```

the later is that only used in the `CustomTaggingMetricIdGenerator` which is pretty much a copy of the built in `TaggingMetricIdGenerator` in the driver , down to copying code we dont use. 

CustomTaggingMetricIdGenerator gets the tenant ID from both the session name (which it ignores) and the startup options and then adds it as a tag to metrics: 

```java
public CustomTaggingMetricIdGenerator(DriverContext context) {
    sessionName = context.getSessionName();
    String prefix =
        Objects.requireNonNull(
            context
                .getConfig()
                .getDefaultProfile()
                .getString(DefaultDriverOption.METRICS_ID_GENERATOR_PREFIX, ""));
    sessionPrefix = prefix.isEmpty() ? "session." : prefix + ".session.";
    nodePrefix = prefix.isEmpty() ? "nodes." : prefix + ".nodes.";
    tenantId =
        ((TenantAwareCqlSessionBuilder.TenantAwareDriverContext) context)
            .getStartupOptions()
            .get(TenantAwareCqlSessionBuilder.TENANT_ID_PROPERTY_KEY);
  }

  @NonNull
  @Override
  public MetricId sessionMetricId(@NonNull SessionMetric metric) {
    return new DefaultMetricId(
        sessionPrefix + metric.getPath(), ImmutableMap.of("tenant", tenantId));
  }

  @NonNull
  @Override
  public MetricId nodeMetricId(@NonNull Node node, @NonNull NodeMetric metric) {
    return new DefaultMetricId(
        nodePrefix + metric.getPath(),
        ImmutableMap.of("node", node.getEndPoint().toString(), "tenant", tenantId));
  }
```

we only need to store the tenant ID in the session name, and we can probably drop the CustomTaggingMetricIdGenerator  because the TaggingMetricIdGenerator adds the session name as a session tag.

amorton:
This PR simplifies how we associate tenantID with the driver session. 

**What this PR does**:

Fixes #2059

**Checklist**
- [ ] Changes manually tested


src/main/resources/application.conf
```diff
@@ -19,7 +19,9 @@ datastax-java-driver {
   }
   advanced.metrics {
     id-generator{
-      class = io.stargate.sgv2.jsonapi.service.cqldriver.CustomTaggingMetricIdGenerator
+      # This will add 'session' tag to session and node level metrics with the session name
+      # that we set to the tenant name. (Also adds a node tag for node level metrics)
+      class = TaggingMetricIdGenerator
```
tatu-at-datastax, line 24: Which package will the class be in when left as unqualified? (I guess it has to be a built-in id generator provided by driver itself)

src/test/java/io/stargate/sgv2/jsonapi/api/v1/AbstractKeyspaceIntegrationTestBase.java
```diff
@@ -234,13 +234,16 @@ public static void checkShouldAbsentMetrics(String commandName) {
...
+    // session_cql_requests_seconds_bucket{module="sgv2-jsonapi",session="3ee89833-3f8c-4140-841f-59ae95bc48e0",le="0.001"} 0.0
+
     Optional<String> sessionLevelDriverMetricTenantId =
         metrics
             .lines()
             .filter(
                 line ->
                     line.startsWith("session_cql_requests_seconds_bucket")
-                        && line.contains("tenant"))
+                        && line.contains("session="))
```
tatu-at-datastax, line 246: Does this mean metric tag changes? Is that ok for production deployments?

tatu-at-datastax (approved):
LGTM: had couple of questions.

tatu-at-datastax:
One thing: would be good to have actual description of main changes, not just brief title (or link to issue that has explanation if there is one).

amorton:
> One thing: would be good to have actual description of main changes, not just brief title (or link to issue that has explanation if there is one).

done, the ticket had a full description. Pointed to that and noted this changes metric tags

src/test/java/io/stargate/sgv2/jsonapi/api/v1/AbstractKeyspaceIntegrationTestBase.java
amorton: yes, this is for the metrics from the driver, I could not see any panels that included those.

PR #2062, opened 2025-04-30
Fixes #2057: findAndRerank, NPE for empty object

### LINKED ISSUE #2057: `NullPointerException` when using `findAndRerank` command without `sort` clause
findAndRerank Command:
```
{
    "findAndRerank": {}
}
```
error:
```
{
    "errors": [
        {
            "message": "Server failed: root cause: (java.lang.NullPointerException) Cannot invoke \"io.stargate.sgv2.jsonapi.api.model.command.clause.sort.FindAndRerankSort.vectorSort()\" because the return value of \"io.stargate.sgv2.jsonapi.api.model.command.impl.FindAndRerankCommand.sortClause()\" is null",
            "id": "fa514317-efba-4bd3-b6ae-a165b9db075c",
            "family": "SERVER",
            "title": "Server failed",
            "scope": "EMPTY"
        }
    ]
}
```

tatu-at-datastax:
**What this PR does**:

Fixes NPE thrown for `findAndRerank` request like:

```
{
    "findAndRerank": {}
}
```

Fixes #2057

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/FindAndRerankCommand.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/FindAndRerankCommand.java
@@ -36,6 +36,9 @@ public record FindAndRerankCommand(
     @Valid @JsonProperty("sort") FindAndRerankSort sortClause,
     @Valid @Nullable Options options)
     implements ReadCommand, Filterable, Projectable, Windowable {
+  public FindAndRerankCommand {
+    sortClause = (sortClause == null) ? FindAndRerankSort.NO_ARG_SORT : sortClause;
+  }
 
   // NOTE: is not VectorSortable because it has its own sort clause.
 
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/resolver/FindAndRerankOperationBuilder.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/resolver/FindAndRerankOperationBuilder.java
@@ -400,14 +400,14 @@ class FindAndRerankOperationBuilder {
   }
 
   private boolean isLexicalSort() {
-    return command.sortClause().lexicalSort() != null;
+    return command.sortClause() != null && command.sortClause().lexicalSort() != null;
   }
 
   private boolean isVectorizeSort() {
-    return command.sortClause().vectorizeSort() != null;
+    return command.sortClause() != null && command.sortClause().vectorizeSort() != null;
   }
 
   private boolean isVectorSort() {
-    return command.sortClause().vectorSort() != null;
+    return command.sortClause() != null && command.sortClause().vectorSort() != null;
   }
 }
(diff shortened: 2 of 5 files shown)


src/main/java/io/stargate/sgv2/jsonapi/service/resolver/FindAndRerankOperationBuilder.java
```diff
@@ -400,14 +400,14 @@ private PathMatchLocator passageLocator() {
   }
 
   private boolean isLexicalSort() {
-    return command.sortClause().lexicalSort() != null;
+    return command.sortClause() != null && command.sortClause().lexicalSort() != null;
```
Hazel-Datastax, line 403: I see adding the null check here could avoid the NPE problem. I am wondering if we could fix the null sortclause problem in the deserializer (`FindAndRerankSortClauseDeserializer`) so that we don’t need to check for null every time? I am concerned about NPE in the future if someone uses it but forgets to check for null. - But I am not sure; maybe it's not easy to add in there. If that's the case, we can have this PR merged first.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/FindAndRerankCollectionIntegrationTest.java
```diff
@@ -34,56 +33,12 @@ public final void createDefaultCollection() {
...
-          .then()
-          .statusCode(200)
-          .body("$", responseIsDDLSuccess());
+      String toDelete = cleanupCollectionName;
+      cleanupCollectionName = null; // clear out just in case
+      deleteCollection(toDelete);
     }
   }
 
-  public void errorOnNotEnabled(
```
tatu-at-datastax, line 55: Not removed, just moved.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/FindAndRerankCommand.java
```diff
@@ -36,6 +36,9 @@ public record FindAndRerankCommand(
     @Valid @JsonProperty("sort") FindAndRerankSort sortClause,
     @Valid @Nullable Options options)
     implements ReadCommand, Filterable, Projectable, Windowable {
+  public FindAndRerankCommand {
+    sortClause = (sortClause == null) ? FindAndRerankSort.NO_ARG_SORT : sortClause;
```
tatu-at-datastax, line 40: Uses "compact constructor" to replace missing/null with empty value to avoid NPE later on.

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/FindAndRerankOperationBuilder.java
tatu-at-datastax: Great idea & good point on nulls possibly being problem some place else.
I had briefly thought about it, but for some reason didn't do it. But looking back I can easily handle this by additional record constructor.

So changed PR to avoid null checks.

Hazel-Datastax (approved):
LGTM!

PR #2067, opened 2025-05-02
Fix #2063: test that max-$lexical length limit is enforced

### LINKED ISSUE #2063: Add tests for max `$lexical` length that can be inserted in a Document
Length of content to store in `$lexical` is limited by setting of `cassandra.sai.max_analyzed_size_kb`, see:

https://github.com/datastax/cassandra/blob/13d57414022131535407c9166619aa4679af85bd/src/java/org/apache/cassandra/index/sai/IndexContext.java#L108

but our code does not yet check for this limit: backend DB does. But we probably should validate the limit on our side for improved dev experience.

And should in general test that limits work the way we expect them to.

tatu-at-datastax:
**What this PR does**:

Adds Integration Tests to verify that maximum length for `$lexical` is enforced.

When writing test it was found out that although there is validation:

1. Exceeding the limit only prevents indexing but DOES NOT FAIL Insert query (by default)
2. Limit checked is NOT simple text length check but calculated on tokenized terms' length: allowing typically somewhat longer content to be indexed (dup tokens removed etc) -- so test needs to use somewhat bigger content to trigger fail.

To solve (1) there is a system property we can set for tests, but also probably want to get added in production (but that may be longer process wrt compatibility constrainsts).

(2) is just a constraints for now.

Further, this PR does not yet add mapping from DB failure into nicer errors -- need a follow-up for that.

Fixes #2063

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/src/test/java/io/stargate/sgv2/jsonapi/api/v1/InsertLexicalInCollectionIntegrationTest.java
+++ b/src/test/java/io/stargate/sgv2/jsonapi/api/v1/InsertLexicalInCollectionIntegrationTest.java
@@ -28,7 +28,7 @@ import org.junit.jupiter.api.condition.DisabledIfSystemProperty;
 @TestClassOrder(ClassOrderer.OrderAnnotation.class)
 public class InsertLexicalInCollectionIntegrationTest
     extends AbstractCollectionIntegrationTestBase {
-  public static final int MAX_LEXICAL_LENGTH = 8192;
+  private static final int MAX_LEXICAL_LENGTH = 8192;
 
   protected InsertLexicalInCollectionIntegrationTest() {
     super("col_insert_lexical_");
--- a/src/test/java/io/stargate/sgv2/jsonapi/api/v1/InsertLexicalInCollectionIntegrationTest.java
+++ b/src/test/java/io/stargate/sgv2/jsonapi/api/v1/InsertLexicalInCollectionIntegrationTest.java
@@ -12,6 +12,7 @@ import static org.hamcrest.Matchers.is;
 import io.quarkus.test.common.WithTestResource;
 import io.quarkus.test.junit.QuarkusIntegrationTest;
 import io.stargate.sgv2.jsonapi.exception.ErrorCodeV1;
+import io.stargate.sgv2.jsonapi.fixtures.TestTextUtil;
 import io.stargate.sgv2.jsonapi.testresource.DseTestResource;
 import org.apache.commons.lang3.RandomStringUtils;
 import org.junit.jupiter.api.ClassOrderer;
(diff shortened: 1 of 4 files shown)


src/test/java/io/stargate/sgv2/jsonapi/api/v1/InsertLexicalInCollectionIntegrationTest.java
```diff
@@ -27,6 +28,8 @@
 @TestClassOrder(ClassOrderer.OrderAnnotation.class)
 public class InsertLexicalInCollectionIntegrationTest
     extends AbstractCollectionIntegrationTestBase {
+  public static final int MAX_LEXICAL_LENGTH = 8192;
```
tatu-at-datastax, line 31: Will initially contain here may move if we start enforcing on Shredder -- but this is not yet certain so did not want to add in non-test constants yet.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/InsertLexicalInCollectionIntegrationTest.java
```diff
@@ -200,6 +227,32 @@ public void failInsertDocWithLexicalIfNotEnabled() {
...
+                "_id": "%s",
+                "$lexical": "%s"
+              }
+              """
+              .formatted(docId, text);
+
+      givenHeadersPostJsonThenOk("{ \"insertOne\": {  \"document\": %s }}".formatted(doc))
+          .body("$", responseIsWritePartialSuccess())
+          .body("errors", hasSize(1))
+          // !!! TODO: not rely on DB error, but enforce our own limits
```
tatu-at-datastax, line 249: Will do a follow-up to convert DB exception, after getting this PR merged.

src/test/java/io/stargate/sgv2/jsonapi/testresource/StargateTestResource.java
```diff
@@ -191,7 +191,9 @@ private GenericContainer<?> baseCassandraContainer(boolean reuse) {
         "-Dcassandra.skip_wait_for_gossip_to_settle=0 -Dcassandra.load_ring_state=false -Dcassandra.initial_token=1 -Dcassandra.sai.max_string_term_size_kb=8"
             // 18-Mar-2025, tatu: to work around [https://github.com/riptano/cndb/issues/13330],
             // need to temporarily add this for HCD:
-            + " -Dcassandra.cluster_version_provider.min_stable_duration_ms=-1";
+            + " -Dcassandra.cluster_version_provider.min_stable_duration_ms=-1"
+            // 02-May-2025, tatu: [data-api#2063] force checking of max analyzed text length
+            + " -Dcassandra.sai.validate_max_term_size_at_coordinator=true";
```
tatu-at-datastax, line 196: This is needed for DB to fail query on indexing fail: not enabled by default.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/InsertLexicalInCollectionIntegrationTest.java
Yuqi-Du: make sense to me

Yuqi-Du (approved):

PR #2069, opened 2025-05-02
Fixes #2068: map INVALID_QUERY into proper LEXICAL_CONTENT_TOO_BIG

### LINKED ISSUE #2068: Convert "lexical too big" `INVALID_QUERY` into proper error (follow-up to #2063)
As tested in #2063, if max-length-for-analyzed-text failure is enabled, we will get error from DB.
We should translate this into proper Data API error, given that we may not be able to reliably handle this problem case before sending insert query.

Message we get is like:

```
Term's analyzed size for column query_lexical_value exceeds the cumulative limit for index. Max allowed size 8.000KiB.
```

for `com.datastax.oss.driver.api.core.servererrors.InvalidQueryException` / errorCode: `INVALID_QUERY`

tatu-at-datastax:
**What this PR does**:

Maps `QueryValidationException` with specific message into `LEXICAL_CONTENT_TOO_BIG` to give better message to caller, in case where `$lexical` value exceeds maximum indexable size.

Fixes #2068

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCodeV1.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCodeV1.java
@@ -166,6 +166,8 @@ public enum ErrorCodeV1 {
 
   LEXICAL_NOT_AVAILABLE_FOR_DATABASE("Lexical search is not available on this database"),
   LEXICAL_NOT_ENABLED_FOR_COLLECTION("Lexical search is not enabled for the collection"),
+  LEXICAL_CONTENT_TOO_BIG(
+      "Lexical content is too big, please use a smaller value for the $lexical field"),
 
   HYBRID_FIELD_CONFLICT(
       "The '$hybrid' field cannot be used with '$lexical', '$vector', or '$vectorize'."),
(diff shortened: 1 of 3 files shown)


src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCodeV1.java
```diff
@@ -166,6 +166,8 @@ public enum ErrorCodeV1 {
 
   LEXICAL_NOT_AVAILABLE_FOR_DATABASE("Lexical search is not available on this database"),
   LEXICAL_NOT_ENABLED_FOR_COLLECTION("Lexical search is not enabled for the collection"),
+  LEXICAL_CONTENT_TOO_BIG(
+      "Lexical content is too big, please use a smaller value for the $lexical field"),
```
tatu-at-datastax, line 170: Will need to figure out how to indicate maximum length constraints: 8kB value default does NOT refer to raw byte or character length of input but something like total length of all tokens indexed (which typically is somewhat smaller than original input byte length)

Yuqi-Du (approved):

Issue #2070, opened 2025-05-05
Refactor SchemaCache so it only has 1 level of cache


amorton:
Currently the SchemaCache caches keyspaces, and then under that tables - so there are two caffeine caches. 

This means we cannot effectly set an overall management of TTL and size etc, because there is a cache of caches / map of maps. 

Refactor so there is a single level of (keyspace, table) cache.

PR #2071, opened 2025-05-05
Fix #2048 refactor CqlSessionCache & add callbacks

### LINKED ISSUE #2048: Add Callback capability to the CQLSessionCache to support tenant metrics eviction
We want to be able to remove metrics for tenants from the MetricsRegistry when the tenant becomes inactive. Currently we do not do this, and there is a lot of stale tenants we are emitting 0 metrics for.

The CQLSessionCache alredy has something like this for removing the tenant from the Schema Cache. 

This ticket is to create a more formal Callback structure to so that the schema cache and metrics can register to know when a tenant becomes inactive.

amorton:
**What this PR does**:

Note: designed to apply after #2059 

refactor the session cache to split out responsibilities, and add unit tests for the parts. Now has the cache, credentials creation, and session creation in different classes.

Add a callback mechanism that is used when a session is removed from the cache, so we can remove metrics that are no longer needed. Also the process now used for managing the schema cache.

Fixes #2048

**Checklist**
- [ ] Changes manually tested


amorton:
Actions not running automatically, this is the most recent run https://github.com/stargate/data-api/actions/runs/14850802371

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/CqlSessionCacheFactory.java
```diff
@@ -0,0 +1,101 @@
...
+/**
+ * Factory for creating a singleton {@link CQLSessionCache} instance that is configured via CDI.
+ *
+ * <p>We use this factory so the cache itself is not a CDI bean. For one so it does not have all
+ * that extra overhead, and because there construction is a bit more complicated.
+ */
+@ApplicationScoped
+public class CqlSessionCacheFactory {
+
+  private CQLSessionCache singleton = null;
```
tatu-at-datastax, line 22: I am not sure I understand why the whole lazy construction logic here -- seems like all configuration is passed via constructor and construction of the cache could be made eager and access simpler?

Adding comments on why (assuming there are solid reasons -- some cyclic dependency) would be good. Or just make construction eager.

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/CqlSessionFactory.java
```diff
@@ -0,0 +1,157 @@
...
+
+  private final String applicationName;
+
+  private final DatabaseType databaseType;
+  private final String localDatacenter;
+  private final Collection<InetSocketAddress> contactPoints;
+  private final List<SchemaChangeListener> schemaChangeListeners;
+  private final Supplier<CqlSessionBuilder> sessionBuilderSupplier;
+
+  CqlSessionFactory(
```
tatu-at-datastax, line 38: Should perhaps add brief Javadoc mentioning this is the main non-test Constructor to use?
(since the other with `Supplier<CqlSessionBuilder>` is for use by test as per comments)

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/CqlSessionFactory.java
```diff
@@ -0,0 +1,157 @@
...
+        databaseType,
+        localDatacenter,
+        cassandraEndPoints,
+        cassandraPort,
+        schemaChangeListeners,
+        CqlSessionBuilder::new);
+  }
+
+  /**
+   * Constructor for the CqlSessionFactory, normally use the overload without the
```
tatu-at-datastax, line 56: Since this is for tests (I think?) could explicitly say so, maybe even tag with `@VisibleForTesting`

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/CqlSessionFactory.java
```diff
@@ -0,0 +1,157 @@
...
+      if (cassandraEndPoints.isEmpty()) {
+        throw new IllegalArgumentException(
+            "Database type is %s but cassandraEndPoints is empty.".formatted(databaseType));
+      }
+      contactPoints =
+          cassandraEndPoints.stream()
+              .map(host -> new InetSocketAddress(host, cassandraPort))
+              .toList();
+    } else {
+      contactPoints = List.of();
```
tatu-at-datastax, line 105: Should this actually verify caller does NOT pass non-null/non-empty `cassandraEndPoints` -- given it is ignored completely for Astra/offline cases?

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/SchemaCache.java
```diff
@@ -1,93 +1,313 @@
...
+  @Inject
+  public SchemaCache(
+      CqlSessionCacheFactory sessionCacheFactory,
+      ObjectMapper objectMapper,
+      OperationsConfig operationsConfig) {
+    this(sessionCacheFactory, objectMapper, operationsConfig, TableBasedSchemaCache::new);
+  }
+
+  /**
+   * NOTE: must not use the sessionCacheFactory in the ctor or because it will create a circular
```
tatu-at-datastax, line 61: I assume this is used by (unit) tests as well (`@Inject`) for non-unit-tests. Could then add `@VisibleForTesting` too.

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/SchemaCache.java
```diff
@@ -1,93 +1,313 @@
...
       String collectionName,
       boolean forceRefresh) {
 
-    // TODO: refactor, this has duplicate code, the only special handling the OSS has is the tenant
-    // check
+    Objects.requireNonNull(namespace, "namespace must not be null");
+
+    // Aaron 1 may 2025 - Based on existing logic I thought we should have a non null tenantId in
+    // all cases other than
+    // when using cassandra as a backend.
```
tatu-at-datastax, line 115: Does off-line case (if still supported) have required tenant id?

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/SchemaCache.java
```diff
@@ -1,93 +1,313 @@
...
+    if (LOGGER.isTraceEnabled()) {
+      LOGGER.trace("onLoad() - tenantId: {}, keyspace: {}", key.tenantId(), key.keyspace());
     }
 
-    final NamespaceCache namespaceCache =
-        schemaCache.get(new CacheKey(tenant, namespace), this::addNamespaceCache);
-    return namespaceCache.getSchemaObject(dataApiRequestInfo, collectionName, forceRefresh);
+    // MUST not use the session cache in the ctor, OK to create a query executor for each inner
+    // cache here
+    var queryExecutor = new QueryExecutor(sessionCacheFactory.create(), operationsConfig);
```
tatu-at-datastax, line 135: Now realizing "SessionCacheFactory" is slightly misleading as it only creates and shares a single(ton) instance. So it's almost as it should be called "SessionCacheProvider" or such.
Just because here it might read like we create separate cache for each executor (which we don't).

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/SchemaCache.java
```diff
@@ -1,93 +1,313 @@
...
+
+  private TableBasedSchemaCache onLoad(SchemaCache.KeyspaceCacheKey key) {
+    if (LOGGER.isTraceEnabled()) {
+      LOGGER.trace("onLoad() - tenantId: {}, keyspace: {}", key.tenantId(), key.keyspace());
     }
 
-    final NamespaceCache namespaceCache =
-        schemaCache.get(new CacheKey(tenant, namespace), this::addNamespaceCache);
-    return namespaceCache.getSchemaObject(dataApiRequestInfo, collectionName, forceRefresh);
+    // MUST not use the session cache in the ctor, OK to create a query executor for each inner
```
tatu-at-datastax, line 133: Not sure I fully understand this comment...

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/SchemaCache.java
```diff
@@ -1,93 +1,313 @@
...
+    // cache here
+    var queryExecutor = new QueryExecutor(sessionCacheFactory.create(), operationsConfig);
+    return tableCacheFactory.create(key.keyspace(), queryExecutor, objectMapper);
+  }
+
+  Optional<SchemaObject> peekSchemaObject(String tenantId, String keyspaceName, String tableName) {
```
tatu-at-datastax, line 139: Would benefit from brief Javadoc (needed for testing?)

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/SchemaCache.java
```diff
@@ -1,93 +1,313 @@
...
+   *
+   * <p>A new schema change listener should be created for each CQL {@link Session} when it is
+   * created because the listener will first listen for {@link SchemaChangeListener#onSessionReady}
+   * and get the tenantID from the session name via {@link Session#getName}.
+   *
+   * <p>If the tenantId is not set, null or blank, we log at ERROR rather than throw because the
+   * callback methods are called on driver async threads and exceptions there are unlikely to be
+   * passed back in the request response.
+   *
+   * <p>This could be non-static inner, but statis to make testing easier so we can pass in the
```
tatu-at-datastax, line 212: typo: statis -> static

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/SchemaCache.java
```diff
@@ -1,93 +1,313 @@
...
+
+    private final SchemaCache schemaCache;
+
+    private String tenantId = null;
+
+    public SchemaCacheSchemaChangeListener(SchemaCache schemaCache) {
+      this.schemaCache = Objects.requireNonNull(schemaCache, "schemaCache must not be null");
+    }
+
+    private boolean isTenantId(String context) {
```
tatu-at-datastax, line 228: Was confused by naming -- looks like should be `hasTenantId()` instead.

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/TableBasedSchemaCache.java
```diff
@@ -10,12 +10,13 @@
 import io.stargate.sgv2.jsonapi.service.schema.collections.CollectionSchemaObject;
 import io.stargate.sgv2.jsonapi.service.schema.collections.CollectionTableMatcher;
 import java.time.Duration;
+import java.util.Optional;
 
 /** Caches the vector enabled status for the namespace */
 // TODO: what is the vector status of a namespace ? vectors are per collection
```
tatu-at-datastax, line 16: Should remove these TODOs?

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/TableBasedSchemaCache.java
```diff
@@ -84,6 +86,10 @@ protected Uni<SchemaObject> getSchemaObject(
     }
   }
 
+  Optional<SchemaObject> peekSchemaObject(String tableName) {
+    return Optional.ofNullable(schemaObjectCache.getIfPresent(tableName));
+  }
+
   private Uni<SchemaObject> loadSchemaObject(RequestContext requestContext, String collectionName) {
```
tatu-at-datastax, line 93: Should change "collection" references in names to "table" since at this level they are tables of different types, not just collections.

src/test/java/io/stargate/sgv2/jsonapi/service/cqldriver/CqlSessionCacheFactoryTests.java
```diff
@@ -0,0 +1,46 @@
...
+import io.stargate.sgv2.jsonapi.service.cqldriver.executor.SchemaCache;
+import java.util.List;
+import org.junit.jupiter.api.Test;
+
+/** Tests for {@link CqlSessionCacheFactory}. */
+public class CqlSessionCacheFactoryTests {
+
+  @Test
+  public void testSingleton() {
+    // Not a lot to test, just checking it always returns the same instance.
```
tatu-at-datastax, line 20: Not sure this test provides lots of value considering all mocking going on; looks like it's mostly scaffolding.

src/test/java/io/stargate/sgv2/jsonapi/service/cqldriver/CqlSessionCacheTests.java
```diff
@@ -0,0 +1,345 @@
...
+
+    var actualSession = fixture.cache.getSession(tenantId, authToken);
+    assertThat(actualSession)
+        .as("Session from cache is instance from factory")
+        .isSameAs(fixture.expectedSession);
+
+    // wait for the session to expire
+    try {
+      Thread.sleep(ttl.toMillis() * 2);
+    } catch (InterruptedException e) {
```
tatu-at-datastax, line 157: Or could have made test method `throws Exception` and let it go through.

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/SchemaCache.java
```diff
@@ -1,93 +1,313 @@
...
+
+    private final SchemaCache schemaCache;
+
+    public SchemaCacheDeactivatedTenantConsumer(SchemaCache schemaCache) {
+      this.schemaCache = Objects.requireNonNull(schemaCache, "schemaCache must not be null");
+    }
+
+    @Override
+    public void accept(String tenantId, RemovalCause cause) {
+      // the sessions are keyed on the tenantID and the crednetials, and one session can work with
```
tatu-at-datastax, line 301: typo: crednetials -> credentials

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/SchemaCache.java
```diff
@@ -1,93 +1,313 @@
...
+      // multiple keyspaces. So we need to evict all the keyspaces for the tenantId
+      schemaCache.evictAllKeyspaces(tenantId);
```
tatu-at-datastax, line 303: Question on this: since there can be concurrent sessions by same tenant, is there any problem (besides inefficiency) in evicting all schema info -- for (all sessions of same) tenant? Is this done because there isn't an efficient way to figure out when the last session for given tenant is evicted so we just want to drop it make sure no schema info ever remains after tenant is gone?
For metrics we do need to figure out "last session" aspect tho don't we.

Conversely only evicting table/keyspace info when TTL is exhausted or cache size limit is reached would work fine without tenant-based eviction?

(I may have misunderstood something -- just thinking out aloud)

tatu-at-datastax (approved):
Looks good -- have lots of questions, nitpicky typo info, but nothing plainly wrong.

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/SchemaCache.java
amorton: A single session can use multiple keyspaces / tables.

the session is keyed on the tenantId + Credentials, and then the {keyspace} and {collection/table} are pulled from the URL path and used in the CQL we generate with fully qualified keyspace.table names.

So the thinking was there is no way to know what keyspaces the tenant touched, just that they may have touched all of them. And then on a miss we have to reload.

> Conversely only evicting table/keyspace info when TTL is exhausted or cache size limit is reached would work fine without tenant-based eviction?
(I may have misunderstood something -- just thinking out aloud)

ah, had not thought of that - I bascially refactored how it worked now. Though thinking it through further - with an SLA checked visiting, that will push things into schema cache, and that may result in non SLA checker users having schema removed.

The next PR is to have differentiated TTL in the session cache based on user agent, so how about we keep it as is now. Then either in the next PR for agent based TTL or when we refactor this into a single level we think about it then ?

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/SchemaCache.java
amorton: Will rename to SessionCacheProvider after reviewing all the notes.

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/CqlCredentials.java
```diff
@@ -71,8 +46,9 @@ public boolean isAnonymous() {
     }
 
     @Override
-    public void addToSessionBuilder(CqlSessionBuilder builder) {
-      // Do nothing
+    public CqlSessionBuilder addToSessionBuilder(CqlSessionBuilder builder) {
```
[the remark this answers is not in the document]
amorton: yeah the idea was the client code for the CqlCredentials does not know how to pass those credentials to the session builder

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/CqlCredentials.java
```diff
@@ -140,4 +113,76 @@ public static UsernamePasswordCredentials fromToken(String encodedCredentials) {
...
+     *     easier. When not null, this token is used to validate the authToken provided in the
+     *     request.
+     * @param fixedUserName the username to use if the fixedToken is set, this is from config
+     *     usually
+     * @param fixedPassword the password to use if the fixedToken is set, this is from config
+     *     usually
+     * @param databaseType the type of database, used to determine if anonymous access is allowed.
+     */
+    public CqlCredentialsFactory(
+        String fixedToken, String fixedUserName, String fixedPassword, DatabaseType databaseType) {
```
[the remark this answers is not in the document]
amorton: yeah I thought that as well, and then wondered just how pedantic could I be :) If you also think that, it must be correct !

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/CqlCredentials.java
```diff
@@ -140,4 +113,76 @@ public static UsernamePasswordCredentials fromToken(String encodedCredentials) {
       }
     }
   }
+
+  /** Factory to create the CqlCredentials based on the provided tokens. */
+  class CqlCredentialsFactory implements CQLSessionCache.CredentialsFactory {
```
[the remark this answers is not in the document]
amorton: I think it did it with the intention of making the implementations less visible, but they can be package scope. Will change

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/CQLSessionCache.java
```diff
@@ -1,283 +1,318 @@
...
-                          host, operationsConfig.databaseConfig().cassandraPort()))
-              .toList();
-      builder.addContactPoints(seeds);
+        Caffeine.newBuilder()
+            .expireAfterAccess(cacheTTL)
+            .expireAfterWrite(cacheTTL)
+            .maximumSize(cacheMaxSize)
+            .removalListener(this::onKeyRemoved)
+            .recordStats();
+    if (asyncTaskOnCaller) {
```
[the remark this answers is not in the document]
amorton: I had one and then realised the log line said most of what I wanted to say. is that enough ?

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/CQLSessionCache.java
```diff
@@ -1,283 +1,318 @@
...
+      // sanity check
+      throw new IllegalStateException("credentialsFactory returned null");
     }
 
-    return switch (operationsConfig.databaseConfig().type()) {
+    return switch (databaseType) {
       case CASSANDRA, OFFLINE_WRITER ->
           new SessionCacheKey(
               tenantId == null || tenantId.isBlank() ? DEFAULT_TENANT : tenantId, credentials);
       case ASTRA -> new SessionCacheKey(tenantId, credentials);
```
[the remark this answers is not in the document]
amorton: The SessionCacheKey normalises null into "" . 

Our handling for tenantId is somewhat flakey in terms of null, empty string, optional string, or string with something like "default_tenant" in it. 

What if afer this PR we normalise the tenantId into something more reliable ? Removing the "offline mode" will make that easier I think

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/CQLSessionCache.java
```diff
@@ -1,283 +1,318 @@
...
    *
-   * @param tenantId optional tenantId, if null converted to empty string
+   * @param tenantId optional tenantId, if null or blank converted to empty string
    * @param credentials Required, credentials for the session
    */
   public record SessionCacheKey(String tenantId, CqlCredentials credentials) {
 
     public SessionCacheKey {
-      if (tenantId == null) {
+      if (tenantId == null || tenantId.isBlank()) {
```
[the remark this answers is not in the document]
amorton: per comments above (left open) I'm really not sure :| was trying to keep the exisitng logic, we need to lock this down.

amorton:
Added https://github.com/stargate/data-api/issues/2082 for refactoring the tenantId

amorton:
Added this for removing the offline mode https://github.com/stargate/data-api/issues/2083

amorton:
Added this for always using the config endpoint https://github.com/stargate/data-api/issues/2084

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/TableBasedSchemaCache.java
amorton: will leave them until we refactor the schema caching so we think about them when doing it see https://github.com/stargate/data-api/issues/2070

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/TableBasedSchemaCache.java
amorton: I was trying  to leave this class as much as possible , will cleanup in https://github.com/stargate/data-api/issues/2070

tatu-at-datastax: makes sense.

PR #2076, opened 2025-05-07
Fix #2073 and #2081: Make sure the error from `HybridFieldExpander` can be captured, and the logging works when executions fail

### LINKED ISSUE #2073: Make sure the error from `HybridFieldExpander.expandHybridField()` is captured
### Analysis of `HybridFieldExpander.expandHybridField(commandContext, command)` call:

- This code is called synchronously.

- If `expandHybridField` throws an exception, the execution of `commandProcessor.processCommand` stops right there and exception propagates out of the `commandProcessor.processCommand` method call synchronously.

- The call stack returns to `MeteredCommandProcessor.processCommand`  and it will not continue `.onItem().invoke().onFailure().invoke()` chain - because it didn’t get a `Uni` :

```
// THIS LINE NOW THROWS THE SYNCHRONOUS EXCEPTION from HybridFieldExpander
  return commandProcessor
      .processCommand(commandContext, command) // <--- Exception happens HERE

      // THESE LINES ARE NEVER REACHED because the previous call threw synchronously
      .onItem()
      .invoke(result -> { /* ... */ })
      .onFailure()
      .invoke(throwable -> {  /* ... */ });
```

- The synchronous exception continues propagating up the call stack, and will be caught by `GenericExceptionMapper` and handled by `ThrowableCommandResultSupplier`  and `ThrowableToErrorMapper`.

### Problem:

- No functional problem, we just cannot capture the error from `expandHybridField` in metrics and logs

### Possible solution

- Update it to async call

Hazel-Datastax:
**What this PR does**:
1. Make sure the error from `HybridFieldExpander` can be captured by log and metrics
2. Fix the logging error in `logger.error(buildCommandLog(commandContext, command, null), throwable);` in `MeteredCommandProcessor.processCommand`
3. Reorganize the code in `CommandProcessor.processCommand`

Fixes #2073 #2081 

**Checklist**
- [x] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/config/CommandLevelLoggingConfig.java
```diff
@@ -30,7 +30,7 @@ public interface CommandLevelLoggingConfig {
   /**
    * @return If request info logging is enabled.
    */
-  @WithDefault("false")
+  @WithDefault("true")
```
Hazel-Datastax, line 33: We enable this in the [deployment environment](https://github.com/riptano/serverless-platform-charts/blob/master/stargate-apis/chart/charts/jsonapi/templates/deployment.yaml#L170). 

I think it's also worth enabling here, otherwise, it may hide some problems - f.ex. see https://github.com/stargate/data-api/issues/2073

src/main/java/io/stargate/sgv2/jsonapi/service/processor/CommandProcessor.java
```diff
@@ -24,13 +24,9 @@
...
- * Processes valid document {@link Command} to read, write, schema change, etc. This is a single
- * entry to run a Command without worrying how to.
- *
- * <p>Called from the API layer which deals with public JSON data formats, the command layer
- * translates from the JSON models to internal and back (shredding and de-shredding).
- *
- * <p>May provide a thread or resource boundary from calling API layer.
+ * Processes a {@link Command} by taking it through a series of transformations: expansion,
+ * vectorization, resolution to an {@link Operation}, and finally execution. It also handles error
+ * recovery and result post-processing.
```
Hazel-Datastax, line 29: I don't really understand the Java doc here. Maybe it's outdated, LMK if we don't want to change

src/main/java/io/stargate/sgv2/jsonapi/service/processor/CommandProcessor.java
```diff
@@ -49,109 +45,147 @@ public CommandProcessor(
...
-                        Operation<SchemaT> operation =
-                            resolver.resolveCommand(commandContext, vectorizedCommand);
-                        return Uni.createFrom().item(operation);
-                      });
+
+        // Step 1: Expand any hybrid fields in the command (synchronous, in-place modification)
+        .invoke(
+            commandToExpand -> {
+              HybridFieldExpander.expandHybridField(commandContext, commandToExpand);
             })
```
Hazel-Datastax, line 79: This is the only critical change in this file - move `HybridFieldExpander.expandHybridField` from the outside to the reactive chain. Others are just extracting the code in the chain into methods to make the chain nicer to read.

src/main/java/io/stargate/sgv2/jsonapi/service/processor/MeteredCommandProcessor.java
```diff
@@ -100,9 +100,21 @@ public <CommandT extends Command, SchemaT extends SchemaObject> Uni<CommandResul
...
+    //    within it before its own reactive chain fully forms and handles errors) throws a
+    //    synchronous exception, this MeteredCommandProcessor can still catch it in its
+    //    `.onFailure()` block for consistent logging (no metrics currently). This acts as a safety
+    //    net.
+    // 2. Lazy Execution (Benefit of `deferred`):
+    //    The `commandProcessor.processCommand` method, which kicks off potentially significant
+    //    work and returns a Uni, will only be invoked when this resulting Uni is actually
+    //    subscribed to.
+    return Uni.createFrom()
+        .deferred(() -> commandProcessor.processCommand(commandContext, command))
```
Hazel-Datastax, line 117: The issue of `HybridFieldExpander.expandHybridField` has already been handled in the `CommandProcessor.processCommand.` This change is just in case a similar problem happens in the future.

Why not use `Uni.createFrom().item(Supplier<T> supplier)`?
Yes, we can. `commandProcessor.processCommand(...)` returns `Uni<CommandResult>`, then `T` would be `Uni<CommandResult>`. So we'd get a `Uni<Uni<CommandResult>>`, which is a "nested Uni". We need to use `flatMap` to handle that:
```
Uni.createFrom()
        .item(() -> commandProcessor.processCommand(commandContext, command))
        .flatMap(innerUni -> innerUni)
```
But I don't think it's better

src/main/java/io/stargate/sgv2/jsonapi/service/processor/CommandProcessor.java
tatu-at-datastax: Change looks like an improvement to me; old description does not make sense, agreed.

tatu-at-datastax (commented):
Looks good to me -- I like refactorings. But would like someone else (Aaron?) to have a look as well just to make sure.

PR #2092, opened 2025-05-16
Remove metrics for tenants from the MeterRegistry when the tenant is evicted from the CQLSessionCache

### LINKED ISSUE #2096: Remove metrics for tenants from the `MeterRegistry` when the tenant is evicted from the `CQLSessionCache`
This ticket is directly related to https://github.com/stargate/data-api/issues/2048. 2048 supported a formal Callback structure when the tenant is evicted from the CQLSessionCache.

This ticket utilizes that structure and evicts tenants' related metrics from `MeterRegistry` when they become inactive

Hazel-Datastax:
**What this PR does**:
Remove metrics for tenants from the MeterRegistry when the tenant is evicted from the CQLSessionCache

Fixes #2096 

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/metrics/MetricsTenantDeactivationConsumer.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/metrics/MetricsTenantDeactivationConsumer.java
@@ -1,17 +1,14 @@
 package io.stargate.sgv2.jsonapi.metrics;
 
+import static io.stargate.sgv2.jsonapi.metrics.MetricsConstants.MetricTags.SESSION_TAG;
+import static io.stargate.sgv2.jsonapi.metrics.MetricsConstants.MetricTags.TENANT_TAG;
+
 import com.github.benmanes.caffeine.cache.RemovalCause;
 import io.micrometer.core.instrument.Meter;
 import io.micrometer.core.instrument.MeterRegistry;
 import io.stargate.sgv2.jsonapi.service.cqldriver.CQLSessionCache;
 import io.stargate.sgv2.jsonapi.service.processor.MeteredCommandProcessor;
-import jakarta.enterprise.context.ApplicationScoped;
-import jakarta.inject.Inject;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.List;
-import java.util.Map;
-import java.util.concurrent.ConcurrentHashMap;
+import java.util.*;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -30,49 +27,14 @@ import org.slf4j.LoggerFactory;
  * crucially, attempts to remove them from the {@code MeterRegistry}, thus stopping the reporting of
  * stale metrics.
  */
-@ApplicationScoped
 public class MetricsTenantDeactivationConsumer
     implements CQLSessionCache.DeactivatedTenantConsumer {
   private static final Logger LOGGER =
       LoggerFactory.getLogger(MetricsTenantDeactivationConsumer.class);
   private final MeterRegistry meterRegistry;
 
-  private final Map<String, List<Meter.Id>> tenantMetrics = new ConcurrentHashMap<>();
-
-  @Inject
   public MetricsTenantDeactivationConsumer(MeterRegistry meterRegistry) {
     this.meterRegistry = meterRegistry;
-  }
-
-  /**
-   * Tracks a specific {@link Meter.Id} for a given tenant. This method should be called by
-   * components that create tenant-specific metrics that need to be cleaned up when the tenant
-   * becomes inactive.
-   *
-   * @param tenantId The identifier of the tenant.
-   * @param meterId The {@link Meter.Id} of the metric to track.
-   * @throws IllegalArgumentException if {@code tenantId} or {@code meterId} is null.
-   */
-  public void trackMeterId(String tenantId, Meter.Id meterId) {
-    if (tenantId == null || meterId == null) {
-      LOGGER.error(
-          "Attempted to track meter with null tenantId or meterId. TenantId: {}, MeterId: {}",
-          tenantId,
-          meterId);
-      // TODO: should we throw an exception here?
-      throw new IllegalArgumentException("Tenant ID and Meter ID must not be null");
-    }
-    tenantMetrics
-        .computeIfAbsent(tenantId, k -> Collections.synchronizedList(new ArrayList<>()))
-        .add(meterId);
-  }
-
-  /**
-   * Returns an unmodifiable view of the currently tracked tenant metrics. Intended primarily for
-   * testing purposes.
-   */
-  public Map<String, List<Meter.Id>> getTenantMetrics() {
-    return Collections.unmodifiableMap(tenantMetrics);
   }
 
   /**
@@ -85,10 +47,17 @@ public class MetricsTenantDeactivationConsumer
    */
   @Override
   public void accept(String tenantId, RemovalCause cause) {
-    List<Meter.Id> meterIdsForTenant = tenantMetrics.remove(tenantId);
-    if (meterIdsForTenant != null && !meterIdsForTenant.isEmpty()) {
-      for (Meter.Id meterId : meterIdsForTenant) {
-        meterRegistry.remove(meterId);
+    if (tenantId == null) {
+      LOGGER.info("Received null tenantId for deactivation");
+      return;
+    }
+
+    // iterate all the metrics in the meterRegistry
+    for (Meter meter : meterRegistry.getMeters()) {
+      String tenantTagValue = meter.getId().getTag(TENANT_TAG);
+      String sessionTagValue = meter.getId().getTag(SESSION_TAG);
+      if (Objects.equals(tenantTagValue, tenantId) || Objects.equals(sessionTagValue, tenantId)) {
+        meterRegistry.remove(meter.getId());
       }
     }
   }
--- a/src/main/java/io/stargate/sgv2/jsonapi/metrics/MetricsTenantDeactivationConsumer.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/metrics/MetricsTenantDeactivationConsumer.java
@@ -44,12 +44,11 @@ public class MetricsTenantDeactivationConsumer
     }
 
     for (Meter meter : meterRegistry.getMeters()) {
-      String tenantTagValue = meter.getId().getTag(TENANT_TAG);
-      String sessionTagValue = meter.getId().getTag(SESSION_TAG);
-      if (Objects.equals(tenantTagValue, tenantId) || Objects.equals(sessionTagValue, tenantId)) {
-        Meter removedMeter = meterRegistry.remove(meter.getId());
-        if (removedMeter == null) {
-          LOGGER.warn(
+      // Check TENANT_TAG first, if not found, check SESSION_TAG
+      if (Objects.equals(meter.getId().getTag(TENANT_TAG), tenantId)
+          || Objects.equals(meter.getId().getTag(SESSION_TAG), tenantId)) {
+        if (meterRegistry.remove(meter.getId()) == null) {
+          LOGGER.debug(
               "Attempted to remove metric with ID {} for tenant {} but it was not found in the registry during the removal phase.",
               meter.getId(),
               tenantId);
--- a/src/main/java/io/stargate/sgv2/jsonapi/metrics/MetricsTenantDeactivationConsumer.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/metrics/MetricsTenantDeactivationConsumer.java
@@ -4,6 +4,8 @@ import com.github.benmanes.caffeine.cache.RemovalCause;
 import io.micrometer.core.instrument.Meter;
 import io.micrometer.core.instrument.MeterRegistry;
 import io.stargate.sgv2.jsonapi.service.cqldriver.CQLSessionCache;
+import io.vertx.core.impl.logging.Logger;
+import io.vertx.core.impl.logging.LoggerFactory;
 import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.inject.Inject;
 import java.util.ArrayList;
@@ -15,6 +17,8 @@ import java.util.concurrent.ConcurrentHashMap;
 @ApplicationScoped
 public class MetricsTenantDeactivationConsumer
     implements CQLSessionCache.DeactivatedTenantConsumer {
+  private static final Logger LOGGER =
+      LoggerFactory.getLogger(MetricsTenantDeactivationConsumer.class);
   private final MeterRegistry meterRegistry;
   private final Map<String, List<Meter.Id>> tenantMetrics = new ConcurrentHashMap<>();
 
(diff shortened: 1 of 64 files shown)


src/main/java/io/stargate/sgv2/jsonapi/metrics/MetricsTenantDeactivationConsumer.java
```diff
@@ -0,0 +1,56 @@
...
+import jakarta.enterprise.context.ApplicationScoped;
+import jakarta.inject.Inject;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+@ApplicationScoped
+public class MetricsTenantDeactivationConsumer
```
Hazel-Datastax, line 18: I know we don't want to use injection but we need this consumer to be a single instance globally. Another consumer `SchemaCacheDeactivatedTenantConsumer` is the inner class of `SchemaCache`. LMK if we want it to be an inner class of something.

src/main/java/io/stargate/sgv2/jsonapi/metrics/MetricsTenantDeactivationConsumer.java
```diff
@@ -0,0 +1,56 @@
+package io.stargate.sgv2.jsonapi.metrics;
+
+import com.github.benmanes.caffeine.cache.RemovalCause;
+import io.micrometer.core.instrument.Meter;
+import io.micrometer.core.instrument.MeterRegistry;
+import io.stargate.sgv2.jsonapi.service.cqldriver.CQLSessionCache;
+import io.vertx.core.impl.logging.Logger;
```
tatu-at-datastax, line 7: I think we should be using

```
import org.slf4j.Logger;
```

src/main/java/io/stargate/sgv2/jsonapi/metrics/MetricsTenantDeactivationConsumer.java
```diff
@@ -0,0 +1,56 @@
...
+    this.meterRegistry = meterRegistry;
+  }
+
+  public void trackMeterId(String tenantId, Meter.Id meterId) {
+    if (tenantId == null || meterId == null) {
+      LOGGER.error(
+          "Attempted to track meter with null tenantId or meterId. TenantId: {}, MeterId: {}");
+      // TODO: should we throw an exception here?
+      throw new IllegalArgumentException("Tenant ID and Meter ID must not be null");
+    }
```
Hazel-Datastax, line 36: Not sure if there is any concern case that may have tenantId or meterId as null, and thus not sure if it's good to throw an exception in this case

src/main/java/io/stargate/sgv2/jsonapi/metrics/MetricsTenantDeactivationConsumer.java
```diff
@@ -0,0 +1,95 @@
...
+   */
+  public void trackMeterId(String tenantId, Meter.Id meterId) {
+    if (tenantId == null || meterId == null) {
+      LOGGER.error(
+          "Attempted to track meter with null tenantId or meterId. TenantId: {}, MeterId: {}",
+          tenantId,
+          meterId);
+      // TODO: should we throw an exception here?
+      throw new IllegalArgumentException("Tenant ID and Meter ID must not be null");
+    }
```
Hazel-Datastax, line 64: Not sure if there is any concern case that may have tenantId or meterId as null, and thus not sure if it's good to throw an exception in this case

src/main/java/io/stargate/sgv2/jsonapi/metrics/MetricsTenantDeactivationConsumer.java
tatu-at-datastax: I think use of injection is ok for this case, as it needs to be passed to `CommandContext`, and its injected through Resources that are DI-managed anyway.

src/main/java/io/stargate/sgv2/jsonapi/metrics/MetricsTenantDeactivationConsumer.java
tatu-at-datastax: I assume meterId can never be null so at least that makes sense. But I think we also pass something like "unknown" as tenant-id if one not otherwise given. So null checking both makes sense to me.

src/main/java/io/stargate/sgv2/jsonapi/metrics/MetricsTenantDeactivationConsumer.java
```diff
@@ -0,0 +1,95 @@
...
+    tenantMetrics
+        .computeIfAbsent(tenantId, k -> Collections.synchronizedList(new ArrayList<>()))
+        .add(meterId);
+  }
+
+  /**
+   * Returns an unmodifiable view of the currently tracked tenant metrics. Intended primarily for
+   * testing purposes.
+   */
+  public Map<String, List<Meter.Id>> getTenantMetrics() {
```
tatu-at-datastax, line 74: If used only for testing, maybe add `@VisibleForTesting`?

Depending on how tests check for things, could also instead return a `Stream` of entries or something (but one that does allow modifying Map contents).

src/main/java/io/stargate/sgv2/jsonapi/metrics/MetricsTenantDeactivationConsumer.java
```diff
@@ -0,0 +1,95 @@
...
+   * tracked metrics for the specified tenant from the internal tracking map and attempts to remove
+   * them from the {@link MeterRegistry}.
+   *
+   * @param tenantId The ID of the tenant whose session was deactivated.
+   * @param cause The reason for the removal from the cache.
+   */
+  @Override
+  public void accept(String tenantId, RemovalCause cause) {
+    List<Meter.Id> meterIdsForTenant = tenantMetrics.remove(tenantId);
+    if (meterIdsForTenant != null && !meterIdsForTenant.isEmpty()) {
```
tatu-at-datastax, line 89: Second check is unnecessary; you can just iterate over empty `List` (also: I may be wrong but it doesn't seem like there should ever be empty Lists if the only way they are removed as by this mechanism)

src/main/java/io/stargate/sgv2/jsonapi/metrics/MetricsTenantDeactivationConsumer.java
```diff
@@ -0,0 +1,95 @@
...
+      LOGGER.error(
+          "Attempted to track meter with null tenantId or meterId. TenantId: {}, MeterId: {}",
+          tenantId,
+          meterId);
+      // TODO: should we throw an exception here?
+      throw new IllegalArgumentException("Tenant ID and Meter ID must not be null");
+    }
+    tenantMetrics
+        .computeIfAbsent(tenantId, k -> Collections.synchronizedList(new ArrayList<>()))
+        .add(meterId);
```
tatu-at-datastax, line 67: Do we know that this call is only made once per metric and not -- like it seems looking at `MeteredCommandProcessor` -- every time we update metric?
If latter, this `List` will explode pretty quickly because `List.add()` adds duplicates just fine (unlike `Set.add()`).

And if so, would need to change to either only add to `List` after checking Meter.Id not yet contained or, probably better, by using a `Set` (but one that is synchronized/concurrent). 
So like `Collections.synchronizedSet(new HashSet<>());` or something.

src/main/java/io/stargate/sgv2/jsonapi/service/processor/MeteredCommandProcessor.java
```diff
@@ -121,7 +121,12 @@ public <CommandT extends Command, SchemaT extends SchemaObject> Uni<CommandResul
             result -> {
               // --- Metrics Recording ---
               Tags tags = getCustomTags(commandContext, command, result);
-              sample.stop(meterRegistry.timer(jsonApiMetricsConfig.metricsName(), tags));
+              Timer timer = meterRegistry.timer(jsonApiMetricsConfig.metricsName(), tags);
+              sample.stop(timer);
+              String tenantId = commandContext.requestContext().getTenantId().orElse(UNKNOWN_VALUE);
+              commandContext
+                  .metricsTenantDeactivationConsumer()
+                  .trackMeterId(tenantId, timer.getId());
```
tatu-at-datastax, line 129: Ok this is one place where metrics entries are produced (via `Timer`). But there are others too, like `CommandResolver` (line 223):

```
    meterRegistry.counter(jsonApiMetricsConfig.indexUsageCounterMetrics(), tags).increment();
```

and `TenantRequestMetricsFilter` (line 113)

```
      meterRegistry.counter(config.metricName(), tags).increment();
```

and likely others. Shouldn't they also call the same method?

tatu-at-datastax (commented):
Makes sense, but I have 2 main concerns:

1. Possible duplicate Meter.Ids in per-tenant `List` (need to use `Set`?)
2. Metric ids only added from one place, but I think metrics are created from other locations too

For (2), I wonder if `MeterRegistry`'s accessors (like `forEachMeter()`) could be used to find all meters for given tenant? That would mean iterating over all meters every time a tenant is evicted, but if tenant-id can be matched with `Meter` (tags are available via `Meter.Id`) it would be more reliably than relying on meters getting added.

Another thing I am not sure about is whether there is any problem throwing out metrics on eviction: that is, is that possible too aggressive? One possibility would be to only start throwing away metrics once total count of metrics exceeds some threshold.
Although I guess that could be problematic unless we did keep track of tenants that are evicted earlier, to have FIFO removal, adding another complexity.

Hazel-Datastax:
@tatu-at-datastax I changed the eviction logic, the two main concerns should be addressed. For another one:

> Another thing I am not sure about is whether there is any problem throwing out metrics on eviction: that is, is that possible too aggressive? One possibility would be to only start throwing away metrics once total count of metrics exceeds some threshold.
Although I guess that could be problematic unless we did keep track of tenants that are evicted earlier, to have FIFO removal, adding another complexity.

Good question, I am not sure either. My thought is, maybe we can try first and see how it goes.

src/main/java/io/stargate/sgv2/jsonapi/metrics/MetricsTenantDeactivationConsumer.java
```diff
@@ -0,0 +1,61 @@
...
+ */
+public class MetricsTenantDeactivationConsumer
+    implements CQLSessionCache.DeactivatedTenantConsumer {
+  private static final Logger LOGGER =
+      LoggerFactory.getLogger(MetricsTenantDeactivationConsumer.class);
+  private final MeterRegistry meterRegistry;
+
+  public MetricsTenantDeactivationConsumer(MeterRegistry meterRegistry) {
+    Objects.requireNonNull(meterRegistry, "MeterRegistry cannot be null.");
+    this.meterRegistry = meterRegistry;
```
tatu-at-datastax, line 27: suggestion: can combine 2 lines into:

   this.meterRegistry = Objects.requireNonNull(meterRegistry, "MeterRegistry cannot be null");

src/test/java/io/stargate/sgv2/jsonapi/metrics/MetricsTenantDeactivationConsumerTests.java
```diff
@@ -0,0 +1,147 @@
...
+  @Test
+  void removeFromEmptyMeterRegistry() {
+    assertThat(meterRegistry.getMeters()).isEmpty();
+
+    assertThatCode(() -> consumer.accept("any-tenant", RemovalCause.EXPIRED))
+        .doesNotThrowAnyException();
+  }
+
+  @Test
+  void testTenantMetricRemovalLifecycle() {
```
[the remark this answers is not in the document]
tatu-at-datastax: It's still only 150 lines in total -- shouldn't be problematic as-is. Can refactor if we feel otherwise in future.

tatu-at-datastax (approved):
LGTM -- way simpler now, I like it.

src/main/java/io/stargate/sgv2/jsonapi/metrics/MetricsTenantDeactivationConsumer.java
```diff
@@ -0,0 +1,60 @@
...
+   */
+  @Override
+  public void accept(String tenantId, RemovalCause cause) {
+    if (tenantId == null) {
+      LOGGER.warn("Received null tenantId for deactivation");
+      return;
+    }
+
+    for (Meter meter : meterRegistry.getMeters()) {
+      String tenantTagValue = meter.getId().getTag(TENANT_TAG);
```
amorton, line 47: nit: get the tenant tag first, check it, and then only get the session one if needed. 

we know this is going to be a look of a few 1ks of metrics, which is fine, but we can also reduce the work we do in the loop.

src/main/java/io/stargate/sgv2/jsonapi/metrics/MetricsTenantDeactivationConsumer.java
```diff
@@ -0,0 +1,60 @@
...
+      String sessionTagValue = meter.getId().getTag(SESSION_TAG);
+      if (Objects.equals(tenantTagValue, tenantId) || Objects.equals(sessionTagValue, tenantId)) {
+        Meter removedMeter = meterRegistry.remove(meter.getId());
+        if (removedMeter == null) {
```
amorton, line 51: we can probably log this as debug, rather than warn, and check the logging level before calling.

src/main/java/io/stargate/sgv2/jsonapi/metrics/MetricsTenantDeactivationConsumer.java
```diff
@@ -0,0 +1,60 @@
...
+    if (tenantId == null) {
+      LOGGER.warn("Received null tenantId for deactivation");
+      return;
+    }
+
+    for (Meter meter : meterRegistry.getMeters()) {
+      String tenantTagValue = meter.getId().getTag(TENANT_TAG);
+      String sessionTagValue = meter.getId().getTag(SESSION_TAG);
+      if (Objects.equals(tenantTagValue, tenantId) || Objects.equals(sessionTagValue, tenantId)) {
+        Meter removedMeter = meterRegistry.remove(meter.getId());
```
amorton, line 50: nit: removedMeter is unused, you can collapse this into the if

src/test/java/io/stargate/sgv2/jsonapi/metrics/MetricsTenantDeactivationConsumerTests.java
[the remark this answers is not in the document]
amorton: if you want to make it look nicer:

* this is a little confusing, we have to metrics with the same name and tenant tag, but different type ? The tests would be easier to understand if you had metric names that were unique. Use variables for these names, you have them for the tenants
* change the code to use .meter() at the end of the find rathter than .counter()  etc. 
* with the above, you can now make a function called assertMetricExists() that takes the names + tags . This will make things like all the code repition in case 1 easier to read
* make a function for creating the metrics, this code is repeated multiple times. The idea here is that in the test function it is easier to read createTimer() than the raw metrics calls you have. 

So your function may look like 

createTimer(...)
createCounter(...)
createSummary(...)
# call consumer 
assertMetricExists(...)
assertMetricNotExists(...)

amorton (approved):
+1

PR #2100, opened 2025-05-22
Add `authTokenPassThrough` property for Embedding Providers

### LINKED ISSUE #2099: Pass the API Authentication token through to Embedding Providers if no other auth available
We do something like this already for the RerankingProvider, so we can look as `NvidiaRerankingProvider` for prior art.

We can tweak it a little for the embedding.

The embedding providers have config like such in `embedding-providers-config.yaml` :  

```
        nvidia:
          # see https://docs.api.nvidia.com/nim/reference/nvidia-embedding-2b-infer
          display-name: Nvidia
          enabled: true
          url: https://ai.api.nvidia.com/v1/retrieval/nvidia/embeddings
          supported-authentications:
            NONE:
              enabled: true
          properties:
            max-batch-size: 8
          models:
            - name: NV-Embed-QA
              vector-dimension: 1024
              properties:
                max-tokens: 512
```

We want some sort of positive config setting that needs to be set, to avoid sending any auth info without explicit config. 

So, lets add to the properties for the NONE auth config (similar to how there are settings for the `HEADER`) mode. Add a property called something like `auth-token-passthrough` that if true we will do the pass through as the `Authorization` header.

Hazel-Datastax:
**What this PR does**:
Add `authTokenPassThrough` property for Embedding Providers

Fixes #2099

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/embedding/configuration/EmbeddingProvidersConfig.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/embedding/configuration/EmbeddingProvidersConfig.java
@@ -56,6 +56,7 @@ public interface EmbeddingProvidersConfig {
       @JsonProperty
       boolean enabled();
 
+      /** a boolean to flag if the Astra token should be passed through to the provider. */
       @JsonProperty
       @WithDefault("false")
       boolean authTokenPassThrough();
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/embedding/configuration/EmbeddingProvidersConfig.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/embedding/configuration/EmbeddingProvidersConfig.java
@@ -57,6 +57,10 @@ public interface EmbeddingProvidersConfig {
       boolean enabled();
 
       @JsonProperty
+      @WithDefault("false")
+      boolean authTokenPassThrough();
+
+      @JsonProperty
       List<TokenConfig> tokens();
     }
 
--- a/src/main/proto/embedding_gateway.proto
+++ b/src/main/proto/embedding_gateway.proto
@@ -107,8 +107,8 @@ message GetSupportedProvidersResponse {
 
     message AuthenticationConfig{
       bool enabled = 1;
-      bool authTokenPassThrough = 2;
-      repeated TokenConfig tokens = 3;
+      repeated TokenConfig tokens = 2;
+      bool authTokenPassThrough = 3;
 
       message TokenConfig{
         string accepted = 1;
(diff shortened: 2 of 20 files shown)


src/main/proto/embedding_gateway.proto
```diff
@@ -107,7 +107,8 @@ message GetSupportedProvidersResponse {
 
     message AuthenticationConfig{
       bool enabled = 1;
-      repeated TokenConfig tokens = 2;
+      bool authTokenPassThrough = 2;
+      repeated TokenConfig tokens = 3;
```
Yuqi-Du, line 111: I did something similar in a previous PR.
Tatu reminded me that it's better not to change the field order in the proto file, as it could cause backward compatibility issues.

ChatGPT:
` Safe changes:
Adding a new field with a new unique number.

Renaming fields (if all consumers regenerate code).

Reordering fields only if field numbers stay the same (but not generally recommended for public APIs).`

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/configuration/EmbeddingProvidersConfig.java
```diff
@@ -56,6 +56,10 @@ interface AuthenticationConfig {
       @JsonProperty
       boolean enabled();
 
+      @JsonProperty
```
tatu-at-datastax, line 59: Javadoc?

src/main/proto/embedding_gateway.proto
tatu-at-datastax: Yes, NEVER change existing id of field, ALWAYS add new ones with new id.

So let's chnage this to keep `tokens = 2` and use `3` for `authTokenPassThrough`

src/main/resources/embedding-providers-config.yaml
```diff
@@ -306,6 +306,7 @@ stargate:
           supported-authentications:
             NONE:
               enabled: true
+              auth-token-pass-through: true
```
tatu-at-datastax, line 309: I am confused here: default in code is `false`, but here `true`. Why?

Hazel-Datastax: We may need to send Astra token to the provider if the provider supports `NONE` authentication type. By default, we will not send and most of the providers do not support `NONE` - currently only Nvidia supports `NONE` and we want to send it to the GPU plane. Hope it makes sense.

tatu-at-datastax: Ok. I think may have misread this. Thanks!

tatu-at-datastax (approved):
LGTM

Hazel-Datastax:
Close, the support is in another PR #2104

PR #2104, opened 2025-05-28
Pass the API Authentication token through to Embedding Providers

### LINKED ISSUE #2099: Pass the API Authentication token through to Embedding Providers if no other auth available

Hazel-Datastax:
**What this PR does**:
Pass the API Authentication token through to Embedding Providers

Fixes #2099 

**Checklist**
- [x] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/api/request/EmbeddingCredentialsSupplier.java
```diff
@@ -0,0 +1,78 @@
...
+   *
+   * @param requestContext The current request context containing HTTP headers.
+   * @param providerConfig The configuration for the embedding provider.
+   * @return An instance of {@link EmbeddingCredentials} with the appropriate credentials.
+   */
+  public EmbeddingCredentials create(
+      RequestContext requestContext,
+      EmbeddingProvidersConfig.EmbeddingProviderConfig providerConfig) {
+
+    var authToken = requestContext.getHttpHeaders().getHeader(this.authTokenHeaderName);
```
amorton, line 49: not: dont need to use this. here

src/main/java/io/stargate/sgv2/jsonapi/api/request/EmbeddingCredentialsSupplier.java
```diff
@@ -0,0 +1,78 @@
...
+    var embeddingApi = requestContext.getHttpHeaders().getHeader(this.embeddingApiKeyHeaderName);
+    var accessId = requestContext.getHttpHeaders().getHeader(this.embeddingAccessIdHeaderName);
+    var secretId = requestContext.getHttpHeaders().getHeader(this.embeddingsecretIdHeaderName);
+
+    // If these three conditions are met, we use the auth token as the embeddingApiKey:
+    // 1. user didn't provide x-embedding-api-key
+    // 2. Collection supports NONE auth (in the yaml config and authConfigFromCollection is null)
+    // 3. Provider has authTokenPassThroughForNoneAuth set to true
+    if (embeddingApi == null
```
amorton, line 58: this is hard to understand, can we split this into creating variables for the three steps above pls.

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/DataVectorizerService.java
```diff
@@ -86,7 +86,14 @@ public <T extends SchemaObject> DataVectorizer constructDataVectorizer(
     return new DataVectorizer(
         embeddingProvider,
         objectMapper.getNodeFactory(),
-        commandContext.requestContext().getEmbeddingCredentials(),
+        commandContext
```
amorton, line 89: if we are here and commandContext.embeddingProvider() is null that feels like an IllegalStateException. i.e. we should not be vectorizing if this is null. 

can we check before this statement ?

also, we are getting the embeddingProvider above here, but this is looking at the command context

amorton (approved):
some nit picks for cody standards, otherwise good to go :)

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/DataVectorizerService.java
Hazel-Datastax: > if we are here and commandContext.embeddingProvider() is null that feels like an IllegalStateException. i.e. we should not be vectorizing if this is null.

I know, but it needs more changes to achieve that. `commandContext.embeddingProvider()` could be null if the command is from `GeneralResource` or `KeyspaceResource` or `CollectionResource` but the collection doesn't have the vectorize config. 

"i.e. we should not be vectorizing if this is null." Agree, but the current code design is we will "vectorize" all the commands (in `CommandProcessor.processCommand -> dataVectorizerService.vectorize(commandContext, cmd)`) and ignore if no need to vectorize.

PR #2118, opened 2025-06-09
Update values.yaml

phact:
fix image repository, should be data-api not jsonapi


tatu-at-datastax:
Ok no. Will add a bogus commit to see if last commit author matters (vaguely recall it does).

tatu-at-datastax:
Ok. Will just re-create as branch in main repo.

tatu-at-datastax:
Re-created as #2517 to get CI to work.

Issue #2054, opened 2025-04-29
Add support for BM25/Lexical sort on "analyzed text" indexed Table columns


tatu-at-datastax:
In addition to BM25/Lexical filtering (#2053) for Tables, we also need to enable BM25/Lexical sorting, similar to how Collections supported it (and somewhat similar to Vector/ANN sorting).

tatu-at-datastax:
Not yet completed, re-opening.

tatu-at-datastax:
Incremental work but not yet implemented.

PR #2123, opened 2025-06-10
Refactor SortClause handling to prep for #2054

tatu-at-datastax:
**What this PR does**:

Refactors building of `SortClause` to allow for Collection/Table handling specialization

N/A

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/builders/CollectionSortClauseBuilder.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/builders/CollectionSortClauseBuilder.java
@@ -1,8 +1,16 @@
 package io.stargate.sgv2.jsonapi.api.model.command.builders;
 
+import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import io.stargate.sgv2.jsonapi.api.model.command.clause.sort.SortClause;
+import io.stargate.sgv2.jsonapi.api.model.command.clause.sort.SortExpression;
+import io.stargate.sgv2.jsonapi.config.constants.DocumentConstants;
+import io.stargate.sgv2.jsonapi.exception.ErrorCodeV1;
 import io.stargate.sgv2.jsonapi.service.schema.collections.CollectionSchemaObject;
+import io.stargate.sgv2.jsonapi.service.schema.collections.DocumentPath;
+import io.stargate.sgv2.jsonapi.service.schema.naming.NamingRules;
+import io.stargate.sgv2.jsonapi.util.JsonUtil;
+import java.util.Collections;
 
 /** {@link SortClauseBuilder} to use with Collections. */
 public class CollectionSortClauseBuilder extends SortClauseBuilder<CollectionSchemaObject> {
@@ -12,6 +20,55 @@ public class CollectionSortClauseBuilder extends SortClauseBuilder<CollectionSch
 
   @Override
   public SortClause buildAndValidate(ObjectNode sortNode) {
+    // $lexical is only special for Collections: handle first
+    JsonNode lexicalValue = sortNode.get(DocumentConstants.Fields.LEXICAL_CONTENT_FIELD);
+    if (lexicalValue != null) {
+      if (sortNode.size() > 1) {
+        throw ErrorCodeV1.INVALID_SORT_CLAUSE.toApiException(
+            "if sorting by '%s' no other sort expressions allowed",
+            DocumentConstants.Fields.LEXICAL_CONTENT_FIELD);
+      }
+      if (!lexicalValue.isTextual()) {
+        throw ErrorCodeV1.INVALID_SORT_CLAUSE.toApiException(
+            "if sorting by '%s' value must be String, not %s",
+            DocumentConstants.Fields.LEXICAL_CONTENT_FIELD,
+            JsonUtil.nodeTypeAsString(lexicalValue));
+      }
+      // We cannot yet determine if lexical sort supported by the collection, just
+      // construct clause
+      return new SortClause(
+          Collections.singletonList(SortExpression.bm25Search(lexicalValue.textValue())));
+    }
+
     return defaultBuildAndValidate(sortNode);
+  }
+
+  @Override
+  protected String validateSortClausePath(String path) {
+    if (!NamingRules.FIELD.apply(path)) {
+      if (path.isEmpty()) {
+        throw ErrorCodeV1.INVALID_SORT_CLAUSE_PATH.toApiException(
+            "path must be represented as a non-empty string");
+      }
+      // But allow "well-known" fields
+      switch (path) {
+        case DocumentConstants.Fields.LEXICAL_CONTENT_FIELD,
+            DocumentConstants.Fields.VECTOR_EMBEDDING_FIELD,
+            DocumentConstants.Fields.VECTOR_EMBEDDING_TEXT_FIELD -> {}
+        default ->
+            throw ErrorCodeV1.INVALID_SORT_CLAUSE_PATH.toApiException(
+                "path ('%s') cannot start with '$' (except for pseudo-fields '$lexical', '$vector' and '$vectorize')",
+                path);
+      }
+    }
+
+    try {
+      path = DocumentPath.verifyEncodedPath(path);
+    } catch (IllegalArgumentException e) {
+      throw ErrorCodeV1.INVALID_SORT_CLAUSE_PATH.toApiException(
+          "sort clause path ('%s') is not a valid path. " + e.getMessage(), path);
+    }
+
+    return path;
   }
 }
--- a/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/builders/SortClauseBuilder.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/builders/SortClauseBuilder.java
@@ -14,11 +14,7 @@ import io.stargate.sgv2.jsonapi.exception.ErrorCodeV1;
 import io.stargate.sgv2.jsonapi.service.cqldriver.executor.SchemaObject;
 import io.stargate.sgv2.jsonapi.service.cqldriver.executor.TableSchemaObject;
 import io.stargate.sgv2.jsonapi.service.schema.collections.CollectionSchemaObject;
-import io.stargate.sgv2.jsonapi.service.schema.collections.DocumentPath;
-import io.stargate.sgv2.jsonapi.service.schema.naming.NamingRules;
-import io.stargate.sgv2.jsonapi.util.JsonUtil;
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
@@ -64,31 +60,13 @@ public abstract class SortClauseBuilder<T extends SchemaObject> {
 
   protected abstract SortClause buildAndValidate(ObjectNode sortNode);
 
+  protected abstract String validateSortClausePath(String path);
+
   protected SortClause defaultBuildAndValidate(ObjectNode sortNode) {
     // safe to iterate, we know it's an Object
     Iterator<Map.Entry<String, JsonNode>> fieldIter = sortNode.fields();
     int totalFields = sortNode.size();
     List<SortExpression> sortExpressions = new ArrayList<>(sortNode.size());
-
-    // $lexical is only allowed alone: handle first
-    JsonNode lexicalValue = sortNode.get(DocumentConstants.Fields.LEXICAL_CONTENT_FIELD);
-    if (lexicalValue != null) {
-      if (sortNode.size() > 1) {
-        throw ErrorCodeV1.INVALID_SORT_CLAUSE.toApiException(
-            "if sorting by '%s' no other sort expressions allowed",
-            DocumentConstants.Fields.LEXICAL_CONTENT_FIELD);
-      }
-      if (!lexicalValue.isTextual()) {
-        throw ErrorCodeV1.INVALID_SORT_CLAUSE.toApiException(
-            "if sorting by '%s' value must be String, not %s",
-            DocumentConstants.Fields.LEXICAL_CONTENT_FIELD,
-            JsonUtil.nodeTypeAsString(lexicalValue));
-      }
-      // We cannot yet determine if lexical sort supported by the collection, just
-      // construct clause
-      return new SortClause(
-          Collections.singletonList(SortExpression.bm25Search(lexicalValue.textValue())));
-    }
 
     while (fieldIter.hasNext()) {
       Map.Entry<String, JsonNode> inner = fieldIter.next();
@@ -209,33 +187,5 @@ public abstract class SortClauseBuilder<T extends SchemaObject> {
       }
     }
     return new SortClause(sortExpressions);
-  }
-
-  private String validateSortClausePath(String path) {
-    if (!NamingRules.FIELD.apply(path)) {
-      if (path.isEmpty()) {
-        throw ErrorCodeV1.INVALID_SORT_CLAUSE_PATH.toApiException(
-            "path must be represented as a non-empty string");
-      }
-      // But allow "well-known" fields
-      switch (path) {
-        case DocumentConstants.Fields.LEXICAL_CONTENT_FIELD,
-            DocumentConstants.Fields.VECTOR_EMBEDDING_FIELD,
-            DocumentConstants.Fields.VECTOR_EMBEDDING_TEXT_FIELD -> {}
-        default ->
-            throw ErrorCodeV1.INVALID_SORT_CLAUSE_PATH.toApiException(
-                "path ('%s') cannot start with '$' (except for pseudo-fields '$lexical', '$vector' and '$vectorize')",
-                path);
-      }
-    }
-
-    try {
-      path = DocumentPath.verifyEncodedPath(path);
-    } catch (IllegalArgumentException e) {
-      throw ErrorCodeV1.INVALID_SORT_CLAUSE_PATH.toApiException(
-          "sort clause path ('%s') is not a valid path. " + e.getMessage(), path);
-    }
-
-    return path;
   }
 }
--- a/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/builders/TableSortClauseBuilder.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/builders/TableSortClauseBuilder.java
@@ -2,7 +2,9 @@ package io.stargate.sgv2.jsonapi.api.model.command.builders;
 
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import io.stargate.sgv2.jsonapi.api.model.command.clause.sort.SortClause;
+import io.stargate.sgv2.jsonapi.exception.ErrorCodeV1;
 import io.stargate.sgv2.jsonapi.service.cqldriver.executor.TableSchemaObject;
+import io.stargate.sgv2.jsonapi.service.schema.collections.DocumentPath;
 
 /** {@link SortClauseBuilder} to use with Tables. */
 public class TableSortClauseBuilder extends SortClauseBuilder<TableSchemaObject> {
@@ -13,5 +15,23 @@ public class TableSortClauseBuilder extends SortClauseBuilder<TableSchemaObject>
   @Override
   public SortClause buildAndValidate(ObjectNode sortNode) {
     return defaultBuildAndValidate(sortNode);
+  }
+
+  @Override
+  protected String validateSortClausePath(String path) {
+    // Tables have few rules: but cannot be empty
+    if (path.isEmpty()) {
+      throw ErrorCodeV1.INVALID_SORT_CLAUSE_PATH.toApiException(
+          "path must be represented as a non-empty string");
+    }
+    // Do we use escaping for Tables? For sake of consistency I think so
+    try {
+      path = DocumentPath.verifyEncodedPath(path);
+    } catch (IllegalArgumentException e) {
+      throw ErrorCodeV1.INVALID_SORT_CLAUSE_PATH.toApiException(
+          "sort clause path ('%s') is not a valid path. " + e.getMessage(), path);
+    }
+
+    return path;
   }
 }
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/resolver/sort/TableCqlSortClauseResolver.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/resolver/sort/TableCqlSortClauseResolver.java
@@ -77,7 +77,6 @@ public class TableCqlSortClauseResolver<CmdT extends Command & Filterable & Sort
     var whereCQLClause =
         TableWhereCQLClause.forSelect(
             commandContext.schemaObject(), tableFilterResolver.resolve(commandContext, command));
-
     return vectorAndVectorizeSorts.isEmpty()
         ? resolveNonVectorSort(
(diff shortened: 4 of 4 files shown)


src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/sort/SortClause.java
```diff
@@ -91,14 +91,9 @@ public SortExpression bm25SearchExpression() {
...
   public void validate(CollectionSchemaObject collection) {
-    // First things first: BM25 search may or may not be available
-    SortExpression bm25Expr = bm25SearchExpression();
-    if (bm25Expr != null) {
-      if (!collection.lexicalConfig().enabled()) {
-        throw ErrorCodeV1.LEXICAL_NOT_ENABLED_FOR_COLLECTION.toApiException(
-            "Lexical search is not enabled for collection '%s'", collection.name());
-      }
-      // But it must be the only sort expression so we can stop here
+    // First things first: BM25 search uses its own index; also the only expression
```
tatu-at-datastax, line 94: Was moved to building part, caught earlier.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/builders/TableSortClauseBuilder.java
```diff
@@ -14,4 +16,22 @@ public TableSortClauseBuilder(TableSchemaObject table) {
...
   }
+
+  @Override
+  protected String validateSortClausePath(String path) {
+    // Tables have few rules: but cannot be empty
+    if (path.isEmpty()) {
+      throw ErrorCodeV1.INVALID_SORT_CLAUSE_PATH.toApiException(
+          "path must be represented as a non-empty string");
+    }
+    // Do we use escaping for Tables? For sake of consistency I think so
```
tatu-at-datastax, line 27: This is how code was working, just adding comment to raise the point we don't have to, really (since "."s have no special meaning for Column names.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/builders/SortClauseBuilder.java
```diff
@@ -64,178 +57,59 @@ public SortClause build(JsonNode node) {
...
 
+  protected abstract String validateSortClausePath(String path);
+
   protected SortClause defaultBuildAndValidate(ObjectNode sortNode) {
     // safe to iterate, we know it's an Object
     Iterator<Map.Entry<String, JsonNode>> fieldIter = sortNode.fields();
     int totalFields = sortNode.size();
     List<SortExpression> sortExpressions = new ArrayList<>(sortNode.size());
 
-    // $lexical is only allowed alone: handle first
```
tatu-at-datastax, line 73: Moved from shared base class to Collection-specific (see above)

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/builders/SortClauseBuilder.java
```diff
@@ -64,178 +57,59 @@ public SortClause build(JsonNode node) {
...
-          Collections.singletonList(SortExpression.bm25Search(lexicalValue.textValue())));
-    }
-
     while (fieldIter.hasNext()) {
       Map.Entry<String, JsonNode> inner = fieldIter.next();
       final String path = inner.getKey().trim();
       // Validation will check against invalid paths, as well as decode "amp-escaping"
-      final String validatedPath = validateSortClausePath(path);
-
-      float[] vectorFloats = null;
```
tatu-at-datastax, line 99: This whole code extracted into per-expression method, separate for Table/Collection builders.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/builders/CollectionSortClauseBuilder.java
```diff
@@ -12,6 +21,106 @@ public CollectionSortClauseBuilder(CollectionSchemaObject collection) {
...
+        throw ErrorCodeV1.INVALID_SORT_CLAUSE.toApiException(
+            "if sorting by '%s' value must be String, not %s",
+            DocumentConstants.Fields.LEXICAL_CONTENT_FIELD,
+            JsonUtil.nodeTypeAsString(lexicalValue));
+      }
+      // We can also check if lexical sort supported by the collection:
+      if (!schema.lexicalConfig().enabled()) {
+        throw ErrorCodeV1.LEXICAL_NOT_ENABLED_FOR_COLLECTION.toApiException(
+            "Lexical search is not enabled for collection '%s'", schema.name());
+      }
```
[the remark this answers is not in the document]
tatu-at-datastax: Yes, that actually makes sense, will move.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/builders/SortClauseBuilder.java
```diff
@@ -64,178 +57,59 @@ public SortClause build(JsonNode node) {
...
-        sortExpressions.add(SortExpression.tableVectorizeSort(path, inner.getValue().textValue()));
-      } else {
-        if (!inner.getValue().isInt()
-            || !(inner.getValue().intValue() == 1 || inner.getValue().intValue() == -1)) {
-          throw ErrorCodeV1.INVALID_SORT_CLAUSE_VALUE.toApiException(
-              "Sort ordering value can only be `1` for ascending or `-1` for descending (not `%s`)",
-              inner.getValue());
-        }
+  protected abstract SortExpression buildAndValidateExpression(
+      String path, String validatedPath, JsonNode innerValue, int totalFields);
```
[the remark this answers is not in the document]
tatu-at-datastax: Actually you are right; I should pass one: validation does not decode escaping.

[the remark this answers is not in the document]
tatu-at-datastax: Removed "extra" path, not passing 2 separate ones.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/builders/SortClauseBuilder.java
```diff
@@ -64,178 +57,58 @@ public SortClause build(JsonNode node) {
...
-        }
+  protected abstract SortExpression buildAndValidateExpression(
+      String validatedPath, JsonNode innerValue, int totalFields);
 
-        boolean ascending = inner.getValue().intValue() == 1;
-        SortExpression exp = SortExpression.sort(validatedPath, ascending);
-        sortExpressions.add(exp);
-      }
+  protected SortExpression defaultBuildAndValidateExpression(
+      String validatedPath, JsonNode innerValue) {
```
amorton, line 82: have not read the whole PR yet - but having a default* in the base feels odd. is there a reason we don't use the same name and have the sub call the super if it wants to ?

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/builders/SortClauseBuilder.java
```diff
@@ -64,178 +57,59 @@ public SortClause build(JsonNode node) {
...
-        SortExpression exp = SortExpression.sort(validatedPath, ascending);
-        sortExpressions.add(exp);
-      }
+  protected SortExpression defaultBuildAndValidateExpression(
+      String path, String validatedPath, JsonNode innerValue) {
+    if (!innerValue.isInt()) {
+      throw ErrorCodeV1.INVALID_SORT_CLAUSE_VALUE.toApiException(
+          "Sort ordering value should be integer `1` or `-1`; or Array of numbers (Vector); or String but was: %s",
+          JsonUtil.nodeTypeAsString(innerValue));
     }
```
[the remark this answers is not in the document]
tatu-at-datastax: Added separate reporting: I think error messaging needs to be further improved but hopefully acceptable for refactoring.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/builders/SortClauseBuilder.java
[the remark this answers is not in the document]
amorton: to add, it feels odd that the function called buildAndValidateExpression() has a `validatedPath` param, at first glance seems like the function would do that.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/builders/SortClauseBuilder.java
```diff
@@ -64,178 +57,67 @@ public SortClause build(JsonNode node) {
...
-            "path must be represented as a non-empty string");
+  protected abstract SortExpression buildAndValidateExpression(
+      String validatedPath, JsonNode innerValue, int totalFields);
+
+  protected SortExpression defaultBuildAndValidateExpression(
+      String validatedPath, JsonNode innerValue) {
+    if (!innerValue.isInt()) {
+      // Special checking for String and ArrayNode to give less confusing error messages
+      if (innerValue.isArray()) {
+        throw ErrorCodeV1.INVALID_SORT_CLAUSE_VALUE.toApiException(
```
amorton, line 86: I'll let those old V1 errors pass, assuming we are lift-and-shift from previous logic

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/builders/SortClauseBuilder.java
amorton: reading further, this looks like handling for a regular sort expression so maybe rename ?

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/builders/SortClauseBuilder.java
```diff
@@ -64,178 +57,67 @@ public SortClause build(JsonNode node) {
...
-          "sort clause path ('%s') is not a valid path. " + e.getMessage(), path);
+    if (!(innerValue.intValue() == 1 || innerValue.intValue() == -1)) {
+      throw ErrorCodeV1.INVALID_SORT_CLAUSE_VALUE.toApiException(
+          "Sort ordering value can only be `1` for ascending or `-1` for descending (not `%s`)",
+          innerValue);
     }
 
-    return path;
+    boolean ascending = innerValue.intValue() == 1;
+    return SortExpression.sort(validatedPath, ascending);
```
amorton, line 104: Just a suggestion - and in line with my confusion about the function name above. Is this the time to introduce a SortType enum and add it to the SortExpression ? Feels like we are doing a lot in the SortExpression what would be simplier with a type field 

Or subclassing to avoid chameleon class

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/builders/SortClauseBuilder.java
```diff
@@ -64,178 +57,67 @@ public SortClause build(JsonNode node) {
...
+  }
+
+  protected float[] tryDecodeBinaryVector(String path, JsonNode innerValue) {
```
amorton, line 107: +1 - at first it looks like duplicating, but i see the diff errors here

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/builders/CollectionSortClauseBuilder.java
```diff
@@ -12,6 +21,107 @@ public CollectionSortClauseBuilder(CollectionSchemaObject collection) {
 
   @Override
   public SortClause buildAndValidate(ObjectNode sortNode) {
+    // $lexical is only special for Collections: handle first
+    JsonNode lexicalValue = sortNode.get(DocumentConstants.Fields.LEXICAL_CONTENT_FIELD);
+    if (lexicalValue != null) {
+      // We can also check if lexical sort supported by the collection:
+      if (!schema.lexicalConfig().enabled()) {
+        throw ErrorCodeV1.LEXICAL_NOT_ENABLED_FOR_COLLECTION.toApiException(
```
amorton, line 29: is there a V2 error for this ?

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/builders/CollectionSortClauseBuilder.java
```diff
@@ -12,6 +21,107 @@ public CollectionSortClauseBuilder(CollectionSchemaObject collection) {
...
+    }
+
+    // Otherwise, use shared default processing
     return defaultBuildAndValidate(sortNode);
   }
+
+  @Override
+  protected String validateSortClausePath(String path) {
+    if (!NamingRules.FIELD.apply(path)) {
+      if (path.isEmpty()) {
```
amorton, line 56: nit:  isBlank

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/builders/CollectionSortClauseBuilder.java
```diff
@@ -12,6 +21,107 @@ public CollectionSortClauseBuilder(CollectionSchemaObject collection) {
...
+        throw ErrorCodeV1.INVALID_SORT_CLAUSE_PATH.toApiException(
+            "path must be represented as a non-empty string");
+      }
+      // But allow "well-known" fields
+      switch (path) {
```
amorton, line 61: nit: switch and if can be a single switch statement. use a guard on the case 

`case String s && s.isBlank():`

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/builders/SortClauseBuilder.java
```diff
@@ -64,178 +57,67 @@ public SortClause build(JsonNode node) {
...
     return new SortClause(sortExpressions);
   }
 
-  private String validateSortClausePath(String path) {
-    if (!NamingRules.FIELD.apply(path)) {
-      if (path.isEmpty()) {
-        throw ErrorCodeV1.INVALID_SORT_CLAUSE_PATH.toApiException(
-            "path must be represented as a non-empty string");
+  protected abstract SortExpression buildAndValidateExpression(
+      String validatedPath, JsonNode innerValue, int totalFields);
```
amorton, line 79: nit: totalFields -> sortFieldCount or sortExpressionCount ???

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/builders/CollectionSortClauseBuilder.java
```diff
@@ -12,6 +21,107 @@ public CollectionSortClauseBuilder(CollectionSchemaObject collection) {
...
   public SortClause buildAndValidate(ObjectNode sortNode) {
+    // $lexical is only special for Collections: handle first
+    JsonNode lexicalValue = sortNode.get(DocumentConstants.Fields.LEXICAL_CONTENT_FIELD);
+    if (lexicalValue != null) {
+      // We can also check if lexical sort supported by the collection:
+      if (!schema.lexicalConfig().enabled()) {
+        throw ErrorCodeV1.LEXICAL_NOT_ENABLED_FOR_COLLECTION.toApiException(
+            "Lexical search is not enabled for collection '%s'", schema.name());
+      }
+      if (sortNode.size() > 1) {
```
amorton, line 32: buildAndValidateExpression gets the totalFields count and does checks like this for $vectorize, unclear why the lexical is handled differently

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/builders/CollectionSortClauseBuilder.java
```diff
@@ -12,6 +21,107 @@ public CollectionSortClauseBuilder(CollectionSchemaObject collection) {
...
+        vectorFloats = new float[arrayNode.size()];
+        if (arrayNode.isEmpty()) {
+          throw ErrorCodeV1.SHRED_BAD_VECTOR_SIZE.toApiException();
+        }
+        for (int i = 0; i < arrayNode.size(); i++) {
+          JsonNode element = arrayNode.get(i);
+          if (!element.isNumber()) {
+            throw ErrorCodeV1.SHRED_BAD_VECTOR_VALUE.toApiException();
+          }
+          vectorFloats[i] = element.floatValue();
```
amorton, line 105: do we have re-usable code in shredding for moving an array into float[]?

the table builder calls arrayNodeToVector()

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/builders/TableSortClauseBuilder.java
```diff
@@ -14,4 +21,50 @@ public TableSortClauseBuilder(TableSchemaObject table) {
   public SortClause buildAndValidate(ObjectNode sortNode) {
     return defaultBuildAndValidate(sortNode);
   }
+
+  @Override
+  protected String validateSortClausePath(String path) {
+    // Tables have few rules: but cannot be empty
+    if (path.isEmpty()) {
```
amorton, line 28: the collection builder runs naming rules, why dont we here ? e.g. the TableNamingRule is pretty basic. and reading that class and here, should it be isEmpty or isBlank ?

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/builders/TableSortClauseBuilder.java
amorton: sure, maybe reword the comment to be more definitive

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/builders/TableSortClauseBuilder.java
```diff
@@ -14,4 +21,50 @@ public TableSortClauseBuilder(TableSchemaObject table) {
...
+          "sort clause path ('%s') is not a valid path. " + e.getMessage(), path);
+    }
+
+    return path;
+  }
+
+  @Override
+  protected SortExpression buildAndValidateExpression(
+      String validatedPath, JsonNode innerValue, int totalFields) {
+    float[] vectorFloats = tryDecodeBinaryVector(validatedPath, innerValue);
```
amorton, line 46: a bit unclear here, comments would help.Are we testing for a vector sort first because it is easier to test for ?

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/builders/TableSortClauseBuilder.java
```diff
@@ -14,4 +21,50 @@ public TableSortClauseBuilder(TableSchemaObject table) {
...
+
+    // handle table vector sort
+    if (vectorFloats != null) {
+      return SortExpression.tableVectorSort(validatedPath, vectorFloats);
+    }
+    if (innerValue instanceof ArrayNode innerArray) {
+      // TODO: HACK: quick support for tables, if the value is an array we will assume the
```
amorton, line 53: we can prob remove the hack comments if they are not needed. 

and we could do the check for the number of fields now with totalFields passed in, This happens in the collection builder.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/sort/SortExpression.java
```diff
@@ -18,7 +18,7 @@ public record SortExpression(
     boolean ascending,
     @Nullable float[] vector,
     @Nullable String vectorize,
-    @Nullable String bm25Query) {
+    @Nullable String lexicalQuery) {
```
amorton, line 21: I made this comment elsewhere as well, now that we have better handle on how and when the SortExpression objects are created it feels like time to have either a typenum or a subclass structure. 

there are hidden behaviour in here: if the instance is created via the static factories than the is* functions work. But there is a public constructor that allows us to break all the rules.

amorton (approved):
I cannot see any logical problems, and I know this is a painful area to crack open, lots of nit picking :) 

Main themes: 
* function names on SortClauseBuilder and subclasses are a bit confusing 
* there may be some refactoring in SortClauseBuilder and subclasses so all casses are handled the same, e.g. lexical is in a diff place. 
* IMHO we should refactor the SortExpression to add a type enum or class structure now that we have better control on how instances are created

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/clause/sort/SortExpression.java
tatu-at-datastax: Yeah Records are just open structs (although immutable).

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/builders/SortClauseBuilder.java
```diff
@@ -64,178 +57,59 @@ public SortClause build(JsonNode node) {
 
   protected abstract SortClause buildAndValidate(ObjectNode sortNode);
 
+  protected abstract String validateSortClausePath(String path);
```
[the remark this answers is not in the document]
tatu-at-datastax: Internal method, but will add a check since those are cheap to have.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/builders/SortClauseBuilder.java
tatu-at-datastax: Yeah it's bit awkward; results of incremental conversion.
Let me think about how to change.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/builders/SortClauseBuilder.java
tatu-at-datastax: Exactly. Validation logic will (need to) change, too, as will exceptions.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/builders/TableSortClauseBuilder.java
tatu-at-datastax: `isEmpty()` works here since caller does `trim()`. Naming rules avoided (but not commented) just to allow leading `$`. But should we? If I remember correctly, they would cause issues elsewhere.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/builders/SortClauseBuilder.java
tatu-at-datastax: and yes, I should rename "validatedPath" as just "path"; no need to document in name (if need be, in Javadoc param definition)

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/builders/CollectionSortClauseBuilder.java
```diff
@@ -12,6 +21,107 @@ public CollectionSortClauseBuilder(CollectionSchemaObject collection) {
...
+              SortExpression.collectionLexicalSort(lexicalValue.textValue())));
+    }
+
+    // Otherwise, use shared default processing
     return defaultBuildAndValidate(sortNode);
   }
+
+  @Override
+  protected String validateSortClausePath(String path) {
+    if (!NamingRules.FIELD.apply(path)) {
```
amorton, line 55: form discussion - consider not validating in the DML commands because the rules are for creating DDL , and if they changed OR it was made in CQL , they would stop access

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/builders/CollectionSortClauseBuilder.java
tatu-at-datastax: Good catch, changed.

PR #2131, opened 2025-06-17
[Tables UDT] CreateType, DropType, AlterType.

Yuqi-Du:
**What this PR does**:
This PR has first 3 DDL commands for UDT. 

# CreateType
One thing to note is for the UDT type definition, we re-use ColumnDesc for CreateType/AlterType, so you will see the custom deserializer TypeFieldsContainerDeserializer, and it passes a boolean flag into ColumnDescDeserializer to indicate we are doing deserializing for UDT definition not table definition. For regular CreateTable/AlterTable, deserializer is singleton instantiated by Quarkus, the boolean flag set as false.

```json
  {
      "createType": {
          "name": "address",
          "definition": {
              "fields": {
                  "city": "text",
                  "postcode": "int"
              }
          }
      }
  }
```

# AlterType.

- This command does not support options like IF NOT EXISTS or IF EXISTS, as these are not available in Astra, even though they are supported in Cassandra 5.
- Note, unlike AlterTable, this command handle both renaming and adding fields in a UDT. It builds a {@link TaskGroup} containing multiple {@link AlterTypeDBTask} instances, they are executed in parallel, so the command may have partial success and return errors for failed tasks.

```json
   {
        "alterType": {
            "name": "address",
            "rename": {
                "fields" : {
                    "country": "countryCode"
                 }
            },
            "add": {
                "fields": {
                    "city": "text",
                    "post_code": "int"
                }
            }
        }
    }
```

# DropType
This one is clear, with ifExists option in-place.

```json

  {
      "dropType": {
          "name": "address",
          "options":{
             "ifExists": true
          }
      }
  }

```

**Checklist**
- [ ] Changes manually tested


Yuqi-Du:
will close, since https://github.com/stargate/data-api/pull/2145 has it

PR #2133, opened 2025-06-18
Fixes the response from `findEmbeddingProviders`: the parameters type - from "NUMBER" to "number"

Hazel-Datastax:
**What this PR does**:
Our previous format is 
```
                        "parameters": [
                            {
                                "name": "vectorDimension",
                                "type": "number",
                                "required": true,
                                "defaultValue": null,
                                "validation": {
                                    "numericRange": [
                                        2,
                                        3072
                                    ]
                                },
                                "help": "Vector dimension to use in the database, should be the same as the model used by the endpoint."
                            }
                        ]
```

now the format is:
```
                        "parameters": [
                            {
                                "name": "vectorDimension",
                                "type": "NUMBER",
                                "required": true,
                                "defaultValue": null,
                                "validation": {
                                    "numericRange": [
                                        2,
                                        3072
                                    ]
                                },
                                "help": "Vector dimension to use in the database, should be the same as the model used by the endpoint."
                            }
                        ]
```

The "number" is all uppercase

Fixes #<issue number>

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/embedding/configuration/EmbeddingProvidersConfig.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/embedding/configuration/EmbeddingProvidersConfig.java
@@ -279,15 +279,14 @@ public interface EmbeddingProvidersConfig {
       NUMBER("number"),
       BOOLEAN("boolean");
 
-      private final String type;
+      private final String apiName;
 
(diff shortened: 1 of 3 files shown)


tatu-at-datastax (commented):
Would this need a test to verify? Although it looks like there is an IT fail to fix, so maybe this is already covered.

PR #2154, opened 2025-07-15
(WIP) Fix #2053: add Lexical Filter support for Tables

### LINKED ISSUE #2053: Add support for "$match" BM25-filter operation on Table columns
Part of BM25 support for Tables is ability to use Lexical/BM25 filtering on `text` / `varchar` columns with "analyzed text" inde (see #2051). This should work using new "$match" operator.

Note that implementation is separate from Collections one (tracked via #2052).

tatu-at-datastax:
**What this PR does**:

Adds support for Lexical filters ("BM25", operator "$match") for Tables

Fixes #2053 

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/src/test/java/io/stargate/sgv2/jsonapi/api/v1/tables/FindWithLexicalFilterTableIntegrationTest.java
+++ b/src/test/java/io/stargate/sgv2/jsonapi/api/v1/tables/FindWithLexicalFilterTableIntegrationTest.java
@@ -1,0 +1,93 @@
+package io.stargate.sgv2.jsonapi.api.v1.tables;
+
+import static io.stargate.sgv2.jsonapi.api.v1.util.DataApiCommandSenders.assertNamespaceCommand;
+import static io.stargate.sgv2.jsonapi.api.v1.util.DataApiCommandSenders.assertTableCommand;
+
+import io.quarkus.test.common.WithTestResource;
+import io.quarkus.test.junit.QuarkusIntegrationTest;
+import io.stargate.sgv2.jsonapi.testresource.DseTestResource;
+import java.util.Map;
+import org.junit.jupiter.api.ClassOrderer;
+import org.junit.jupiter.api.Nested;
+import org.junit.jupiter.api.Order;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestClassOrder;
+import org.junit.jupiter.api.condition.DisabledIfSystemProperty;
+
+@QuarkusIntegrationTest
+@WithTestResource(value = DseTestResource.class, restrictToAnnotatedClass = false)
+@TestClassOrder(ClassOrderer.OrderAnnotation.class)
+public class FindWithLexicalFilterTableIntegrationTest extends AbstractTableIntegrationTestBase {
+  private static final String TABLE_NAME = "lexicalFilterTableTest";
+
+  @DisabledIfSystemProperty(named = TEST_PROP_LEXICAL_DISABLED, matches = "true")
+  @Nested
+  @Order(1)
+  class Setup {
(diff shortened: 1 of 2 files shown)


src/main/java/io/stargate/sgv2/jsonapi/api/model/command/builders/FilterClauseBuilder.java
```diff
@@ -150,6 +150,8 @@ private void populateExpression(
             case "$or" -> LogicalExpression.or();
             case DocumentConstants.Fields.VECTOR_EMBEDDING_FIELD,
                     DocumentConstants.Fields.VECTOR_EMBEDDING_TEXT_FIELD ->
+                // TODO: (21-Jul-2025) Should be refactored to CollectionFilterClauseBuilder as it
+                // only applies to Collections
```
Hazel-Datastax, line 154: Maybe create an issue for it?

src/test/java/io/stargate/sgv2/jsonapi/api/v1/tables/FindWithLexicalFilterTableIntegrationTest.java
```diff
@@ -0,0 +1,148 @@
...
+
+@QuarkusIntegrationTest
+@WithTestResource(value = DseTestResource.class)
+@TestClassOrder(ClassOrderer.OrderAnnotation.class)
+public class FindWithLexicalFilterTableIntegrationTest extends AbstractTableIntegrationTestBase {
+  private static final String TABLE_NAME = "lexicalFilterTableTest";
+
+  @DisabledIfSystemProperty(named = TEST_PROP_LEXICAL_DISABLED, matches = "true")
+  @Nested
+  @Order(1)
```
Hazel-Datastax, line 25: I guess using `@Order` here (and below) is for the setup, can we use `@BeforeAll` so we can remove `@Order`? It may be hard to maintain in the future with the `@Order`

Hazel-Datastax (approved):
LGTM

src/test/java/io/stargate/sgv2/jsonapi/api/v1/tables/FindWithLexicalFilterTableIntegrationTest.java
tatu-at-datastax: Agreed wrt ordering. One possible problem is whether `@DisabledIfSystemProperty` works for `@BeforeAll` -- I can try it.

tatu-at-datastax: As per ChatGpt:

-----
Good question — this touches on how JUnit 5 conditional test execution works versus Quarkus’s own test lifecycle.

@DisabledIfSystemProperty (from org.junit.jupiter.api.condition) is a JUnit 5 annotation. It’s only effective on:

Test methods (@Test)

Test classes

👉 It does not apply to lifecycle methods like @BeforeAll, @AfterAll, @BeforeEach, @AfterEach.
If you put it on a @BeforeAll method, JUnit will ignore it — the method will still run if the containing class itself is enabled.

So, if you want to conditionally disable a whole test class (including its @BeforeAll setup), the correct approach is to annotate the class itself with... "

-----

So, I cannot use that method directly, unless I change the whole test class.

But I will try to do just that. Thanks!

tatu-at-datastax: Looks good in result logs too (proper skipping of tests on DSE-6.9 vs HCD).

PR #2165, opened 2025-07-29
Fix #2164: Map<K, Boolean> into Map<K, String> + post-process for config handling

### LINKED ISSUE #2164: Change to fix breaking Quarkus config (SmallRye): "" no longer acceptable as `null` for `Boolean`
Looks like something in SmallRye config library broke such that our use of "" for "undefined" (`null`) for `ApiFeature` throws exception.

This is pretty nasty regression, but not much we can do except to use `Optional<Boolean>` instead of `Boolean`.

**EDIT**: Change to use `Optional<Boolean>` does NOT WORK as value of `Map`: `Map<ApiType, Optional<Boolean>>` fails!!!

tatu-at-datastax:
**What this PR does**:

Uses `Map<ApiFeature, String>` instead of `Map<ApiFeature, Boolean>` for configuration, with bit of post-processing.

Fixes #2164

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/config/feature/ApiFeatures.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/config/feature/ApiFeatures.java
@@ -3,7 +3,6 @@ package io.stargate.sgv2.jsonapi.config.feature;
 import io.stargate.sgv2.jsonapi.api.request.RequestContext;
 import java.util.Collections;
 import java.util.Map;
-import java.util.Optional;
 
 /**
  * Accessor for combined state of feature flags; typically based on static configuration (with its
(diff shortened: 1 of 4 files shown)


tatu-at-datastax:
Third attempt: bind to `Map<ApiFeature, String>` at first: this SHOULD work.
And then finagle into `Boolean` manually.

src/main/java/io/stargate/sgv2/jsonapi/config/feature/ApiFeatures.java
```diff
@@ -51,4 +51,18 @@ public boolean isFeatureEnabled(ApiFeature flag) {
     }
     return flag.enabledByDefault();
   }
+
+  private Boolean booleanFromString(String str) {
+    if (str == null || str.isBlank()) {
+      return null; // no value, so not enabled
+    }
+    if ("true".equals(str)) {
```
amorton, line 60: should this be `equalsIgnoreCase()`

amorton (approved):
Approved so it can merge, but checking should it be equalsIgnoreCase()

src/main/java/io/stargate/sgv2/jsonapi/config/feature/ApiFeatures.java
tatu-at-datastax: Matter of taste: I think by default only "true" and "false" accepted for boolean/Boolean in Java.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/RerankFeatureDisabledIntegrationTest.java
```diff
@@ -30,9 +30,10 @@ public TestResource() {}
 
     @Override
     public String getFeatureFlagReranking() {
-      // return empty to leave feature "undefined" (disabled unless per-request header override)
+      // return BLANK String to leave feature "undefined" (disabled unless per-request header
+      // override)
       // ("false" would be "disabled" for all tests, regardless of headers)
-      return "";
+      return " ";
```
tatu-at-datastax, line 36: Using non-empty because apparently SmallRye fails on Empty String directly. What a piece-of-work library :-(

PR #2166, opened 2025-07-29
Fix #2164 by adding custom Boolean converter which accepts "" as `null`

### LINKED ISSUE #2164: Change to fix breaking Quarkus config (SmallRye): "" no longer acceptable as `null` for `Boolean`

tatu-at-datastax:
**What this PR does**:

Adds custom `Converter<Boolean>` to accept empty String as `null` value

Fixes #2164

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/config/NullableBooleanConverter.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/config/NullableBooleanConverter.java
@@ -1,0 +1,23 @@
+package io.stargate.sgv2.jsonapi.config;
+
+import jakarta.enterprise.context.ApplicationScoped;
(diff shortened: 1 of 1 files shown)


tatu-at-datastax:
Alas, does not work either!!!! Custom `Converter`s not used for structured types.

Issue #1672, opened 2024-11-08
ordered insertMany incorrectly fails if an row has pre-execution error


hemidactylus:
Symptom: insertMany on tables, when `ordered=True` and there's a faulty row _in mid-list_, does not insert even those coming prior to that one.

This is at odds with Collections (ordered=true), where the docs are inserted sequentially until the erroring one.

Also in that case the response has no `"status"` field at all.

(first reported [here](https://datastax.slack.com/archives/C06P0MPSKLZ/p1731084786393449)).

hemidactylus:
It was later clarified that if by "faulty row" one means "a row with a wrong type", then the fact that no insertions occur is by design: Tables do preemptive all-rows data validation and this is the expected behaviour.

amorton:
re-open as we are discussing

amorton:
moved out of the dec-hotfix because it is not a blocking problem, and though small the change is deep and I want to make sure we get it correct.

amorton:
moved to To-Do because we are working on it. 

@Yuqi-Du  I think the fix is like you have, but simplier. 

When the processing is sequential the only test is that we fail if the current attempt is in error state.

PR #2193, opened 2025-09-16
[Tables] insertMany ordered as true, middle faulty row should not fail the whole batch.

Yuqi-Du:
What this PR does:
When there is a faulty row in the entire batch of insertMany, and ordered as true.
Every row before the faulty row should succeed, and all the faulty row errors should propagate to the result.

Which issue(s) this PR fixes:
fixes: https://github.com/stargate/data-api/issues/1672

**Checklist**
- [x] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/service/operation/tasks/TaskGroup.java
```diff
@@ -81,17 +81,50 @@ public void throwIfNotAllTerminal() {
   }
 
   /**
-   * Checks if, given the config of the container and the current state of the tasks, the container
-   * should fail fast and stop processing any further tasks.
+   * Determines whether the given {@code targetTask} should fail fast based on the container's
```
amorton, line 84: nit: replace container with TaskGroup

amorton (changes requested):
We do not have stict ordering guarantees in the TaskGroup, and the predicate for finding the tasks before the current task can be stricter.

PR #2206, opened 2025-09-25
Sample program for YAML file patching abilities

Hazel-Datastax:
**What this PR does**:
Sample program for YAML file patching abilities

Fixes #NA

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/util/YamlMerger.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/util/YamlMerger.java
@@ -8,7 +8,6 @@ import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.fasterxml.jackson.databind.node.POJONode;
 import com.fasterxml.jackson.databind.node.ValueNode;
 import com.fasterxml.jackson.dataformat.yaml.YAMLFactory;
-import com.fasterxml.jackson.datatype.jdk8.Jdk8Module;
 import java.io.IOException;
 import java.io.InputStream;
 import java.util.Iterator;
@@ -22,8 +21,6 @@ import java.util.Map;
  *   <li>Arrays: replaced entirely when present in patch
  *   <li>Scalars (string/number/boolean/null): replaced by patch value
  * </ul>
- *
- * This class is independent and can be used without other framework pieces.
  */
 public final class YamlMerger {
 
@@ -31,8 +28,6 @@ public final class YamlMerger {
 
   public YamlMerger() {
     this.yamlMapper = new ObjectMapper(new YAMLFactory());
-    // Register commonly used modules; harmless if not needed
-    this.yamlMapper.registerModule(new Jdk8Module());
   }
 
   /** Merge two YAML strings and return the merged YAML string. */
@@ -90,7 +85,7 @@ public final class YamlMerger {
 
     // If both are objects, merge field-by-field
     if (base.isObject() && patch.isObject()) {
-      ObjectNode result = ((ObjectNode) base.deepCopy());
+      ObjectNode result = base.deepCopy();
       Iterator<Map.Entry<String, JsonNode>> fields = patch.fields();
       while (fields.hasNext()) {
         Map.Entry<String, JsonNode> entry = fields.next();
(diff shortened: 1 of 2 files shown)


src/main/java/io/stargate/sgv2/jsonapi/util/YamlMerger.java
```diff
@@ -0,0 +1,139 @@
...
+ *   <li>Arrays: replaced entirely when present in patch
+ *   <li>Scalars (string/number/boolean/null): replaced by patch value
+ * </ul>
+ */
+public final class YamlMerger {
+
+  private final ObjectMapper yamlMapper;
+
+  public YamlMerger() {
+    this.yamlMapper = new ObjectMapper(new YAMLFactory());
```
tatu-at-datastax, line 30: Works, but there's convenience type `YAMLMapper` to use ("this.yamlMapper = new YAMLMapper()")

src/main/java/io/stargate/sgv2/jsonapi/util/YamlMerger.java
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  private JsonNode deepCopy(JsonNode node) {
+    if (node == null) {
+      return null;
+    }
+    if (node.isObject() || node.isArray()) {
+      return node.deepCopy();
+    }
+    if (node instanceof ValueNode) {
```
amorton, line 99: this feels like a future bug, when if the caller changed the vlue of the returned ValueNode ? we shoud not return any of the same objects.

src/main/java/io/stargate/sgv2/jsonapi/util/YamlMerger.java
```diff
@@ -0,0 +1,105 @@
...
+    // If both are arrays, replace entirely with patch (no element-wise merging)
+    if (base.isArray() && patch.isArray()) {
+      return ((ArrayNode) patch).deepCopy();
+    }
+
+    // Otherwise scalars or differing types: patch overrides
+    return deepCopy(patch);
+  }
+
+  private JsonNode deepCopy(JsonNode node) {
```
amorton, line 92: this can be cleaner using a modern switch on node. 

case for null, case for ValueNode, and default is call deepCopy()

src/test/java/io/stargate/sgv2/jsonapi/util/YamlMergerTest.java
```diff
@@ -0,0 +1,416 @@
...
+import com.fasterxml.jackson.dataformat.yaml.YAMLFactory;
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+class YamlMergerTest {
```
amorton, line 17: we should simplify these tests to just be about testing the merging without the complexity of the embedding config or other faked examples or reading other config files. 

so very simple tests that make it easy to see the yaml changes. 

will need to load the yaml into a JsonNode in this test class because of other changes on the YamlMerger - see comments there.

amorton (approved):
ok, lets make some changes to this is more a basic JsonMerger. 

We will sync on the next ticket to handle the Yaml merging

amorton (changes requested):
hit wrong button, we want to make changes before merging

src/main/java/io/stargate/sgv2/jsonapi/util/YamlMerger.java
```diff
@@ -0,0 +1,139 @@
...
+  public String toYaml(JsonNode node) {
+    try {
+      return yamlMapper.writeValueAsString(node);
+    } catch (JsonProcessingException e) {
+      throw new IllegalArgumentException("Failed to write YAML", e);
+    }
+  }
+
+  /** Core merge logic following the documented semantics. */
+  public JsonNode mergeNodes(JsonNode base, JsonNode patch) {
```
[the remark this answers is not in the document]
tatu-at-datastax: I think use JSON Patch would not help here -- it's standard for defining stand-alone patch definitions to use for merging changes (diffs) across documents, bit like unix `diff` and `patch` commands. But it would not help here where task is updating one document with another
(not sure I am explaining this well).

Renaming of second document as `overrides` or something (instead of `patch`) would help understand the difference.

src/main/java/io/stargate/sgv2/jsonapi/util/YamlMerger.java
tatu-at-datastax: `ValueNode`s are fully immutable, cannot be changed and meant to be shared/reused.
(I agree with not sharing mutable `ArrayNode`s and `ObjectNode`s of course)

PR #2234, opened 2025-10-23
[Issue 2112]  do NOT allow create index for scalar column with index function

### LINKED ISSUE #2112: [Tables] Preferred representation for indexes is the shorthand for scalar columns, the $values one for lists/sets
Regardless of whether the shorthand `"column": name` or the expanded form `"column": {name: "$values"}` is used, indexes created for scalars are then described (by listIndexes) using the short form, while those on maps and sets are described using the longer form.

This seems unnecessarily surprising to the user.

## Reproduction

Given the following table:

```
{
    "createTable": {
        "definition": {
            "columns": {
                "scalar_i": {"type": "int"},
                "scalar_t": {"type": "text"},
                "list_i": {
                    "type": "list",
                    "valueType": "int"
                },
                "list_t": {
                    "type": "list",
                    "valueType": "text"
                },
                "set_i": {
                    "type": "set",
                    "valueType": "int"
                },
                "set_t": {
                    "type": "set",
                    "valueType": "text"
                },
                "map_e_ii": {
                    "type": "map",
                    "keyType": "int",
                    "valueType": "int"
                },
                "map_e_it": {
                    "type": "map",
                    "keyType": "int",
                    "valueType": "text"
                },
                "map_e_ti": {
                    "type": "map",
                    "keyType": "text",
                    "valueType": "int"
                },
                "map_e_tt": {
                    "type": "map",
                    "keyType": "text",
                    "valueType": "text"
                },
                "map_k_ii": {
                    "type": "map",
                    "keyType": "int",
                    "valueType": "int"
                },
                "map_k_it": {
                    "type": "map",
                    "keyType": "int",
                    "valueType": "text"
                },
                "map_k_ti": {
                    "type": "map",
                    "keyType": "text",
                    "valueType": "int"
                },
                "map_k_tt": {
                    "type": "map",
                    "keyType": "text",
                    "valueType": "text"
                },
                "map_v_ii": {
                    "type": "map",
                    "keyType": "int",
                    "valueType": "int"
                },
                "map_v_it": {
                    "type": "map",
                    "keyType": "int",
                    "valueType": "text"
                },
                "map_v_ti": {
                    "type": "map",
                    "keyType": "text",
                    "valueType": "int"
                },
                "map_v_tt": {
                    "type": "map",
                    "keyType": "text",
                    "valueType": "text"
                }
            },
            "primaryKey": {
                "partitionBy": ["scalar_i"],
                "partitionSort": {}
            }
        }
    }
}
```

create two indexes using either the long or short form (both are reported in the payloads here -- choose one as you prefer).

```
{
    "createIndex": {
        "name": "idx_scalar_t",
        "definition": {
            "column": "scalar_t"
            // OR ALTERNATIVELY "column": {"scalar_t": "$values"}
        }
    }
}
```

and

```
{
    "createIndex": {
        "name": "idx_list_t",
        "definition": {
            "column": "list_t"
            // OR ALTERNATIVELY "column": {"list_t": "$values"}
        }
    }
}
```

Now run a `listIndexes(explain=true)`. The result has the column specifications always as follows (one short, one long):

```
{
    "status": {
        "indexes": [
            {
                "name": "idx_list_t",
                "definition": {
                    "column": {
                        "list_t": "$values"
                    },
                    "options": {}
                },
                "indexType": "regular"
            },
            {
                "name": "idx_scalar_t",
                "definition": {
                    "column": "scalar_t",
                    "options": {
                        "ascii": false,
                        "caseSensitive": true,
                        "normalize": false
                    }
                },
                "indexType": "regular"
            }
        ]
    }
}
```

Yuqi-Du:
This PR will add a guardrail to not allow index creation for scalar column with an index function.

E.G. If we have scalar column 'scalar_t'.

Valid: plain column string value
```json
{
    "createIndex": {
        "name": "idx_scalar_t",
        "definition": {
            "column": "scalar_t"
        }
    }
}
```

Invalid: $keys/$values indexFuntion
```json
{
    "createIndex": {
        "name": "idx_scalar_t",
        "definition": {
            "column": {"scalar_t": "$values"}
        }
    }
}
```

Fixes #2112 

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -293,6 +293,12 @@
       <scope>test</scope>
       <version>${wiremock.version}</version>
     </dependency>
+    <dependency>
+      <groupId>uk.org.webcompere</groupId>
+      <artifactId>system-stubs-jupiter</artifactId>
+      <version>2.1.8</version>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
   <build>
     <plugins>
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/CqlSessionFactory.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/CqlSessionFactory.java
@@ -5,7 +5,11 @@ import com.datastax.oss.driver.api.core.CqlSessionBuilder;
 import com.datastax.oss.driver.api.core.config.DefaultDriverOption;
 import com.datastax.oss.driver.api.core.config.DriverConfigLoader;
 import com.datastax.oss.driver.api.core.metadata.schema.SchemaChangeListener;
+import com.datastax.oss.driver.internal.core.config.typesafe.DefaultDriverConfigLoader;
 import com.google.common.annotations.VisibleForTesting;
+import com.typesafe.config.Config;
+import com.typesafe.config.ConfigFactory;
+import com.typesafe.config.ConfigRenderOptions;
 import io.stargate.sgv2.jsonapi.config.DatabaseType;
 import io.stargate.sgv2.jsonapi.service.cqldriver.executor.optvector.SubtypeOnlyFloatVectorToArrayCodec;
 import java.net.InetSocketAddress;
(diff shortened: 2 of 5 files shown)


src/main/java/io/stargate/sgv2/jsonapi/service/schema/tables/ApiRegularIndex.java
```diff
@@ -122,6 +122,10 @@ public ApiRegularIndex create(
 
       // create ApiRegularIndex for the target primitive column
       if (apiColumnDef.type().isPrimitive()) {
+        if (mapComponentDesc != null) {
+          // if user specified $keys/$values for a scalar primitive column, error out
+          throw SchemaException.Code.INVALID_FORMAT_FOR_INDEX_CREATION_COLUMN.get();
```
tatu-at-datastax: Would be nice to include name of index that has invalid definition.

tatu-at-datastax (approved):
LGTM - one suggestion if easy enough to do.

hemidactylus:
LGTM, thank you for this!

src/main/java/io/stargate/sgv2/jsonapi/service/schema/tables/ApiRegularIndex.java
Yuqi-Du: Thanks!
This will only fail for createIndexCommand, and it is only one index that been created.
Since our deserializer also uses this error and it does not have the context, so appending an index name is hard for the deserializer part.

I will see if I can put accurate info into the error msg though!

tatu-at-datastax: Ok that is true: index is known in context.

PR #2239, opened 2025-10-28
Fix #1883: enable Tables feature by default

### LINKED ISSUE #1883: Default enable the tables feature flag
Default enable tables feature flag in data-api so the below override flag is not needed on startup. 
`-Dstargate.feature.flags.tables=true`

tatu-at-datastax:
**What this PR does**:

Changes defaults for Tables feature to enabled (was disabled)

Fixes #1883

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/docker-compose/.env
+++ b/docker-compose/.env
@@ -3,7 +3,7 @@ LOGLEVEL=INFO
 REQUESTLOG=false
 DATAAPITAG=v1
 DATAAPIIMAGE=stargateio/jsonapi
-DSETAG=6.9.14
+DSETAG=6.9.15
 DSEIMAGE="cr.dtsx.io/datastax/dse-server"
 HCDTAG=1.2.3
 # 18-Aug-2025, tatu: latest official release; staging may be used pre-release
--- a/docker-compose/dse.yaml
+++ b/docker-compose/dse.yaml
@@ -1,4 +1,4 @@
-# DSE Config Version: 6.9.14
+# DSE Config Version: 6.9.15
 
 # Memory limit for DSE In-Memory tables as a fraction of system memory. When not set,
 # the default is 0.2 (20% of system memory).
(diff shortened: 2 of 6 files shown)


src/test/java/io/stargate/sgv2/jsonapi/api/v1/tables/TableFeatureDisabledIntegrationTest.java
```diff
@@ -29,53 +28,33 @@ public TestResource() {}
...
+  // With explicit config disable, header cannot override (config takes precedence)
   @Order(2)
   @Test
-  public void okCreateWithFeatureEnabledViaHeader() {
+  public void failCreateEvenWithHeaderWhenExplicitlyDisabled() {
     assertNamespaceCommand(keyspaceName)
         .header(ApiFeature.TABLES.httpHeaderName(), "true")
         .postCreateTable(simpleTableDef(TABLE_TO_CREATE))
-        .wasSuccessful();
-  }
```
Hazel-Datastax: nit: maybe we have a new error code for this case? I understand if we disable it in the config(maybe due to the underlying infrastructure not supporting?), the user cannot override. But it seems weird to user that they are trying to enable but get `TABLE_FEATURE_NOT_ENABLED` response. Considering this is a rare case and we may never disable it, I'm fine with this error.

Just curious, you said `With explicit config disable, header cannot override (config takes precedence)`. Our current prod config (in argocd) is explicitly enabled. That means with this change, the user cannot disable it (even using Feature-Flag-Tables=false in the header). I know it's a strange and rare case, just raise and can be ignored :)

Hazel-Datastax (approved):
LGTM

src/test/java/io/stargate/sgv2/jsonapi/api/v1/tables/TableFeatureDisabledIntegrationTest.java
tatu-at-datastax: Hmmmh. I can see the potential confusion. The ability to use headers to override/enable feature is mostly meant for testing/early-access phase. But I think the main/only real usage is to enable something dynamically, not to disable.

So I think it is fine not to allow disabling. But maybe like you said, we should not even accept "false" case.

Let me think about this a bit.

PR #2244, opened 2025-11-04
Evict session in cache when encountering AllNodesFailedException

### LINKED ISSUE #2246: CQLSessionCache: Failure to evict stale sessions on `AllNodesFailedException` prevents automatic recovery from cluster restarts
# TL;DR

# 1. Problem Description

# 4. Proposed Solution

Hazel-Datastax:
**What this PR does**:
Evict session in cache when encountering AllNodesFailedException

Fixes #2246 

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -8,8 +8,9 @@
     <!-- Quarkus -->
     <quarkus.platform.artifact-id>quarkus-bom</quarkus.platform.artifact-id>
     <quarkus.platform.group-id>io.quarkus.platform</quarkus.platform.group-id>
-    <!-- 14-oct-2025 , aaron: Updated to latest Quarkus for security fixes in netty -->
-    <quarkus.platform.version>3.28.3</quarkus.platform.version>
+    <!-- 14-oct-2025, aaron: Updated to latest Quarkus for security fixes in netty -->
+    <!-- 10-nov-2025, tatu: Updated to later Quarkus (issues with 3.29.x tho -->
+    <quarkus.platform.version>3.28.5</quarkus.platform.version>
     <!-- miscellaneous -->
     <wiremock.version>3.4.2</wiremock.version>
     <!-- 17-Oct-2025, tatu: Although quarkus-bom provides for Mockito version,
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/CqlCredentials.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/CqlCredentials.java
@@ -74,11 +74,11 @@ public interface CqlCredentials {
     @Override
     public String toString() {
       // Don't log the full token, just the first 4 chars
+      final String tokenDesc = (token.length() <= 4) ? token : (token.substring(0, 4) + "...");
       return new StringBuilder("TokenCredentials{")
           .append("token='")
-          .append(token.substring(0, 4))
-          .append("...'")
-          .append(", isAnonymous=")
+          .append(tokenDesc)
+          .append("', isAnonymous=")
           .append(isAnonymous())
           .append('}')
           .toString();
@@ -120,20 +120,21 @@ public interface CqlCredentials {
     }
 
     static UsernamePasswordCredentials fromToken(String encodedCredentials) {
-      String[] parts = encodedCredentials.split(":");
-      if (parts.length != 3) {
-        throw new UnauthorizedException(
-            "Invalid credentials format, expected `Cassandra:Base64(username):Base64(password)`");
+      if (encodedCredentials.startsWith(USERNAME_PASSWORD_TOKEN_PREFIX)) {
+        String[] parts = encodedCredentials.split(":");
+        if (parts.length == 3) {
+          try {
+            String userName = new String(Base64.getDecoder().decode(parts[1]));
+            String password = new String(Base64.getDecoder().decode(parts[2]));
+            return new UsernamePasswordCredentials(userName, password);
+          } catch (IllegalArgumentException e) {
+            ; // fall down to failure case
+          }
+        }
       }
-
-      try {
-        String userName = new String(Base64.getDecoder().decode(parts[1]));
-        String password = new String(Base64.getDecoder().decode(parts[2]));
-        return new UsernamePasswordCredentials(userName, password);
-      } catch (IllegalArgumentException e) {
-        throw new UnauthorizedException(
-            "Invalid credentials format, expected `Cassandra:Base64(username):Base64(password)`");
-      }
+      throw new UnauthorizedException(
+          "Invalid credentials format, expected `%sBase64(username):Base64(password)`"
+              .formatted(USERNAME_PASSWORD_TOKEN_PREFIX));
     }
   }
 }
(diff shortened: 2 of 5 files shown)


src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/CQLSessionCache.java
```diff
@@ -213,6 +213,41 @@ public CqlSession getSession(String tenantId, String authToken, String userAgent
...
+  /**
+   * Evicts a session from the cache based on the provided {@link RequestContext}.
+   *
+   * @see #evictSession(String, String, String) for details on eviction.
+   * @param requestContext The request context containing tenant, auth, and user agent info.
+   */
+  public void evictSession(RequestContext requestContext) {
+    Objects.requireNonNull(requestContext, "requestContext must not be null for eviction");
+
+    // Validation happens when creating the credentials and session key
```
tatu-at-datastax: If this means missing values cannot happen, should a simple `get()` instead of `orElse()?

src/test/java/io/stargate/sgv2/jsonapi/service/cqldriver/CqlSessionCacheTests.java
```diff
@@ -582,23 +695,6 @@ private Fixture newFixture() {
...
-      boolean setExpected) {
-    return newFixture(
-        databaseType,
-        consumers,
-        cacheTTL,
-        cacheSize,
-        SLA_USER_AGENT,
-        SLA_USER_TTL,
-        false,
-        setExpected);
```
Hazel-Datastax: No usage, remove

src/main/java/io/stargate/sgv2/jsonapi/service/operation/OperationAttempt.java
```diff
@@ -127,6 +128,7 @@ protected SubT downcast() {
    *
    * @param queryExecutor The {@link CommandQueryExecutor} to use for executing the query, this
    *     handles interacting with the driver.
+   * @param commandContext The {@link CommandContext} to provide context for the operation attempt.
```
tatu-at-datastax: Usually Context objects are the first arguments to methods, maybe do that here?

src/test/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/DefaultDriverExceptionHandlerTest.java
```diff
@@ -212,6 +276,9 @@ private static Stream<TestArguments> tableDriverErrorHandledData() {
         new TestArguments(
             new UnsupportedProtocolVersionException(null, "unsupported protocol", List.of()),
             Assertions.isUnexpectedDriverException()),
+        new TestArguments(
+            allFailedTwoNodesOneWriteTimeout(),
+            Assertions.of(DatabaseException.Code.TIMEOUT_WRITING_DATA)),
```
Hazel-Datastax: I found that `allFailedTwoNodesOneWriteTimeout` has no usage, maybe forgot to add it in the test case. So I add one here

src/test/java/io/stargate/sgv2/jsonapi/service/cqldriver/CqlSessionCacheTests.java
```diff
@@ -512,6 +514,117 @@ public void expireKeepsSessionAfterNonSLAUserRead() {
...
+            false,
+            true);
+
+    // Verify cache is empty
+    assertThat(fixture.cache.peekSession(TENANT_ID, AUTH_TOKEN, null))
+        .as("Cache is empty before eviction")
+        .isNotPresent();
+
+    // Evict a non-existent session - should not throw
+    fixture.cache.evictSession(TENANT_ID, AUTH_TOKEN, null);
```
tatu-at-datastax: If method can be made to return `boolean` can verify `false` being returned.

src/test/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/DefaultDriverExceptionHandlerTest.java
```diff
@@ -45,6 +49,66 @@ public class DefaultDriverExceptionHandlerTest {
...
+    verify(TEST_DATA.SESSION_CACHE, times(1)).evictSession(TEST_DATA.REQUEST_CONTEXT);
+    verifyNoMoreInteractions(TEST_DATA.SESSION_CACHE);
+
+    // Test `tableDriverErrorHandled` will verify the exception is handled properly, will not test
+    // here
+  }
+
+  @Test
+  public void handleAllNodesFailedExceptionEvictsSessionEvenWhenEvictionFails() {
+    // Make evictSession throw an exception
```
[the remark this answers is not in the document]
Hazel-Datastax: Under most/normal cases, `evictSession` shouldn't fail; this is just in case some rare bugs in Caffeine cache or some unpredictable runtime issues

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/DefaultDriverExceptionHandler.java
```diff
@@ -463,17 +485,25 @@ public RuntimeException handle(WriteTimeoutException exception) {
    * io.stargate.sgv2.jsonapi.service.operation.tables.TableDriverExceptionHandler}
    */
   @FunctionalInterface
-  public interface Factory<T extends SchemaObject>
-      extends BiFunction<T, SimpleStatement, DriverExceptionHandler> {
+  public interface Factory<T extends SchemaObject> {
+
+    /** Signature of the constructor for the DefaultDriverExceptionHandler. */
+    DriverExceptionHandler apply(
```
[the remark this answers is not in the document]
tatu-at-datastax: `apply()` is really only used because of JDK's functional interfaces: for specific use it has no meaning... it's about same as method `do()` :) (if `do` was not a Java keyword and could be used).
So it's not used because it makes sense but because `Function` and others chose it.

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/DefaultDriverExceptionHandler.java
```diff
@@ -152,9 +165,23 @@ public RuntimeException handle(InvalidKeyspaceException exception) {
 
   @Override
   public RuntimeException handle(AllNodesFailedException exception) {
+    // Evict the session from the cache, as it's likely in a "zombie" state
```
amorton: what happens if this error is handled by any subclass ?

src/main/java/io/stargate/sgv2/jsonapi/service/operation/tasks/BaseTask.java
```diff
@@ -59,6 +59,9 @@ public abstract class BaseTask<
   /** The current result supplier, supplied by the subclass every time we try / re-try */
   private ResultSupplierT resultSupplier = null;
 
+  /** The command context for the current execution, set when execute() is called */
```
amorton: This is passed as a param to execute(), seems odd to then store it as instance state.

If it is not set in the ctor, then it adds risks that we may use it when we do not know if it has been set. So we either need to check if it is null or rely on the idea that code always runs in a particular way. 

I see we are using this because it is used when building the error handler, I think this is another sign that we need to look at the idea of passing this info into the error handler.

Hazel-Datastax:
Closed - implement it in another PR#2260

PR #2247, opened 2025-11-05
Fix #1911: allow empty `options` for createIndex, ENTRY

### LINKED ISSUE #1911: [Table] Creating an entries index on a table fails if `options: {}` passed
I am reporting this because I remember the motto that _passing empty/null objects in payloads should always have the same effects as omitting them altogether_. Not an immediate issue but still the following breaks the motto.

**summary**: When creating an `ENTRIES` index on a `MAP<...>` column with `createIndex`, providing an empty `options: {}` in the payload generates an error, whereas omitting it works just fine.

(On the contrary, when calling `listIndexes(explain=True)`, all indexes come back with their `options`, even those having `options: {}`.)

_As tested locally against `main`, commit hash `64b063664c89331e0a16656c31e58730af749726` dated Wed Mar 5 14:27:15 2025 -0800 ._

## Reproduction

Given

```
CREATE TABLE default_keyspace.t (
    id text PRIMARY KEY,
    list_int list<int>,
    map_text_int_e map<text, int>,
    map_text_int_k map<text, int>,
    map_text_int_v map<text, int>,
    set_int set<int>,
    t text
)
```

### Failing case

```
{
    "createIndex": {
        "name": "idx_t_map_text_int_e",
        "definition": {
            "column": "map_text_int_e",
            "options": {}
        }
    }
}
```

results in

```
{
    "errors": [
        {
            "message": "Index function `entries` can not apply to map column when analyze options are specified.\n\nThe command contains analyze options: ascii=null, caseSensitive=null, normalize=null.\nThe table default_keyspace.t defines the columns: id(text), list_int(UNSUPPORTED CQL type: list<int>), map_text_int_e(UNSUPPORTED CQL type: map<text, int>), map_text_int_k(UNSUPPORTED CQL type: map<text, int>), map_text_int_v(UNSUPPORTED CQL type: map<text, int>), set_int(UNSUPPORTED CQL type: set<int>), t(text).\n\nResend the command without options.",
            "scope": "SCHEMA",
            "errorCode": "CANNOT_ANALYZE_ENTRIES_ON_MAP_COLUMNS",
            "id": "99dc6a19-5570-427f-8408-e6cc3b7acc1e",
            "family": "REQUEST",
            "title": "Can not create index on map column with index function entries and specify analyze options"
        }
    ]
}

```

and no index being created.

### Successful (baseline) case

Using 

```
{
    "createIndex": {
        "name": "idx_t_map_text_int_e",
        "definition": {
            "column": "map_text_int_e"
        }
    }
}
```

returns

```
{
    "status": {
        "ok": 1
    }
}

```

and indeed CQL (describe table) tells me that

```
[...]
CREATE CUSTOM INDEX idx_t_map_text_int_e ON default_keyspace.t (entries(map_text_int_e)) USING 'StorageAttachedIndex';
[...]
```

tatu-at-datastax:
**What this PR does**:

Allows use of empty (`{ }`) "options" for `createIndex` of type "ENTRIES" (similar to missing `options`)

Fixes #1911

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/table/definition/indexes/RegularIndexDefinitionDesc.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/table/definition/indexes/RegularIndexDefinitionDesc.java
@@ -1,5 +1,6 @@
 package io.stargate.sgv2.jsonapi.api.model.command.table.definition.indexes;
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.annotation.JsonPropertyOrder;
@@ -102,6 +103,7 @@ public record RegularIndexDefinitionDesc(
           @JsonProperty(TableDescConstants.RegularIndexDefinitionDescOptions.NORMALIZE)
           Boolean normalize) {
 
+    @JsonIgnore
     public boolean isEmpty() {
       return ascii == null && caseSensitive == null && normalize == null;
     }


src/main/java/io/stargate/sgv2/jsonapi/api/model/command/table/definition/indexes/RegularIndexDefinitionDesc.java
```diff
@@ -102,6 +103,11 @@ public record RegularIndexDescOptions(
           @JsonProperty(TableDescConstants.RegularIndexDefinitionDescOptions.NORMALIZE)
           Boolean normalize) {
 
+    @JsonIgnore
```
tatu-at-datastax: Helper method we do not want to be part of JSON serialization.

PR #2250, opened 2025-11-07
Add driver config `advanced.resolve-contact-points = false`

### LINKED ISSUE #2246: CQLSessionCache: Failure to evict stale sessions on `AllNodesFailedException` prevents automatic recovery from cluster restarts

Hazel-Datastax:
**What this PR does**:
As detailed in issue #2246, the Data API can fail to recover after its backend Cassandra cluster's pods are recycled. The application-level fix is in PR #2244. This PR just adds a driver config to make the initial connection process more robust.

**Why this Change is Important**
By default, the Java driver resolves a contact point's hostname to an IP address once and then caches that IP internally for all subsequent connection attempts during the session bootstrap.

In a dynamic environment like Kubernetes, where service IPs can change (e.g., during pod restarts or network events), this default behavior can cause the driver to get "stuck" trying to connect to a stale, non-existent IP addres

Setting `resolve-contact-points = false` instructs the driver to re-resolve the DNS name for each connection attempt during the bootstrap phase.

Fixes #2246

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/src/main/resources/application.conf
+++ b/src/main/resources/application.conf
@@ -52,6 +52,7 @@ datastax-java-driver {
   }
   basic.request.timeout = 20 seconds
 
+  # see https://github.com/stargate/data-api/pull/2250 description
   advanced.resolve-contact-points = false
 
 


tatu-at-datastax (approved):
LGTM: could use a comment.

PR #2260, opened 2025-11-17
Evict session in cache when encountering AllNodesFailedException

### LINKED ISSUE #2246: CQLSessionCache: Failure to evict stale sessions on `AllNodesFailedException` prevents automatic recovery from cluster restarts

Hazel-Datastax:
**What this PR does**:
Evict session in cache when encountering AllNodesFailedException

Fixes #2246

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/src/test/java/io/stargate/sgv2/jsonapi/exception/APIExceptionTest.java
+++ b/src/test/java/io/stargate/sgv2/jsonapi/exception/APIExceptionTest.java
@@ -74,7 +74,7 @@ public class APIExceptionTest extends ConfiguredErrorTest {
   }
 
   @Test
-  public void defaultExceptionActionsFieldEmpty() {
+  public void defaultExceptionFlagsFieldEmpty() {
     var exception =
         TestScopeException.Code.SCOPED_REQUEST_ERROR.get(
             "name", TEST_DATA.VAR_NAME,
@@ -87,7 +87,7 @@ public class APIExceptionTest extends ConfiguredErrorTest {
   }
 
   @Test
-  public void exceptionActionsSetViaConstructor() {
+  public void exceptionFlagsSetViaConstructor() {
     var exception =
         TestScopeException.Code.SCOPED_REQUEST_ERROR.get(
             EnumSet.of(ExceptionFlags.UNRELIABLE_DB_SESSION),
--- a/src/test/java/io/stargate/sgv2/jsonapi/exception/ErrorCodeTest.java
+++ b/src/test/java/io/stargate/sgv2/jsonapi/exception/ErrorCodeTest.java
@@ -54,30 +54,30 @@ public class ErrorCodeTest extends ConfiguredErrorTest {
   }
 
   @Test
-  public void getWithExceptionActions() {
+  public void getWithExceptionFlags() {
     var errorCode = TestScopeException.Code.SCOPED_REQUEST_ERROR;
-    var actions = EnumSet.of(UNRELIABLE_DB_SESSION);
+    var flags = EnumSet.of(UNRELIABLE_DB_SESSION);
 
     var errorFromParamArgs =
         assertDoesNotThrow(
-            () -> errorCode.get(actions, "name", TEST_DATA.VAR_NAME, "value", TEST_DATA.VAR_VALUE));
+            () -> errorCode.get(flags, "name", TEST_DATA.VAR_NAME, "value", TEST_DATA.VAR_VALUE));
 
     var errorFromMap =
         assertDoesNotThrow(
             () ->
                 errorCode.get(
-                    actions,
+                    flags,
                     Map.of(
                         "name", TEST_DATA.VAR_NAME,
                         "value", TEST_DATA.VAR_VALUE)),
             String.format("Creating exception with template %s", errorCode.template()));
 
     assertThat(errorFromParamArgs.exceptionFlags)
-        .as("Exception created with param args carries the given actions")
-        .isEqualTo(actions);
+        .as("Exception created with param args carries the given flags")
+        .isEqualTo(flags);
     assertThat(errorFromMap.exceptionFlags)
-        .as("Exception created with map carries the given actions")
-        .isEqualTo(actions);
+        .as("Exception created with map carries the given flags")
+        .isEqualTo(flags);
   }
 
   @Test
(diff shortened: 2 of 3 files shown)


src/main/java/io/stargate/sgv2/jsonapi/exception/APIException.java
```diff
@@ -97,11 +106,22 @@ public APIException(ErrorInstance errorInstance) {
     this.body = errorInstance.body();
     Objects.requireNonNull(errorInstance.httpStatusOverride(), "httpStatusOverride cannot be null");
     this.httpStatus = errorInstance.httpStatusOverride().orElse(DEFAULT_HTTP_STATUS);
+    Objects.requireNonNull(errorInstance.exceptionActions(), "exceptionActions cannot be null");
+    this.exceptionActions = errorInstance.exceptionActions();
```
tatu-at-datastax: I think you can combine this like:

   this.exceptionActions = Objects.requireNonNull(errorInstance.exceptionActions(), "....");

src/main/java/io/stargate/sgv2/jsonapi/exception/APIException.java
```diff
@@ -97,11 +106,22 @@ public APIException(ErrorInstance errorInstance) {
...
   }
 
   public APIException(
```
tatu-at-datastax: I know Javadoc comment was already missing, but if this is just for testing, maybe add a comment stating that? Otherwise it seems odd to have delegation working in this direction (from individual arguments to wrapper object)

tatu-at-datastax (approved):
LGTM: had a few questions but seems legit.

src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorTemplate.java
```diff
@@ -56,7 +54,8 @@ public record ErrorTemplate<T extends APIException>(
     String code,
     String title,
     String messageTemplate,
-    Optional<Integer> httpStatusOverride) {
+    Optional<Integer> httpStatusOverride,
+    EnumSet<ExceptionAction> exceptionActions) {
```
amorton: rename exceptionFlags

and see below, how should these flags be used differenttly to the ones passed in toException ?

src/main/java/io/stargate/sgv2/jsonapi/exception/ExceptionAction.java
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * <p>These actions are populated by the code that creates the exception (for example {@link
+ * io.stargate.sgv2.jsonapi.service.cqldriver.executor.DefaultDriverExceptionHandler}) and
+ * interpreted when the exception is mapped to a response. This keeps remediation logic close to
+ * where the error originates while still allowing generic handling at the API layer.
+ *
+ * @see io.stargate.sgv2.jsonapi.service.processor.CommandProcessor#handleExceptionActions
+ * @see io.stargate.sgv2.jsonapi.service.cqldriver.executor.DefaultDriverExceptionHandler
+ */
+public enum ExceptionAction {
```
amorton: rename to ExceptionFlags

src/main/java/io/stargate/sgv2/jsonapi/exception/ExceptionAction.java
```diff
@@ -0,0 +1,25 @@
...
+  /**
+   * Evict the session from cache. This is intended for use in scenarios where a session is known to
+   * be in an unrecoverable state (e.g., after all cluster nodes restart) and needs to be forcibly
+   * removed to allow for a fresh connection on the next request.
+   */
+  EVICT_SESSION_CACHE,
```
amorton: nit: lets tweak the name, currently the name and the comments talk about what the outcome we want not the thing that has happened. 

e.g.. 

UNRELIABLE_DB_SESSION - the error indicates that the DB session used is no longer reliable and should be terminated

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/CQLSessionCache.java
```diff
@@ -213,6 +213,46 @@ public CqlSession getSession(String tenantId, String authToken, String userAgent
...
+   * @param userAgent Nullable user agent
+   * @return {@code true} if a session was evicted, {@code false} otherwise.
+   */
+  public boolean evictSession(String tenantId, String authToken, String userAgent) {
+
+    var cacheKey = createCacheKey(tenantId, authToken, userAgent);
+
+    // Invalidate the key. This will trigger the onKeyRemoved listener,
+    // which will close the CqlSession and run other cleanup.
+    if (sessionCache.asMap().remove(cacheKey) != null) {
```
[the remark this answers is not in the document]
amorton: +1 and you can link to the stack overflow thread for why

amorton (changes requested):

src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorCode.java
```diff
@@ -67,6 +77,10 @@ default T get(Map<String, String> values) {
     return template().toException(values);
   }
 
+  default T get(EnumSet<ExceptionFlags> exceptionFlags, Map<String, String> values) {
```
amorton: nit: have `get(Map<String, String> values)` call this overload

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/DefaultDriverExceptionHandler.java
```diff
@@ -162,7 +165,8 @@ public RuntimeException handle(AllNodesFailedException exception) {
       case DriverException e -> maybeHandle(e);
         // this is a non-driver based exception, so map to generic unexpected driver error
       case RuntimeException re ->
-          DatabaseException.Code.UNEXPECTED_DRIVER_ERROR.get(errVars(schemaObject, re));
+          DatabaseException.Code.UNEXPECTED_DRIVER_ERROR.get(
+              EnumSet.of(UNRELIABLE_DB_SESSION), errVars(schemaObject, re));
```
amorton: mmmm, not sure but I dont think it will hurt anything too much

src/test/java/io/stargate/sgv2/jsonapi/exception/APIExceptionTest.java
```diff
@@ -71,4 +72,42 @@ public void toStringFormatted() {
         .contains(exception.title)
         .contains(exception.body);
   }
+
+  @Test
+  public void defaultExceptionActionsFieldEmpty() {
```
amorton: nit: defaultExceptionActionsFieldEmpty to defaultExceptionFlagsFieldEmpty

amorton (approved):
looks good, some nit picks otherwise ship it

PR #2279, opened 2025-12-05
Convert `ErrorV1.SERVER_INTERNAL_ERROR` and usage to new errors

### LINKED ISSUE #2278: Convert `ErrorV1.SERVER_INTERNAL_ERROR` to new ("ErrorsV2") variant

tatu-at-datastax:
**What this PR does**:

Converts `ErrorV1.SERVER_INTERNAL_ERROR` and usage to new `ServerException.Code.INTERNAL_SERVER_ERROR`

Fixes #2278

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/src/test/java/io/stargate/sgv2/jsonapi/service/cqldriver/CqlCredentialsFactoryTests.java
+++ b/src/test/java/io/stargate/sgv2/jsonapi/service/cqldriver/CqlCredentialsFactoryTests.java
@@ -8,7 +8,7 @@ import static org.mockito.Mockito.*;
 import com.datastax.oss.driver.api.core.CqlSessionBuilder;
 import io.quarkus.security.UnauthorizedException;
 import io.stargate.sgv2.jsonapi.config.DatabaseType;
-import io.stargate.sgv2.jsonapi.exception.JsonApiException;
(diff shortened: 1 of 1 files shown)


tatu-at-datastax:
NOTE: reduction on code coverage is not meaningful -- call sequence for uncovered path slightly bigger at bytecode level and we do not test most internal error paths (by definition these should never occur).

src/main/resources/errors.yaml
```diff
@@ -1750,6 +1750,16 @@ request-errors:
...
+  - scope:
+    code: INTERNAL_SERVER_ERROR
+    title: Internal server error
+    body: |-
+      An unexpected internal server error occurred while processing the request. 
+      
+      Error Message: ${errorMessage}
+      
+      ${SNIPPET.RETRY_UNKNOWN}
+
```
Hazel-Datastax: nit: The error message `An unexpected internal server error occurred while processing the request. ` looks very similar to `UNEXPECTED_SERVER_ERROR` and hard to differentiate - makes me think why we need two error codes. Maybe Sarah can have some inputs on the messages?

Also, maybe include `Error Class: ${errorClass}` as well so it would be easier for us to debug?

PR #2307, opened 2025-12-15
SchemaObjectCache refactor

### LINKED ISSUE #1791: Refactor the NamespaceCache and SchemaCache classes
These classes need a large factor to bring the code quality up and make them better handle collections and tables. 

They will then be better at returning errors we need such as in #1769

Yuqi-Du:
**What this PR does**:

# Major change
- Change schemaCache to schemaObjectCache(two level cache -> one level cache), extended as TTL cache
- identified by new SchemaObjectIdentifier as key. 
- Cache access change to async non-blocking style.

# Notable points
- delete files (TableBaseSchemaCache.java and SchemaCache.java), so the old two level cache files
- need to break the circular dependency for both CqlSessionCacheSupplier.java and SchemaObjectCacheSupplier.java
- null handling for possible keyspace and table, since schemaObjectIdentifier are used across DB/Keyspace/Table command.
- update all tests, modify mocks for schemaObject and metadata, use schemaObjectIdentifier instead of names

Fixes #1791 

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/Command.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/api/model/command/Command.java
@@ -1,5 +1,6 @@
 package io.stargate.sgv2.jsonapi.api.model.command;
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonSubTypes;
 import com.fasterxml.jackson.annotation.JsonTypeInfo;
 import io.stargate.sgv2.jsonapi.metrics.CommandFeatures;
@@ -37,6 +38,8 @@ public interface Command {
    */
   CommandName commandName();
 
+  // there are places where we serialize the command, and this looks like a bean property, so ignore
+  @JsonIgnore
   default boolean isForceSchemaRefresh() {
     return commandName().getCommandType().isForceSchemaRefresh();
   }


src/main/java/io/stargate/sgv2/jsonapi/api/model/command/CommandContext.java
```diff
@@ -47,11 +57,15 @@ public class CommandContext<SchemaT extends SchemaObject> {
...
   private final SchemaT schemaObject;
+  private final RequestTracing requestTracing;
+  private final RequestContext requestContext;
   private final EmbeddingProvider
       embeddingProvider; // to be removed later, this is a single provider
   private final String commandName; // TODO: remove the command name, but it is used in 14 places
-  private final RequestContext requestContext;
-  private RequestTracing requestTracing;
+
+  // both per request list of objects that want to update the logging MDC context,
```
tatu-at-datastax: both "per request..." and ?

src/main/java/io/stargate/sgv2/jsonapi/service/schema/SchemaObjectIdentifier.java
```diff
@@ -0,0 +1,228 @@
...
+  private SchemaObjectIdentifier(
+      SchemaObjectType type, Tenant tenant, CqlIdentifier keyspace, CqlIdentifier table) {
+
+    this.type = type;
+    this.tenant = tenant;
+    this.keyspace = keyspace;
+    this.table = table;
+
+    // this is used internally, so using internal name for the type not apiName()
+    this.fullName =
```
tatu-at-datastax: Doing this eagerly can be bit expensive if `fullName` is not used for every id instance (if and when we create lots of throw-away instances). But probably not worth worrying about unless it shows in profiler.

tatu-at-datastax (approved):
I have to admit I did not get super deep, but I did skim through it all and for what it is worth nothing seemed off.
Added some notes, fixed couple of typos, and I think it's ready to go.

PR #2400, opened 2026-02-18
Refactor `createKeyspace` command

Hazel-Datastax:
**What this PR does**:
While working on the MCP support, I found the createKeyspace` command is confusing. This PR is trying to make it clear.

Fixes NA

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,18 @@
 # Changelog
 
+## [v1.0.39](https://github.com/stargate/data-api/tree/v1.0.39) (2026-02-17)
+
+[Full Changelog](https://github.com/stargate/data-api/compare/v1.0.38...v1.0.39)
+
+**Merged pull requests:**
+
+- Refactor `MeteredEmbeddingProvider` to be only wrapper, not sub-class, renamed as `MeteredEmbeddingProviderWrapper` [\#2397](https://github.com/stargate/data-api/pull/2397) ([tatu-at-datastax](https://github.com/tatu-at-datastax))
+- Fix Docker/test-containers version problem by raising client version from 1.32 to 1.44 [\#2395](https://github.com/stargate/data-api/pull/2395) ([tatu-at-datastax](https://github.com/tatu-at-datastax))
+- Fix \#2392: `EmbeddingProvider` and `RerankingProvider` `decideRetry\(\)` buggy [\#2393](https://github.com/stargate/data-api/pull/2393) ([tatu-at-datastax](https://github.com/tatu-at-datastax))
+- Fix \#2378 use DocumentId when checking replacement [\#2390](https://github.com/stargate/data-api/pull/2390) ([amorton](https://github.com/amorton))
+- Minor IDE warnings clean up [\#2389](https://github.com/stargate/data-api/pull/2389) ([tatu-at-datastax](https://github.com/tatu-at-datastax))
+- Fix \#2336: update DSE 6.9 to 6.9.18 [\#2388](https://github.com/stargate/data-api/pull/2388) ([tatu-at-datastax](https://github.com/tatu-at-datastax))
+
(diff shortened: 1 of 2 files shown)


src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/CreateKeyspaceCommand.java
```diff
@@ -27,31 +28,32 @@ public record Options(@Nullable @Valid Replication replication) {}
...
-   * @param strategyOptions Options for each strategy. For <code>SimpleStrategy</code>,
-   *     `replication_factor` is optional. For the <code>NetworkTopologyStrategy</code> each data
-   *     center with replication.
+   * @param strategyOptions Additional strategy options as a flat map. For {@code SimpleStrategy},
+   *     use {@code replication_factor} (integer). For {@code NetworkTopologyStrategy}, use
+   *     datacenter names as keys with replication factor as values (e.g. {@code "dc1": 3}).
    */
-  @Schema(description = "Cassandra based replication settings.")
-  // no record due to the @JsonAnySetter, see
-  // https://github.com/FasterXML/jackson-databind/issues/562
```
Hazel-Datastax: I checked that this issue has been closed, change to record

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/CreateKeyspaceCommand.java
```diff
@@ -27,31 +28,32 @@ public record Options(@Nullable @Valid Replication replication) {}
...
-  public static class Replication {
-    @NotNull()
-    @Pattern(regexp = "SimpleStrategy|NetworkTopologyStrategy")
-    @JsonProperty("class")
-    private String strategy;
-
-    @JsonAnySetter
-    @Schema(hidden = true)
```
Hazel-Datastax: If I understand it correctly, we used `@JsonAnySetter` and `@Schema(hidden = true)` was that we want to capture all properties that do not match a declared field (like class) and put them into the `strategyOptions` map. The reason to do this is:
1. match the CQL structure - see [doc](https://docs.datastax.com/en/cql-oss/3.3/cql/cql_reference/cqlCreateKeyspace.html)
```
CREATE  KEYSPACE [IF NOT EXISTS] keyspace_name 
   WITH REPLICATION = { 
      'class' : 'SimpleStrategy', 'replication_factor' : N } 
     | 'class' : 'NetworkTopologyStrategy', 
       'dc1_name' : N [, ...] 
   }
   [AND DURABLE_WRITES =  true|false] ;
```
2. Users won't see a confusing `strategyOptions` field in the documentation. If it weren't hidden, users might mistakenly think they need to send a nested JSON object like:
```
// INCORRECT interpretation if not hidden
{
    "class": "SimpleStrategy",
    "strategyOptions": { "replication_factor": 1 }
}
```
Instead of the correct flat structure:
```
// CORRECT usage handled by @JsonAnySetter
{
    "class": "SimpleStrategy",
    "replication_factor": 1
}
```

^^^^ That's from my guess, not necessarily true.

But I think the schema description and the code structure are confusing to me. I wouldn't know that I can use `"replication_factor": 1 ` or `"dc_1" : 3` as `strategyOptions` without referencing the CQL documentation.  

So I added more description and removed `@Schema(hidden = true)`

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/CreateKeyspaceCommand.java
tatu-at-datastax: Good follow up! Looks like issue fixed in Jackson 2.18, and Quarkus we use brings in 2.20(.1)

Issue #2387, opened 2026-02-11
Merge EGW functionality into Data API repo


tatu-at-datastax:
This is umbrella issue for work to deprecate separate https://github.com/Riptano/embedding-gateway repo, by merging its functionality into Data API, run within same process.

At high level, EGW exposes 5 RPC methods, 3 for embeddings, 2 for reranking:

## EmbeddingService (EmbeddingServiceImpl)

1. `embed` — Vectorize text. Takes a provider name, model, list of input texts, and auth tokens; returns float embeddings.                                                                                                      
2. `getSupportedProviders` — Returns the full catalog of configured embedding providers, their models, auth requirements, parameters, and request properties.
3. `validateCredential` — Validates a credential/API key against SyncService for a given provider and tenant.

## RerankingService (RerankingServiceImpl)

4. `rerank` — Re-scores a list of text passages against a query using a reranking model. Returns ranked indices  with scores.
5. `getSupportedRerankingProviders` — Returns the catalog of configured reranking providers, their models, auth configs, and request properties.

Of these, Data API can already use in "embedded mode" (1) and (4) -- calling providers directly -- as well as (2) and (5) (Load local configs) -- as long as setting --`stargate.jsonapi.operations.enableEmbeddingGateway`  -- is disabled (default setting); already done when running ITs. This leaves (3), `validateCredential` as one thing to port over.

So

* Config loading (EmbeddingProvidersConfigProducer): when disabled, uses the default embedded config rather than fetching from the gateway.
* Credential validation (VectorizeConfigValidator:192): credential validation against the gateway only runs when enabled.
* Reranking (RerankingProviderFactory): same flag controls whether reranking uses the gateway or direct HTTP.

In addition, retry logic and extra logic may need to be added on Data API side.

So: tentative work:

1. Add `io.stargate.embedding.gateway.secrets` package and esp `SyncServiceClient` on Data API codebase, change code in `VectorizeConfigValidator` (around line 192) to use that directly, instead of gRPC
      * Also need to implement Credential access (see `EmbeddingServiceImpl.getAuthTokens()`) from within `EmbeddingProvider` call sequence
2. Check if retry logic needs to be improved
      * See `io.stargate.embedding.gateway.ratelimit` for error-based rate-limiting in particular
3. Check if extra logging needs adding (see `io.stargate.embedding.gateway.logging.EmbeddingGatewayLog`

tatu-at-datastax:
Completed as of v1.0.44

PR #2405, opened 2026-02-21
Add SyncService client from EGW for embedded calling from Data API

tatu-at-datastax:
**What this PR does**:

Embeds sync-service validation call code from EGW into Data API, to prepare for removal of EGW

Part of #2387

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## [v1.0.43](https://github.com/stargate/data-api/tree/v1.0.43) (2026-03-31)
+
+[Full Changelog](https://github.com/stargate/data-api/compare/v1.0.42...v1.0.43)
+
+**Merged pull requests:**
+
+- Bumping version for next data-api release [\#2443](https://github.com/stargate/data-api/pull/2443) ([github-actions[bot]](https://github.com/apps/github-actions))
+
 ## [v1.0.42](https://github.com/stargate/data-api/tree/v1.0.42) (2026-03-31)
 
 [Full Changelog](https://github.com/stargate/data-api/compare/v1.0.41...v1.0.42)
--- a/pom.xml
+++ b/pom.xml
@@ -3,7 +3,7 @@
   <modelVersion>4.0.0</modelVersion>
   <groupId>io.stargate</groupId>
   <artifactId>sgv2-jsonapi</artifactId>
-  <version>1.0.43-SNAPSHOT</version>
+  <version>1.0.44-SNAPSHOT</version>
   <properties>
     <!-- Quarkus -->
     <quarkus.platform.artifact-id>quarkus-bom</quarkus.platform.artifact-id>


src/main/java/io/stargate/sgv2/jsonapi/syncservice/SyncServiceClient.java
```diff
@@ -0,0 +1,154 @@
...
+    return syncService
+        .credential(
+            "Bearer " + token, UUID.randomUUID().toString(), tenant.toString(), provider, cred)
+        .onFailure(
+            error -> {
+              return error instanceof APIException apiException
+                  && Objects.equals(
+                      apiException.code,
+                      EmbeddingProviderException.Code
+                          .EMBEDDING_GATEWAY_UNABLE_RESOLVE_AUTHENTICATION_TYPE
```
Hazel-Datastax: nit: `EMBEDDING_GATEWAY`

src/main/java/io/stargate/sgv2/jsonapi/syncservice/SyncServiceErrorMessageMapper.java
```diff
@@ -0,0 +1,52 @@
...
+   * @param response
+   * @return
+   */
+  public static RuntimeException getDefaultException(Response response) {
+
+    // Status code == 408 and 504 for timeout
+    if (response.getStatus() == Response.Status.REQUEST_TIMEOUT.getStatusCode()
+        || response.getStatus() == Response.Status.GATEWAY_TIMEOUT.getStatusCode()) {
+
+      return EmbeddingProviderException.Code.EMBEDDING_GATEWAY_UNABLE_RESOLVE_AUTHENTICATION_TYPE
```
Hazel-Datastax: nit: EMBEDDING_GATEWAY in this file

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/VectorizeConfigValidator.java
```diff
@@ -189,8 +193,12 @@ private void validateAuthentication(
 
         // Validate the credential name from secret service
         // already append the .providerKey to the value in CreateCollectionCommand
+        // !!! 24-Feb-2026, tatu: NOTE: method seems to be expected to signal failure by
+        //   throwing Exception as return value is discarded here. Is this correct?
         if (operationsConfig.enableEmbeddingGateway()) {
           validateCredentials.validate(userConfig.provider(), userAuth.getValue());
+        } else {
+          syncServiceClient.validateKey(userConfig.provider(), userAuth.getValue());
```
Hazel-Datastax: This `Uni<Boolean>` is never subscribed to, so the validation will not actually execute. 

The original `validateCredentials.validate()` works because it uses a gRPC blocking stub (`EmbeddingServiceGrpc.newBlockingStub`), which is a synchronous call that blocks until completion. The new `validateKey()` returns a `Uni` (async/reactive), so just calling it without subscribing is a no-op.

Two possible fixes:
1. Subscribe at the call site: `syncServiceClient.validateKey(...).await().atMost(Duration.ofSeconds(10));`
2. Or update`validateKey`'s signature to return `boolean` directly (with the internal private method still using `Uni` for the retry chain), so the call site stays clean and consistent with the original `validate()`.
```
public boolean validateKey(String provider, String key) {
    return validateKey(
        dataApiRequestInfo.authToken(), dataApiRequestInfo.tenant().toString(), provider, key)
        .await().atMost(Duration.ofSeconds(10));
}
```

Another problem, which has been mentioned in the comment: both the original `validateCredentials.validate()` and the new `validateKey()` discard their return value here. Looking at `validateKey`'s implementation, it can return `false` when `res.credentials()` is null or empty (without errors) — in that case the invalid credential would silently pass validation. 

Maybe consider throwing an exception on false as well, which always returns true or throws.

src/main/java/io/stargate/sgv2/jsonapi/syncservice/SyncServiceClient.java
```diff
@@ -0,0 +1,154 @@
...
+  private Uni<Boolean> validateKey(String token, String tenant, String provider, String key) {
+    return syncService
+        .valid("Bearer " + token, UUID.randomUUID().toString(), tenant, provider, key)
+        .onFailure(
+            error -> {
+              return error instanceof APIException apiException
+                  && Objects.equals(
+                      apiException.code,
+                      EmbeddingProviderException.Code
+                          .EMBEDDING_GATEWAY_UNABLE_RESOLVE_AUTHENTICATION_TYPE
```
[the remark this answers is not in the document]
tatu-at-datastax: Yes -- there is `SyncServiceErrorMessageMapper` that needs to be rewritten to replace this exception type.

PR #2460, opened 2026-04-27
Fix #2459 allow overrides for "findAndRerank" command

### LINKED ISSUE #2459: Allow overriding Reranking options (provider, model, auth, parameters) on per-request basis for `findAndRerank` command
As per title, if reranking is enabled for a Collection, it should be possible to pass overrides to use (f.ex) different model for configured provider -- or even full provider/model override.

Override is specified as `options` entry with key `rerank`, like so:

```
        {"findAndRerank": {
            "sort": {"$hybrid": {"$vectorize": "search text"}},
            "options": {
                "limit": 10,
                "rerank": {
                  "provider": "nvidia",
                  "modelName": "nvidia/llama-3.2-nv-rerankqa-1b-v2"
                }
            }
        }}
```

Notes:

* Override not allowed if Reranking Feature not enabled (for Data API instance)
* Override is allowed evenn if Collection does not have it enabled
* Override definition must be complete (replacing Collection's definition, if any), not partial ("all or nothing") -- except that "auth" and "parameters" are optional
* Should make use of existing types where possible: for example `CreateCollectionCommand.RerankServiceDesc` for override definition

tatu-at-datastax:
**What this PR does**:

Adds ability to provide rerank provider/model overrides for findAndRerank; see https://github.com/stargate/data-api/issues/2459 for details.

Fixes #2459

**Checklist**
- [ ] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/FindAndRerankCommand.java
```diff
@@ -103,7 +108,55 @@ public record Options(
...
+                  "Optional per-request override for the reranking service. Overrides the collection-level reranking provider, model, and/or authentication.",
+              implementation = RerankServiceOverride.class)
+          @JsonProperty("rerank")
+          RerankServiceOverride rerankServiceOverride) {}
+
+  /**
+   * Optional per-request override for the reranking service configuration. All fields are optional;
+   * unspecified fields fall back to the collection's defaults.
+   */
+  public record RerankServiceOverride(
```
amorton: -1 can we reuse the record from `CreateCollectionCommand.RerankServiceDesc`

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/FindAndRerankOperationBuilder.java
```diff
@@ -164,32 +170,51 @@ private void checkSupported() {
       // TODO: more info in the error
       throw RequestException.Code.UNSUPPORTED_RERANKING_COMMAND.get();
```
amorton, line 219: I would guess that if the collection as created without reranking enabled the reranking config for the collection will say it is disabled. We need a different path if there is a reranking override.

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/FindAndRerankOperationBuilder.java
```diff
@@ -164,32 +170,51 @@ private void checkSupported() {
...
 
+  /**
+   * Returns true when the per-request override specifies a different provider or model. Auth-only
+   * overrides return false because the collection's own provider/model is unchanged and should go
+   * through the normal (EOL-only) validation path rather than the stricter override validation.
+   */
+  private boolean hasRerankOverride() {
+    var override =
+        getOrDefault(command.options(), FindAndRerankCommand.Options::rerankServiceOverride, null);
+    return override != null && (override.provider() != null || override.modelName() != null);
```
amorton: there is an isEmpty() on the record, the one on the create collection command is better. we should use that.

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/FindAndRerankOperationBuilder.java
```diff
@@ -238,6 +263,107 @@ private void checkSupported() {
             .collect(java.util.stream.Collectors.toUnmodifiableList()));
   }
 
+  /**
+   * Resolves the effective RerankServiceDef by merging any per-request override from the command
+   * options with the collection's configured defaults. Result is memoized for the lifetime of this
+   * builder.
+   */
+  private CollectionRerankDef.RerankServiceDef resolveEffectiveRerankServiceDef(
```
amorton: -1 - we do not want to merge the two settings, that leads to more complexity than is needed. If there is any per request setting it should override the collection settings

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/FindAndRerankOperationBuilder.java
```diff
@@ -238,6 +263,107 @@ private void checkSupported() {
...
+        new CollectionRerankDef.RerankServiceDef(
+            effectiveProvider, effectiveModelName, effectiveAuth, collectionDef.parameters());
+    return effectiveRerankServiceDef;
+  }
+
+  /**
+   * Validates that the overridden provider and model exist and are usable in the reranking
+   * providers configuration.
+   */
+  private void validateRerankOverride(
```
amorton: -1 as I assume this is duplicated from the place we do this in the create collection code, reuse that code.

amorton (changes requested):

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/FindAndRerankOperationBuilder.java
tatu-at-datastax: My assumption was that override would not work to change disabled to enable, hence generated this way.

If we want different logic that is possible... but do we?

tatu-at-datastax: As per discussions, we may want to allow overrides as long as Data API itself has feature enabled.

tatu-at-datastax: Changed.

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/FindAndRerankOperationBuilder.java
tatu-at-datastax: Changed to not merge settings.

tatu-at-datastax:
Re-prompted:

```
Requirements changed -- re-plan based on issue https://github.com/stargate/data-api/issues/2459 that this PR is for
...
Review your work
...
Now, can you add one or more unit tests
...
Add one or more ITs as well
...
How about tests with auth override?
```

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/FindAndRerankOperationBuilder.java
```diff
@@ -160,45 +165,63 @@ private void checkSupported() {
...
-                  .message()
-                  .orElse("The model is no longer supported (reached its end-of-life).")));
+
+    // Resolve effective rerank service def: per-request override replaces collection config
+    // entirely; otherwise fall back to the collection's configured defaults.
+    if (hasOverride) {
+      var rerankingProvidersConfig = commandContext.rerankingProviderFactory().getRerankingConfig();
+      validateRerankOverride(
+          rerankingProvidersConfig, rerankOverride.provider(), rerankOverride.modelName());
+      effectiveRerankServiceDef =
```
amorton: there is no protection in this builder to make sure it only runs once, this is mutable shared state. If not null, throw IllegalState for the builder reuse

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/FindAndRerankOperationBuilder.java
```diff
@@ -238,6 +261,63 @@ private void checkSupported() {
             .collect(java.util.stream.Collectors.toUnmodifiableList()));
   }
 
+  /**
+   * Validates that the overridden provider and model exist and are usable in the reranking
+   * providers configuration.
+   */
+  // package-private for unit testing
+  void validateRerankOverride(
```
amorton: -1 for this function, it is duplicating code and breaking encapsulation. 

CollectionRerankDef.fromApiDesc() is the existing funciton we use to validate the reranking provided by the user. It is designed to be called when from create collection. It has parts we can re-use for vliadating the service description, the only difference is that we may use a different error code

e.g. INVALID_CREATE_COLLECTION_OPTIONS for create and RequestException.Code.INVALID_RERANK_OVERRIDE for findAndRerank. 

AND theat we may have different rules for the model support.

When we do createCollection the model cannot be DEPRECATED or END_OF_LIFE 

When we do findAndReRank using the model from the collection schema it cannot be END_OF_LIFE 

when we do findAndRerank using an overload it cannot be DEPRECATED or END_OF_LIFE 

Refactor CollectionRerankDef as needed so we can all it to create the CollectionRerankDef for the overridden service def, and centralize error throwing into it for checking the model has the correct support status. We should update all places we check model status to be in this class.

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/FindAndRerankOperationBuilder.java
```diff
@@ -160,45 +165,63 @@ private void checkSupported() {
...
+      }
     }
   }
 
   private TaskGroupAndDeferrables<RerankingTask<CollectionSchemaObject>, CollectionSchemaObject>
       rerankTasks(List<RerankingTask.DeferredCommandWithSource> deferredCommandResults) {
 
-    // Previous code will check reranking is supported
-    var providerConfig = commandContext.schemaObject().rerankingConfig().rerankServiceDef();
+    // checkSupported() already resolved effectiveRerankServiceDef
```
amorton: Add a null check using Objects , like a gentleman

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/FindAndRerankOperationBuilder.java
```diff
@@ -160,45 +165,63 @@ private void checkSupported() {
...
-              "message",
-              modelConfig
-                  .apiModelSupport()
-                  .message()
-                  .orElse("The model is no longer supported (reached its end-of-life).")));
+
+    // Resolve effective rerank service def: per-request override replaces collection config
+    // entirely; otherwise fall back to the collection's configured defaults.
+    if (hasOverride) {
+      var rerankingProvidersConfig = commandContext.rerankingProviderFactory().getRerankingConfig();
```
amorton: see comments on validateRerankOverride() for a better way to do this

src/main/resources/errors.yaml
```diff
@@ -265,11 +265,21 @@ request-errors:
     code: UNSUPPORTED_RERANKING_COMMAND
     title: Reranking not enabled for collection
     body: |-
-      The `findAndRerank` is not enabled for the collection.
-      
-      The Command is only supported for collections that have `vector`, `lexical`, and `rerank` configured. They additionally require a vectorize configuration if `$vectorize` is used.
-      
-      Resend using a supported collection. 
+      The `findAndRerank` is not enabled for the collection and no per-request reranking service override was provided.
```
amorton: use of caveman like language, "and no thing make grug mad"

use ", and a reranking service override was not provided with the command."

src/main/resources/errors.yaml
```diff
@@ -265,11 +265,21 @@ request-errors:
...
+
+      The Command is only supported for collections that have `vector`, `lexical`, and `rerank` configured, or when a per-request reranking service override is provided via the `rerank` option. They additionally require a vectorize configuration if `$vectorize` is used.
```
amorton: ", or when a reranking service override is provided via the `rerank` command option."

src/main/resources/errors.yaml
```diff
@@ -265,11 +265,21 @@ request-errors:
...
+
+      Resend using a supported collection or provide a reranking service override in the command options.
```
amorton: "or provide a `rerank` override in the command options"

src/main/resources/errors.yaml
```diff
@@ -265,11 +265,21 @@ request-errors:
...
+
+  - scope:
+    code: INVALID_RERANK_OVERRIDE
+    title: Invalid reranking service override
+    body: |-
+      The per-request reranking service override in the findAndRerank command is invalid.
```
amorton: I dont like the use of "per-request" is is for the command, see other comments and re-work.

amorton (dismissed):
see comments, too much code duplication going on.

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/FindAndRerankOperationBuilder.java
tatu-at-datastax: Ok. Good clarification: I was not sure whether `DEPRECATED` should be allowed (like from defaults) for overrides or not -- could see either way being possible.

But decision is that overrides are not allowed to use DEPRECATED model.

src/main/resources/errors.yaml
tatu-at-datastax: Not sure I read it as being primitive, but extended version does sound better.

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/FindAndRerankOperationBuilder.java
tatu-at-datastax: Added check in latest commit.

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/FindAndRerankOperationBuilder.java
tatu-at-datastax: Added check.

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/FindAndRerankOperationBuilder.java
tatu-at-datastax: Refactored to reduce code duplication, centralize handling. Have a look if this makes sense, @amorton.

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/FindAndRerankOperationBuilder.java
```diff
@@ -195,51 +199,59 @@ private void checkSupported() {
...
+  /**
+   * Resolve the effective {@link CollectionRerankDef.RerankServiceDef} for this command: a
+   * non-empty command override replaces the collection-level config entirely; otherwise fall back
+   * to the collection's configured defaults. Throws {@link
+   * RequestException.Code#UNSUPPORTED_RERANKING_COMMAND} if neither source provides a service.
+   */
+  private CollectionRerankDef.RerankServiceDef resolveRerankServiceDef() {
+    var rerankOverride =
+        getOrDefault(command.options(), FindAndRerankCommand.Options::rerankServiceOverride, null);
+    boolean hasOverride = rerankOverride != null && !rerankOverride.isEmpty();
```
erichare: NIT: I believe this would mean that:

`"rerank": {"provider": null, "modelName": null}`

Would be treated as hasOverride=FALSE (if i'm thinking of isEmpty() correctly). Just wondering if we want to be more explicit there - that's a case where (potentially) the client might want to provide an override, but made a mistake that led to nulls for the values. In which case, maybe we'd want to either document this, or reject these as invalid?

erichare (approved):
LGTM @tatu-at-datastax ! One NIT that i found.

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/FindAndRerankOperationBuilder.java
tatu-at-datastax: Hmmmh. Good question. Let me see.

tatu-at-datastax: Ok. Looks like it is rather difficult to both allow `{ }` AND exclude `{"provider": null, "modelName": null}` -- especially when we are re-using (for better or worse) `RerankingServiceDesc` for both reranking definition (Collection creation) and for reranking overrides.
But I'll see if I can figure out something. :)

tatu-at-datastax: @erichare Changed so that Overrides with both `{ }` and `{"provider": null, "modelName": null}` will now fail: only explicit `null` (and missing override) are accepted as valid "no overrides".

tatu-at-datastax:
> LGTM @tatu-at-datastax ! One NIT that i found.

This has now been resolved: no more quiet ignoral of non-null overrides.

PR #2461, opened 2026-04-27
feat: Increase page sizes (default 50, max 500)

### LINKED ISSUE #2196: Allow pagination with page sizes bigger than 20
The current implementation of the Data API has a fixed page size of 20. 

It would be desirable for paginated results to either increase the page size (to for example 100 ) or put an option somewhere in the request structure where the user can modify the default page size of 20 up to a certain hard upper limit value (e.g. 100 or 1000).

When retrieving many small documents having such a small page size makes the retrieval somehow slow. 
The bigger/customizable page size should be generally made useable regardless of the backend (Cassandra, DSE or Astra DB)

erichare:
**What this PR does**:

Fixes #2196

**Checklist**
- [x] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/config/OperationsConfig.java
```diff
@@ -35,18 +35,24 @@ public interface OperationsConfig {
...
    * defaults to 100
    */
   int DEFAULT_MAX_DOCUMENT_INSERT_COUNT = 100;
 
   /**
-   * @return Defines the default document page size, defaults to <code>20</code>.
+   * @return Defines the default document page size, defaults to <code>50</code>.
    */
-  @Max(500)
+  @Max(MAX_PAGE_SIZE)
```
tatu-at-datastax: Not sure we can lower maximum allowed from 500 to 100, potentially breaking change?

erichare: Good point. Callers might already be using up to 500. I will set the max back to 500

src/main/java/io/stargate/sgv2/jsonapi/config/OperationsConfig.java
```diff
@@ -35,18 +35,27 @@ public interface OperationsConfig {
...
    * defaults to 100
    */
   int DEFAULT_MAX_DOCUMENT_INSERT_COUNT = 100;
 
   /**
-   * @return Defines the default document page size, defaults to <code>20</code>.
+   * @return Defines the default document page size, defaults to <code>50</code>.
    */
-  @Max(500)
+  @Max(MAX_CONFIGURABLE_PAGE_SIZE)
```
erichare, line 72: @tatu-at-datastax sorry, accidentally wiped out your comment lol, but now the max configurable size is set back to 500

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/FindAndRerankCommand.java
```diff
@@ -112,7 +113,11 @@ public record HybridLimits(
       @JsonProperty(DocumentConstants.Fields.LEXICAL_CONTENT_FIELD) int lexicalLimit,
       CommandFeatures commandFeatures)
       implements Recordable {
-    public static final HybridLimits DEFAULT = new HybridLimits(50, 50, CommandFeatures.EMPTY);
+    public static final HybridLimits DEFAULT =
```
amorton, line 118: plus set the default for limit above to be 10, and update FindAndRerankOperationBuilder line 207 and update the comment on the field. 

And can we update the comment on the find command limit and add the default.

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/UpdateManyCommandResolver.java
```diff
@@ -101,7 +101,7 @@ private FindCollectionOperation getFindOperation(
         DocumentProjector.includeAllProjector(),
         null != command.options() ? command.options().pageState() : null,
         Integer.MAX_VALUE,
-        operationsConfig.defaultPageSize(),
+        operationsConfig.maxDocumentUpdateCount(),
```
amorton: -1, this is the page size for the read part of the update command. Revert to previous.

src/main/resources/application.conf
```diff
@@ -78,7 +78,7 @@ datastax-java-driver {
       basic.request.timeout = 10 seconds
       // this will be the number of rows we return in a single page of results to the user.
       // It may be overridded for in memory sorting, but this is the default.
```
amorton: nit: fix typo overridded > overridden

src/test/java/io/stargate/sgv2/jsonapi/api/model/command/impl/HybridLimitsDeserializerTest.java
```diff
@@ -85,6 +96,12 @@ private static Stream<Arguments> validLimitsTestCases() {
             """,
             new FindAndRerankCommand.HybridLimits(
                 99, 99, CommandFeatures.of(CommandFeature.HYBRID_LIMITS_NUMBER))),
+        Arguments.of(
```
amorton: from discussions - prob not needed

src/test/java/io/stargate/sgv2/jsonapi/api/v1/tables/FindWithRegularSortTableIntegrationTest.java
```diff
@@ -74,8 +76,8 @@ class FindSortSuccess {
     public void sortByTextAndNullValue() throws Exception {
       sortByName(testDatas, true);
       JsonNodeFactory nodefactory = objectMapper.getNodeFactory();
-      final ArrayNode arrayNode = nodefactory.arrayNode(20);
-      for (int i = 0; i < 20; i++) {
+      final ArrayNode arrayNode = nodefactory.arrayNode(DEFAULT_PAGE_SIZE);
```
amorton: we can tidy this up, pls move the code to create the test data into a single function

src/test/java/io/stargate/sgv2/jsonapi/api/v1/FindCollectionWithSortIntegrationTest.java
```diff
@@ -48,8 +51,8 @@ public void setUp() {
     public void sortByTextAndNullValue() throws Exception {
       sortByUserName(testDatas, true);
       JsonNodeFactory nodefactory = objectMapper.getNodeFactory();
-      final ArrayNode arrayNode = nodefactory.arrayNode(20);
-      for (int i = 0; i < 20; i++) {
+      final ArrayNode arrayNode = nodefactory.arrayNode(DEFAULT_PAGE_SIZE);
```
amorton: pls move this into a function if we can for the test data

src/test/java/io/stargate/sgv2/jsonapi/config/OperationsConfigTest.java
```diff
@@ -0,0 +1,34 @@
...
+import io.quarkus.test.junit.QuarkusTest;
+import io.quarkus.test.junit.TestProfile;
+import io.stargate.sgv2.jsonapi.testresource.NoGlobalResourcesTestProfile;
+import jakarta.inject.Inject;
+import jakarta.validation.constraints.Max;
+import org.junit.jupiter.api.Test;
+
+@QuarkusTest
+@TestProfile(NoGlobalResourcesTestProfile.Impl.class)
+class OperationsConfigTest {
```
amorton: nice, but not needed :)

src/test/java/io/stargate/sgv2/jsonapi/service/operation/collections/FindCollectionOperationTest.java
```diff
@@ -1307,16 +1303,25 @@ public void findOrWithAllFilter() throws Exception {
...
+      when(queryExecutor.executeRead(
+              eq(requestContext), any(SimpleStatement.class), any(), anyInt()))
           .then(
               invocation -> {
+                SimpleStatement actualStatement = invocation.getArgument(1, SimpleStatement.class);
+                assertThat(actualStatement.getQuery())
+                    .contains(
+                        "SELECT key, tx_id, doc_json FROM \"%s\".\"%s\" WHERE"
+                            .formatted(KEYSPACE_NAME, COLLECTION_NAME))
+                    .contains("array_contains CONTAINS ?")
```
amorton: -1 this changes the nature of the test, need to check the exact cql structure because the order of the comparisons is important

src/test/java/io/stargate/sgv2/jsonapi/service/operation/collections/FindCollectionOperationTest.java
```diff
@@ -1375,16 +1376,25 @@ public void findOrWithAllFilterWithNegation() throws Exception {
...
       List<Row> rows = List.of(resultRow(0, "doc1", UUID.randomUUID(), doc1));
       AsyncResultSet results = new MockAsyncResultSet(KEY_TXID_JSON_COLUMNS, rows, null);
       final AtomicInteger callCount = new AtomicInteger();
       QueryExecutor queryExecutor = mock(QueryExecutor.class);
-      when(queryExecutor.executeRead(eq(requestContext), eq(stmt), any(), anyInt()))
+      when(queryExecutor.executeRead(
+              eq(requestContext), any(SimpleStatement.class), any(), anyInt()))
           .then(
               invocation -> {
+                SimpleStatement actualStatement = invocation.getArgument(1, SimpleStatement.class);
```
amorton: -1, this changes the nature of the test. It does not confirm the query is structured correctly. We should revert.

src/test/java/io/stargate/sgv2/jsonapi/service/resolver/FindAndRerankOperationBuilderTest.java
```diff
@@ -0,0 +1,140 @@
...
+class FindAndRerankOperationBuilderTest {
+
+  @Inject ObjectMapper objectMapper;
+  @Inject OperationsConfig operationsConfig;
+  @Inject FindCommandResolver findCommandResolver;
+
+  private final TestConstants testConstants = new TestConstants();
+
+  @Test
+  void defaultsHybridLimitsOnCommandContextForInnerFinds() throws Exception {
```
amorton: tests on the deserialisation of options.hybridLimits are done in HybridLimitsDeserializerTest. Have a look at updating that to handle the `hybridLimits` is undefined. 

That has the advantage of running as a unit test. 

We can remove this test.

src/test/java/io/stargate/sgv2/jsonapi/service/resolver/UpdateManyCommandResolverTest.java
```diff
@@ -96,7 +96,8 @@ public void idFilterCondition() throws Exception {
 
                         assertThat(find.objectMapper()).isEqualTo(objectMapper);
                         assertThat(find.commandContext()).isEqualTo(commandContext);
-                        assertThat(find.pageSize()).isEqualTo(operationsConfig.defaultPageSize());
+                        assertThat(find.pageSize())
+                            .isEqualTo(operationsConfig.maxDocumentUpdateCount());
```
amorton: tied to the UpdateManyCommandResolver change to be rolled back, this can be rolled back as well.,

Multiple places in the file.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/tables/FindPaginationTableIntegrationTest.java
```diff
@@ -53,26 +54,35 @@ public void findWithNoSortAndPartitionedPageState() throws Exception {
 
       String pageState = null;
 
-      // each partition should have 3 pages
-      for (int page = 0; page < 3; page++) {
+      int pageCount =
```
amorton, line 57: lets dive into this on the next review, just need to think it through more.

amorton (changes requested):
some changes , as discussed in person

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/FindAndRerankCommand.java
```diff
@@ -38,6 +39,10 @@ public record FindAndRerankCommand(
     @Valid @JsonProperty("sort") FindAndRerankSort sortClause,
     @Valid @Nullable Options options)
     implements ReadCommand, Filterable, Projectable, Windowable {
+
+  public static final int DEFAULT_LIMIT = 10;
+  public static final int MAX_HYBRID_LIMIT = 100;
```
tatu-at-datastax: Shouldn't these be defined in `OperationsConfig` like other limits?

erichare: Just updated this @tatu-at-datastax - and found a few other places where that could be refactored. PTAL when you get a chance, thank you!

src/main/java/io/stargate/sgv2/jsonapi/config/OperationsConfig.java
```diff
@@ -35,18 +35,40 @@ public interface OperationsConfig {
...
   int DEFAULT_MAX_FILTER_SIZE = 64;
 
+  /** Defines the default page size for read queries. */
+  int DEFAULT_PAGE_SIZE = 50;
+
+  /** Defines the default number of documents returned by findAndRerank. */
+  int DEFAULT_FIND_AND_RERANK_LIMIT = 10;
+
+  /** Defines the minimum hybrid search read limit. */
+  int MIN_HYBRID_SEARCH_LIMIT = 0;
```
tatu-at-datastax: Does minimum of 0 make sense? Maybe it should be 1, for reading 0 candidates seems useless.

src/main/java/io/stargate/sgv2/jsonapi/config/OperationsConfig.java
```diff
@@ -35,18 +35,40 @@ public interface OperationsConfig {
...
+
   /**
    * Defines the default maximum documents to insert setting for {@code InsertMany} command;
    * defaults to 100
    */
   int DEFAULT_MAX_DOCUMENT_INSERT_COUNT = 100;
 
   /**
-   * @return Defines the default document page size, defaults to <code>20</code>.
+   * @return Defines the default document page size, defaults to <code>50</code>.
```
tatu-at-datastax, line 70: Shouldn't this be 50 (see above for `MAX_CONFIGURABLE_PAGE_SIZE`)?

**EDIT**: never mind, this refers to "@WithDefault" part.

tatu-at-datastax:
I think the big problem here is that validation of configuration settings uses static constants and not dynamic `OperationsConfig` (dynamic wrt config file and System prop/env variable overrides).
To fix that would require bigger changes: validation can NOT be done on `FindAndRerankCommand` deserializer(s) -- those can only bind values from JSON but do not have access to `OperationsConfig` -- instead, it needs to be done in `FindAndRerankOperationBuilder` as follow-up step, when constructing operation with `build()` method.

tatu-at-datastax (commented):
As per my comment above, I think this needs to be rewritten unfortunately... validation and application of default values cannot be done from Deserializers.

erichare:
> As per my comment above, I think this needs to be rewritten unfortunately... validation and application of default values cannot be done from Deserializers.

@tatu-at-datastax makes sense, can you look at my most recent commit, is this more what you had in mind?

tatu-at-datastax:
> > As per my comment above, I think this needs to be rewritten unfortunately... validation and application of default values cannot be done from Deserializers.
> 
> @tatu-at-datastax makes sense, can you look at my most recent commit, is this more what you had in mind?

Yes, with a quick look this looks exactly like what I had in mind!

Will try to go over all changes later on, but these changes look good.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/FindAndRerankCommand.java
```diff
@@ -63,14 +65,22 @@ public void addCommandFeatures(CommandFeatures commandFeatures) {
   public record Options(
       @Positive(message = "limit should be greater than `0`")
           @Schema(
-              description = "The maximum number of documents to return after reranking.",
+              description =
+                  "The maximum number of documents to return after reranking. Defaults to "
+                      + OperationsConfig.DEFAULT_FIND_AND_RERANK_LIMIT
```
tatu-at-datastax: This is challenging... description must be static, but the setting can be overridden by `application.yaml`. System property and/or Env variable. So description may indicate wrong setting.

Maybe we should just leave it out instead of use possibly wrong value?

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/FindAndRerankCommand.java
```diff
@@ -63,14 +65,22 @@ public void addCommandFeatures(CommandFeatures commandFeatures) {
...
-              implementation = Integer.class)
+              implementation = Integer.class,
+              defaultValue = "" + OperationsConfig.DEFAULT_FIND_AND_RERANK_LIMIT)
           Integer limit,
       /** ---- */
       @Schema(
               description =
-                  "The maximum number of documents to read for the vector and lexical queries that feed into the reranking. May be an number or an object with $vector and $lexical fields.",
+                  "The maximum number of documents to read for the vector and lexical queries that feed into the reranking. Defaults to "
+                      + OperationsConfig.DEFAULT_HYBRID_SEARCH_LIMIT
```
tatu-at-datastax: Same as above.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/FindCommand.java
```diff
@@ -32,9 +33,12 @@ public record Options(
       @Positive(message = "limit should be greater than `0`")
           @Schema(
               description =
-                  "Maximum number of document that can be fetched for the command. If value is higher than the default page size, amount of returned documents will be limited to the default page size and page state will be returned in the response, so a caller can to continue paging through documents.",
+                  "Maximum number of documents to return. Defaults to the default page size, "
+                      + OperationsConfig.DEFAULT_PAGE_SIZE
```
tatu-at-datastax: Same problem wrt static description.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/FindCommand.java
```diff
@@ -32,9 +33,12 @@ public record Options(
...
+                      + ". If the limit is higher than the default page size, the response will include up to the default page size of documents and a page state so the caller can continue paging through documents.",
               type = SchemaType.INTEGER,
-              implementation = Integer.class)
+              implementation = Integer.class,
+              defaultValue = "" + OperationsConfig.DEFAULT_PAGE_SIZE)
```
tatu-at-datastax: Same problem wrt static default value.

src/main/java/io/stargate/sgv2/jsonapi/config/OperationsConfig.java
```diff
@@ -35,18 +35,40 @@ public interface OperationsConfig {
   /** Defines the default max size of filter fields. */
   int DEFAULT_MAX_FILTER_SIZE = 64;
 
+  /** Defines the default page size for read queries. */
```
tatu-at-datastax: Technically should be "DEFAULT_DEFAULT_PAGE_SIZE", I think, given it may be overridden with sys props, env var, application.yaml

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/FindAndRerankOperationBuilder.java
```diff
@@ -124,6 +127,40 @@ public Operation<CollectionSchemaObject> build() {
...
+      return;
+    }
+    throw RequestException.Code.COMMAND_FIELD_VALUE_INVALID.get(
+        Map.of(
+            "field",
+            field,
+            "value",
+            String.valueOf(value),
+            "message",
+            "must be between %d and %d".formatted(bounds.min(), bounds.max())));
```
tatu-at-datastax: Should indicate that limits are inclusive (if they are, like I think they are; if not, exclusive)

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/FindAndRerankCommand.java
erichare: I agree - i removed it!

src/main/java/io/stargate/sgv2/jsonapi/config/OperationsConfig.java
erichare: I kept the naming but made clear that this can be overridden in the javadocs, let me know if this looks okay

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/FindAndRerankOperationBuilder.java
erichare: Yes, they are inclusive, and good call, i added this.

erichare:
@tatu-at-datastax Thank you again for the review, i pushed a commit to address them i hope, let me know how this looks now!

src/main/java/io/stargate/sgv2/jsonapi/config/OperationsConfig.java
tatu-at-datastax: Yeah, double DEFAULT not great either..

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/FindAndRerankCommand.java
```diff
@@ -112,7 +114,11 @@ public record HybridLimits(
       @JsonProperty(DocumentConstants.Fields.LEXICAL_CONTENT_FIELD) int lexicalLimit,
       CommandFeatures commandFeatures)
       implements Recordable {
-    public static final HybridLimits DEFAULT = new HybridLimits(50, 50, CommandFeatures.EMPTY);
+    public static final HybridLimits DEFAULT =
```
tatu-at-datastax, line 118: Is this constant still needed/useful? Or could it be removed?

erichare: My bad, good catch. It's unused now thanks to defaultHybridLimits and only used in a test that basically is just a tautology. I just removed it @tatu-at-datastax

tatu-at-datastax (approved):
Ok, looks good to me, so approving. But next @amorton can indicate if further changes needed, or approve as-is.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/FindAndRerankCommand.java
```diff
@@ -157,13 +157,12 @@ public HybridLimits deserialize(
       };
     }
 
-    private HybridLimits deserialize(JsonParser jsonParser, NumericNode limitsNumber)
-        throws JsonMappingException {
-
+    // Range validation against dynamic OperationsConfig bounds happens later in
+    // FindAndRerankOperationBuilder.build(); the deserializer only handles JSON shape.
+    private HybridLimits deserialize(JsonParser jsonParser, NumericNode limitsNumber) {
+      int limit = limitsNumber.asInt();
```
amorton: nit: use var when type if obvious

src/main/java/io/stargate/sgv2/jsonapi/config/OperationsConfig.java
```diff
@@ -156,6 +185,14 @@ public interface OperationsConfig {
   @WithDefault("100")
   int defaultCountPageSize();
 
+  /**
+   * @return Defines the default number of documents returned by {@code findAndRerank}, defaults to
+   *     <code>10</code>.
```
amorton: nit: link to the const {@link #DEFAULT_FIND_AND_RERANK_LIMIT}

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/FindAndRerankOperationBuilder.java
```diff
@@ -382,6 +419,13 @@ private FindCommand.Options buildFindOptions(boolean forVectorRead) {
         getOrDefault(command.options(), FindAndRerankCommand.Options::includeSortVector, false));
   }
 
+  private FindAndRerankCommand.HybridLimits defaultHybridLimits() {
```
amorton: can we move this back to a const in the Hybrid Limits class 

`public static final HybridLimits DEFAULT = new HybridLimits(50, 50, CommandFeatures.EMPTY);`

erichare:
@amorton Please check my latest commit when you can! To avoid hardcoding 50 on that line i restored, i added the default hybrid limits to Reranking constants. That way, the operations config doesnt need to be imported, and it (like the const in FindAndRerank impl) use that. Let me know if that looks good, the rest of the code should be as it was prior now.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/tables/FindWithRegularSortTableIntegrationTest.java
```diff
@@ -73,39 +74,21 @@ class FindSortSuccess {
...
-      JsonNodeFactory nodefactory = objectMapper.getNodeFactory();
-      final ArrayNode arrayNode = nodefactory.arrayNode(20);
-      for (int i = 0; i < 20; i++) {
-        arrayNode.add(
-            objectMapper.readTree(
-                objectMapper
-                    .writerWithDefaultPrettyPrinter()
-                    .writeValueAsString(testDatas.get(i))));
-      }
+      final ArrayNode arrayNode = testDataArray(testDatas, 0, DEFAULT_PAGE_SIZE);
```
amorton: nit: does not need final, can use var

src/test/java/io/stargate/sgv2/jsonapi/api/v1/tables/FindWithRegularSortTableIntegrationTest.java
```diff
@@ -351,6 +289,17 @@ public void sortFailDueToTooMany() {
 
   private static final ObjectMapper objectMapper = new ObjectMapper();
 
+  private static ArrayNode testDataArray(List<Object> datas, int offset, int limit)
+      throws JsonProcessingException {
+    final ArrayNode arrayNode = objectMapper.getNodeFactory().arrayNode(limit);
```
amorton: nit: dont need final, can use var

src/test/java/io/stargate/sgv2/jsonapi/api/v1/FindAndRerankCollectionIntegrationTest.java
```diff
@@ -98,6 +98,60 @@ void failOnLexicalDisabled() {
...
+                }
+            }
+            """;
+
+    givenHeadersPostJsonThen(keyspaceName, collectionName, rerank)
+        .body("$", responseIsError())
+        .body("errors[0].errorCode", is(RequestException.Code.COMMAND_FIELD_VALUE_INVALID.name()))
+        .body("errors[0].message", containsString("hybridLimits.$vector"))
+        .body("errors[0].message", containsString("101"))
+        .body("errors[0].message", containsString("must be between 1 and 100"));
```
amorton, line 152: nit: can this use the constants for min and max ?

amorton, line 152: nit: can this use the constants for min and max ?

src/test/java/io/stargate/sgv2/jsonapi/api/v1/FindAndRerankCollectionIntegrationTest.java
```diff
@@ -98,6 +98,60 @@ void failOnLexicalDisabled() {
         "only be used on collections for which Lexical feature is enabled");
   }
 
+  @Test
+  void failOnHybridLimitsAboveMax() {
```
amorton, line 102: I know we have unit tests for the deserialisation, but for integration tests should kind of ignore / forgot those. So it would be good to check some other cases like negative values, and using the hybridLimits as an option. 

we can keep them in this function if it do not get too big, and reuse the table def 

HOWEVER - I think this test needs tighening up, it only works because the default for the reranker is that it is enabled. If the reranking was set to enabled=false then a different error would be raised. 

So update the collection definition to set the lexical && rerank to be enabled explicity. The other tests in the class check the collection has the features enabled

src/test/java/io/stargate/sgv2/jsonapi/api/v1/FindCollectionWithSortIntegrationTest.java
```diff
@@ -550,6 +455,17 @@ public void sortFailDueToTooMany() {
 
   private static final ObjectMapper objectMapper = new ObjectMapper();
 
+  private static ArrayNode testDataArray(List<Object> datas, int offset, int limit)
+      throws JsonProcessingException {
+    final ArrayNode arrayNode = objectMapper.getNodeFactory().arrayNode(limit);
```
amorton, line 460: nit: dont need final, can use var.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/FindCollectionWithSortIntegrationTest.java
```diff
@@ -47,15 +49,7 @@ public void setUp() {
...
-      JsonNodeFactory nodefactory = objectMapper.getNodeFactory();
-      final ArrayNode arrayNode = nodefactory.arrayNode(20);
-      for (int i = 0; i < 20; i++) {
-        arrayNode.add(
-            objectMapper.readTree(
-                objectMapper
-                    .writerWithDefaultPrettyPrinter()
-                    .writeValueAsString(testDatas.get(i))));
-      }
+      final ArrayNode arrayNode = testDataArray(testDatas, 0, DEFAULT_PAGE_SIZE);
```
amorton: nit: dont need final, can use var.

src/test/java/io/stargate/sgv2/jsonapi/api/v1/PaginationIntegrationTest.java
```diff
@@ -18,8 +19,8 @@ public class PaginationIntegrationTest extends AbstractCollectionIntegrationTest
   @TestMethodOrder(MethodOrderer.OrderAnnotation.class)
   @Order(1)
   class NormalFunction {
-    private static final int defaultPageSize = 20;
-    private static final int documentAmount = 50;
+    private static final int defaultPageSize = OperationsConfig.DEFAULT_PAGE_SIZE;
+    private static final int documentAmount = 120;
```
amorton: nit: I know you did not write this, but this var name is not great and it is not great.  Lets make some tidy up in var names, and comments if needed to make it clear we want to test getting into a 3rd page that is partial in size.

Can we
- rename `documentAmount` to `totalDocuments`, and change it to calculated as (2 * pageSize) + lastPageSize , make lastPageSize an int = 20 
- change the last test to just check against lastPageSize

src/test/java/io/stargate/sgv2/jsonapi/api/v1/UpdateManyIntegrationTest.java
```diff
@@ -182,7 +182,7 @@ public void limit() {
 
     @Test
     public void limitMoreDataFlag() {
-      insert(25);
+      insert(55);
```
amorton: nit: lets pull this into a class level variable, make it DEFAULT_PAGE_SIZE + someVarName where someVarName has some meaning like "partialPageSize"

src/test/java/io/stargate/sgv2/jsonapi/api/v1/UpdateManyIntegrationTest.java
```diff
@@ -165,8 +165,8 @@ public void limit() {
           .body("$", responseIsStatusOnly())
           .body("status.matchedCount", is(20))
           .body("status.modifiedCount", is(20))
-          .body("status.moreData", is(true))
-          .body("status.nextPageState", not(emptyOrNullString()));
+          .body("status.moreData", nullValue())
```
amorton, line 174: I *think* this test is doing the correct thing, but why was the previous version of this code working ? 

The previous default page size was 20, so I wonder why it said there was more data ? Perhaps because we did not know and it's not authorative , i.e. we can get a false positive 

This test is now checking that the moreData is false / not present when we expect. Can we update so there is a class level var called something like `partialPageSize` = DEFAULT_PAGE_SIZE - 20 . See if it makes sense to do this as a local var, or class level, and if there is other places where a good var name makes it clearer. 

That may make it clearer that we are checking what happens when clearly less than one page is there for updates. Prob also want to change the function name. 

BUT that also makes me wonder what do we expect to happen when the DEFAULT_PAGE_SIZE number of docuemnts are there

src/test/java/io/stargate/sgv2/jsonapi/service/resolver/FindAndRerankOperationBuilderTest.java
```diff
@@ -0,0 +1,248 @@
...
+import io.stargate.sgv2.jsonapi.service.schema.collections.CollectionSchemaObject;
+import io.stargate.sgv2.jsonapi.service.schema.collections.IdConfig;
+import io.stargate.sgv2.jsonapi.testresource.NoGlobalResourcesTestProfile;
+import jakarta.inject.Inject;
+import java.util.List;
+import java.util.Optional;
+import org.junit.jupiter.api.Test;
+
+@QuarkusTest
+@TestProfile(NoGlobalResourcesTestProfile.Impl.class)
```
amorton, line 28: I think we discussed this was not needed if all it was doing was testing the deserailisation and the errors. 

we have the deserislisation as unit tests, and do we have the errors in integration tests ? 

We can keep it if it is not covered in other places

amorton (changes requested):
some changes as discussed, getting into the weeds with things now :)

src/test/java/io/stargate/sgv2/jsonapi/api/v1/UpdateManyIntegrationTest.java
erichare: That was my thought. I was thinking that the code might assume that, if we return up to the page size, return that there is more data just so the user is forced to check. But that seems odd. I've made the changes like you said to hopefully make this more clear - and it checks the right behavior both for a partial page, and for a full page + some amount. Let me know how this looks?

src/test/java/io/stargate/sgv2/jsonapi/service/resolver/FindAndRerankOperationBuilderTest.java
erichare: Yeah, deserialization is covered separately, and there’s integration coverage for the public error path. I think this test is still worth keeping because it covers the builder-specific piece: validating against OperationsConfig bounds and setting the hybrid limits on the command context.

erichare:
@amorton I believe i just pushed fixes / comments for all your review points!

src/main/java/io/stargate/sgv2/jsonapi/config/OperationsConfig.java
```diff
@@ -32,21 +33,50 @@
...
+
+  /**
+   * Built-in default number of documents returned by findAndRerank. The effective default may be
+   * overridden via application.yaml, system properties, or environment variables (see {@link
+   * #defaultFindAndRerankLimit()}).
+   */
+  int DEFAULT_FIND_AND_RERANK_LIMIT = 10;
+
+  /** Built-in minimum hybrid search read limit (may be overridden via configuration). */
+  int MIN_HYBRID_SEARCH_LIMIT = RerankingConstants.HybridSearchLimits.MIN;
```
amorton: these three vars have a single use, and all these are doing is pulling from anotehr cons. Can we cut out the middle man and just reference RerankingConstants.HybridSearchLimits

src/main/resources/application.conf
```diff
@@ -77,8 +77,8 @@ datastax-java-driver {
     table-read {
       basic.request.timeout = 10 seconds
       // this will be the number of rows we return in a single page of results to the user.
-      // It may be overridded for in memory sorting, but this is the default.
-      basic.request.page-size = 20
+      // It may be overridden for in memory sorting, but this is the default.
+      basic.request.page-size = 50
```
amorton, line 82: we should also set this after line 53, at the same level they have there. 

that will be setting it for the default profile

src/test/java/io/stargate/sgv2/jsonapi/service/resolver/FindAndRerankOperationBuilderTest.java
```diff
@@ -0,0 +1,248 @@
...
+        .withCommand(commandHigh)
+        .withFindCommandResolver(findCommandResolver)
+        .build();
+  }
+
+  private FindAndRerankCommand command(String json) throws Exception {
+    return objectMapper.readValue(json, FindAndRerankCommand.class);
+  }
+
+  private CommandContext<CollectionSchemaObject> commandContext() {
```
amorton, line 202: -1 on using the string name for the command, there is an enum of CommandName. 

If you jump into TestConstants make an overload there such as 


``` public CommandContext<CollectionSchemaObject> collectionContext(
      CommandName  commandName,
      CollectionSchemaObject schema){}
 ```

File: src/test/java/io/stargate/sgv2/jsonapi/service/resolver/FindAndRerankOperationBuilderTest.java [hunk 41]
```diff
@@ -0,0 +1,248 @@
...
+    var commandContext =
+        testConstants.collectionContext("findAndRerank", vectorLexicalRerankSchema(), null, null);
+
+    var rerankingProvidersConfig = mock(RerankingProvidersConfig.class);
+    var modelConfig = mock(RerankingProvidersConfig.RerankingProviderConfig.ModelConfig.class);
+    when(modelConfig.apiModelSupport())
```
amorton, line 210: is there some whay we can refactor to not need to do all this mocking override ? It's kind of fragile, and suggests the class model is not made for testing. 

IMHO all we are testing in here is user input validation for the limit, but we are mocking and pretendsing models are supporting. Something is harder that it needs to be. And it's partially because the RerankingProviderConfig is a pain of a data model. 

If it works, we can get this merged in and then do a fast follow to try to make this more testable. That was one nice thing about the limits being checked in the custom deserializer before, it was an easy unit test. 

lets merge and noodle on this

src/test/java/io/stargate/sgv2/jsonapi/service/resolver/FindAndRerankOperationBuilderTest.java
```diff
@@ -0,0 +1,248 @@
...
+    when(rerankingProvidersConfig.filterByRerankServiceDef(any())).thenReturn(modelConfig);
+    when(commandContext.rerankingProviderFactory().getRerankingConfig())
+        .thenReturn(rerankingProvidersConfig);
+    when(commandContext.rerankingProviderFactory().create(any(), any(), any(), any(), any(), any()))
+        .thenReturn(mock(RerankingProvider.class));
+
+    return commandContext;
+  }
+
+  private CollectionSchemaObject vectorLexicalRerankSchema() {
```
amorton: -1 on this function  have a look at how the TestConstants class works and follow the pattern there. Otherwise we end up with too much code duplication and fragile tests.

i.e. add another schema object like the VECTOR_COLLECTION_SCHEMA_OBJECT - follow the code layout

amorton (approved):
there are a few things to tweak, can you do those and then merge. Including in FindAndRerankCollectionIntegrationTest

we can then look at the issues of need for basic testing of the limits needing so much boiler plate in the FindAndReRankOperationsBuilder.

src/test/java/io/stargate/sgv2/jsonapi/service/resolver/FindAndRerankOperationBuilderTest.java
erichare: Agreed. I’ll leave this as-is for this PR

PR #2471, opened 2026-05-14
Fix #2468 alterCollection, approach 2 (like "alterTable") -- PRIMARY option

### LINKED ISSUE #2468: Add command `alterCollection`, initially to allow enabling "lexical" feature after creation
Currently settings of Collections are immutable after `createCollection` has created them.

But we need to be able to change some aspects, starting with ability to enable support for "lexical" (addition of column "$lexical" with analyzed text index).

For enabling lexical, no re-indexing occurs: it just allows use of `$lexical` field (via matching column) and indexing of newly inserted or updated Documents.

Command should be `CollectionCommand` to connect at same URL path as `AlterTable`, linked from `CollectionResource`, and take Collection Name as Path parameter.

Command should initially accept an Object argument with single entry -- "lexical" -- accepting same value type as `createCollection`s "options.lexical" field. So:

```
  {
    "alterCollection": {
      "lexical": { "enabled": true, "analyzer": "standard" }
    }
  }
```

Q? Should this instead look more like "alterTable"? So:

```
{
  "alterCollection": {
    "operation": {
      "enableLexical": { "analyzer": "standard" }
    }
  }
}
```

For now, go with Option 1.

Only `{"enabled": true}` case is to be supported in the first phase.

Return value should be simple "ok":

```
"status": {
  "ok": 1
}
```

Command should:

1. Fail if unrecognized root-level fields encountered (only "lexical" recognized)
2. Fail if "lexical" was to be disabled
3. Fail if "lexical" feature not enabled for the Database (same as for "CreateCollectionCommand")
4. Succeed without action if "lexical" feature already enabled for Collection with same settings; fail if settings differ
5. Try to:
      1. create "$lexical" column, create text-analyzed index for it (same as with "CreateCollectionCommand"), update "comment" of underlying Table
      2. Succeed if all above steps succeeded; fail otherwise

Note that this command does not have counterpart in MongoDB.

tatu-at-datastax:
**What this PR does**:

Implements "alterCollection" command to allow enabling of "lexical" indexing, with "alterTable" style

Fixes #2468

**Checklist**
- [x] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/AlterCollectionCommand.java
```diff
@@ -0,0 +1,25 @@
...
+import io.stargate.sgv2.jsonapi.api.model.command.CollectionCommand;
+import io.stargate.sgv2.jsonapi.api.model.command.CommandName;
+import io.stargate.sgv2.jsonapi.api.model.command.NoOptionsCommand;
+import org.eclipse.microprofile.openapi.annotations.media.Schema;
+
+@Schema(
+    description =
+        "Command that alters mutable settings of an existing collection. Currently supports enabling the 'lexical' feature.")
+@JsonTypeName(CommandName.Names.ALTER_COLLECTION)
+public record AlterCollectionCommand(AlterCollectionOperation operation)
```
tatu-at-datastax, line 13: Modeled after `AlterTableCommand`, same approach with polymorphich operation types.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/AlterCollectionOperation.java
```diff
@@ -0,0 +1,14 @@
...
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+
+/**
+ * Polymorphic operation payload for {@link AlterCollectionCommand}. Each operation is represented
+ * by a record implementing this interface; Jackson selects the concrete subtype by the wrapper key
+ * (e.g. {@code "enableLexical"}). Mirrors {@link AlterTableOperation}.
+ */
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, include = JsonTypeInfo.As.WRAPPER_OBJECT)
+@JsonSubTypes({@JsonSubTypes.Type(value = AlterCollectionOperationImpl.EnableLexical.class)})
+public sealed interface AlterCollectionOperation
```
tatu-at-datastax, line 13: Modeled after `AlterTableOperation`, but uses sealed types to reduce likelihood of missing handlers

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/AlterCollectionOperationImpl.java
```diff
@@ -0,0 +1,27 @@
...
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonTypeName;
+import com.fasterxml.jackson.databind.JsonNode;
+import java.util.Map;
+import javax.annotation.Nullable;
+import org.eclipse.microprofile.openapi.annotations.media.Schema;
+
+/** Each operation that {@link AlterCollectionCommand} understands is represented by a record. */
+public class AlterCollectionOperationImpl {
```
tatu-at-datastax, line 12: and this is modeled after `AlterTableOperationImpl`.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/collections/CreateCollectionOperation.java
```diff
@@ -666,4 +664,42 @@ private SimpleStatement buildSaiIndex(
...
+    var analyzerDef = lexicalConfig.analyzerDefinition();
+    // Note: needs to be either plain (unquoted) String (NOT quoted JSON String) OR JSON Object
+    final String analyzerString =
+        analyzerDef.isTextual() ? analyzerDef.asText() : analyzerDef.toString();
+    final String lexicalCol = DocumentConstants.Columns.LEXICAL_INDEX_COLUMN_NAME;
+    final String prefix = ifNotExists ? "CREATE CUSTOM INDEX IF NOT EXISTS" : "CREATE CUSTOM INDEX";
+    return SimpleStatement.newInstance(
+            """
+                %s "%s" ON "%s"."%s" (%s)
+                  USING 'StorageAttachedIndex' WITH OPTIONS = { 'index_analyzer': '%s' }
```
erichare, line 686: Its probably not a huge concern, but my mind is still thinking about CQL injection from the keyspace work... we're not escaping `'` in the analyzer string here, which i guess could be a risk? maybe we can just defensively do something like

`analyzerString.replace("'", "''")`

before it goes into the statement?

src/main/resources/errors.yaml
```diff
@@ -2025,13 +2033,13 @@ request-errors:
 
   - scope: SCHEMA
     code: TOO_MANY_INDEXES_FOR_COLLECTION
-    title: Cannot create collection due to number of existing indexes
+    title: Cannot create indexes for collection due to number of existing indexes
```
erichare, line 2036: Do we want this change generally? Even when the user tries to create a collection directly?

erichare (approved):
LGTM @tatu-at-datastax . I approved it, with two comments left, one that maybe could be addressed and one that's just more of a question. Looks really well done, nice tests.

tatu-at-datastax:
> LGTM @tatu-at-datastax . I approved it, with two comments left, one that maybe could be addressed and one that's just more of a question. Looks really well done, nice tests.

Thank you @erichare -- good points esp. on quoting/escaping.

I'll leave this open for @amorton and @Hazel-Datastax to consider too, from strategic perspective (if it makes sense).

PR #2472, opened 2026-05-18
security: validate keyspace and datacenter names to prevent CQL injection (GHSA-p64p-96pw-mxwv)

erichare:
Fix for [GHSA-p64p-96pw-mxwv](https://github.com/stargate/data-api/security/advisories/GHSA-p64p-96pw-mxwv).

> Originally opened as [stargate/data-api-ghsa-p64p-96pw-mxwv#1](https://github.com/stargate/data-api-ghsa-p64p-96pw-mxwv/pull/1) in the private advisory fork; re-opened here on a branch in the main repo per team request.

## Summary

Two unrelated CQL-injection sinks share a common root cause: user-supplied identifiers are interpolated into CQL via `String.format` without validation or escaping.

1. **`dropKeyspace` / `dropNamespace`** — the request `name` is passed directly into `DROP KEYSPACE IF EXISTS "%s"`. The analogous Create resolvers already validate via `NamingRules.KEYSPACE.checkRule(name)`; the Drop resolvers do not. A double-quote character breaks out of the quoted identifier.
2. **NetworkTopologyStrategy replication map** — `CreateNamespaceKeyspaceCommandResolver.networkTopologyStrategyMap` interpolates user-supplied datacenter-name keys via `", '%s': %d"` with no validation. A single-quote in the key breaks out of the literal and lets an attacker inject CQL into the `WITH REPLICATION = {…}` clause.

## Changes

- `DropKeyspaceCommandResolver`, `DropNamespaceCommandResolver` — apply `NamingRules.KEYSPACE.checkRule(name)`, mirroring the Create path.
- `CreateNamespaceKeyspaceCommandResolver` — validate each NetworkTopologyStrategy datacenter-name key against an allowlist (`[A-Za-z0-9_-]+`, max 48 chars). The character set accepts realistic cloud DC names like `us-east-1` while rejecting any character that could break the single-quoted CQL literal (notably `'`, `\`, whitespace).
- New error code `SchemaException.Code.UNSUPPORTED_REPLICATION_DATA_CENTER_NAME` with template in `errors.yaml`.

`Integer` replication-factor values are unaffected — Jackson type-coercion rejects non-numeric input and `%d` only formats numerics.

## Test plan

- [x] New unit tests in `DropKeyspaceCommandResolverTest`:
  - rejects double-quote injection payload
  - rejects empty name
  - rejects hyphen / special chars
  - same protection on the deprecated `dropNamespace` alias
- [x] New unit tests in `CreateKeyspaceCommandResolverTest`:
  - rejects single-quote injection payload in DC-name key
  - rejects DC name with whitespace
  - rejects DC name over the 48-char limit
  - accepts hyphenated cloud-style DC name (`us-east-1`) — regression guard for legitimate users
- [x] `./mvnw test -Dtest='DropKeyspaceCommandResolverTest,CreateKeyspaceCommandResolverTest'` — 21 / 21 pass.
- [ ] CI: full test + integration-test suite on this branch.
- [ ] Manual smoke: `createKeyspace` + `dropKeyspace` happy path against a real cluster.
- [ ] Confirm `vulnerable_version_range` on the advisory before publishing the GHSA (currently `<= 1.0.x` as a placeholder; current `pom.xml` is `1.0.47-SNAPSHOT`).

## Notes for reviewers

- The DC allowlist intentionally includes `-` so Astra / AWS-region style DC names (`dc-westus2`, `us-east-1`) keep working. If you'd prefer the stricter `\w+` rule used for keyspace identifiers, flag it and I can tighten.
- I did not change `DropKeyspaceOperation.DROP_KEYSPACE_CQL` itself; the validated identifier is still interpolated. The fix relies on the allowlist being CQL-safe. Switching to a `QueryBuilder` API or driver-side `CqlIdentifier` quoting would be a larger refactor and is out of scope for this patch.


tatu-at-datastax (approved):
LGTM

PR #2473, opened 2026-05-18
Add billing event logging for embedding and reranking token usage

### LINKED ISSUE #2137: Emit model usage events
Now we have #2008 add feature flags to configure emitting model usage events in headers. 

We want billing events for the egress bytes, ingress bytes, and token usage. 

Use feature flags to enable / disable at config time and at request time. 

As part of this, also add metrics for ModelUsage

Hazel-Datastax:
**What this PR does**:
Add billing event logging for embedding and reranking token usage.

Event definition below

All providers configured as "inhouse" (?) are treated diff to others which are assumed to be external.
Sample logs:
```
// internal model usage
{
	"id": "f6716d5a-7ca4-4afa-9826-a14561b8c0b7",
	"timestamp": "2026-05-22T00:43:30.549312Z",
	"product": "serverless",
	"event_type": "internal_model_total_tokens",
	"properties": {
		"usage": 7,
		"region": "CASSANDRA_REGION",
		"resource_type": "serverless_database",
		"resource_id": "SINGLE-TENANT",
		"provider" : "nvidia",
		"model" : "nvidia/llama-3.2-nv-rerankqa-1b-v2"
	}
}

// external model usage
{
	"id": "98cb204d-8f8b-4784-aab6-0b7e96b9695d",
	"timestamp": "2026-05-22T00:43:30.549339Z",
	"product": "serverless",
	"event_type": "external_model_total_tokens",
	"properties": {
		"usage": 21832,
		"region": "CASSANDRA_REGION",
		"resource_type": "serverless_database",
		"resource_id": "SINGLE-TENANT",
		"provider" : "openai",
		"model" : "text-embedding-3-small"
	}
}

// byte used in the request, split for internal and external 

{
	"id": "aede7d63-eb64-4c86-9dc3-a8432c07f120",
	"timestamp": "2026-05-22T00:43:30.549335Z",
	"product": "serverless",
	"event_type": "internal_model_egress_bytes OR external_model_egress_bytes",
	"properties": {
		"usage": 84,
		"region": "CASSANDRA_REGION",
		"resource_type": "serverless_database",
		"resource_id": "SINGLE-TENANT",
		"provider" : "nvidia",
		"model" : "nvidia/llama-3.2-nv-rerankqa-1b-v2"
	}
}

// byte used in the response, split for internal and external 

{
	"id": "aede7d63-eb64-4c86-9dc3-a8432c07f120",
	"timestamp": "2026-05-22T00:43:30.549335Z",
	"product": "serverless",
	"event_type": "internal_model_ingress_bytes OR external_model_ingress_bytes",
	"properties": {
		"usage": 84,
		"region": "CASSANDRA_REGION",
		"resource_type": "serverless_database",
		"resource_id": "SINGLE-TENANT",
		"provider" : "nvidia",
		"model" : "nvidia/llama-3.2-nv-rerankqa-1b-v2"
	}
}
```

Event Def :

```json
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "type": "object",
  "properties": {
    "id": {
      "type": "string",
      "format": "uuid"
    },
    "timestamp": {
      "type": "string",
      "format": "date-time"
    },
    "event_type": {
      "type": "string",
      "enum": ["third_party_embeddings_egress"]
    },
    "product": {
      "type": "string",
      "enum": ["serverless", "managed_cluster", "pcu", "streaming"]
    },
    "properties": {
      "type": "object",
      "properties": {
        "usage": {
          "type": "number"
        },
        "org_id": {
          "type": "string"
        },
        "region": {
          "type": "string",
          "enum": [
            "us-east-1", "us-east-2", "us-west-1", "us-west-2",
            "af-south-1", "ap-east-1", "ap-south-1", "ap-south-2", 
            "ap-southeast-1", "ap-southeast-2", "ap-southeast-3", "ap-northeast-1",
            "ap-northeast-2", "ap-northeast-3", "ca-central-1", "eu-central-1",
            "eu-central-2", "eu-west-1", "eu-west-2", "eu-west-3", "eu-south-1",
            "eu-south-2", "eu-north-1", "me-south-1", "me-central-1", "sa-east-1",
            "eastus", "eastus2", "southcentralus", "westus", "westus2", 
            "westus3", "centralus", "northcentralus", "southafricanorth", 
            "southafricawest", "australiaeast", "australiasoutheast", 
            "australiacentral", "australiacentral2", "brazilsouth", "brazilsoutheast", 
            "canadacentral", "canadaeast", "centralindia", "japaneast", "japanwest", 
            "koreacentral", "koreasouth", "norwayeast", "norwaywest", "switzerlandnorth", 
            "switzerlandwest", "uaecentral", "uaenorth", "eastasia", "southeastasia", 
            "westeurope", "northeurope", "francecentral", "francesouth", 
            "germanynorth", "germanywestcentral", "uksouth", "ukwest", "swedencentral", 
            "swedensouth", "polandcentral",
            "us-central1", "us-east1", "us-east4", "us-west1", "us-west2", 
            "us-west3", "us-west4", "northamerica-northeast1", "northamerica-northeast2", 
            "southamerica-east1", "southamerica-west1", "europe-west1", 
            "europe-west2", "europe-west3", "europe-west4", "europe-west6", 
            "europe-central2", "asia-east1", "asia-east2", "asia-northeast1", 
            "asia-northeast2", "asia-northeast3", "asia-south1", "asia-south2", 
            "asia-southeast1", "asia-southeast2", "australia-southeast1", 
            "australia-southeast2", "me-west1", "me-central1"
          ]
        },
        "resource_id": {
          "type": "string"
        },
        "resource_type": {
          "type": "string",
          "enum": ["serverless_database", "streaming_tenant", "managed_cluster_database", "pcu_group"]
        },
        "provider_id": {
          "type": "string"
        }
      },
      "required": [
        "usage",
        "region",
        "resource_id",
        "resource_type"
      ]
    }
  },
  "required": [
    "id",
    "timestamp",
    "event_type",
    "product",
    "properties"
  ]
}
```
Fixes #2137

**Checklist**
- [x] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/service/provider/BillingEventLogger.java
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  /**
+   * Builds the billing event JSON string, or returns {@code null} if the tenant is not Astra.
+   * Package-private for testing.
+   */
+  static String buildBillingEventJson(ModelUsage modelUsage) {
+    if (modelUsage.tenant().databaseType() != DatabaseType.ASTRA) {
+      return null;
+    }
```
Hazel-Datastax: Not sure if we want this. Can remove if we want to log it anyway

src/main/java/io/stargate/sgv2/jsonapi/service/provider/ModelProvider.java
```diff
@@ -46,6 +46,14 @@ public String apiName() {
     return apiName;
   }
 
+  /** Returns the billing region for this provider, or {@code null} if not applicable. */
+  public String billingRegion() {
+    return switch (this) {
+      case NVIDIA -> "us-west-2";
+      default -> null;
```
Hazel-Datastax: I assume only Nvidia needs "region"?

src/main/java/io/stargate/sgv2/jsonapi/service/provider/BillingEventLogger.java
```diff
@@ -0,0 +1,53 @@
...
+ * for invoicing.
+ */
+public final class BillingEventLogger {
+
+  private static final Logger BILLING_LOGGER = LoggerFactory.getLogger("billing.events");
+  private static final Logger LOGGER = LoggerFactory.getLogger(BillingEventLogger.class);
+  private static final ObjectWriter OBJECT_WRITER = new ObjectMapper().writer();
+
+  private static final String PRODUCT = "serverless";
+  private static final String RESOURCE_TYPE = "serverless_database";
```
Hazel-Datastax: Not sure about the value, just hardcoded the value here

src/main/java/io/stargate/sgv2/jsonapi/service/provider/BillingEvent.java
```diff
@@ -0,0 +1,93 @@
...
+import java.util.UUID;
+
+/**
+ * A billing event emitted as a structured JSON log line for downstream billing pipelines.
+ *
+ * <p>Each event represents the aggregated token usage from a single user request to a model
+ * provider (embedding or reranking). Created via {@link #from(ModelUsage, String, String)} after
+ * batch aggregation.
+ *
+ * @param id Unique UUID v4 identifier for this event.
```
tatu-at-datastax: suggestion: "UUID v4 (random-based)"

(I always forget which is which :) )

src/main/java/io/stargate/sgv2/jsonapi/service/provider/BillingEventLogger.java
tatu-at-datastax: Was wondering about same: for now hard-coding may actually make most sense -- only make configurable if/when needed.

src/main/java/io/stargate/sgv2/jsonapi/service/provider/BillingEventLogger.java
tatu-at-datastax: No strong opinion, but I think check for whether to log should be in method above (`logBillingEvent()`), not here.

src/main/java/io/stargate/sgv2/jsonapi/service/provider/ModelProvider.java
tatu-at-datastax: I do not know, better check with Aaron.

src/main/java/io/stargate/sgv2/jsonapi/service/reranking/operation/RerankingProvider.java
```diff
@@ -204,6 +204,10 @@ private RerankingResponse aggregateRanks(List<BatchedRerankingResponse> batchRes
         finalRanks.add(new Rank(batchStartIndex + rank.index(), rank.score()));
       }
     }
+    // Emit billing event with aggregated token usage across all batches
+    if (aggregatedModelUsage != null) {
```
tatu-at-datastax: Can this be `null`? If not, remove check.

src/main/java/io/stargate/sgv2/jsonapi/service/provider/BillingEvent.java
```diff
@@ -0,0 +1,93 @@
...
+ * @param properties Usage details including token count, region, and resource identifiers.
+ */
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public record BillingEvent(
+    String id,
+    String timestamp,
+    String product,
+    @JsonProperty("event_type") String eventType,
+    BillingProperties properties)
+    implements Recordable {
```
tatu-at-datastax, line 51: Not made any use of yet, so could leave out. But I guess it's ok to add as it's not a lot of code.

src/main/java/io/stargate/sgv2/jsonapi/service/provider/BillingEventLogger.java
```diff
@@ -0,0 +1,53 @@
...
+
+  private BillingEventLogger() {}
```
tatu-at-datastax: I know we have various opinions on usefulness of Quarkus DI but I wonder if `BillingEventLogger` should be injectable (if easy enough to achieve) since any configurability would require passing it `OperationsConfig` or one of other config objects.

We can do it later if need be, but if easy enough to do now, would be preferable.

But even if not injected, I think that we should avoid `static` call pattern and pass `BillingEventLogger` instance (even if instance for now is global singleton) to callers.

tatu-at-datastax (commented):
Looks good, some thoughts added.

src/main/java/io/stargate/sgv2/jsonapi/service/provider/BillingEvent.java
amorton: is used by tracing, which is not wide spread but there

src/main/java/io/stargate/sgv2/jsonapi/service/provider/BillingEvent.java
```diff
@@ -0,0 +1,93 @@
...
+ *
+ * @param id Unique UUID v4 identifier for this event.
+ * @param timestamp ISO 8601 timestamp of when the event was created.
+ * @param product Product identifier, e.g. {@code "serverless"}.
+ * @param eventType Event type in the format {@code {provider}_{modelType}_tokens}, e.g. {@code
+ *     "nvidia_embeddings_tokens"}.
+ * @param properties Usage details including token count, region, and resource identifiers.
+ */
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public record BillingEvent(
```
amorton, line 45: add a canonical constructor to validate required fields are set, can we update the fields to be strongly typed I am guessing that impacts the JSON serialisation as well. 

e.g. id as uuid, timestamp as instant

src/main/java/io/stargate/sgv2/jsonapi/service/provider/BillingEvent.java
```diff
@@ -0,0 +1,93 @@
...
+   * Billing event properties containing usage metrics and resource identifiers.
+   *
+   * @param usage Total token count from the provider response.
+   * @param region Deployment region for the provider, or {@code null} if not applicable (omitted
+   *     from JSON).
+   * @param resourceType Type of the billed resource, e.g. {@code "serverless_database"}.
+   * @param resourceId The tenant/database ID used to infer the billing tenant.
+   */
+  @JsonInclude(JsonInclude.Include.NON_NULL)
+  public record BillingProperties(
```
amorton, line 82: nit: move inner to the bottomo of class

src/main/java/io/stargate/sgv2/jsonapi/service/provider/BillingEvent.java
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  /**
+   * Creates a billing event from an aggregated {@link ModelUsage}.
+   *
+   * @param modelUsage Aggregated model usage from a completed provider request.
+   * @param product Product identifier from configuration.
+   * @param resourceType Resource type from configuration.
+   */
+  public static BillingEvent from(ModelUsage modelUsage, String product, String resourceType) {
```
amorton: can we move the logic for this into the model usage, making the BillingEvent more reusable / less bound just to model usage . Or move it into the BillingEventLogger class, so the logic is in one place

src/main/java/io/stargate/sgv2/jsonapi/service/provider/BillingEvent.java
```diff
@@ -0,0 +1,93 @@
...
+    String eventType =
+        modelUsage.modelProvider().apiName()
+            + "_"
+            + modelUsage.modelType().billingName()
+            + "_tokens";
+
+    var properties =
```
amorton: we should have a BillingEvent for the egress / request bytes, called something like "${MODEL_PROVIDER_API_NAME}_egress_bytes"

src/main/java/io/stargate/sgv2/jsonapi/service/provider/BillingEvent.java
```diff
@@ -0,0 +1,93 @@
...
+        new BillingProperties(
+            modelUsage.totalTokens(),
+            modelUsage.modelProvider().billingRegion(),
```
amorton: this should be the region the DB is in, there may be an ENV var already set but I dont think so. 

so we may need a new configuration option for the billing region

src/main/java/io/stargate/sgv2/jsonapi/service/provider/BillingEventLogger.java
```diff
@@ -0,0 +1,53 @@
...
+    if (json != null) {
+      BILLING_LOGGER.info(json);
+    }
+  }
+
+  /**
+   * Builds the billing event JSON string, or returns {@code null} if the tenant is not Astra.
+   * Package-private for testing.
+   */
+  static String buildBillingEventJson(ModelUsage modelUsage) {
```
amorton: use the ApiFeature to control if we want to do billing, so we can have it set at startup and also controlled at request time. 

NOTE: please check that if enabled at startup a request CANNOT turn it off :)

src/main/java/io/stargate/sgv2/jsonapi/service/provider/BillingEventLogger.java
```diff
@@ -0,0 +1,53 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Logs billing events as structured JSON via a dedicated {@code billing.events} logger.
+ *
+ * <p>Only emits events when the tenant is Astra. Downstream billing pipelines parse these log lines
+ * for invoicing.
+ */
+public final class BillingEventLogger {
```
amorton: Consider calling this something less tied to how the billing is handled , like just `Billing` 

e.g. then you have `Blling.bill(ModelUsage)` or `bill(Foo)` and the user of the class does not care what happens

src/main/java/io/stargate/sgv2/jsonapi/service/provider/BillingEventLogger.java
amorton: we will need instance if we can use the ApiFeatures, so maybe hand on instance off the CommandContent because it will need to know the request context.

src/main/java/io/stargate/sgv2/jsonapi/service/provider/BillingEventLogger.java
amorton: from discussions - tatu point is do not build the billing event unless you need to, we need to check a feature flag AND if the logger is enabled at the level

src/main/java/io/stargate/sgv2/jsonapi/service/provider/ModelProvider.java
amorton: from discussing, in other parts, this is the region of the db, prob needs a new config

src/main/java/io/stargate/sgv2/jsonapi/service/reranking/operation/RerankingProvider.java
amorton: from discussion, logBillingEvent() may be the best place if we want to check this

src/main/resources/application.yaml
```diff
@@ -127,6 +127,8 @@ quarkus:
         level: INFO
       'io.stargate':
         level: DEBUG
+      'billing.events':
+        level: INFO
```
amorton, line 142: for info, maybe make a change. Below is an example of setting a logger category to use a different appender / handler

PLAIN_CONSOLE - defines a new appender / handler that will write using a different format. It is a handler of type console , named PLAIN_CONSOLE, with a different formatt than the default handler called "console"

NOTE: we will want to check with billing peeps what format they want. 

this `%c %m$n` from your code would log 

```
billing.event {"the": json"}
```

take away %c to just log the json on a line by itself. 

the category: `io.stargate.sgv2.jsonapi.testbench.reporting.TestBenchConsoleWriter` says for this named logger, its level is INFO and it would use the PLAIN_CONSOLE appender / handler, AND importantly DO NOT inherit from above - so do not also use the `console` appender  

```yaml
 log:
    console:
      format: "%-5p [%t] %d{yyyy-MM-dd HH:mm:ss,SSS} %F:%L - %m%n"
    handler:
      console:
        PLAIN_CONSOLE:
          format: "%m%n"
    category:
      # production log level for this category is DEBUG, way too noisy for tests
      'io.stargate':
      # Fine for JVM mode; if building native images, use index-dependency instead
      'io.quarkus.deployment.steps.ReflectiveHierarchyStep':
        level: ERROR
      'io.stargate.sgv2.jsonapi.testbench.reporting.TestBenchConsoleWriter':
        level: INFO
        handlers:
          - PLAIN_CONSOLE
        use-parent-handlers: false
```

amorton (approved):
some tweaks as discussed, heading in the right direction

amorton (changes requested):
hit wrong button, tweaks as discussed

tatu-at-datastax:
I think issue for this is #2137 fwtw.

src/main/java/io/stargate/sgv2/jsonapi/api/request/tenant/Tenant.java
```diff
@@ -29,30 +29,63 @@ public class Tenant implements Recordable {
...
+   * don't have a per-request region concept, so all Cassandra tenants report this fixed string.
+   */
+  public static final String CASSANDRA_REGION_DEFAULT = "CASSANDRA_REGION";
+
+  /**
+   * Fallback region value for multi-tenant databases (e.g. Astra) when the request URL doesn't
+   * carry a parseable region segment. Downstream billing pipelines can detect and investigate these
+   * events.
+   */
+  public static final String UNKNOWN_REGION = "unknown";
```
amorton: from discussion - lets go with `unknown-region`

src/main/java/io/stargate/sgv2/jsonapi/api/request/tenant/Tenant.java
```diff
@@ -29,30 +29,63 @@ public class Tenant implements Recordable {
...
+      new Tenant(DatabaseType.CASSANDRA, SINGLE_TENANT_ID, CASSANDRA_REGION_DEFAULT);
 
   private final DatabaseType databaseType;
   private final String tenantId;
+  private final String region;
 
-  private Tenant(DatabaseType databaseType, String tenantId) {
+  private Tenant(DatabaseType databaseType, String tenantId, String region) {
     this.databaseType = databaseType;
     // this is just for sanity checking, keep it lower case
```
amorton: nit - remove the comment on lower case, that was for old code

src/main/java/io/stargate/sgv2/jsonapi/api/request/tenant/Tenant.java
```diff
@@ -98,11 +137,15 @@ public boolean equals(Object obj) {
       return false;
     }
     return Objects.equals(databaseType, that.databaseType)
-        && Objects.equals(tenantId, that.tenantId);
+        && Objects.equals(tenantId, that.tenantId)
```
amorton: BUG - the same fields should be used in hash as in equals

From discussions, I think we can exclude the region from the equals() and hash() and comment that we dont use it because the tenant is the same in every region. Include that the tenant is used as part of the SchemaObjectIdentifier and that is used in the schema object cache key

also, pls update the function comment and remove case insensitive part

src/main/java/io/stargate/sgv2/jsonapi/config/feature/ApiFeature.java
```diff
@@ -69,7 +69,20 @@ public enum ApiFeature {
   /**
    * The request will return a trace of the processing that includes both the message and the data.
    */
-  REQUEST_TRACING_FULL("request-tracing-full", false);
+  REQUEST_TRACING_FULL("request-tracing-full", false),
+
+  /**
+   * Billing feature flag: if enabled, the API will emit per-request billing events (token usage and
```
amorton: :)

src/main/java/io/stargate/sgv2/jsonapi/service/provider/Billing.java
```diff
@@ -0,0 +1,131 @@
...
+ *   <li>(NVIDIA only) a {@code nvidia_gpu_plane_egress_bytes} event with {@link
+ *       ModelUsage#responseBytes()} — bytes coming back from the GPU plane
+ * </ul>
+ *
+ * <p>Eligibility requires all of: {@link ApiFeature#BILLING} is enabled, the {@code billing.events}
+ * logger is enabled, and the {@link ModelUsage} is non-null. The region for each event is read from
+ * {@link ModelUsage#tenant()} ({@code Tenant.region()}).
+ */
+@ApplicationScoped
+public class Billing {
```
amorton, line 40: rather than inject, we will put this on the RequestContext after #2479 is merged. 

sing the CDI, and then requiring the APiFeatures to be passed on the method call is a pain and has added 2 params in places were we dont need any

src/main/java/io/stargate/sgv2/jsonapi/service/provider/Billing.java
```diff
@@ -0,0 +1,131 @@
...
+      } catch (JacksonException e) {
+        LOGGER.error("Failed to serialize billing event of type {}", event.eventType(), e);
+      }
+    }
+  }
+
+  /**
+   * Whether a billing event should be emitted for the given request. Package-private for testing.
+   */
+  boolean shouldEmit(ModelUsage modelUsage, ApiFeatures apiFeatures) {
```
amorton: private ?

src/main/java/io/stargate/sgv2/jsonapi/service/provider/Billing.java
```diff
@@ -0,0 +1,131 @@
...
+  /**
+   * Builds the list of billing events for one {@link ModelUsage}: a {@code *_tokens} event with
+   * {@link ModelUsage#totalTokens()}, a {@code {provider}_egress_bytes} event with {@link
+   * ModelUsage#requestBytes()}, and for NVIDIA also a {@code nvidia_gpu_plane_egress_bytes} event
+   * with {@link ModelUsage#responseBytes()}. Package-private for testing.
+   *
+   * <p>E.g. for an NVIDIA embeddings call with 7 tokens, 512 request bytes and 1024 response bytes,
+   * this returns three events of the form:
+   *
+   * <pre>{@code
```
amorton: nit: is {@code meant to be there ?

src/main/java/io/stargate/sgv2/jsonapi/service/provider/Billing.java
```diff
@@ -0,0 +1,131 @@
...
+   *  "properties":{"usage":512,"region":"us-west-2",
+   *                "resource_type":"serverless_database","resource_id":"<tenant>"}}
+   *
+   * {"id":"...","timestamp":"...","product":"serverless",
+   *  "event_type":"nvidia_gpu_plane_egress_bytes",
+   *  "properties":{"usage":1024,"region":"us-west-2",
+   *                "resource_type":"serverless_database","resource_id":"<tenant>"}}
+   * }</pre>
+   */
+  List<BillingEvent> buildEvents(ModelUsage modelUsage) {
```
amorton, line 103: private ?

src/main/java/io/stargate/sgv2/jsonapi/service/provider/Billing.java
```diff
@@ -0,0 +1,131 @@
...
+      events.add(
+          newEvent(
+              "nvidia_gpu_plane_egress_bytes", modelUsage.responseBytes(), region, resourceId));
+    }
+    return List.copyOf(events);
+  }
+
+  private BillingEvent newEvent(String eventType, long usage, String region, String resourceId) {
+    var properties =
+        new BillingEvent.BillingProperties(usage, region, config.resourceType(), resourceId);
```
amorton: we should check these are set in the ctor, and we can just reference them as strings

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/FindAndRerankCommandResolver.java
```diff
@@ -22,20 +23,23 @@ public class FindAndRerankCommandResolver implements CommandResolver<FindAndRera
...
 
   @Inject
   public FindAndRerankCommandResolver(
       OperationsConfig operationsConfig,
       ObjectMapper objectMapper,
       MeterRegistry meterRegistry,
       JsonApiMetricsConfig jsonApiMetricsConfig,
-      FindCommandResolver findCommandResolver) {
+      FindCommandResolver findCommandResolver,
+      Billing billing) {
```
amorton: -1, we will move this into the requestcontext

src/main/java/io/stargate/sgv2/jsonapi/api/request/tenant/Tenant.java
```diff
@@ -66,13 +99,19 @@ public static Tenant create(DatabaseType databaseType, String tenantId) {
                 "Unsupported single tenant database type: " + databaseType);
       };
     }
-    return new Tenant(databaseType, normalizedValidated.toLowerCase());
+    String normalizedRegion = (region == null || region.isBlank()) ? UNKNOWN_REGION : region;
```
amorton: nit: String not needed here, use var when the typeis obvious for reading

src/main/java/io/stargate/sgv2/jsonapi/api/request/tenant/SubdomainTenantResolver.java
```diff
@@ -78,7 +79,15 @@ public Tenant resolve(RoutingContext context, SecurityContext securityContext) {
         }
       }
 
-      return TenantFactory.instance().create(tenantId);
+      // Astra subdomains follow the pattern {tenantId}-{region}, so anything past the
+      // tenantId + '-' separator is the region. Empty / missing region is fine — Tenant.create
+      // will fall back to Tenant.UNKNOWN_REGION.
+      String region = null;
```
amorton, line 69: we should add a test for SubdomainTenantResolver - it has been working for a long time, but the new work for the region is new. Just a couple of tests that show it extracting from a full endpoint url

src/main/java/io/stargate/sgv2/jsonapi/service/provider/BillingEvent.java
```diff
@@ -0,0 +1,84 @@
...
+ * bytes for an embedding or reranking call). Constructed and emitted by {@link Billing}.
+ *
+ * @param id Unique random-based (UUID v4) identifier for this event.
+ * @param timestamp ISO 8601 timestamp of when the event was created.
+ * @param product Product identifier, e.g. {@code "serverless"}.
+ * @param eventType Event type, e.g. {@code "nvidia_embeddings_tokens"} or {@code
+ *     "nvidia_egress_bytes"}.
+ * @param properties Usage details including the billable amount, region, and resource identifiers.
+ */
+@JsonInclude(JsonInclude.Include.NON_NULL)
```
amorton, line 44: nit: can we include a nicelt formatted example of the JSON

amorton (changes requested):
couple of changes, and the bug in the equals method. 

we will hold on this until we get #2479 in and I will then move the billing instance to be on the command context so you do not need to pass it around.

src/main/java/io/stargate/sgv2/jsonapi/service/provider/Billing.java
Hazel-Datastax: need it for testing, add `@VisibleForTesting`

src/main/java/io/stargate/sgv2/jsonapi/service/provider/Billing.java
Hazel-Datastax: need it for testing, add `@VisibleForTesting`

src/main/java/io/stargate/sgv2/jsonapi/api/request/tenant/SubdomainTenantResolver.java
```diff
@@ -78,7 +79,15 @@ public Tenant resolve(RoutingContext context, SecurityContext securityContext) {
...
+      if (subdomain.length() > tenantId.length() + 1) {
```
tatu-at-datastax: Should this verify "-" separator, just in case?

src/main/java/io/stargate/sgv2/jsonapi/api/request/tenant/Tenant.java
```diff
@@ -29,30 +28,60 @@ public class Tenant implements Recordable {
...
   private static final Tenant SINGLE_TENANT_CASSANDRA =
-      new Tenant(DatabaseType.CASSANDRA, SINGLE_TENANT_ID);
+      new Tenant(DatabaseType.CASSANDRA, SINGLE_TENANT_ID, CASSANDRA_REGION_DEFAULT);
 
   private final DatabaseType databaseType;
   private final String tenantId;
+  private final String region;
 
-  private Tenant(DatabaseType databaseType, String tenantId) {
+  private Tenant(DatabaseType databaseType, String tenantId, String region) {
```
tatu-at-datastax, line 51: As per comment later on:

  /** The deployment region for this tenant. Never null. */
  public String region() {
    return region;
  }

shouldn't we either fail on `null`, or assign default (`CASSANDRA_REGION_DEFAULT`)?

src/main/java/io/stargate/sgv2/jsonapi/api/request/tenant/Tenant.java
```diff
@@ -29,30 +28,60 @@ public class Tenant implements Recordable {
...
+  }
+
+  /**
+   * Factory method to create a Tenant instance with no explicit region (region defaults per
+   * database type — Cassandra gets {@link #CASSANDRA_REGION_DEFAULT}, other types get {@link
+   * #UNKNOWN_REGION}).
+   */
+  @VisibleForTesting
+  public static Tenant create(DatabaseType databaseType, String tenantId) {
+    return create(databaseType, tenantId, null);
```
tatu-at-datastax, line 64: Similarly, shouldn't we choose the default as per comment above?

src/main/java/io/stargate/sgv2/jsonapi/api/request/tenant/Tenant.java
```diff
@@ -66,28 +95,40 @@ public static Tenant create(DatabaseType databaseType, String tenantId) {
...
   public int hashCode() {
     return Objects.hash(databaseType, tenantId);
   }
 
   /**
-   * Equals based on the combined values of {@link #databaseType} and a CASE INSENSITIVE {@link
-   * #tenantId}.
+   * Equals based on the combined values of {@link #databaseType} and {@link #tenantId}.
+   *
+   * <p>{@link #region} is intentionally excluded: a tenant is the same logical tenant in every
```
tatu-at-datastax, line 129: This seems dangerous to me and something I'd be hesitant to do. `region` is not a transient aspect but logical part of `Tenant` definition.

But that almost suggests that adding "region" in `Tenant` is wrong and there should be different type (TenantInRegion?).

Not sure what the right thing to do is, just surfacing my concerns with this aspect.

src/main/java/io/stargate/sgv2/jsonapi/config/feature/ApiFeature.java
```diff
@@ -69,7 +69,20 @@ public enum ApiFeature {
...
+   * egress bytes) on the dedicated {@code billing.events} logger for Astra tenants. If left as
+   * default, no billing events are emitted.
+   *
+   * <p>Set via {@code stargate.feature.flags.billing=true} at startup (authoritative; request
+   * headers cannot disable a startup-enabled flag) or per-request via {@code Feature-Flag-billing}
+   * header when not configured at startup.
+   *
+   * <p>Disabled by default.
+   */
+  BILLING("billing", false);
```
tatu-at-datastax: naming: should this be "BILLING_EVENTS" / `billing-events`?

src/main/java/io/stargate/sgv2/jsonapi/service/provider/Billing.java
```diff
@@ -0,0 +1,144 @@
...
+    this.product = requireNonBlank(config.product(), "billing.product");
+    this.resourceType = requireNonBlank(config.resourceType(), "billing.resource_type");
+    this.objectWriter = new ObjectMapper().writer();
+  }
+
+  /**
+   * Emits billing events for the given aggregated model usage, if the request and configuration
+   * allow it. No-op otherwise (feature disabled, logger disabled, null usage).
+   */
+  public void bill(ModelUsage modelUsage) {
```
tatu-at-datastax: Naming: maybe `emit()` or `emitEvent()` ?

src/main/java/io/stargate/sgv2/jsonapi/service/provider/Billing.java
```diff
@@ -0,0 +1,144 @@
...
+  private final String resourceType;
+  private final ApiFeatures apiFeatures;
+  private final ObjectWriter objectWriter;
+
+  public Billing(BillingConfig config, ApiFeatures apiFeatures) {
+    Objects.requireNonNull(config, "config must not be null");
+    this.apiFeatures = Objects.requireNonNull(apiFeatures, "apiFeatures must not be null");
+    this.product = requireNonBlank(config.product(), "billing.product");
+    this.resourceType = requireNonBlank(config.resourceType(), "billing.resource_type");
+    this.objectWriter = new ObjectMapper().writer();
```
tatu-at-datastax: This is rather expensive: single `ObjectMapper` (or `ObjectWriter` constructed) should be reused for multiple calls.
(much of cost for mappers is for construction or first use (serializers created then, cached).

src/main/java/io/stargate/sgv2/jsonapi/service/provider/Billing.java
```diff
@@ -0,0 +1,144 @@
...
+            region,
+            resourceId));
+    events.add(
+        newEvent(providerApi + "_egress_bytes", modelUsage.requestBytes(), region, resourceId));
+    if (provider == ModelProvider.NVIDIA) {
+      events.add(
+          newEvent(
+              "nvidia_gpu_plane_egress_bytes", modelUsage.responseBytes(), region, resourceId));
+    }
+    return List.copyOf(events);
```
tatu-at-datastax: This is internal usage: no need to create defensive copy.

src/main/java/io/stargate/sgv2/jsonapi/service/provider/BillingEvent.java
```diff
@@ -0,0 +1,101 @@
...
+ * bytes for an embedding or reranking call). Constructed and emitted by {@link Billing}.
+ *
+ * <p>Example serialized form:
+ *
+ * <pre>
+ * {
+ *   "id": "8c0e9b8a-1d3a-4f6b-9c0d-1234567890ab",
+ *   "timestamp": "2026-05-20T14:23:11.482Z",
+ *   "product": "serverless",
+ *   "event_type": "nvidia_embeddings_tokens",
```
tatu-at-datastax: Is this still naming of event_types we are going for? (yesterday's call changed it I thought)

src/test/java/io/stargate/sgv2/jsonapi/service/provider/BillingEventTest.java
```diff
@@ -0,0 +1,66 @@
...
+class BillingEventTest {
+
+  private static final ObjectMapper MAPPER = new ObjectMapper();
+
+  private static BillingEvent.BillingProperties props() {
+    return new BillingEvent.BillingProperties(123L, "us-west-2", "serverless_database", "tenant-x");
+  }
+
+  @Test
+  void canonicalConstructorRejectsNulls() {
```
tatu-at-datastax: I guess it's ok but these are sort of low value tests.

tatu-at-datastax (commented):
Looks mostly good, some comments.

src/main/java/io/stargate/sgv2/jsonapi/api/request/tenant/Tenant.java
Hazel-Datastax: By the time region reaches the private ctor it's already non-null: multi-tenant goes through `(region == null || region.isBlank()) ? UNKNOWN_REGION : region` in 3-arg, and single-tenant returns the static `SINGLE_TENANT_CASSANDRA` outright (region arg is ignored). Picking the default in 2-arg would just duplicate the fallback. Added `Objects.requireNonNull(region)` on the private ctor as the defensive guard.

src/main/java/io/stargate/sgv2/jsonapi/api/request/tenant/Tenant.java
Hazel-Datastax: add `Objects.requireNonNull` in ctor as https://github.com/stargate/data-api/pull/2473#discussion_r3326742442

src/main/java/io/stargate/sgv2/jsonapi/service/provider/BillingEvent.java
Hazel-Datastax: will have another pr to clean up those naming things, event types/format etc

src/test/java/io/stargate/sgv2/jsonapi/service/provider/BillingEventTest.java
Hazel-Datastax: removed

src/main/java/io/stargate/sgv2/jsonapi/api/request/tenant/Tenant.java
Hazel-Datastax: Aaron explicitly asked to exclude region in his earlier review. @amorton WDYT?

src/main/java/io/stargate/sgv2/jsonapi/config/feature/ApiFeature.java
Hazel-Datastax: Name came out of earlier rounds with @amorton, can change if we align

src/main/java/io/stargate/sgv2/jsonapi/service/provider/Billing.java
Hazel-Datastax: same, name came out of earlier rounds with @amorton, can change if we align

src/main/java/io/stargate/sgv2/jsonapi/config/feature/ApiFeature.java
amorton: Happy to go with BILLING_EVENTS , perhaops per the discussion to also have an option to return them in headers it should be `BILLING_EVENTS_LOGGING` and then  when we can return them it would be `BILLING_EVENTS_RESPONSE`

src/main/java/io/stargate/sgv2/jsonapi/service/provider/Billing.java
amorton: emitEvent() sounds good

src/main/java/io/stargate/sgv2/jsonapi/service/provider/BillingEventType.java
```diff
@@ -0,0 +1,70 @@
...
+  private final boolean internal;
+  private final Dimension dimension;
+
+  BillingEventType(String eventName, boolean internal, Dimension dimension) {
+    this.eventName = eventName;
+    this.internal = internal;
+    this.dimension = dimension;
+  }
+
+  /** Lower-case event_type string used in the JSON billing event. */
```
amorton, line 49: nit: if we are saying lower case, we may want something to makes sure that is the case like lcase () in the ctor

src/main/java/io/stargate/sgv2/jsonapi/service/provider/BillingEventType.java
```diff
@@ -0,0 +1,70 @@
...
+public enum BillingEventType {
+  INTERNAL_MODEL_TOTAL_TOKENS("internal_model_total_tokens", true, Dimension.TOTAL_TOKENS),
+  EXTERNAL_MODEL_TOTAL_TOKENS("external_model_total_tokens", false, Dimension.TOTAL_TOKENS),
+  INTERNAL_MODEL_EGRESS_BYTES("internal_model_egress_bytes", true, Dimension.EGRESS_BYTES),
+  EXTERNAL_MODEL_EGRESS_BYTES("external_model_egress_bytes", false, Dimension.EGRESS_BYTES),
+  INTERNAL_MODEL_INGRESS_BYTES("internal_model_ingress_bytes", true, Dimension.INGRESS_BYTES),
+  EXTERNAL_MODEL_INGRESS_BYTES("external_model_ingress_bytes", false, Dimension.INGRESS_BYTES);
+
+  /** The billable dimension a {@link BillingEventType} measures. */
+  public enum Dimension {
```
amorton: I like what this enum is for, but I think Dimension is overloaded with the dimension of the vector. 

Can we find another name, such as  `Metric` or `Meter`

src/main/java/io/stargate/sgv2/jsonapi/api/request/tenant/Tenant.java
```diff
@@ -66,28 +95,40 @@ public static Tenant create(DatabaseType databaseType, String tenantId) {
...
+   * region it is reached from. Tenant is used as part of {@code SchemaObjectIdentifier} which keys
```
amorton, line 130: 👍

src/main/java/io/stargate/sgv2/jsonapi/config/BillingConfig.java
```diff
@@ -0,0 +1,36 @@
...
+   */
+  @WithDefault("nvidia")
+  List<String> internalModelProviders();
+
+  /**
+   * The set of {@link BillingEventType} values to emit. Events whose type is not in this set are
+   * dropped. When the property is not configured, all values of {@link BillingEventType} are
+   * enabled. Explicitly setting it to an empty value disables all billing events.
+   */
+  Optional<Set<BillingEventType>> enabledEventTypes();
```
amorton, line 60: took me some thnking, but I get this is the best approach now - given that the default needs to be a constant string. 

Could you just give some examples like below: 

```
// Not specified (Optional.empty() → all enabled):
stargate:
  jsonapi:
    billing:
      product: serverless
      
// Specified but empty (Optional of empty set → none enabled):
stargate:
  jsonapi:
    billing:
      enabled-event-types: []
      
// specified and enabling three
stargate:
  jsonapi:
    billing:
      enabled-event-types:
        - INTERNAL_MODEL_TOTAL_TOKENS
        - EXTERNAL_MODEL_TOTAL_TOKENS
        - INTERNAL_MODEL_EGRESS_BYTES
```

src/main/java/io/stargate/sgv2/jsonapi/service/provider/Billing.java
```diff
@@ -0,0 +1,158 @@
...
+ * <p>Eligibility requires all of: {@link ApiFeature#BILLING_EVENTS_LOGGING} is enabled, the {@code
+ * billing.events} logger is enabled, and the {@link ModelUsage} is non-null. The region for each
+ * event is read from {@link ModelUsage#tenant()} ({@code Tenant.region()}).
+ */
+public class Billing {
+
+  private static final Logger BILLING_LOGGER = LoggerFactory.getLogger("billing.events");
+  private static final Logger LOGGER = LoggerFactory.getLogger(Billing.class);
+
+  // ObjectMapper / ObjectWriter construction is expensive (serializers are built and cached on
```
amorton: nit: comment like this is not needed, please trim unecessary comments.

src/main/java/io/stargate/sgv2/jsonapi/JsonApiStartUp.java
```diff
@@ -12,7 +13,8 @@ public class JsonApiStartUp {
   private final OperationsConfig operationsConfig;
 
   @Inject
-  public JsonApiStartUp(DebugModeConfig config, OperationsConfig operationsConfig) {
+  public JsonApiStartUp(
+      DebugModeConfig config, OperationsConfig operationsConfig, BillingConfig billingConfig) {
```
amorton, line 29: billingConfig - is not used

src/test/java/io/stargate/sgv2/jsonapi/api/request/tenant/SubdomainTenantResolverTests.java
```diff
@@ -0,0 +1,140 @@
...
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
+
+/** Tests for {@link SubdomainTenantResolver}, focused on Astra-style endpoint URLs. */
+public class SubdomainTenantResolverTests {
+
+  // Astra tenant ids are 36-char UUIDs; production config caps tenant extraction at the UUID
+  // length. See
+  // https://github.com/riptano/serverless-platform-charts/blob/master/stargate-apis/chart/charts/jsonapi/templates/deployment.yaml#L120
```
amorton: -1 this is a link to a private repo

src/test/java/io/stargate/sgv2/jsonapi/api/request/tenant/SubdomainTenantResolverTests.java
```diff
@@ -0,0 +1,140 @@
...
+  private static final int TENANT_UUID_LENGTH = 36;
+  private static final String TENANT_UUID = "11111111-2222-3333-4444-555555555555";
+  // https://github.com/riptano/serverless-platform-charts/blob/master/stargate-apis/chart/charts/jsonapi/templates/deployment.yaml#L122
```
amorton: -1 this is a link to a private repo

src/test/java/io/stargate/sgv2/jsonapi/api/request/tenant/SubdomainTenantResolverTests.java
```diff
@@ -0,0 +1,140 @@
...
+  private static final String SUBDOMAIN_REGEX =
```
amorton, line 29: Can we add a comment to explain what this reges is doing ? looks like a UUID match regex

src/test/java/io/stargate/sgv2/jsonapi/api/request/tenant/SubdomainTenantResolverTests.java
```diff
@@ -0,0 +1,140 @@
...
+import org.junit.jupiter.params.provider.ValueSource;
+
+/** Tests for {@link SubdomainTenantResolver}, focused on Astra-style endpoint URLs. */
+public class SubdomainTenantResolverTests {
+
+  // Astra tenant ids are 36-char UUIDs; production config caps tenant extraction at the UUID
+  // length. See
+  // https://github.com/riptano/serverless-platform-charts/blob/master/stargate-apis/chart/charts/jsonapi/templates/deployment.yaml#L120
+  private static final int TENANT_UUID_LENGTH = 36;
+  private static final String TENANT_UUID = "11111111-2222-3333-4444-555555555555";
```
amorton: this is a valid UUID, but the version is a weird non RFC standard one. It also is uses a small set of the avail chars for a uuid, so to test the text matching we may want something more real life.

may be a good idea to get a v4 uuid

src/test/java/io/stargate/sgv2/jsonapi/api/request/tenant/SubdomainTenantResolverTests.java
```diff
@@ -0,0 +1,140 @@
...
+    var context = routingContextWithHost(TENANT_UUID + "-" + region + ".apps.astra.datastax.com");
+
+    Tenant tenant = prodConfigResolver().resolve(context, null);
+
+    assertThat(tenant.toString()).isEqualTo(TENANT_UUID);
+    assertThat(tenant.region()).isEqualTo(region);
+  }
+
+  @Test
+  public void noRegionInSubdomainFallsBackToUnknown() {
```
amorton, line 97: what if the full URL is `11111111-2222-3333-4444-555555555555-.apps.astra.datastax.com`

has the dash but no region

src/test/java/io/stargate/sgv2/jsonapi/api/request/tenant/SubdomainTenantResolverTests.java
```diff
@@ -0,0 +1,140 @@
...
+
+    assertThat(tenant.toString()).isEqualTo(TENANT_UUID);
+    assertThat(tenant.region()).isEqualTo(Tenant.UNKNOWN_REGION);
+  }
+
+  @Test
+  public void invalidTenantSubdomainRejectedByRegex() {
+    // Subdomain doesn't match the UUID regex — resolver passes null to TenantFactory, which for
+    // ASTRA surfaces as IllegalArgumentException ("tenantId must not be empty").
+    var context = routingContextWithHost("not-a-uuid.apps.astra.datastax.com");
```
amorton, line 124: I think if the TenatFactory is told the db type is CASSANDRA it would still result in the same error, becase we do not expect a tenant. Can we also test that

src/test/java/io/stargate/sgv2/jsonapi/api/request/tenant/SubdomainTenantResolverTests.java
```diff
@@ -0,0 +1,140 @@
...
+  public void hostWithoutSubdomainRejected() {
+    // No '.' in host — no subdomain to parse → resolver passes null to TenantFactory.
+    var context = routingContextWithHost("localhost");
+
+    assertThatThrownBy(() -> prodConfigResolver().resolve(context, null))
+        .isInstanceOf(IllegalArgumentException.class);
+  }
+
+  @Test
+  public void noMaxCharsAndNoRegexAcceptsRawSubdomain() {
```
amorton, line 169: I like this test :) want to make it even more pedantic ? if there is a uPPer case in the domain that should still be in the tenant, remember I broke this when I lowercased

src/test/java/io/stargate/sgv2/jsonapi/api/request/tenant/TenantFactoryTests.java
```diff
@@ -49,4 +49,24 @@ public void shouldCreateTenantWithCorrectDatabaseType() {
     assertThat(tenant.databaseType()).isEqualTo(DatabaseType.ASTRA);
     assertThat(tenant.toString()).isEqualTo("abc");
   }
+
+  @Test
+  public void shouldCreateTenantWithRegion() {
```
amorton, line 54: Can we also include DB type is CASSANDRA

src/test/java/io/stargate/sgv2/jsonapi/api/request/tenant/TenantTests.java
```diff
@@ -86,6 +86,17 @@ public void equalsAndHashCode() {
     assertThat(tenantB.hashCode()).isEqualTo(tenantC.hashCode());
   }
 
+  @Test
```
amorton, line 89: 👍

src/test/java/io/stargate/sgv2/jsonapi/service/provider/BillingEventTest.java
```diff
@@ -0,0 +1,66 @@
...
+    UUID id = UUID.randomUUID();
+    Instant now = Instant.parse("2026-05-19T00:29:21.506481Z");
+    BillingEvent event =
+        new BillingEvent(
+            id, now, "serverless", BillingEventType.INTERNAL_MODEL_TOTAL_TOKENS, props());
+
+    String json = MAPPER.writeValueAsString(event);
+    JsonNode node = MAPPER.readTree(json);
+
+    assertThat(node.get("id").asText()).isEqualTo(id.toString());
```
amorton: See examples in FindOneWithSortIntegrationTest for how this can be made to use a string with the full JSON in it, to make it easier to understand. 

Then run a json sensitive (not just string matching) using the `net.javacrumbs.jsonunit` package we already us. 

e.g. 
```java
assertThatJson(actualString).isEqualTo(expectedString);
```

src/main/java/io/stargate/sgv2/jsonapi/service/provider/Billing.java
```diff
@@ -0,0 +1,158 @@
...
+            .enabledEventTypes()
+            .map(Set::copyOf)
+            .orElseGet(() -> EnumSet.allOf(BillingEventType.class));
+  }
+
+  /**
+   * Emits billing events for the given aggregated model usage, if the request and configuration
+   * allow it. No-op otherwise (feature disabled, logger disabled, null usage).
+   */
+  public void emitEvent(ModelUsage modelUsage) {
```
amorton, line 74: this will silently swallow a null modelUsage, even if billing is enabled. This can result in a bug in the calling code not being noticed. 

I think it would be safer to make modelUsage non-nullable, throw an error if it is, so that the calling code needs to understand if it has model usage. 

Do not need to change the code in MeteredEmbeddingProviderWrapper , it assumes there is billing and sends it. this would throw null pointer if there is no billing

src/main/java/io/stargate/sgv2/jsonapi/service/provider/Billing.java
```diff
@@ -0,0 +1,158 @@
...
+      String resourceId,
+      String providerName,
+      String modelName) {
+    if (!enabledEventTypes.contains(eventType)) {
+      return;
+    }
+    var properties =
+        new BillingEvent.BillingProperties(
+            usage, region, resourceType, resourceId, providerName, modelName);
+    events.add(new BillingEvent(UUID.randomUUID(), Instant.now(), product, eventType, properties));
```
amorton: I think it would be best if all the events from the same modelUsage had the same timestamp, it will make it a little bit easier to if we ever need to go through logs

src/test/java/io/stargate/sgv2/jsonapi/TestConstants.java
```diff
@@ -316,6 +323,22 @@ public RequestContext requestContext() {
     return new RequestContext(TENANT, AUTH_TOKEN, USER_AGENT);
   }
 
+  /**
+   * BILLING-off, fully-mocked-config Billing instance for tests that mock the {@link
+   * RequestContext} and need {@code requestContext.billing()} to return a usable no-op rather than
+   * null.
+   */
+  public static Billing noOpBilling() {
```
amorton, line 331: it seems unecessary to mock the billing config like this, with actual settings, when it is a no-op. 

it can be a fast follow or in the next PR, but we should either change the ctor for billing so it ignores the config when disabled OR extract billing into an interface where there is a NO-OP implementation. 

current implementation is a little confusing

amorton (approved):
NOTE: -1 on some code in here because the comments have links to private repos.

some changes I trust you to make or put in a fast follow. 

a good foundation to build on

src/main/java/io/stargate/sgv2/jsonapi/JsonApiStartUp.java
Hazel-Datastax: add java doc to explain:
```
  /**
   * {@link DebugModeConfig} and {@link BillingConfig} are injected here purely to anchor them as
   * CDI beans, even though we never read them through this constructor — they are consumed
   * elsewhere via {@link io.stargate.sgv2.jsonapi.api.model.command.CommandConfig#get(Class)}.
   *
   * <p>Quarkus ARC removes unused beans at build time, which also drops the SmallRye mapping
   * registration for any {@code @ConfigMapping} interface that has no injection point. When that
   * happens, {@link io.stargate.sgv2.jsonapi.ConfigPreLoader#onStart} fails at startup with
   * {@code SRCFG00027: Could not find a mapping for ...} and every {@code @QuarkusTest} blows up
   * with "Failed to start quarkus". Listing the config interface as a constructor parameter is
   * the minimal way to pin it.
   */
```

src/test/java/io/stargate/sgv2/jsonapi/api/request/tenant/SubdomainTenantResolverTests.java
Hazel-Datastax: Added `invalidTenantSubdomainOnCassandraReturnsSingleTenant` — but CASSANDRA doesn't throw the same error. Because CASSANDRA is single-tenant, `Tenant.create` ignores the tenantId entirely and returns the `SINGLE_TENANT_CASSANDRA` singleton (and normalizeValidateTenantId only throws for multi-tenant types). 

So the new test reflects the actual behavior and doesn't throw `IllegalArgumentException`

src/test/java/io/stargate/sgv2/jsonapi/TestConstants.java
Hazel-Datastax: will have it in another PR

PR #2475, opened 2026-05-19
security: build keyspace DDL via SchemaBuilder/CqlIdentifier (follow-up to #2472)

erichare:
## Summary

Follow-up to [#2472](https://github.com/stargate/data-api/pull/2472) (GHSA-p64p-96pw-mxwv). After [#2474](https://github.com/stargate/data-api/pull/2474) reverted #2472 with the reasoning "the actual fix is to use a query builder," this PR is the query-builder fix.

- `CreateKeyspaceOperation` and `DropKeyspaceOperation` now build their `SimpleStatement` via the driver's `SchemaBuilder` with `CqlIdentifier.fromInternal(name)`. `CqlIdentifier` doubles embedded quote characters, so the keyspace identifier stays well-formed regardless of upstream validation — no `String.format` interpolation remains.
- `CreateKeyspaceOperation` takes `(name, strategy, strategyOptions)` directly; the manual replication-map string-building in `CreateNamespaceKeyspaceCommandResolver` is gone.
- The drop resolvers do **not** apply `NamingRules.KEYSPACE.checkRule` — addressing the concern in #2474 about blocking drops of keyspaces created outside our rules. The `SchemaBuilder` + `CqlIdentifier` path makes that resolver-level check unnecessary for safety.

## Driver-gap finding for datacenter names

While building this PR I verified — by both decompiling `OptionsUtils.extractOptionValue` and running a live probe — that the driver's `SchemaBuilder.withNetworkTopologyStrategy(Map)` does **not** escape map keys. Repro:

```java
SchemaBuilder.createKeyspace(CqlIdentifier.fromInternal("ks"))
    .withNetworkTopologyStrategy(Map.of("dc'); EVIL --", 1)).asCql();
// → CREATE KEYSPACE ks WITH replication={'class':'NetworkTopologyStrategy','dc'); EVIL --':1}
```

The DC-name allowlist applied in `CreateNamespaceKeyspaceCommandResolver.validateStrategyOptions(...)` is therefore the **actual security control** for the replication map, not defense-in-depth. The resolver comment is updated to flag this so the allowlist isn't accidentally removed in a later cleanup. A regression-pinning test (`CreateKeyspaceOperationTest.driverDoesNotEscapeDataCenterMapKeys`) documents the driver behaviour so we get a clear signal if a future driver upgrade starts escaping.

## Files changed

- `CreateKeyspaceOperation` / `DropKeyspaceOperation` → `SchemaBuilder` + `CqlIdentifier`
- `CreateKeyspaceCommandResolver` / `CreateNamespaceCommandResolver` → pass strategy + options to the operation; call `validateStrategyOptions`
- `CreateNamespaceKeyspaceCommandResolver` → drop `getReplicationMap`; keep DC-name validation with an updated rationale comment
- `SchemaException` + `errors.yaml` → re-add `UNSUPPORTED_REPLICATION_DATA_CENTER_NAME` (reverted in #2474)
- New `CreateKeyspaceOperationTest` and `DropKeyspaceOperationTest` exercise the SchemaBuilder behaviour directly
- `CreateKeyspaceCommandResolverTest` exercises both the resolver fields and DC-name injection rejection

## Test plan

- [x] `./mvnw test -Dtest='CreateKeyspaceCommandResolverTest,DropKeyspaceCommandResolverTest,CreateKeyspaceOperationTest,DropKeyspaceOperationTest'` → 27/27 passing (JDK 21)
- [x] `./mvnw compile` → clean, including google-java-format check (JDK 21)
- [ ] Full CI suite (run as part of PR checks)


src/main/java/io/stargate/sgv2/jsonapi/service/operation/keyspaces/CreateKeyspaceOperation.java
```diff
@@ -1,38 +1,71 @@
...
+ *
  * @param name Name of the keyspace to create.
- * @param replicationMap A replication json, see
- *     https://docs.datastax.com/en/cql-oss/3.3/cql/cql_reference/cqlCreateKeyspace.html#Table2.Replicationstrategyclassandfactorsettings.
+ * @param strategy Replication strategy name. {@code "NetworkTopologyStrategy"} selects the
+ *     network-topology strategy; any other value (including {@code null}) selects the simple
+ *     strategy.
+ * @param strategyOptions Strategy options. For {@code NetworkTopologyStrategy} this is the
+ *     datacenter-name to replication-factor map. For the simple strategy the {@code
+ *     replication_factor} entry is used (defaulting to 1). May be {@code null}.
```
amorton: -1  - how can the replication factor for a DC be null ?

src/main/java/io/stargate/sgv2/jsonapi/service/operation/keyspaces/DropKeyspaceOperation.java
```diff
@@ -12,24 +14,26 @@
 /**
  * Operation that drops a Cassandra keyspace if it exists.
  *
+ * <p>The CQL statement is constructed via the driver's {@link SchemaBuilder} using a {@link
```
amorton: this comment to not contributing to the understanding of this class can be removed

src/main/java/io/stargate/sgv2/jsonapi/service/operation/keyspaces/CreateKeyspaceOperation.java
```diff
@@ -1,38 +1,71 @@
...
 import io.stargate.sgv2.jsonapi.service.cqldriver.executor.QueryExecutor;
 import io.stargate.sgv2.jsonapi.service.operation.Operation;
 import io.stargate.sgv2.jsonapi.service.operation.collections.SchemaChangeResult;
+import java.util.Map;
 import java.util.function.Supplier;
 
 /**
  * Operation that creates a new Cassandra keyspace that serves as a namespace for the Data API.
  *
+ * <p>The CQL statement is constructed via the driver's {@link SchemaBuilder} using a {@link
```
amorton: pls read this comment and make it sound more human

src/main/java/io/stargate/sgv2/jsonapi/service/operation/keyspaces/CreateKeyspaceOperation.java
```diff
@@ -1,38 +1,71 @@
...
+ * double quotes, removing the previous keyspace-name interpolation sink.
+ *
+ * <p>The driver's {@code withNetworkTopologyStrategy(Map)} does <em>not</em> escape map keys, so
+ * datacenter-name map keys must still be validated by callers (see the resolver's allowlist).
+ *
  * @param name Name of the keyspace to create.
- * @param replicationMap A replication json, see
- *     https://docs.datastax.com/en/cql-oss/3.3/cql/cql_reference/cqlCreateKeyspace.html#Table2.Replicationstrategyclassandfactorsettings.
+ * @param strategy Replication strategy name. {@code "NetworkTopologyStrategy"} selects the
+ *     network-topology strategy; any other value (including {@code null}) selects the simple
```
amorton: "any other value (including {@code null}) selects the simple" feels weird, is this a change from existing functionality ?

src/main/java/io/stargate/sgv2/jsonapi/service/operation/keyspaces/CreateKeyspaceOperation.java
```diff
@@ -1,38 +1,71 @@
...
-    // execute
+    SimpleStatement createKeyspace = buildStatement();
     return queryExecutor
         .executeCreateSchemaChange(dataApiRequestInfo, createKeyspace)
 
         // if we have a result always respond positively
         .map(any -> new SchemaChangeResult(any.wasApplied()));
   }
+
+  SimpleStatement buildStatement() {
```
amorton: should be private

src/main/java/io/stargate/sgv2/jsonapi/service/operation/keyspaces/CreateKeyspaceOperation.java
```diff
@@ -1,38 +1,71 @@
...
+ * @param strategy Replication strategy name. {@code "NetworkTopologyStrategy"} selects the
+ *     network-topology strategy; any other value (including {@code null}) selects the simple
+ *     strategy.
+ * @param strategyOptions Strategy options. For {@code NetworkTopologyStrategy} this is the
+ *     datacenter-name to replication-factor map. For the simple strategy the {@code
+ *     replication_factor} entry is used (defaulting to 1). May be {@code null}.
  */
-public record CreateKeyspaceOperation(String name, String replicationMap) implements Operation {
+public record CreateKeyspaceOperation(
+    String name, String strategy, Map<String, Integer> strategyOptions) implements Operation {
```
amorton: change the name to be a CqlIdentifer, and then use CqlIdentifierUtil.cqlIdentifierFromUserInput() in the resolver

src/main/java/io/stargate/sgv2/jsonapi/service/operation/keyspaces/CreateKeyspaceOperation.java
```diff
@@ -1,38 +1,71 @@
...
         // if we have a result always respond positively
         .map(any -> new SchemaChangeResult(any.wasApplied()));
   }
+
+  SimpleStatement buildStatement() {
+    CreateKeyspaceStart start =
+        SchemaBuilder.createKeyspace(CqlIdentifier.fromInternal(name)).ifNotExists();
+
+    CreateKeyspace withReplication;
+    if (NETWORK_TOPOLOGY_STRATEGY.equals(strategy)) {
```
amorton, line 46: use case insensitive ?

src/main/java/io/stargate/sgv2/jsonapi/service/operation/keyspaces/CreateKeyspaceOperation.java
```diff
@@ -1,38 +1,71 @@
...
+      withReplication =
+          start.withNetworkTopologyStrategy(strategyOptions != null ? strategyOptions : Map.of());
+    } else {
+      int replicationFactor =
+          (strategyOptions != null)
```
amorton: I am not sure if we actually accept null strategyptions from the command and resolver

src/main/java/io/stargate/sgv2/jsonapi/service/operation/keyspaces/DropKeyspaceOperation.java
```diff
@@ -12,24 +14,26 @@
...
+ * CqlIdentifier}, which handles identifier quoting/escaping. This removes the previous {@code
+ * String.format} interpolation sink so CQL safety no longer depends on upstream name validation.
+ *
  * @param name Name of the keyspace to drop.
  */
 public record DropKeyspaceOperation(String name) implements Operation {
```
amorton: change to accept CqlIdentifier and use use CqlIdentifierUtil.cqlIdentifierFromUserInput() in the resolver

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/CreateKeyspaceCommandResolver.java
```diff
@@ -10,8 +10,8 @@
 import java.util.Map;
 
 /**
- * Command resolver for {@link CreateKeyspaceCommand}. Responsible for creating the replication map.
- * Resolve a {@link CreateKeyspaceCommand} to a {@link CreateKeyspaceOperation}
+ * Command resolver for {@link CreateKeyspaceCommand}. Resolves to a {@link CreateKeyspaceOperation}
```
amorton: this comment is not needed, why should the Resolver class have a comment about how another class works ?

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/CreateKeyspaceCommandResolver.java
```diff
@@ -38,7 +38,8 @@ public Operation resolveDatabaseCommand(
         (command.options() != null && command.options().replication() != null)
```
amorton: change to use ApiOptionUtils.getOrDefault() , do a function import, so it looks like this in usage :

```java
    var docIdDesc =
        getOrDefault(command.options(), CreateCollectionCommand.Options::idConfig, null);
```

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/CreateNamespaceKeyspaceCommandResolver.java
```diff
@@ -1,44 +1,50 @@
...
-    } else {
-      return simpleStrategyMap(strategyOptions);
-    }
-  }
+  // Allowlist for datacenter names. The driver's SchemaBuilder.withNetworkTopologyStrategy(Map)
+  // does NOT escape map keys (see OptionsUtils.extractOptionValue in java-driver-query-builder),
+  // so a hostile DC-name key passed to the driver would still produce broken CQL. This allowlist
+  // is therefore the actual security control for the replication map, not just API policy —
+  // do not remove without first ensuring DC names are escaped before reaching the driver.
+  private static final Pattern VALID_DATA_CENTER_NAME = Pattern.compile("[A-Za-z0-9_\\-]+");
```
amorton: -1 have we found some code as to what a valid DC name is in Cassandra ?

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/CreateNamespaceKeyspaceCommandResolver.java
```diff
@@ -1,44 +1,50 @@
...
-        map.append(", '%s': %d".formatted(dcEntry.getKey(), dcEntry.getValue()));
-      }
+  /**
+   * Validate datacenter-name map keys when the strategy is {@code NetworkTopologyStrategy}. For
+   * other strategies the {@code strategyOptions} map is interpreted by the driver as simple-
+   * strategy options (e.g. {@code replication_factor}) and is not validated here.
+   */
+  protected static void validateStrategyOptions(
+      String strategy, Map<String, Integer> strategyOptions) {
+    if (!NETWORK_TOPOLOGY_STRATEGY.equals(strategy) || strategyOptions == null) {
```
amorton: does this need to be case insensitive ? 

i am a little confused what this check is doing ?  If the network topology stragy is not being used, why are we checking strategyOptions ?

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/CreateNamespaceKeyspaceCommandResolver.java
```diff
@@ -1,44 +1,50 @@
...
-      return DEFAULT_REPLICATION_MAP;
+  private static void checkDataCenterName(String name) {
+    if (name == null
+        || name.isEmpty()
+        || name.length() > MAX_DATA_CENTER_NAME_LENGTH
+        || !VALID_DATA_CENTER_NAME.matcher(name).matches()) {
+      throw SchemaException.Code.UNSUPPORTED_REPLICATION_DATA_CENTER_NAME.get(
+          Map.of(
+              "maxNameLength",
+              String.valueOf(MAX_DATA_CENTER_NAME_LENGTH),
```
amorton: the overload that takes a map is Map<String, String> so is String.valueOf() needed ?

amorton (changes requested):
I am still not convinced the rules for the DC are actually correct. That needs to be validated in the C* code 

We can do some validation on the dropKeyspace to check the identifier is a regex word char, but not to check the length 

some more tweaks to bring the old collection code up to a better standard. 

and the comments, um, need a human touch

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/KeyspaceCommandResolverSupport.java
```diff
@@ -0,0 +1,29 @@
...
+  private static final Pattern DROP_KEYSPACE_NAME = Pattern.compile("\\w+");
+
+  private KeyspaceCommandResolverSupport() {}
+
+  static CqlIdentifier keyspaceIdentifierForCreate(String name) {
+    return cqlIdentifierFromUserInput(NamingRules.KEYSPACE.checkRule(name));
+  }
+
+  static CqlIdentifier keyspaceIdentifierForDrop(String name) {
+    if (name == null || !DROP_KEYSPACE_NAME.matcher(name).matches()) {
```
amorton: yes we should validate,but we cannot contrain the length because in HCD and C* the keyspac ename can be longer than astra. And it could be made in CQL .

So we need perhaps something on the nameing rules that is `sanitizeInput()` ? that will check the chars are from the allowed set of char, but not the length

it should prob throw the same or a similar exception inside `sanitizeInput()` as we do in `checkRule()`

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/CreateKeyspaceCommand.java
```diff
@@ -39,7 +39,7 @@ public record Options(@Nullable @Valid Replication replication) {}
               + "For NetworkTopologyStrategy, use {\"class\": \"NetworkTopologyStrategy\", \"datacenter_name\": N, ...}.")
   public record Replication(
       @NotNull
-          @Pattern(regexp = "SimpleStrategy|NetworkTopologyStrategy")
+          @Pattern(regexp = "(?i)(SimpleStrategy|NetworkTopologyStrategy)")
           @JsonProperty("class")
```
amorton: the code in cassandra for getting the replication strategy is case sensitive, see AbstractReplicationStrategy.getClass(). 

SO we should either keep it that way here, OR we need to fix it later. 

lets go with leaving this without the (?i) because we do not know how to capitalize, and tehcnically this can be any fully qualified path.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/CreateKeyspaceCommand.java
```diff
@@ -53,7 +53,7 @@ public record Replication(
                       + "For NetworkTopologyStrategy, use datacenter names as keys with replication factor as values "
                       + "(e.g. 'dc1': 3, 'dc2': 2).",
               type = SchemaType.OBJECT)
-          Map<String, Integer> strategyOptions) {}
+          Map<String, @NotNull Integer> strategyOptions) {}
```
amorton, line 56: ensure we have a test for this, I think you do now

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/CreateNamespaceCommand.java
```diff
@@ -53,7 +53,7 @@ public record Options(@Nullable @Valid Replication replication) {}
       deprecated = true)
   public record Replication(
       @NotNull()
-          @Pattern(regexp = "SimpleStrategy|NetworkTopologyStrategy")
+          @Pattern(regexp = "(?i)(SimpleStrategy|NetworkTopologyStrategy)")
```
amorton: same as keyspace, -1 on making case sensitive

src/main/java/io/stargate/sgv2/jsonapi/exception/SchemaException.java
```diff
@@ -85,6 +85,7 @@ public enum Code implements ErrorCode<SchemaException> {
     UNSUPPORTED_JSON_TYPE_FOR_TEXT_INDEX,
     UNSUPPORTED_LIST_DEFINITION,
     UNSUPPORTED_MAP_DEFINITION,
+    UNSUPPORTED_REPLICATION_DATA_CENTER_NAME,
```
amorton: I think this may actually be INVALID, for unsupported it is "we understand what you want to do, but you cannot do that" , e.g.  see  INVALID_USER_DEFINED_TYPE_NAME above

src/main/java/io/stargate/sgv2/jsonapi/service/operation/keyspaces/CreateKeyspaceOperation.java
amorton: ok, we marked the options as @NotNull in the command class, but this class does not know and needs to be defensive in it's inputs. So we shoud ensure that all of the values in the map or not null for this check and above.

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/CreateNamespaceKeyspaceCommandResolver.java
```diff
@@ -1,44 +1,38 @@
...
-      "{'class': 'SimpleStrategy', 'replication_factor': 1}";
+  private static final String NETWORK_TOPOLOGY_STRATEGY = "NetworkTopologyStrategy";
 
-  // resolve the replication map
-  String getReplicationMap(String strategy, Map<String, Integer> strategyOptions) {
-    if (strategy == null && strategyOptions == null) {
-      return DEFAULT_REPLICATION_MAP;
+  // The driver writes NetworkTopologyStrategy map keys as CQL string literals without escaping
+  // them.
+  private static final Pattern VALID_DATA_CENTER_NAME = Pattern.compile("[^']+");
```
amorton: as discussed, we need a thought out reason why this is the correct pattern to sanitize with. The code to reference from C* will be the NetworkTopologyStrategy

for example, a CQL quoted identifier is delim't with `"`  so should we care about that ? 

is this a better idea ?
[\"\'`\u2018\u2019\u201C\u201D]
That covers: " double, ' single, ` backtick, ' ' curly single, " " curly double.

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/KeyspaceCommandResolverSupport.java
```diff
@@ -0,0 +1,29 @@
...
+import static io.stargate.sgv2.jsonapi.util.CqlIdentifierUtil.cqlIdentifierFromUserInput;
+
+import com.datastax.oss.driver.api.core.CqlIdentifier;
+import io.stargate.sgv2.jsonapi.exception.ErrorTemplate;
+import io.stargate.sgv2.jsonapi.exception.SchemaException;
+import io.stargate.sgv2.jsonapi.service.schema.naming.NamingRules;
+import java.util.Map;
+import java.util.regex.Pattern;
+
+final class KeyspaceCommandResolverSupport {
```
amorton: unsure on why we need a new class here, and genally a "Support" class with static functions would need to be more widely usable. The pattern with the CreateKeyspace/Namespace is a base class. 

We also now have logic spread out for the DDL operations for a Keyspace. 

Consider renaming `CreateNamespaceKeyspaceCommandResolver` to `KeyspaceDDLCommandResolver` and move all the helper functions into there. 

Sublass all the Create and Delete for keyspace / namespace off the same class

amorton (changes requested):
some iteration, particularly around the santizing Vs checking name rules for deletion

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/CreateKeyspaceCommand.java
erichare: Yup, the test is CreateKeyspaceCommandTest.java (line 69): it sends "dc1": null in replication and asserts the validator returns "must not be null".

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/CreateNamespaceKeyspaceCommandResolver.java
erichare: https://cassandra.apache.org/doc/stable/cassandra/developing/cql/ddl.html#create-keyspace

NTS option keys are data-center names in a CQL map, not CQL identifiers, and the driver emits them as raw single-quoted CQL string keys. So ASCII ' is the delimiter we must reject; "/backtick/curly quotes do not close the string literal in that position.

PR #2479, opened 2026-05-22
Versioned Schema for lexical & rerank features

### LINKED ISSUE #2466: createCollection fails with EXISTING_COLLECTION_DIFFERENT_SETTINGS after lexical/rerank feature is enabled on a kube with pre-existing disabled-lexical collections
# What happened ? 

A `createCollection` command that had previously being worked for a user started failing after the lexcial and rerank features were enabled in their kube by the change https://github.com/riptano/serverless-argocd/pull/10615

The command started failing and returning errors such as:

```json
{
	"id": "5383de4c-925e-4673-804a-0d8b4c0671ed",
	"family": "REQUEST",
	"scope": "SCHEMA",
	"errorCode": "EXISTING_COLLECTION_DIFFERENT_SETTINGS",
	"title": "Collection already exists with different settings",
	"message": "Collection 'chat_messages' already exists but with settings different from ones passed with 'createCollection' command.\n\nIf you need to change collection settings you will need to 'deleteCollection', then re-create with new settings."
}
``` 

This was not the users fault, there was nothing they could do to fix this. 

The fix was to revert the change introduced by the PR above. 

## Why did the user start getting EXISTING_COLLECTION_DIFFERENT_SETTINGS error ? 

Whern creating a collection, code in the `CreateCollectionOperation` class failed to reconcile the configuration of an existing collection with the proposed new collection sent by the user. 

There is code written specifically to do this reconcilation, it failed. 

It detects if the two specifications are different, and works out if the difference is only that lexical search or reranking are now enabled. 

## If the collection existed, why did the user try to recreate it ? 

The Data API tries to follow behaviour of MongoDB, and so allows a `createCollection` command to silently fail if the user tries to create a collection that is the same as an existing on. If the user tries to create a collection with a different config they get the `EXISTING_COLLECTION_DIFFERENT_SETTINGS` error the customer saw. 

We know that many customers use this method, such as the one who had this issue, where they create the collection many times, perhaps as instances of code come and god. 

## Why did the config reconcilation fail ? 

The reconcilation failed because of the use of reference quality rather than value equality. 

Below is the relevant code from `CreateCollectionOperation`: 

```java

  boolean settingsAreEqual = existingCollectionSettings.equals(newCollectionSettings);

  if (!settingsAreEqual) {
    final var oldLexical = existingCollectionSettings.lexicalConfig();
    final var newLexical = lexicalConfig();
    final var oldReranking = existingCollectionSettings.rerankingConfig();
    final var newReranking = rerankDef();

    // So: for backwards compatibility reasons we may need to override settings if
    // (and only if) the collection was created before lexical and reranking.
    // In addition, we need to check that new lexical settings are for defaults
    // (difficult to check the lexicalConfigsame for reranking; for now assume that if lexical
    // is default, reranking is also default).
    if (oldLexical == CollectionLexicalConfig.configForPreLexical()
        && newLexical == CollectionLexicalConfig.configForDefault()
        && oldReranking == CollectionRerankDef.configForPreRerankingCollection()
        && newReranking == CollectionRerankDef.configForDefault()) {
      var originalNewSettings = newCollectionSettings;
      newCollectionSettings =
          newCollectionSettings.withLexicalAndRerankOverrides(
              oldLexical, existingCollectionSettings.rerankingConfig());
      // and now re-check if settings are the same
      settingsAreEqual = existingCollectionSettings.equals(newCollectionSettings);
      logger.info(
          "CreateCollectionOperation for {}.{} with existing legacy lexical/reranking settings, new settings differ. Tried to unify, result: {}"
              + " Old settings: {}, New settings: {}",
          commandContext.schemaObject().identifier().keyspace(),
          name,
          settingsAreEqual,
          existingCollectionSettings,
          originalNewSettings);
    } else {
      logger.info(
          "CreateCollectionOperation for {}.{} with different settings (but not old legacy lexical/reranking settings), cannot unify."
              + " Old settings: {}, New settings: {}",
          commandContext.schemaObject().identifier().keyspace(),
          name,
          existingCollectionSettings,
          newCollectionSettings);
    }
  }
```

**NOTE:** the use of `==` in the if statemeent after detecting the settings were different. This was intended to identify if the old settings for lexical and rerank were what they would be **pre** these features being added, and if the new settings are the same as the current defaults. Meaning the new settings came in from defaults. 

If this check works out the "new" settings are just the defaults then replace the lexical and reranking config in the new settings with the "old" ones and re-compare. 

**It should have used value equality to compare the values of these config items, rather then object reference quality.** 
 
We configed this check returned false by finding the second log in splunk for this tenant: 

(some reformatting)
```
CreateCollectionOperation for default_keyspace.project_documents with different settings (but not old legacy lexical/reranking settings), cannot unify. 

Old settings:
 CollectionSchemaObject[identifier=COLLECTION:default_keyspace.project_documents, idConfig=IdConfig[idType=], vectorConfig=VectorConfig[vectorEnabled=true, columnVectorDefinitions={$vectorize=VectorColumnDefinition[fieldName=$vectorize, vectorSize=3072, similarityFunction=COSINE, sourceModel=OTHER, vectorizeDefinition=null]}], indexingConfig=CollectionIndexingConfig[allowed=[documentId, projectId, userId], denied=[], indexedProject=Suppliers.memoize(io.stargate.sgv2.jsonapi.service.schema.collections.CollectionIndexingConfig$$Lambda/0x00000003018b8468@4399c3b8)], lexicalConfig=CollectionLexicalConfig[enabled=false, analyzerDefinition=null], rerankDef=CollectionRerankDef[enabled=false, rerankServiceDesc=null]], 

New settings: 

CollectionSchemaObject[identifier=COLLECTION:default_keyspace.project_documents, idConfig=IdConfig[idType=], vectorConfig=VectorConfig[vectorEnabled=true, columnVectorDefinitions={$vectorize=VectorColumnDefinition[fieldName=$vectorize, vectorSize=3072, similarityFunction=COSINE, sourceModel=OTHER, vectorizeDefinition=null]}], indexingConfig=CollectionIndexingConfig[allowed=[documentId, projectId, userId], denied=[], indexedProject=Suppliers.memoize(io.stargate.sgv2.jsonapi.service.schema.collections.CollectionIndexingConfig$$Lambda/0x00000003018b8468@169de8c0)], lexicalConfig=CollectionLexicalConfig[enabled=true, analyzerDefinition="standard"], rerankDef=CollectionRerankDef[enabled=true, rerankServiceDesc=RerankServiceDef[provider=nvidia, modelName=nvidia/llama-3.2-nv-rerankqa-1b-v2, authentication=null, parameters=null]]]
```

## Why did testing not detect this situation ? 

We have integration tests for this code, in `CreateCollectionBackwardCompatibilityIntegrationTest`, that creates a an existing collection manually using CQL and then runs a command to create a new collection using the API and checks that the user does not get an error.

We used CQL because this test was written with the lexical and reranking code in the code base, so we needed a way to create collections from **before** the code was in the code base. 

It created a collection via CQL as below:  

```
  String collectionWithoutLexicalRerank =
      """
                CREATE TABLE IF NOT EXISTS "%s"."%s" (
                    key frozen<tuple<tinyint, text>> PRIMARY KEY,
                    array_contains set<text>,
                    array_size map<text, int>,
                    doc_json text,
                    exist_keys set<text>,
                    query_bool_values map<text, tinyint>,
                    query_dbl_values map<text, decimal>,
                    query_null_values set<text>,
                    query_text_values map<text, text>,
                    query_timestamp_values map<text, timestamp>,
                    query_vector_value vector<float, 123>,
                    tx_id timeuuid
                ) WITH comment = '{"collection":{"name":"%s","schema_version":1,"options":{"defaultId":{"type":""}}}}';
                """;
```

The schema of the collection is in the comments of the table. 

It then checked that a collection with the same name could be created, and no error returned: 

```
// create the same collection using API - should not get
// COLLECTION_EXISTS_WITH_DIFFERENT_SETTINGS error
givenHeadersPostJsonThenOkNoErrors(
          """
              {
                  "createCollection": {
                      "name": "%s"
                  }
              }
      """
          .formatted(PRE_LEXICAL_RERANK_COLLECTION_NAME))
  .body("$", responseIsStatusOnly())
  .body("status.ok", is(1));
```

For two reasons this test did not surface the use of reference equality. 

First, the `createCollection` command that was sent did not define an `options` member, and so code selected to use the singleton default objects. When we updated the test to provide some options (like the user request) the test did not fail like we expected it to. So while we reconize this as a fault in the code it is probably not the only one. 

Second, the "old" table created by the test did not match what the customer has. From looking at the customers schema we can see their table as below: 

```
CREATE TABLE IF NOT EXISTS "34613964383738372d353061632d343961352d613661322d393435623534623934363165_default_keyspace".project_documents (
    key tuple<tinyint, text> PRIMARY KEY,
    doc_json text,
    query_vector_value vector<float, 3072>,
    tx_id timeuuid,
    array_contains set<text>,
    array_size map<text, int>,
    exist_keys set<text>,
    query_bool_values map<text, tinyint>,
    query_dbl_values map<text, decimal>,
    query_null_values set<text>,
    query_text_values map<text, text>,
    query_timestamp_values map<text, timestamp>
) WITH ID = cafcdb41-93c3-11f0-b5d2-a9bcf953d305
    AND additional_write_policy = '99p'
    AND bloom_filter_fp_chance = 0.01
    AND caching = {'keys': 'ALL', 'rows_per_partition': 'NONE'}
    AND cdc = false
    AND comment = '{"collection":{"name":"project_documents","schema_version":1,"options":{"indexing":{"allow":["projectId","userId","documentId"]},"vector":{"dimension":3072,"metric":"cosine","sourceModel":"OTHER"},"defaultId":{"type":""},"lexical":{"enabled":false},"rerank":{"enabled":false}}}}'
    AND compaction = {'class': 'org.apache.cassandra.db.compaction.UnifiedCompactionStrategy'}
    AND compression = {'chunk_length_in_kb': '16', 'class': 'org.apache.cassandra.io.compress.LZ4Compressor'}
    AND memtable = {}
    AND crc_check_chance = 1.0
    AND default_time_to_live = 0
    AND extensions = {}
    AND gc_grace_seconds = 864000
    AND max_index_interval = 2048
    AND memtable_flush_period_in_ms = 0
    AND min_index_interval = 128
    AND read_repair = 'BLOCKING'
    AND speculative_retry = '99p';
```

The collection definition in the table comment is: 

```json
{
	"collection": {
		"name": "project_documents",
		"schema_version": 1,
		"options": {
			"indexing": {
				"allow": [
					"projectId",
					"userId",
					"documentId"
				]
			},
			"vector": {
				"dimension": 3072,
				"metric": "cosine",
				"sourceModel": "OTHER"
			},
			"defaultId": {
				"type": ""
			},
			"lexical": {
				"enabled": false
			},
			"rerank": {
				"enabled": false
			}
		}
	}
}
```

**NOTE:** The customers table has values for the lexcial and rerank config settings, while the table created for the test above **did not**. While we did not trace the code, we believe the missing values for these settings caused the code to pull the singleton defaults similar to the fault for options not being present in the request. 

After updating the test to send options and it still not failing, we updated the test to use the same "on disk" collection config as the customer and were able to get the test to fail. 

With this in mind we consider the presence in the "on disk" config of the `lexical` and `rerank` fields to be the cause of failure for this user.

## Why did the test not cover the case the user had ? 

The lexcial and reranking code were added to the API together a number of months ago. At the time we added it we considered the backwards compatiblity as is evidenced by the testing and code above. However we did not consider all cases, and the length of time between the original release and expanding availability now exposed that. 

When the code was created we considered the backwards case where: the lexcial and rerank code is deployed with the features enabled, to existing kubes with existing collections created before the code was released. 

Because the BM25 indexes and access to the reranking models was limited, the code was deployed to some kubes with the features enabled and to others with the features disabled. 

We **did not** consider the backwards case where: the lexical and rerank code is deployed but disabled and then later enabled, and when enabled it would handle collections that were created after the initial code was released but before it was ena bled as a feature.

amorton:
hazel/fix-lexical-rerank-disable-then-enable-backward-compatibility

**What this PR does**:

Fixes #2466 

**Checklist**
- [ ] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/CreateCollectionCommand.java
```diff
@@ -23,59 +23,57 @@ public record CreateCollectionCommand(
...
         @JsonInclude(JsonInclude.Include.NON_NULL)
         @Nullable
-        @Schema(
-            description = "Configuration options for the collection",
-            type = SchemaType.OBJECT,
-            implementation = Options.class)
+        @Schema(description = "Configuration options for the collection", type = SchemaType.OBJECT)
         Options options)
     implements CollectionOnlyCommand {
+
```
amorton, line 29: Remove unnecessary implementation= when the record property is clearly defined

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/CreateCollectionCommand.java
```diff
@@ -133,25 +135,18 @@ public record VectorSearchConfig(
...
-        this.metric = metric;
-        this.sourceModel = sourceModel;
-        this.vectorizeConfig = vectorizeConfig;
-      }
-    }
+            VectorizeConfig vectorizeConfig) {}
 
-    public record IndexingConfig(
+    /** --- */
+    public record IndexingDesc(
```
amorton, line 133: tweak naming to match standard: Desc for public side Def for internal

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/CreateCollectionCommand.java
```diff
@@ -238,7 +233,8 @@ public void validateIndexingPath(List<String> paths) {
       }
     }
 
-    public record LexicalConfigDefinition(
+    /** --- */
+    public record LexicalDesc(
```
amorton, line 229: tweak naming to match standard: Desc for public side Def for internal

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/CommandConfig.java
```diff
@@ -34,7 +34,7 @@ public class CommandConfig {
   private final ConcurrentMap<Class<?>, Object> configCache = new ConcurrentHashMap<>();
 
   // use getConfigProvider()
-  private SmallRyeConfig configProvider;
+  private volatile SmallRyeConfig configProvider;
```
amorton, line 37: should have had volatile because is lazy init

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/CommandContext.java
```diff
@@ -86,7 +82,6 @@ private CommandContext(
       JsonProcessingMetricsReporter jsonProcessingMetricsReporter,
       CQLSessionCache cqlSessionCache,
       CommandConfig commandConfig,
-      ApiFeatures apiFeatures,
```
amorton, line 89: API features moved to the RequestContext because we need it before we get the command context, we need it to know what schema is supported when we get the shcmea from the DB

src/main/java/io/stargate/sgv2/jsonapi/config/feature/ApiFeature.java
```diff
@@ -19,7 +19,7 @@ public enum ApiFeature {
   /**
    * Lexical search/sort feature flag: if enabled, the API will allow construction of
    * "$lexical"-enabled Collections. If disabled, those operations will fail with {@link
-   * SchemaException.Code#LEXICAL_NOT_AVAILABLE_FOR_DATABASE}).
+   * SchemaException.Code#LEXICAL_FEATURE_NOT_ENABLED}).
```
amorton, line 22: rename to match REANK anbd MCP

src/main/java/io/stargate/sgv2/jsonapi/util/CqlIdentifierUtil.java
```diff
@@ -13,19 +13,51 @@ public abstract class CqlIdentifierUtil {
...
+   * Call to get the decription to use in a message like an error rmessage.
+   *
+   * <p>Use {@link CqlIdentifier#asInternal()} to get the internal representation which should never
+   * have double quotes on it. The internal will maintain if the case if created via {@link
+   * #cqlIdentifierFromUserInput(String)} because it forces double quotes, which {@link
+   * CqlIdentifier#fromCql(String)} strips but leaves the case intact.
+   */
   public static String cqlIdentifierToMessageString(CqlIdentifier identifier) {
-    return identifier == null ? "null" : identifier.asCql(true);
+    return identifier == null ? "null" : identifier.asInternal();
```
amorton, line 60: NOTE: previously the quotes in our messages were coming from the fact we had quotes in the internal representation in the CqlIdentifer. The few changes and comments in here are to make that more understandable. 

THIS IS THE REASON for the test changes that no longer need to escape quotes 

IF WE WANT TO - after this PR I can change this function so it puts back tickets around the name and will update the tests

I changed the CreateCollectionOperation to take an identifier rather than string, and it went bonkers.

tatu-at-datastax:
I think this PR needs to be split into multiple ones, is getting humongous.

src/main/java/io/stargate/sgv2/jsonapi/service/schema/SchemaHolder.java
```diff
@@ -0,0 +1,108 @@
...
+   * <p>This is useful when comparing schema from a user and what is already on disk. i.e. if the
+   * user gave as a null value for schema, then this instance will have a null persisted value, and
+   * when replacement is the value we got from disk we will use that. This allows for accurate
+   * comparision of a createCollection command schema to an existing collection schema.
+   *
+   * @param replacement The value to use if this instance does not have a persisted value.
+   * @return A decision on whether to use the replacement or this instance.
+   */
+  public ReplaceDecision<T> replaceIfMissing(SchemaHolder<T> replacement) {
+    Objects.requireNonNull(replacement, "replacement must be null");
```
tatu-at-datastax: nit: must NOT be null.

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/CreateCollectionCommandResolver.java
```diff
@@ -18,17 +18,21 @@
...
 import java.util.Map;
 
+/**
+ * Resolves the {@link CreateCollectionCommand } command into a {@link CreateCollectionOperation}
+ */
 @ApplicationScoped
 public class CreateCollectionCommandResolver implements CommandResolver<CreateCollectionCommand> {
 
-  private final ObjectMapper objectMapper;
+  private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
```
tatu-at-datastax: Why not injected when injection is used anyway? (specifically removed from being injected)
We have some specific configuration for mappers (in `ObjectMapperConfiguration`)

src/main/java/io/stargate/sgv2/jsonapi/service/schema/SchemaHolder.java
tatu-at-datastax: I'll just fix it.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/CreateCollectionCommand.java
tatu-at-datastax: I assume you verified that works? (just curious as to why `implementation` was added in the first place)

Asked LLM and it confirms that should work. So should be fine.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/CreateCollectionCommand.java
```diff
@@ -23,59 +23,57 @@ public record CreateCollectionCommand(
...
+      // -----
       @Valid
           @Nullable
           @JsonInclude(JsonInclude.Include.NON_NULL)
           @Schema(
               description = "Vector search configuration for the collection",
               type = SchemaType.OBJECT,
-              implementation = VectorSearchConfig.class)
-          VectorSearchConfig vector,
+              implementation = VectorSearchDesc.class)
```
tatu-at-datastax: Could drop this too.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/CreateCollectionCommand.java
```diff
@@ -23,59 +23,57 @@ public record CreateCollectionCommand(
...
           @JsonInclude(JsonInclude.Include.NON_NULL)
           @Nullable
           @Schema(
               description =
                   "Optional indexing configuration to provide allow/deny list of fields for indexing",
-              type = SchemaType.OBJECT,
-              implementation = IndexingConfig.class)
-          IndexingConfig indexing,
+              type = SchemaType.OBJECT)
+          CreateCollectionCommand.Options.IndexingDesc indexing,
```
tatu-at-datastax, line 54: and this.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/CreateCollectionCommand.java
```diff
@@ -23,59 +23,57 @@ public record CreateCollectionCommand(
...
           @JsonInclude(JsonInclude.Include.NON_NULL)
           @Nullable
           @Schema(
               description =
                   "Optional configuration defining if and how to support use of '$lexical' field",
-              type = SchemaType.OBJECT,
-              implementation = LexicalConfigDefinition.class)
-          LexicalConfigDefinition lexical,
+              type = SchemaType.OBJECT)
+          CreateCollectionCommand.Options.LexicalDesc lexical,
```
tatu-at-datastax, line 63: and this.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/CreateCollectionCommand.java
tatu-at-datastax: and this while removing unnecessary things.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/CreateCollectionCommand.java
```diff
@@ -88,7 +86,8 @@ public record IdConfig(
             @JsonProperty("type")
             String idType) {}
 
-    public record VectorSearchConfig(
+    /** --- */
```
tatu-at-datastax, line 87: Remove, fluff.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/CreateCollectionCommand.java
```diff
@@ -98,6 +97,7 @@ public record VectorSearchConfig(
             @JsonProperty("dimension")
             @JsonAlias("size") // old name
             Integer dimension,
+        // -----
```
tatu-at-datastax, line 95: fluff to remove.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/CommandContext.java
tatu-at-datastax: Ideally this change would have been merged first as a prep step.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/CommandContext.java
```diff
@@ -161,31 +142,44 @@ public String commandName() {
     return commandName;
   }
 
+  // Lazy init: apiFeatures config is accessed too early in unit tests if done at construction time
   public RequestTracing requestTracing() {
+    if (requestTracing == null) {
+      synchronized (this) {
```
tatu-at-datastax, line 148: Possible, but I think this is unnecessary optimization; could just make method itself `synchronized`. It's never going to be contested.

src/main/java/io/stargate/sgv2/jsonapi/api/request/RequestContext.java
```diff
@@ -168,6 +181,33 @@ public void removeFromMDC() {
...
+                  commandConfig.get(FeaturesConfig.class), getHttpHeaders());
+        }
+      }
+    }
+    return apiFeatures;
+  }
+
+  public VersionedSchema versionedSchema() {
+    // using a sync block here because the context can be accessed by multiple tasks concurrently
+    if (versionedSchema == null) {
```
tatu-at-datastax: As with earlier, construction very cheap so could just construct on-demand and neither cache nor synchronize (just needs "apiFeatures" and does almost no processing).

src/main/java/io/stargate/sgv2/jsonapi/service/operation/collections/CreateCollectionOperation.java
```diff
@@ -38,240 +45,237 @@
 public record CreateCollectionOperation(
     CommandContext<KeyspaceSchemaObject> commandContext,
     DatabaseLimitsConfig dbLimitsConfig,
-    ObjectMapper objectMapper,
```
tatu-at-datastax, line 41: Here too we may want to still inject, to get default configuration for mapper.

src/main/java/io/stargate/sgv2/jsonapi/service/schema/SchemaVersion.java
```diff
@@ -0,0 +1,7 @@
+package io.stargate.sgv2.jsonapi.service.schema;
+
+/** A base interface so we can have different schema versions for tables and collections */
+public interface SchemaVersion {
+
+  int ordinalValue();
```
tatu-at-datastax, line 16: Javadoc?

src/main/java/io/stargate/sgv2/jsonapi/service/schema/VersionedSchema.java
```diff
@@ -0,0 +1,36 @@
...
+ * Container for the {@link SchemaFactory} 's that are configured with the APIFeatures instance.
+ *
+ * <p><b>NOTE:</b> Most queries will not need to access schema factories if they get their schema
+ * objects directly from cache, so we do not always need an instance of this class. It is lazy
+ * created by the {@link io.stargate.sgv2.jsonapi.api.request.RequestContext} and would be a waste
+ * to make for every request (but APIFeatures are request scoped so kind of needed).
+ */
+public class VersionedSchema {
+
+  private final CollectionLexicalDefSchemaFactory lexicalDefSchemaValueDef;
```
tatu-at-datastax: Naming confusing: is it factory (as type suggests) or value/definition (like name suggests)?

src/main/java/io/stargate/sgv2/jsonapi/util/CqlIdentifierUtil.java
```diff
@@ -13,19 +13,51 @@ public abstract class CqlIdentifierUtil {
...
 
+  /**
+   * Call this when we get a CQL identifier from the user input.
+   *
+   * <p>We are trying to preserve the original user input, including the case sensitivity. {@link
+   * CqlIdentifier#fromCql(String)} will lower case the identifier if it is not quoted, and this is
+   * kind of unexpected for non cql users.
+   */
   public static CqlIdentifier cqlIdentifierFromUserInput(String name) {
     if (Strings.isDoubleQuoted(name)) {
```
tatu-at-datastax, line 24: But why/how would user ever use quotes around name? (I know it's existing code but confusing still)

src/main/java/io/stargate/sgv2/jsonapi/util/CqlIdentifierUtil.java
```diff
@@ -13,19 +13,51 @@ public abstract class CqlIdentifierUtil {
...
+      //  fromCql will see the double quotes, strip them, and make the internal retrain case.
+      // e.g. `"myCol"` -> Cqlidentifer with internal set to `myCol`
       return CqlIdentifier.fromCql(name);
     }
+    // the identifier does not have a double quote, so we need to double quote it to preserve case.
+    // e.g. `myCol` -> `"myCol"`  -> Cqlidentifer with internal set to `myCol`
```
tatu-at-datastax, line 30: There's `CqlIdentifier.fromInternal()` to avoid whole wrapping/unwrapping fluff, let's use that -- it does what is needed here.

src/main/java/io/stargate/sgv2/jsonapi/util/CqlIdentifierUtil.java
tatu-at-datastax: +1 for adding wrapping back. But I think with `pretty == false` so they are always quoted in message Strings.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/CreateCollectionCommand.java
tatu-at-datastax: Update: LLM tested this and elaborated:

-----

The empirical result actually flips my initial answer in an interesting way.
  
The diff (current PR vs. implementation = Options.class restored):

1. CreateCollectionCommand's options field becomes a $ref instead of an inlined schema.
     - Without implementation: $ref: "#/components/schemas/Options" — clean reference.
     - With implementation = Options.class: the entire Options structure is inlined into the property (~60 lines of inline schema). The named Options component is not produced at all.
     - This is a known SmallRye/MicroProfile OpenAPI quirk: explicitly setting implementation on a property whose declared type already matches that class causes the generator to inline rather than reference.
   
-----

and I suspect non-inlined is what we actually want.

But it does change OpenAPI Schema generated.

src/main/java/io/stargate/sgv2/jsonapi/service/schema/SchemaHolder.java
tatu-at-datastax: done.

src/main/java/io/stargate/sgv2/jsonapi/service/schema/SchemaHolder.java
```diff
@@ -0,0 +1,108 @@
...
+ * An instance of a schema value created by a {@link SchemaFactory} subclass, see that class for
+ * instructions on how to create a new instance.
+ *
+ * <p>Call {@link #runningValue()} to get the value that should be used for operations that need
+ * this value.
+ *
+ * <p>If you have a value of schema from a user, which may be null, and a value from the disk /
+ * existing collection call {@link #replaceIfMissing(SchemaHolder)} to decide which value to use.
+ *
+ * @param <T> The type of the schema value
```
tatu-at-datastax, line 17: Is there no common base type for values?

src/main/java/io/stargate/sgv2/jsonapi/service/schema/SchemaHolder.java
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  /**
+   * Decides if this instance has a persisted value that should be used, or if the replacement
+   * should be used.
+   *
+   * <p>This is useful when comparing schema from a user and what is already on disk. i.e. if the
+   * user gave as a null value for schema, then this instance will have a null persisted value, and
+   * when replacement is the value we got from disk we will use that. This allows for accurate
+   * comparision of a createCollection command schema to an existing collection schema.
```
tatu-at-datastax: typo: comparision -> comparison

src/main/java/io/stargate/sgv2/jsonapi/service/schema/SchemaHolder.java
```diff
@@ -0,0 +1,108 @@
...
+ */
+public class SchemaHolder<T> {
+  private static final Logger LOGGER = LoggerFactory.getLogger(SchemaHolder.class);
+
+  private final SchemaFactory<T> factory;
+
+  private final SchemaVersion persistedVersion;
+
+  // Nullable
+  private final T persistedValue;
```
tatu-at-datastax, line 27: Keeping this `final` seems to complicate the whole design a lot, with odd `ReplaceDecision` type and all -- isn't the idea of holder to allow actual overriding?

I just find this set of related types quite complicated for what is being solved.

src/main/java/io/stargate/sgv2/jsonapi/service/schema/SchemaHolder.java
```diff
@@ -0,0 +1,108 @@
...
+    this.persistedValue = persistedValue;
+    this.factory = Objects.requireNonNull(factory, "factory must not be null");
+  }
+
+  /**
+   * This is the value to use for operations that need this value
+   *
+   * @return the Value of the schema value to use for regular operations.
+   */
+  public T runningValue() {
```
tatu-at-datastax, line 41: maybe "resolvedValue()"?

src/main/java/io/stargate/sgv2/jsonapi/service/schema/SchemaHolder.java
```diff
@@ -0,0 +1,108 @@
...
+   *
+   * <p>This is useful when comparing schema from a user and what is already on disk. i.e. if the
+   * user gave as a null value for schema, then this instance will have a null persisted value, and
+   * when replacement is the value we got from disk we will use that. This allows for accurate
+   * comparision of a createCollection command schema to an existing collection schema.
+   *
+   * @param replacement The value to use if this instance does not have a persisted value.
+   * @return A decision on whether to use the replacement or this instance.
+   */
+  public ReplaceDecision<T> replaceIfMissing(SchemaHolder<T> replacement) {
```
tatu-at-datastax, line 59: If holder was mutable, could just return simple `boolean` on whether replacement occurred?

src/main/java/io/stargate/sgv2/jsonapi/service/schema/VersionedSchema.java
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * Container for the {@link SchemaFactory} 's that are configured with the APIFeatures instance.
+ *
+ * <p><b>NOTE:</b> Most queries will not need to access schema factories if they get their schema
+ * objects directly from cache, so we do not always need an instance of this class. It is lazy
+ * created by the {@link io.stargate.sgv2.jsonapi.api.request.RequestContext} and would be a waste
+ * to make for every request (but APIFeatures are request scoped so kind of needed).
+ */
+public class VersionedSchema {
```
tatu-at-datastax: Naming confusion: `VersionedSchema` vs `SchemaHolder` which refer to 2 different Schemas I think.

tatu-at-datastax (approved):
I find the system quite a bit more complicated than expected, but since there's extensive testing let's just get it in. Added a few notes but nothing related to core logic.

src/main/java/io/stargate/sgv2/jsonapi/service/schema/collections/CollectionSchemaObject.java
Hazel-Datastax, line 454: bug? `equals()` and `hashCode()` are not matched

src/main/java/io/stargate/sgv2/jsonapi/service/schema/collections/CollectionLexicalDef.java
```diff
@@ -52,7 +95,7 @@ public record CollectionLexicalConfig(
    * @throws NullPointerException if lexical search is enabled and analyzerDefinition is null
    * @throws IllegalStateException if lexical search is disabled and analyzerDefinition is not null
```
Hazel-Datastax: nit: should be "IllegalArgumentException" not "IllegalStateException"

src/main/java/io/stargate/sgv2/jsonapi/service/schema/SchemaFactory.java
```diff
@@ -0,0 +1,183 @@
+package io.stargate.sgv2.jsonapi.service.schema;
+
+import java.util.Objects;
+
+/**
+ * A typed factory for creting {@link SchemaHolder} instances, see the subclasses for details on
```
Hazel-Datastax: typo: creting -> creating

src/main/java/io/stargate/sgv2/jsonapi/service/schema/SchemaFactory.java
```diff
@@ -0,0 +1,183 @@
...
+ * their configuration.
+ *
+ * <p>Using the factory to wrap an instance of an implementing <code>T</code> that is from either a
+ * user or from disk, and may be null - this is called the "Persisted Value". Specific rules then
+ * kick in to get a "Running Value" which is the value you are actually going to use in operations.
+ *
+ * <p><b>NOTE:</b> The factory needs to know if the feature is enabled, and this is normally done
+ * via {@link io.stargate.sgv2.jsonapi.config.feature.ApiFeatures} which can be overridden per
+ * reqest. So these factories need to be per request (or smart caching), see {@link
```
Hazel-Datastax: typo: "reqest" -> "request"

src/main/java/io/stargate/sgv2/jsonapi/service/schema/collections/CollectionSchemaObject.java
```diff
@@ -113,13 +99,13 @@ public IndexUsage newIndexUsage() {
...
-  public Recordable.DataRecorder recordTo(Recordable.DataRecorder dataRecorder) {
+  public DataRecorder recordTo(DataRecorder dataRecorder) {
     return super.recordTo(dataRecorder)
         .append("idConfig", idConfig)
         .append("vectorConfig", vectorConfig)
         .append("indexingConfig", indexingConfig)
-        .append("lexicalConfig", lexicalConfig)
-        .append("rerankDef", rerankDef);
+        .append("lexicalDef", lexicalDef.runningValue())
+        .append("rerankDef", rerankDef.runningValue());
```
Hazel-Datastax, line 108: `recordTo` uses lexicalDef.runningValue() (logging), but `equals` compares the whole SchemaHolder → it is difficult to pinpoint in debugging whether the discrepancy lies in the `persistedValue` or the `runningValue`?

src/main/java/io/stargate/sgv2/jsonapi/service/operation/collections/CreateCollectionOperation.java
```diff
@@ -38,240 +45,237 @@
...
+        TableCommentConstants.COLLECTION_LEXICAL_CONFIG_KEY, overrideLexicalDef.runningValue());
+    optionsNode.putPOJO(
+        TableCommentConstants.COLLECTION_RERANKING_CONFIG_KEY, overrideRerankDef.runningValue());
+
+    var collectionNode = OBJECT_MAPPER.createObjectNode();
+    collectionNode.put(
+        TableCommentConstants.COLLECTION_NAME_KEY, cqlIdentifierToJsonKey(collectionName));
+    collectionNode.put(
+        TableCommentConstants.SCHEMA_VERSION_KEY,
+        CollectionSchemaVersion.CURRENT_VERSION.toString());
```
Hazel-Datastax: I know we override the `toString` method, but maybe using `ordinalValue()` is better? Jackson' `put(String, int)` overload will pick the right JSON numeric type, and `asInt()` on the read side still 
works for both new (numeric) and old (numeric-string)

Hazel-Datastax (approved):

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/CreateCollectionCommand.java
amorton: I struggle to read the file with all of the @decoratorsions and no new lines, this is the only way I have worked out to stop the formatter from moveing newlines

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/CommandContext.java
amorton: I was thinking that would syncronize all access, even once the object has been created from the lazy init. Admittedly it's not a hot code path, more a general pattern to avoid sync on a method unless necessary.

src/main/java/io/stargate/sgv2/jsonapi/service/resolver/CreateCollectionCommandResolver.java
amorton: changed, just was not thinking of it

src/main/java/io/stargate/sgv2/jsonapi/service/schema/SchemaHolder.java
amorton: correct, no common base because it would mean everything like CollectionLexicalDef and IdConfig etc would have to have a common base and I think that would be restrictive. 

For one Records cannot extend a class, so it would be an interface. We *could* have a marker interface `SchemaValue` to make it clear, but cannot think of any features for it right now.

src/main/java/io/stargate/sgv2/jsonapi/service/schema/SchemaHolder.java
amorton: OK, am open to talking through making SchemaHolder mutable. 

Current design is prob a hangover from thinking about schema as immutable objects. So if the holder is mutable, it means the schema of a Collection in memory could be mutated.  

But agree, it is a bit clunky the way the ReplaceDecision is used. That is going to be from this decision and how the code is currently written to handle name collision and how we build the CollectionSchemaObject. 

Lets chat, this is first version, can iterate.

src/main/java/io/stargate/sgv2/jsonapi/service/schema/SchemaHolder.java
amorton: will confirm in person

src/main/java/io/stargate/sgv2/jsonapi/util/CqlIdentifierUtil.java
amorton: ok, confirming this is just for this case (last line) . Will sync and spin into another ticket

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/CreateCollectionCommand.java
amorton: I have removed all from this file, other then the two below because am worried the type erasure on the list will confuse things

```
            @Schema(
                description = "List of allowed indexing fields",
                type = SchemaType.ARRAY,
                implementation = String.class) // leaving implementaiton= because of the List
            @Nullable
            List<String> allow,
        // -----
            @Schema(
                description = "List of denied indexing fields",
                type = SchemaType.ARRAY,
                implementation = String.class)  // leaving implementaiton= because of the List
            @Nullable
            List<String> deny) {
```

src/main/java/io/stargate/sgv2/jsonapi/service/schema/collections/CollectionSchemaObject.java
tatu-at-datastax: Should probably augment altho strictly speaking not a bug (hashCode()s of equal instances will be same)

src/main/java/io/stargate/sgv2/jsonapi/util/CqlIdentifierUtil.java
amorton: Keeping open to sync, I think if we want the result of `cqlIdentifierToMessageString()` to have some quoting we should not rely on the CqlIdentifer so we avoid overloading what we use them for. 

We should do something like below, so it's clear "This is extra we are only adding for the human message"

```java 
return identifier == null ? "null" :  "`" + identifier.asInternal() + "`";
```

tatu-at-datastax: As long as we try to use same quoting character everywhere, yes -- whichever we decide on.

src/main/java/io/stargate/sgv2/jsonapi/service/schema/collections/CollectionLexicalDef.java
amorton: fixed

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/CreateCollectionCommand.java
tatu-at-datastax: Right: should keep (I think) for Collections and Maps, only remove from non-generic types.

src/main/java/io/stargate/sgv2/jsonapi/api/model/command/impl/CreateCollectionCommand.java
tatu-at-datastax: ok.

src/main/java/io/stargate/sgv2/jsonapi/service/schema/SchemaHolder.java
tatu-at-datastax: If there is no natural base type, no need to create one just to have one. All good. :)

PR #2484, opened 2026-05-27
Use Schema Builder in CreateCollectionOperation

amorton:
**What this PR does**:

CreateCollectionOperation to user the SchemaBuilder to tidy up the way statements are created. 

Next changes will build on this to centralize the definition of the Table used by a Collection 

Fixes #<issue number>

**Checklist**
- [ ] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/service/operation/collections/CreateCollectionOperation.java
```diff
@@ -529,222 +525,146 @@ TableMetadata findTableAndValidateLimits(
...
-            + "    array_contains      set<text>,"
-            + "    query_bool_values   map<text, tinyint>,"
-            + "    query_dbl_values    map<text, decimal>,"
-            + "    query_text_values   map<text, text>, "
-            + "    query_timestamp_values map<text, timestamp>, "
-            + "    query_null_values   set<text>, "
-            + lexicalField
-            + "    PRIMARY KEY (key))";
+
     if (comment != null) {
```
tatu-at-datastax: Suggestion from LLM: 

"  The trace only fires on the no-comment path — exactly the case you're less likely to need it.
  Restructure so the build/log/return is shared:

  var statement = (comment != null ? create.withComment(comment) : create).build();
  if (LOGGER.isTraceEnabled()) {
    LOGGER.trace("getCreateTable() - created table statement: {}", statement.getQuery());
  }
"

src/main/java/io/stargate/sgv2/jsonapi/service/operation/collections/CreateCollectionOperation.java
```diff
@@ -529,222 +525,146 @@ TableMetadata findTableAndValidateLimits(
...
+    if (ifNotExists) {
+      start = start.ifNotExists();
+    }
+
+    var onTable = start.onTable(keyspace, collectionName);
+    var createIndex = isEntries ? onTable.andColumnEntries(column) : onTable.andColumn(column);
+
+    if (!options.isEmpty()) {
+      // in the CQL statement OPTIONS are the things after WITH, and for the `create index` there is
+      // an option called OPTIONS calling withSASIOptions deals with this.
```
tatu-at-datastax, line 663: "SASI" seems odd wrt naming. Not that we can change that probably some legacy reason.
Comment could mention it's actually for SAI options.

tatu-at-datastax (approved):
LGTM

src/test/java/io/stargate/sgv2/jsonapi/service/operation/collections/CreateCollectionOperationTest.java
Hazel-Datastax, line 251: Could we add a few assertions on the actual CREATE TABLE / CREATE INDEX CQL strings here, not just the statement count and the parsed table-comment JSON? 

This refactor swaps hand-built strings for SchemaBuilder DSL — semantic intent is the same but the textual output is fully driver-controlled now. The existing assertions would still pass if the SchemaBuilder chain accidentally dropped a column, used a slightly wrong DataType (e.g. ASCII vs TEXT), or changed identifier quoting in a way Cassandra silently accepts but treats differently.

A lightweight snapshot would catch that without being brittle — no need for full string equality, just enough substrings to pin the column set and the key index options:
```
    var statements = schemaChangeMemento.cqlStatements; // or however the fixture exposes them
    var createTableCql = statements.get(0).getQuery();
    assertThat(createTableCql)
        .contains("CREATE TABLE IF NOT EXISTS")
        .contains("key frozen<tuple<tinyint, text>>")
        .contains("query_text_values")
        .contains("query_null_values")
        .contains("query_vector_value")
        .contains("vector<float, 5>");
    var vectorIndexCql = statements.stream()
        .map(SimpleStatement::getQuery)
        .filter(q -> q.contains("query_vector_value") && q.contains("CREATE"))
        .findFirst().orElseThrow();
    assertThat(vectorIndexCql)
        .contains("similarity_function")
        .contains("source_model");
```
Same idea for the lexical index (`index_analyzer`) and for `denyAll` (no indexes emitted). The point is to make a future "simplification" of the SchemaBuilder call chain fail loudly here, instead of slipping past unit tests and surfacing as a production incident.

Hazel-Datastax (approved):

src/main/java/io/stargate/sgv2/jsonapi/service/operation/collections/CreateCollectionOperation.java
amorton: twas a bug, honestly not sure how it compiled :) fixed

src/test/java/io/stargate/sgv2/jsonapi/service/operation/collections/CreateCollectionOperationTest.java
amorton: yes we should it is a big hole that we dont test that, I'll want to keep this PR small so @tatu-at-datastax  does not kill me :) 

My next step was to move all the logic for a CollectionTable (e.g. including the CollectionTableMatcher) into a better single location. 

Then I think the validation of "did we create a good collection table" could run on CQL schema we read, and use the same logic as the API uses when working out if the table is good. i.e. Integration test would create collection , and we can then verify how the API sees the collection. 

Will also think about if / how we inspect the CQL commands as part of unit tests

src/main/java/io/stargate/sgv2/jsonapi/service/operation/collections/CreateCollectionOperation.java
amorton: next PR relocated this code, have updated it there

PR #2489, opened 2026-06-01
Fix #2487: add 3 more vector indexing options for CreateCollection (out of 5 possible)

### LINKED ISSUE #2487: Add vector_indexing to CreateCollection command
Currently "CreateCollection" only takes `VectorSearchDesc` that contains 2 things to optionally configure:

1. `source_model`: values like `ada002`, `bert`
2. `metric` (was: "function", in C "similarity_function"): values like `dot_product`, `cosine`, `euclidean`

There are more options that can be added, listed here:

https://github.com/datastax/cassandra/blob/main/src/java/org/apache/cassandra/index/sai/VECTOR.md?plain=1#L18

and some if not all should be added:

* `maximum_node_connections` (default: 16)
    * valid range: 1-512
* `construction_beam_width` (default: 100)
    * valid range: 1-3200
* `neighborhood_overflow` (default: 1.0 in memtable, 1.2 in compaction)
    * valid range: > 0
* `alpha` (default: dimension > 3 gets 1.2. Otherwise, 2.0 in memtable, 1.4 in compaction)
    * valid range: > 0
* `enable_hierarchy` (default: false)
    * valid range: true/false

**EDIT**: looks like DSE-6.9 does NOT support these `maximum_node_connections`, `construction_beam_width`
or `enable_hierarchy` options (and probably not others either), only HCD does?

Not sure how to proceed.

tatu-at-datastax:
**What this PR does**:

Adds 3 new Vector index options:

* `maximum_node_connections`
* `construction_beam_width`
* `enable_hierarchy`

**EDIT**: options only available on HCD, not DSE-6.9. Not sure how to go about that.

Fixes #2487 

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -40,6 +40,8 @@
     <stargate.int-test.cassandra.auth-enabled>true</stargate.int-test.cassandra.auth-enabled>
     <!-- Defaults are for DSE which does not support (full) lexical feature set so: -->
     <stargate.int-test.lexical-disabled>true</stargate.int-test.lexical-disabled>
+    <!-- DSE 6.9 SAI does not support the 'enable_hierarchy' vector index option so: -->
+    <stargate.int-test.vector-hierarchy-disabled>true</stargate.int-test.vector-hierarchy-disabled>
     <!-- 14-Mar-2025, tatu: no more use of Stargate Coordinator -->
     <!-- Cassandra driver -->
     <driver.version>4.17.0</driver.version>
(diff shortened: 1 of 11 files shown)


tatu-at-datastax:
Not sure we want to take this approach -- more likely we would have "option sets" for usage like:

1. Small data, high recall
2. Big data, low latency

or in general named sets where use selects goal, not settings directly.

tatu-at-datastax:
Will not proceed with this approach -- instead, will approach with more abstract setting/goal.

PR #2497, opened 2026-06-09
fix: countDocuments counts at most one page of keys

### LINKED ISSUE #2496: countDocuments caps at DEFAULT_PAGE_SIZE rather than 1000 on v1.0.47?
Tested with Data API 1.0.47 and HCD 2.0.6. _(The behaviour is correct on v1.0.46)_

When a collection holds more than 50 items, the `select count(*)` command returns the right number all the way, while the `countDocuments` API command caps at 50.

Expectation is: same as 1.0.46 and prior versions = a number matching the count from CQL, all the way to 1000.

Screenshots (side-by-side CQL and data api for 49, 51, 251 documents resp.. Images taken with 1.0.47 + HCD 2.0.6):

Image

Image

Image

erichare:
**What this PR does**:

Fixes a regression, #2496, in 1.0.47 where `countDocuments` returned at most one page of keys (50 with the page size defaults from #2461). `countDocumentsByKey` started the `fetchNextPage()` calls without waiting for them, so only the first page was counted and a failure while fetching a page was dropped. Counting now reads every page of the result set, using the same repeating/whilst pattern as `findOrderDocument`, and surfaces paging failures as `COUNT_READ_FAILED`. Also applies the configured `defaultCountPageSize` to the count query statement: `SimpleStatement` is immutable, so the previous `setPageSize()` call was discarded and count queries paged at the driver profile default instead of the configured value. Adds a unit test for counting across pages and a `CountPaginationIntegrationTest` that counts more documents than fit in one page of keys.

Also restores the global `basic.request.page-size = 50` that #2498 removed as a stopgap for the same regression. With counting now paginating through every page of keys, the global cap no longer truncates counts, and it keeps per-page memory bounded for all driver profiles that do not override it (the original intent of #2461).

Fixes the `countDocuments` regression reported against 1.0.47.

**Checklist**
- [x] Changes manually tested


src/test/java/io/stargate/sgv2/jsonapi/api/v1/CountPaginationIntegrationTest.java
```diff
@@ -0,0 +1,96 @@
...
+ * every page of the result set and not just the first one. DseTestResource is updated to have
+ * maxCountLimit above the page size so the count is not capped before a page boundary is crossed.
+ */
+@QuarkusIntegrationTest
+@QuarkusTestResource(
+    value = CountPaginationIntegrationTest.CountPaginationTestResource.class,
+    restrictToAnnotatedClass = true)
+@TestClassOrder(ClassOrderer.OrderAnnotation.class)
+@TestMethodOrder(MethodOrderer.OrderAnnotation.class)
+public class CountPaginationIntegrationTest extends AbstractCollectionIntegrationTestBase {
```
amorton: There is an exsting CountIntegrationTest can we see about adding it to that ?

src/test/java/io/stargate/sgv2/jsonapi/service/operation/collections/CountCollectionOperationTest.java
```diff
@@ -440,5 +445,68 @@ public void error() {
                 assertThat(je.code).isEqualTo(DatabaseException.Code.COUNT_READ_FAILED.name());
               });
     }
+
+    @Test
+    public void countAcrossMultiplePages() {
```
amorton, line 505: can we add a few comments explaining what is happening here, just a bunch going on that we will need to remember later 

Also: remember that using `var` where it is clear makes the code a lot easier to read

src/main/java/io/stargate/sgv2/jsonapi/service/operation/collections/CollectionReadOperation.java
```diff
@@ -388,22 +388,30 @@ default Uni<CountResponse> countDocumentsByKey(
...
+              Uni<AsyncResultSet> page =
+                  previousPage == null
+                      ? queryExecutor.executeCount(dataApiRequestInfo, simpleStatement)
+                      : Uni.createFrom().completionStage(previousPage.fetchNextPage());
+              return page.onItem().invoke(stateRef::set);
+            })
+        // Read keys while more pages exist, the last page is still emitted and counted
+        .whilst(resultSet -> resultSet.hasMorePages())
+        .collect()
+        .in(AtomicLong::new, (counter, resultSet) -> counter.addAndGet(resultSet.remaining()))
```
amorton, line 411: ok, this is nice. one thing that we need to be aware of is that `resultSet.remaining()` is the remaining rows in the driver iterator for that page, not the number of rows returned in the page. 

as far as I can tell there is no way to get the number of rows in the page, other than reading remaining(). 

so we need a comment, that it is important that the resultset is not iterated, otherwise the count will be low.

src/main/java/io/stargate/sgv2/jsonapi/service/operation/collections/CountCollectionOperation.java
```diff
@@ -73,7 +73,7 @@ private SimpleStatement buildSelectQuery() {
               .build();
     }
     SimpleStatement simpleStatement = query.queryToStatement();
-    simpleStatement.setPageSize(pageSize());
-    return simpleStatement;
+    // SimpleStatement is immutable, setPageSize returns a new statement
+    return simpleStatement.setPageSize(pageSize());
```
amorton, line 77: good catch

amorton (changes requested):
looks good, some comments for the tests to avoid creating more code than we need

src/test/java/io/stargate/sgv2/jsonapi/api/v1/CountPaginationIntegrationTest.java
erichare: Done!

erichare:
@amorton tried to address your comments here, let me know if this looks good!

src/test/java/io/stargate/sgv2/jsonapi/service/operation/collections/CountCollectionOperationTest.java
```diff
@@ -440,5 +445,75 @@ public void error() {
...
```
Hazel-Datastax, line 505: The old `getCount` swallowed the failure from `fetchNextPage()` mid pagination and now it surfaces as `COUNT_READ_FAILED`

Our current tests cover:
- countAcrossMultiplePages covers multi-page success
- the existing error() test covers the initial executeCount(...) failing

Can we have a new test that covers "first page succeeds, a later fetchNextPage() fails"?

Hazel-Datastax (approved):
LGTM

src/test/java/io/stargate/sgv2/jsonapi/service/operation/collections/CountCollectionOperationTest.java
erichare: Just added that test @Hazel-Datastax !

amorton (approved):
LGTM

PR #2499, opened 2026-06-11
Refactor Collection Table Matcher

amorton:
**What this PR does**:

Refactor the ColectionTableMatcher into the SuperShreddingTablePredicate , and create a standard way to define what is in a SuperShredding table, via the SuperShreddingBuilder. We can then use this builder to create different components we need, such as predciate and cql statements to create the table. 

This PR sets a foundation by putting the definition of super shredding in one place, and building a way to test / verify we are creating CQL we expect. With the idea that above that low level CQL testing verification we run tests against the tableMetadata etc.  

This PR only updates the table matcher, next PR's will update how CreateCollectionOperation and related classes work to make a super shredding table. 

**Checklist**
- [ ] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/exception/ErrorFormatters.java
```diff
@@ -123,7 +123,7 @@ public static String errFmt(ApiDataType apiDataType) {
   }
 
   public static String errFmt(DataType dataType) {
-    return nullSafe(dataType, d -> d.asCql(true, true));
+    return nullSafe(dataType, d -> d.asCql(false, true));
```
amorton, line 127: change to not force frozen when it was not needed

src/main/java/io/stargate/sgv2/jsonapi/config/constants/DocumentConstants.java
```diff
@@ -36,44 +36,6 @@ interface Fields {
...
-    /** Document field name that will have text value for which vectorize method in called */
-    String BINARY_VECTOR_TEXT_FIELD = "$binary";
-
-    /** Field name used in projection clause to get similarity score in response. */
-    String VECTOR_FUNCTION_SIMILARITY_FIELD = "$similarity";
-
-    /** Physical table column name that stores the lexical content. */
-    String LEXICAL_INDEX_COLUMN_NAME = "query_lexical_value";
-  }
-
```
amorton, line 76: Moved all names of columns to be in SuperShreddingMetadata

src/main/java/io/stargate/sgv2/jsonapi/service/schema/collections/spec/SuperShreddingMetadata.java
```diff
@@ -0,0 +1,485 @@
...
+            QUERY_DOUBLE_VALUES,
+            QUERY_TEXT_VALUES,
+            QUERY_TIMESTAMP_VALUES,
+            QUERY_NULL_VALUES,
+            QUERY_VECTOR_VALUE,
+            QUERY_LEXICAL_VALUE);
+    List<ColumnDef> PARTITION_KEY = List.of(KEY);
+    List<ColumnDef> OPTIONAL = List.of(QUERY_VECTOR_VALUE, QUERY_LEXICAL_VALUE);
+    List<ColumnDef> REQUIRED = listDifference(ALL, OPTIONAL);
+    List<ColumnDef> REQUIRED_NON_PK = listDifference(REQUIRED, OPTIONAL);
```
clun: should be `listDifference(REQUIRED, PARTITION_KEY)`  you want to find what is required simply not in the PK

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/CollectionIndexUsage.java
clun, line 68: I think we are missing the lexicalIndexTag, adding a line 68

```java
this.vectorIndexTag |= other.vectorIndexTag;
this.lexicalIndexTag |= other.lexicalIndexTag;
```

src/main/java/io/stargate/sgv2/jsonapi/service/schema/collections/spec/SuperShreddingBuilder.java
```diff
@@ -0,0 +1,201 @@
...
+  }
+
+  /**
+   * Gets the index definitions and options for the super shredding table based on {@link
+   * SuperShreddingBinding}
+   *
+   * <p>This pulls the options from the {@link SuperShreddingBinding} and puts them into maps of the
+   * values each index definition needs
+   */
+  protected Stream<IndexDef> indexDefs(SuperShreddingBinding binding) {
```
clun: The parameter `binding` is unused, i think there is no need to introduce the parameter except if you want to override the internal one which would be a mutation different for the rest of the class => delete.

src/main/java/io/stargate/sgv2/jsonapi/util/ColumnMetadataPredicate.java
```diff
@@ -0,0 +1,169 @@
...
+
+  // Compare predicates by the identifier name the column must have.
+  public static final Comparator<ColumnMetadataPredicate> IDENTIFIER_COMPARATOR =
+      Comparator.comparing(ColumnMetadataPredicate::name, CQL_IDENTIFIER_COMPARATOR);
+
+  protected final CqlIdentifier name;
+  protected final DataType type;
+
+  protected ColumnMetadataPredicate(CqlIdentifier name, DataType type) {
+    // no null checks in the ctor, so a subclass can fully override if they want to.
```
clun: ok subclasses can fully override but they would also have to override the `ToString` or you raise a NPE.

i would not make `name` nor `type` ` Objects.requireNonNull()` in the toString() or still add the test in the ctor

src/main/java/io/stargate/sgv2/jsonapi/service/schema/collections/spec/SuperShreddingTablePredicate.java
```diff
@@ -0,0 +1,212 @@
...
+
+  private final SuperShreddingBinding superShreddingBinding;
+  private final List<ColumnMetadataPredicate> expectedOptionals;
+
+  // when non null, this is the list of predicates that defines the columns that are ONLY allowed to
+  // exist
+  private final List<ColumnMetadataPredicate> strictMatch;
+
+  // A def that represents the rules used by the old `CollectionTableMatcher`
+  private static final SuperShreddingBinding BACKWARDS_COMPAT =
```
clun, line 49: `SuperShreddingBinding.builder().build()` 

It seems you are putting all default builder values, now i would understand you want to make them explicitely visible at this level

src/main/java/io/stargate/sgv2/jsonapi/service/schema/collections/spec/SuperShreddingTablePredicate.java
```diff
@@ -0,0 +1,212 @@
...
+ *
+ * <p>Uses the shared abstract definition of super shredding in {@link SuperShreddingMetadata}
+ *
+ * <p><b>Note:</b> How we create the statements for, predicate to test for, and test data to use
+ * with code that uses a super shredding table starts with the {@link SuperShreddingBuilder} class
+ * which has some slightly complex tests around it.
+ *
+ * <p>This class used to be called <code>CollectionTableMatcher</code>
+ *
+ * <p><b>NOTE:</b> As of June 2026, there is no check the indexes are valid, this will be future
```
clun, line 35: Got it, but let me see if i can create a baby index validator after the column valudation in `SchemaObjectFactory`

clun: The predicate checks columns but not indexes. A collection with correct columns but missing/broken SAI indexes passes the predicate and is treated as healthy -- downstream queries then fail with confusing CQL errors.

I've prototyped a `SuperShreddingIndexValidator` on branch `clun/review-2499-index-validator`:

- **`SuperShreddingIndexValidator`** -- validates SAI indexes after the predicate passes, returns a `ValidationResult` with missing/present/unexpected indexes
- **`SuperShreddingIndexValidatorTest`** -- 13 tests covering healthy tables, missing required/optional indexes, unexpected indexes, and warn logging
- **`TableMetadataTestUtil`** additions -- `removeIndex`, `removeAllIndexes`, `clearAllIndexes`, `addIndex`
- **`SchemaObjectFactory`** sketch -- shows where to wire `validate()` after `IS_COLLECTION_PREDICATE.test()` (old code kept as comment)

Branch: [`clun/review-2499-index-validator`](https://github.com/stargate/data-api/tree/clun/review-2499-index-validator) (based on this PR's branch)

clun (commented):
Here are some findings and hopefully helpful new classes as a review

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/executor/CollectionIndexUsage.java
amorton: good catch, i barely looked at that that file.

src/main/java/io/stargate/sgv2/jsonapi/util/ColumnMetadataPredicate.java
amorton: udpated the toString() - when it calls errFmt those functions handle null so no need to check there, they will print null. 

removed code duplication by moving the null check into the super class, there are only 2 direct subclasses that need to worry about the checkNulls param

src/main/java/io/stargate/sgv2/jsonapi/service/schema/collections/spec/SuperShreddingTablePredicate.java
amorton: thanks, will have a look after we get this one in. For this PR I wanted to keep the logic of the predicate the same but change implementation - so all previously passed tests keep passing. 

Future work can expand the logic in the predicate

src/main/java/io/stargate/sgv2/jsonapi/service/schema/collections/spec/SuperShreddingTablePredicate.java
amorton: not sure if this is a half finished comment. 

The BackwardsCompat is there to capture past behaviour for the no param ctor to use for backwards compat. 

So the new builder code to create  the Predicate is only used by tests, until we start changing the tests. Such as also tests for indexes.  

The goal in this PR was to change implementation, but leave behaviour the same.

src/main/java/io/stargate/sgv2/jsonapi/service/schema/collections/spec/SuperShreddingBuilder.java
amorton: removed, I think it was just old code

erichare (approved):

tatu-at-datastax (approved):

PR #2505, opened 2026-06-15
feat: expose extended vector indexing options on createVectorIndex

### LINKED ISSUE #2487: Add vector_indexing to CreateCollection command

### LINKED ISSUE #2509: Add vector_indexing to CreateVectorIndex command

erichare:
**What this does**

Adds a `vectorIndexing` field under `definition.options` of `createVectorIndex` so you can tune the SAI vector index beyond `metric` / `sourceModel`.

One or the other, by JSON type:

- **String** — a profile name from the in-code `VectorIndexProfiles` registry (e.g. `"small-high-recall"`) that expands into a set of SAI options.
- **Object** — raw Cassandra SAI options in snake_case, restricted to an allow-list: `maximum_node_connections`, `construction_beam_width`, `neighborhood_overflow`, `alpha`, `enable_hierarchy`. Unknown keys, reserved keys (`similarity_function`, `source_model`), and non-scalar values are rejected.

A profile and raw options are mutually exclusive. `metric` / `sourceModel` stay as their own fields and aren't allowed inside the options object.

The profile name is not persisted. On read-back (`listIndexes`) the index's tuning options are matched against the known profiles and echoed as the profile name when they match exactly, otherwise as the raw options. Detection is a stopgap and will likely be replaced before prod (see the #2508 discussion).

```json
{
  "createVectorIndex": {
    "name": "idx",
    "definition": {
      "column": "v",
      "options": {
        "metric": "cosine",
        "sourceModel": "openai-v3-small",
        "vectorIndexing": "small-high-recall"
      }
    }
  }
}
```

Or with raw options instead:

```json
"vectorIndexing": { "maximum_node_connections": 32, "alpha": 1.2 }
```

**Issues**
- Fixes #2487
- Field naming per #2509, overloaded shape per #2508

**Testing**
- Unit: deserialization (string/object/invalid), profile expansion, allow-list / reserved-key / non-scalar rejection, profile detection on read-back, apply→describe round-trip.
- Integration (`CreateTableIndexIntegrationTest`): API-validation cases, backend-agnostic (DSE + HCD).
- The tuning options need a cluster that allows custom SAI HNSW params; on clusters that don't, the DB error is surfaced.

**Follow-ups**
- Apply `vectorIndexing` to `createCollection`.
- Externalize profiles to config and tune the mappings.
- Revisit profile detection vs a durable store before prod.

**Checklist**
- [x] Changes manually tested
- [x] Automated tests added/updated
- [x] Documentation added/updated (OpenAPI `@Schema` descriptions)


erichare:
@amorton I've updated the PR so that it doesn't try to use table extensions to persist the profile name, nor does it accept both a profile and options.... they're mutually exclusive as per the design. Just opened it up for review!

PR #2513, opened 2026-06-17
Fix #2462: add api support for filtering counter columns

### LINKED ISSUE #2462: Counter table support lists `filter: true`, contrary to actual support
A table with `counter` columns, when described by the Data API, returns e.g. this (note the ` "filter": true` bit):

```
{
    "status": {
        "tables": [
            {
                "name": "c",
                "definition": {
                    "columns": {
                        "a": {
                            "type": "int"
                        },
                        "b": {
                            "type": "counter",
                            "apiSupport": {
                                "createTable": false,
                                "insert": false,
                                "read": true,
                                "filter": true,
                                "cqlDefinition": "counter"
                            }
                        }
                    },
                    "primaryKey": {
                        "partitionBy": [
                            "a"
                        ],
                        "partitionSort": {}
                    }
                }
            }
        ]
    }
}
```

Yet when running a search that filters on that column, such as

```
{
    "findOne": {
        "filter": {
            "b": 1
        }
    }
}
```

this is what one -predictably- gets:

```
{
    "status": {
        "warnings": [
            {
                "id": "f2b94f83-e974-42ec-8d29-3ea0d528d8a8",
                "family": "REQUEST",
                "scope": "WARNING",
                "errorCode": "MISSING_INDEX",
                "title": "Filter includes columns that are not indexed",
                "message": "The filter includes columns that are not indexed. \n\nThe table default_keyspace.c has the primary key: a(int).\nAnd has indexes on the columns: [None].\nThe request filtered on the un-indexed columns: b.\n\nNote: There may be a small delay in newly created indexes propagating through the system, this warning may be ignored if the column(s) were recently indexed.\n\nThe query was executed without taking advantage of the primary key or indexes on the table, this can have performance implications on large tables.\n\nSee documentation for best practices for filtering."
            }
        ]
    },
    "errors": [
        {
            "id": "69bf86d0-c4ba-4916-9833-0c0ff5f2f9a1",
            "family": "REQUEST",
            "scope": "FILTER",
            "errorCode": "INVALID_FILTER_COLUMN_VALUES",
            "title": "Column value in filter is not valid",
            "message": "Only values that are supported by the column data type can be included when filtering.\n\nThe table default_keyspace.c defines the columns: a(int), b(counter).\nThe command included the column b that has COUNTER type defined.\n\nResend the command using only supported column values."
        }
    ]
}
```

erichare:
## Summary

Fixes #2462.

A table with a `counter` column was described by the Data API with `apiSupport.filter: true`, yet filtering on that column failed at runtime with `INVALID_FILTER_COLUMN_VALUES`. Investigation reveals counters *are* filterable in CQL via `ALLOW FILTERING`.

## Root cause

Filter literals arrive from JSON as `BigDecimal`/`BigInteger`, but `COUNTER` only registered a single codec, `COUNTER_FROM_LONG`. Codec lookup therefore found no codec that `handlesJavaValue(BigDecimal)`, threw `ToCQLCodecException`, and surfaced as `INVALID_FILTER_COLUMN_VALUES`.

## Fix

- Add `COUNTER_FROM_BIG_DECIMAL` and `COUNTER_FROM_BIG_INTEGER` (mirroring `BIGINT`'s variabts ub `JSONCodecs`, and register them in `JSONCodecRegistries`.
- Keep `ApiDataTypeDefs.COUNTER` `apiSupport.filter = true`

## Tests

- `JSONCodecRegistryTest`: added `COUNTER` cases covering `Long` / `BigInteger` / `BigDecimal`
- `filterCounter` rewritten from error-expecting to success-expecting: seeds a counter value via raw CQL then filters on equality and asserts the row is returned with a single `MISSING_INDEX` warning.

## Test plan

- [x] `./mvnw test -Dtest=JSONCodecRegistryTest` passes (JDK 21), formatting compliant
- [x] `./mvnw verify -Dit.test='UnsupportedTypeTableIntegrationTest$Counter'` passes against DSE 6.9.21 (all 5 Counter ITs green)
- [ ] CI: full `UnsupportedTypeTableIntegrationTest` against DSE/HCD matrix

### DIFF
--- a/src/test/java/io/stargate/sgv2/jsonapi/api/v1/tables/UnsupportedTypeTableIntegrationTest.java
+++ b/src/test/java/io/stargate/sgv2/jsonapi/api/v1/tables/UnsupportedTypeTableIntegrationTest.java
@@ -161,6 +161,19 @@ public class UnsupportedTypeTableIntegrationTest extends AbstractTableIntegratio
       assertThat(executeCqlStatement(createTable.build())).isTrue();
     }
 
+    // Regression test for #2462: a counter column must report apiSupport.filter == true in the
+    // table description, because counters ARE filterable via ALLOW FILTERING (they just cannot be
+    // indexed). Filtering succeeds with a MISSING_INDEX warning (see filterCounter() below).
+    @Test
+    @Order(2)
+    public final void listTablesReportsCounterFilterable() {
+      assertNamespaceCommand(keyspaceName)
+          .templated()
+          .listTables(true)
+          .wasSuccessful()
+          .hasTableColumnApiSupport(TABLE_COUNTER, "counter", false, false, true, true);
+    }
+
     // In Cassandra, you cannot use an INSERT statement directly for counters.
     // Instead, counter columns require a special kind of update operation.
     // Counters in Cassandra are designed to increment or decrement a value,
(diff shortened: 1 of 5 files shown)


amorton (changes requested):
-1 Blocking this merging. 

There are two possible fixes here: either the API support flag is wrong (we dont support it) OR we actually should support it and the bug is that we dont. 

we need to decide on what we want to do before letting AI write code. Also, PR comments such as these are not very helpful it is just a mash of words

src/test/java/io/stargate/sgv2/jsonapi/api/v1/tables/UnsupportedTypeTableIntegrationTest.java
```diff
@@ -161,6 +161,25 @@ public final void createDefaultTablesAndIndexes() {
...
+    @Order(2)
+    public final void listTablesReportsCounterNotFilterable() {
+      var apiSupportPath =
+          "status.tables.find { it.name == '%s' }.definition.columns.counter.apiSupport"
+              .formatted(TABLE_COUNTER);
+      assertNamespaceCommand(keyspaceName)
+          .templated()
+          .listTables(true)
+          .wasSuccessful()
+          .body(apiSupportPath + ".createTable", is(false))
```
amorton: this is fragile and encourages copy-past coding, we would create assertion functions for this

amorton:
There is an existing test the confirms some code things we should not be able to filter on a counter 

https://github.com/stargate/data-api/blob/3bf486f2975adad37caeca6948cea8e68f78b877/src/test/java/io/stargate/sgv2/jsonapi/api/v1/tables/UnsupportedTypeTableIntegrationTest.java#L212

BUT checking the design docs we said it was possible to filter on a counter. 

we need to work out what is the correct behaviour

erichare:
Good point @amorton, digging into this in more detail (i.e., whether there's a technical reason why we don't support it, and if so, can we overcome that and support it, as the spec indicates that we want to do)

erichare:
@amorton Updated the PR based on our investigation last week!

PR #2527, opened 2026-07-27
feat: Add authenticated database readiness check based on driver session metadata

### LINKED ISSUE #2526: Implement Cassandra-Based Readiness Probe for Data API
## Background

Currently, Data API pods may remain in a Ready state even when they are unable to communicate with Cassandra. This can lead to scenarios where Kubernetes continues routing traffic to unhealthy pods, resulting in request failures and inconsistent behavior across replicas.

In Kubernetes, readiness probes should reflect whether a pod can successfully perform its core function. Since the Data API depends on Cassandra for all data operations, a pod that cannot connect to Cassandra should fail its readiness check and be removed from service until connectivity is restored.

This approach would simplify the consumer experience by making the Data API either fully available or unavailable, rather than partially functional depending on which pod receives the request.

---

## Problem Statement

Several failure scenarios can occur today:
- Network partition where only some Data API pods can reach Cassandra.
- Cassandra outage or degraded connectivity.
- Invalid or expired Cassandra sessions.
- Infrastructure issues affecting pod-to-Cassandra communication.

In these situations, affected Data API pods may still be considered healthy by Kubernetes and continue receiving traffic, leading to failed API requests.

---

## Proposed Solution

Implement a Quarkus MicroProfile Readiness Health Check that validates Cassandra connectivity.

The readiness check should:

1. Obtain a Cassandra `CqlSession` from the existing session cache.
2. Verify that the session is open and usable.
3. Execute a lightweight query against Cassandra (e.g. `SELECT release_version FROM system.local`).
4. Mark the readiness check as:
   - **UP** when the query succeeds.
   - **DOWN** when session acquisition or query execution fails.

When the readiness check reports **DOWN**, Kubernetes will automatically remove the pod from service endpoints until connectivity is restored.

---

## Expected Benefits

- Prevent traffic from being routed to pods that cannot access Cassandra.
- Improve reliability during partial network failures.
- Align Data API behavior with Kubernetes readiness best practices.
- Provide a simpler service contract for consumers where Data API availability accurately reflects Cassandra availability.
- Reduce intermittent and difficult-to-diagnose request failures.

Proposal 
```java
package io.stargate.sgv2.jsonapi.api.health;

import com.datastax.oss.driver.api.core.CqlSession;
import com.datastax.oss.driver.api.core.cql.SimpleStatement;
import io.stargate.sgv2.jsonapi.api.request.RequestContext;
import io.stargate.sgv2.jsonapi.api.request.UserAgent;
import io.stargate.sgv2.jsonapi.api.request.tenant.Tenant;
import io.stargate.sgv2.jsonapi.api.request.tenant.TenantFactory;
import io.stargate.sgv2.jsonapi.config.OperationsConfig;
import io.stargate.sgv2.jsonapi.service.cqldriver.CQLSessionCache;
import jakarta.enterprise.context.ApplicationScoped;
import jakarta.inject.Inject;
import java.time.Duration;
import org.eclipse.microprofile.health.HealthCheck;
import org.eclipse.microprofile.health.HealthCheckResponse;
import org.eclipse.microprofile.health.HealthCheckResponseBuilder;
import org.eclipse.microprofile.health.Readiness;
import org.slf4j.Logger;

/**
 * Health check that verifies Cassandra connectivity for the Data API readiness probe.
 *
 * <p>This check verifies that the Data API can establish and maintain connections to Cassandra by:
 * <ul>
 *   <li>Obtaining a CQL session from the session cache</li>
 *   <li>Verifying the session is not closed</li>
 *   <li>Executing a simple query against system.local to confirm connectivity</li>
 * </ul>
 *
 * <p>In Kubernetes, this readiness check ensures that pods which cannot reach Cassandra
 * are marked as not ready and removed from the service load balancer, preventing
 * traffic from being routed to unhealthy pods.
 *
 * <p><b>Multi-tenancy Note:</b> This health check uses a default tenant configuration.
 * In multi-tenant deployments, this verifies base Cassandra connectivity but not
 * tenant-specific credentials or permissions. The check will pass if the Data API
 * can connect to Cassandra with default/system credentials.
 */
@Readiness
@ApplicationScoped
public class CassandraConnectionHealthCheck implements HealthCheck {

  private static final Logger LOGGER = LoggerFactory.getLogger(CassandraConnectionHealthCheck.class);

  // Simple query to verify connectivity - reads from system.local which always exists
  private static final String HEALTH_CHECK_QUERY = "SELECT release_version FROM system.local";
  private static final Duration HEALTH_CHECK_TIMEOUT = Duration.ofSeconds(5);
  
  // Health check uses a synthetic tenant ID for verification
  private static final String HEALTH_CHECK_TENANT_ID = "health-check";
  private static final String HEALTH_CHECK_AUTH_TOKEN = "";

  @Inject
  CQLSessionCache sessionCache;

  @Inject
  OperationsConfig operationsConfig;

  @Override
  public HealthCheckResponse call() {

    HealthCheckResponseBuilder responseBuilder = HealthCheckResponse.named("cassandra-connection");

    try {
      // Create a tenant for health check purposes
      Tenant healthCheckTenant = TenantFactory.instance().create(HEALTH_CHECK_TENANT_ID);
      
      // Create a minimal RequestContext for the health check
      RequestContext requestContext = new RequestContext(
          healthCheckTenant,
          HEALTH_CHECK_AUTH_TOKEN,
          new UserAgent("DataAPI-HealthCheck/1.0")
      );

      // Get or create a session - this will verify we can connect to Cassandra
      CqlSession session = sessionCache.getSession(requestContext)
          .await()
          .atMost(HEALTH_CHECK_TIMEOUT);

      // Verify the session is not closed
      if (session.isClosed()) {
        LOGGER.warn("Cassandra session is closed during health check");
        return responseBuilder
            .down()
            .withData("reason", "Session is closed")
            .build();
      }

      // Execute a simple query to verify connectivity and query execution
      SimpleStatement statement = SimpleStatement.builder(HEALTH_CHECK_QUERY)
          .setTimeout(HEALTH_CHECK_TIMEOUT)
          .setConsistencyLevel(operationsConfig.queriesConfig().consistency().reads())
          .build();

      var resultSet = session.execute(statement);
      var row = resultSet.one();
      
      // If we got here, the query succeeded
      String version = row != null ? row.getString("release_version") : "unknown";

      LOGGER.trace("Cassandra health check passed, version: {}", version);
      
      return responseBuilder
          .up()
          .withData("cassandra_version", version)
          .withData("session_name", session.getName())
          .build();

    } catch (Exception e) {
      LOGGER.error("Cassandra health check failed", e);
      return responseBuilder
          .down()
          .withData("error", e.getClass().getSimpleName())
          .withData("message", e.getMessage() != null ? e.getMessage() : "Unknown error")
          .build();
    }
  }
}
```

erichare:
**What this PR does**:

Adds `GET /v1/health/ready`, an authenticated readiness endpoint for both Astra and Cassandra.

- Gets a session via the normal `CQLSessionCache` using the request's tenant, token, and User-Agent, and reports `UP` when the session's driver metadata has an `UP` node. No query is issued and no canary table is needed — a pod that cannot connect fails session creation and reports `DOWN`.
- Requires the configured SLA User-Agent (`stargate.jsonapi.operations.sla-user-agent`) so probe sessions get the shorter SLA cache TTL. Fails closed if it is unset.
- Responses: 200 `{"status":"UP"}`, 401 invalid auth, 403 wrong User-Agent, 503 `{"status":"DOWN"}`. Fully async with a 5s timeout. Readiness polling is excluded from tenant request metrics.

This also means a probe token that authenticates but cannot read the schema makes the endpoint report `DOWN` rather than an auth error.

Fixes #2526

**Validation**:

- Full unit suite green (3441 tests); formatting clean.
- `SessionEvictionIntegrationTest` extended to assert readiness `UP`, `DOWN`, `UP` across a container outage.

**Checklist**

- [ ] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/api/health/CassandraConnectionHealthCheck.java
```diff
@@ -0,0 +1,180 @@
...
+import java.util.Base64;
+import java.util.Objects;
+import org.eclipse.microprofile.health.HealthCheck;
+import org.eclipse.microprofile.health.HealthCheckResponse;
+import org.eclipse.microprofile.health.Readiness;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Health check that verifies Cassandra connectivity for the Data API readiness probe.
```
amorton: -1 comments should explain how this is registered, what url it is under etc. 

the AI generated comments dont help

src/main/java/io/stargate/sgv2/jsonapi/api/health/CassandraConnectionHealthCheck.java
```diff
@@ -0,0 +1,180 @@
...
+ *
+ * <p>Readiness polling deliberately keeps the cached session active while polling continues. If
+ * requests use the same credentials, they share that session; otherwise the check maintains a
+ * dedicated session for the configured default credentials.
+ *
+ * <p>The database type is runtime configuration, so the bean remains registered for other database
+ * types. In those deployments it reports UP without accessing the Cassandra session cache.
+ */
+@Readiness
+@ApplicationScoped
```
amorton: we are trying to remove CDI injection, is there a way to do this without it ?

src/main/java/io/stargate/sgv2/jsonapi/api/health/CassandraConnectionHealthCheck.java
```diff
@@ -0,0 +1,180 @@
...
+      CqlSessionCacheSupplier sessionCacheSupplier, OperationsConfig operationsConfig) {
+    this(
+        Objects.requireNonNull(sessionCacheSupplier, "sessionCacheSupplier must not be null").get(),
+        operationsConfig,
+        HEALTH_CHECK_TIMEOUT);
+  }
+
+  @VisibleForTesting
+  CassandraConnectionHealthCheck(
+      CQLSessionCache sessionCache, OperationsConfig operationsConfig, Duration timeout) {
```
amorton: prefer to abstract out the OperationsConfig properties if we can, again trying to remove the CDI things

src/main/java/io/stargate/sgv2/jsonapi/api/health/CassandraConnectionHealthCheck.java
```diff
@@ -0,0 +1,180 @@
...
+        operationsConfig.databaseConfig().type() == DatabaseType.CASSANDRA
+            ? createAuthToken(operationsConfig.databaseConfig())
+            : null;
+  }
+
+  @Override
+  public HealthCheckResponse call() {
+    var responseBuilder = HealthCheckResponse.named(HEALTH_CHECK_NAME);
+
+    if (operationsConfig.databaseConfig().type() != DatabaseType.CASSANDRA) {
```
amorton: this means it will not run for astra, is this correct ? Confusing that we are adding this but not using it

src/main/java/io/stargate/sgv2/jsonapi/api/health/CassandraConnectionHealthCheck.java
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+    var healthCheckTenant = TenantFactory.instance().create(null);
+
+    try {
+      var session =
+          sessionCache
+              .getSession(healthCheckTenant, authToken, HEALTH_CHECK_USER_AGENT)
+              .await()
+              .atMost(timeout);
```
amorton: await() is blocking for the caller thread, we generally want to avoid making blocking calls. This should be using the UNI framework.

src/main/java/io/stargate/sgv2/jsonapi/api/health/CassandraConnectionHealthCheck.java
```diff
@@ -0,0 +1,180 @@
...
+
+      if (session.isClosed()) {
```
amorton: why do we think this is going to happen ?

src/main/java/io/stargate/sgv2/jsonapi/api/health/CassandraConnectionHealthCheck.java
```diff
@@ -0,0 +1,180 @@
...
+ */
+@Readiness
+@ApplicationScoped
+public class CassandraConnectionHealthCheck implements HealthCheck {
+
+  private static final Logger LOGGER =
+      LoggerFactory.getLogger(CassandraConnectionHealthCheck.class);
+
+  private static final String HEALTH_CHECK_NAME = "cassandra-connection";
+  private static final String HEALTH_CHECK_QUERY = "SELECT release_version FROM system.local";
```
amorton: this table is not replicated, the query is handled completly on the local machine so is not really checking the coordinator can read. A better check would read form a dsitributed table, like the `datastax_sla.check` table

src/main/java/io/stargate/sgv2/jsonapi/api/health/CassandraConnectionHealthCheck.java
```diff
@@ -0,0 +1,180 @@
...
+              .await()
+              .atMost(timeout);
+
+      if (session.isClosed()) {
+        LOGGER.warn("Cassandra session is closed during health check");
+        evictSession(healthCheckTenant);
+        return responseBuilder.down().withData("reason", "Session is closed").build();
+      }
+
+      var statement =
```
amorton: statement never changes, this can be created in the ctor

src/main/java/io/stargate/sgv2/jsonapi/api/health/CassandraConnectionHealthCheck.java
```diff
@@ -0,0 +1,180 @@
...
+          SimpleStatement.builder(HEALTH_CHECK_QUERY)
+              .setTimeout(timeout)
+              .setConsistencyLevel(operationsConfig.queriesConfig().consistency().reads())
+              .build();
+
+      var resultSet = session.execute(statement);
```
amorton: -1 for blocking IO call, we need to use async calls in the Uni framework

src/main/java/io/stargate/sgv2/jsonapi/api/health/CassandraConnectionHealthCheck.java
```diff
@@ -0,0 +1,180 @@
...
+          .down()
+          .withData("error", e.getClass().getSimpleName())
+          .withData("message", e.getMessage() != null ? e.getMessage() : "Unknown error")
+          .build();
+    }
+  }
+
+  private static String createAuthToken(OperationsConfig.DatabaseConfig databaseConfig) {
+    return databaseConfig
+        .fixedToken()
```
amorton: -1 - the code comments for fixedToken explain what this is for it's not something we want to include in actual prod code its for testing

src/main/java/io/stargate/sgv2/jsonapi/api/health/CassandraConnectionHealthCheck.java
```diff
@@ -0,0 +1,180 @@
...
+      if (current instanceof AllNodesFailedException
+          || current instanceof ClosedConnectionException) {
+        return true;
+      }
+      current = current.getCause();
+    }
+    return false;
+  }
+
+  private void evictSession(Tenant tenant) {
```
amorton: -1 , the sessionCache evicts sessions when they TTL. It is also designed to evist sessions faster when they use the SLA checker user agent. Because this is not using the sla user agent they will be treated like regular user sessions and last for 10mins (by default I think). 

Better to use the SLA checker user agent and not do this

src/main/java/io/stargate/sgv2/jsonapi/config/OperationsConfig.java
```diff
@@ -220,16 +220,16 @@ interface DatabaseConfig {
     DatabaseType type();
 
     /**
-     * Username when connecting to cassandra database (when type is {@link DatabaseType#CASSANDRA})
-     * and fixedToken is used
+     * Username used for Cassandra connections when fixedToken is configured, and by the Cassandra
```
amorton: -1 - do not want to overload this

src/test/java/io/stargate/sgv2/jsonapi/api/v1/SessionEvictionIntegrationTest.java
```diff
@@ -30,6 +30,8 @@ public class SessionEvictionIntegrationTest extends AbstractCollectionIntegratio
 
   private static final Logger LOGGER =
       LoggerFactory.getLogger(SessionEvictionIntegrationTest.class);
+  private static final String READINESS_PATH = "/stargate/health/ready";
```
amorton: -1 for anything using stargate in a path name

amorton (changes requested):
-1 this only works in non-astra, uses blocking IO, and reads from a non replicated table so why the C* node will need to be marked as UP to respond it will not verify that its able to communicate with enough nodes to achieve quourm  

need to rethink what we are trying to do here

amorton:
this needs a rethink about what we are trying to do

erichare:
> this needs a rethink about what we are trying to do

understood! i'll address the specific comments you made just for the practice, but yeah understood about rethinking the purpose

src/main/java/io/stargate/sgv2/jsonapi/api/health/CassandraConnectionHealthCheck.java
erichare: good point. i dont see why it would. ill remove it

src/main/java/io/stargate/sgv2/jsonapi/api/health/CassandraConnectionHealthCheck.java
erichare: this was my mistake, i was misunderstanding the intent of the original issue. This now does get more complicated a bit though.... ahh, i see you already responded in the original GH issue as to why it gets more complicated. 

I have a plan and ill update and explain it...

src/main/java/io/stargate/sgv2/jsonapi/api/health/CassandraConnectionHealthCheck.java
erichare: yep, good call. updated

src/main/java/io/stargate/sgv2/jsonapi/api/health/CassandraConnectionHealthCheck.java
erichare: I did some reading, correct me if i'm wrong about any of this... but Quarkus’s standard application health-check mechanism is CDI-based. There is an alternative tho, SmallRye’s programmatic HealthRegistry. https://quarkus.io/guides/smallrye-health https://smallrye.io/docs/smallrye-health/3.0.0/health-registry.html. giving it a try

src/main/java/io/stargate/sgv2/jsonapi/api/health/CassandraConnectionHealthCheck.java
erichare: Cleaned it up and talked about how its registered

erichare:
Okay, @amorton , so here's what i did...

First of all, renamed the endpoint to `GET /v1/health/ready`, no stargate in the path lol. Its an authenticated database readiness endpoint used for both Astra and Cassandra deployments. It gets a session through the normal session cache. 

The endpoint executes `SELECT * FROM datastax_sla.check LIMIT 1` with the `table-read` driver
profile. That profile uses `LOCAL_QUORUM`, so an `UP` response confirms that the coordinator can
complete a read from a replicated table at local quorum. It does not validate every tenant's
credentials, write availability, or cross-region availability.

Switched everything to async, response would look like `{"status":"UP"}`... and the readiness probe is pure java, no CDI.

I'll push shortly

src/test/java/io/stargate/sgv2/jsonapi/api/v1/SessionEvictionIntegrationTest.java
erichare: haha fair enough!

src/main/java/io/stargate/sgv2/jsonapi/api/health/CassandraConnectionHealthCheck.java
erichare: Explained the new approach as a separate comment

src/main/java/io/stargate/sgv2/jsonapi/api/health/CassandraConnectionHealthCheck.java
erichare: Got it

PR #2530, opened 2026-08-05
fix: map unexpected reranking gateway error

erichare:
**What this PR does**:

Fixes the reranking gateway failure:

`No enum constant io.stargate.sgv2.jsonapi.exception.SchemaException.Code.UNEXPECTED_SERVER_ERROR`

`RerankingEGWClient` assumed every error code returned by Embedding Gateway was a `SchemaException.Code`. Embedding Gateway uses `ServerException.Code.UNEXPECTED_SERVER_ERROR` for non-API exception failures, so that enum lookup masked the original provider failure with an `IllegalArgumentException`.

This change maps that known server code back to `ServerException`, then leaves the existing schema-error path unchanged. It preserves the public `SERVER_UNEXPECTED_SERVER_ERROR` identity used by responses and metrics while retaining the original gateway error body.

Validation:

- 4 focused gateway-client tests passed (new: `handleUnmappedErrorCode`)
- formatting clean (`fmt:format`, 0 reformatted)

N/A — follow-up to the GPU-plane concurrency investigation.

**Checklist**
- [ ] Changes manually tested after deployment


erichare:
@amorton i know this might be more of a workaround than a solution, so i understand if you just want to discard it and "do it right", and im happy to spend more time. But was wondering if you think doing something like this could be a temporary way to improve the situation, at least allowing us to more easily see the root nvidia error?

Edit: my thought was we could at least perhaps deploy this to dev to help debug the underlying errors we're seeing more easily, but let me know what you think

erichare:
Folded into #2533 (combined reranking-resilience PR) as a single squashed commit — net change is identical.

erichare:
@amorton if its helpful this is the PR i had opened before when i was testing that error.

erichare:
Closing in favor of #2549

PR #2531, opened 2026-08-06
Prime reranking auth before parallel batches

erichare:
**What this PR does**:

Runs reranking batch 0 to completion before constructing and subscribing to the remaining batches. A successful first batch primes the upstream API Gateway authorizer cache; the tail batches then retain their existing concurrent, fail-fast execution.

This reduces the authorization cache-miss fan-out of one multi-batch `findAndRerank` request from as many as one miss per batch to one miss per request. Concurrent top-level requests can still race each other, so this is intentionally paired with the Lambda timeout and resilience changes in riptano/cloud-helm-charts#1095.

The tradeoff is one reranking-batch round trip on the critical path before the concurrent tail starts.

Companion to riptano/cloud-helm-charts#1095.

**Validation**:

- `JAVA_HOME=/Library/Java/JavaVirtualMachines/graalvm-21.jdk/Contents/Home ./mvnw -B -ntp -Dstyle.color=always clean test` (3,424 tests)
- Focused reranking provider and gateway tests (10 tests)
- Added coverage for first-batch sequencing, concurrent tail execution, first-batch failure, single-batch requests, empty passages, result indexes, and model usage aggregation

**Checklist**

- [ ] Changes manually tested in a deployed environment
- [x] Documentation not required (no API or configuration change)

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/reranking/operation/RerankingProvider.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/reranking/operation/RerankingProvider.java
@@ -79,16 +79,40 @@ public abstract class RerankingProvider extends ProviderBase {
   public Uni<RerankingResponse> rerank(
       String query, List<String> passages, RerankingCredentials rerankingCredentials) {
 
-    // TODO: what to do if passages is empty?
     List<List<String>> passageBatches = createPassageBatches(passages);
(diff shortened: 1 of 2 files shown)


amorton (commented):
-1 blocking until we know more , at the surface level this does not feel like the best idea, we should fix the upstream problem

amorton (changes requested):

PR #2533, opened 2026-08-10
fix: don't retry rerank timeouts, raise read timeout to 20s

erichare:
Increase the timeout and remove retries from the `/v1/reranking` call.

## Changes

- Don't retry rerank calls. Timeouts were the only retried failure and saturated an already saturated queue
- Raise the default read timeout from 5s to 20s, roughly the old worst-case budget (4 attempts x 5s). Overridable in `reranking-providers-config.yaml`.
- NOTE: better queue drain algos in 2.0.0 rust NIMs, investigate after upgrade

## Testing

New `RerankingProviderTest#timeoutsAreNotRetried`; all rerank provider tests pass.


amorton (changes requested):
-1 trying to solve too many problems, can we pick one at a time and fix ?

amorton (changes requested):
-1 for overly wide changes to the default and removing the retry on timeout

PR #1826, opened 2025-01-14
Fix #1817: handle null/missing vector value with find/findOne

tatu-at-datastax:
**What this PR does**:

Fixes #1817 wherein missing/null `vector` column value (API Tables) handling throws and returns HTTP 500 fail for `find`/`findOne`

**Which issue(s) this PR fixes**:
Fixes #1817

**Checklist**
- [ ] Changes manually tested
- [x] Automated Tests added/updated
- [ ] Documentation added/updated
- [x] CLA Signed: [DataStax CLA](https://cla.datastax.com/)

Issue #1817, opened 2025-01-08
Reading a null vector column returns `IllegalArgumentException` in find command


Hazel-Datastax:
Create Table:
```json
{
    "createTable": {
        "name": "test",
        "definition": {
            "columns": {
                "id": "text",
                "vector": {
                    "type": "vector",
                    "dimension": 5,
                    "service": {
                        "provider": "openai",
                        "modelName": "text-embedding-3-small"
                    }
                }
            },
            "primaryKey": "id"
        },
        "options": {
            "ifNotExists": true
        }
    }
}
```

Insert:
```json
{
    "insertMany": {
        "documents": [
            {
                "id": "4",
                "vector": ""
            },
            {
                "id": "5",
                "vector": "def"
            }
        ]
    }
}
```

Find:
```
{
    "find": {
    }
}
```

Response on find:
```json
{
    "errors": [
        {
            "message": "Server failed: root cause: (java.lang.IllegalArgumentException) Input ByteBuffer must not be null and must have non-zero remaining bytes",
            "errorCode": "SERVER_UNHANDLED_ERROR",
            "exceptionClass": "JsonApiException",
            "id": "c9e3b865-13f3-4978-84c5-5def3f5c9898",
            "family": "SERVER",
            "title": "Server failed",
            "scope": "DATABASE"
        }
    ]
}
```

What I found in CQL:
```
cassandra@cqlsh:default_keyspace> select * from test;

 id | vector
----+----------------------------------------------------------------
  4 |                                                           null
  5 | b'>\xb2\xf7\xe6=\x12\xe0q>\xf9\x10\xed?\x06\x9c\x0f?\x1aS\x19'

(2 rows)
```

amorton:
from discussions - prob something in the Json Codec converting from the Java value to how we want to show it in JSON

PR #2538, opened 2026-08-10
Update openjdk and quarkus version

Hazel-Datastax:
**What this PR does**:
Update openjdk and quarkus version

Fixes [#<issue number>](https://github.com/riptano/mission-control/issues/3143)

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/metrics/CommandFeatures.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/metrics/CommandFeatures.java
@@ -76,15 +76,17 @@ public final class CommandFeatures {
   }
 
   /**
-   * Generates Micrometer Tags representing the features in this instance. Each feature is
-   * represented as a tag with its name and a value of {@code true}.
+   * Generates Micrometer Tags representing the features in this instance. Every {@link
+   * CommandFeature} is emitted as a tag, with value {@code true} if used and {@code false}
+   * otherwise: Prometheus requires all meters with the same name to have the same tag keys, so the
+   * key set must not vary with which features a command used.
    *
-   * @return A {@link Tags} object containing the tags for each feature.
+   * @return A {@link Tags} object containing a tag for each feature.
    */
   public Tags getTags() {
     return Tags.of(
-        commandFeatures.stream()
-            .map(f -> Tag.of(f.getTagName(), String.valueOf(true)))
+        EnumSet.allOf(CommandFeature.class).stream()
+            .map(f -> Tag.of(f.getTagName(), String.valueOf(commandFeatures.contains(f))))
             .toArray(Tag[]::new));
   }
 
--- a/src/main/java/io/stargate/sgv2/jsonapi/metrics/CommandFeatures.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/metrics/CommandFeatures.java
@@ -12,11 +12,14 @@ import java.util.Objects;
  * It uses an {@link EnumSet} internally for efficient storage and operations on commandFeatures.
  */
 public final class CommandFeatures {
-  private final EnumSet<CommandFeature> commandFeatures;
 
-  /** A instance representing no commandFeatures in use. */
+  /** An instance representing no commandFeatures in use. */
   public static final CommandFeatures EMPTY =
       new CommandFeatures(EnumSet.noneOf(CommandFeature.class));
+
+  private static final EnumSet<CommandFeature> ALL_FEATURES = EnumSet.allOf(CommandFeature.class);
+
+  private final EnumSet<CommandFeature> commandFeatures;
 
   /** Private constructor, use factory methods 'of' or 'create' */
   private CommandFeatures(EnumSet<CommandFeature> commandFeatures) {
@@ -76,16 +79,22 @@ public final class CommandFeatures {
   }
 
   /**
-   * Generates Micrometer Tags representing the features in this instance. Every {@link
-   * CommandFeature} is emitted as a tag, with value {@code true} if used and {@code false}
-   * otherwise: Prometheus requires all meters with the same name to have the same tag keys, so the
-   * key set must not vary with which features a command used.
+   * Generates Micrometer Tags representing the features in this instance.
+   *
+   * <p>Every {@linkCommandFeature} is emitted as a tag, with value {@code true} if used and {@code
+   * false} otherwise. Prometheus / micrometer requires all meters with the same name to have the
+   * same tag keys, so all tags are used even for features not used by the request.
+   *
+   * <p>Not a Micrometer change — the check is identical in 1.14.7 and 1.17.0. Quarkus added the
+   * throw: 3.38.1 vs 3.30.8 Before that it was a silent no-op
+   * (L1240)[https://github.com/micrometer-metrics/micrometer/blob/v1.14.7/micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java#L1240-L1244]
+   * , so these metrics were being dropped since #2058, not just now.
    *
    * @return A {@link Tags} object containing a tag for each feature.
    */
   public Tags getTags() {
     return Tags.of(
-        EnumSet.allOf(CommandFeature.class).stream()
+        ALL_FEATURES.stream()
             .map(f -> Tag.of(f.getTagName(), String.valueOf(commandFeatures.contains(f))))
             .toArray(Tag[]::new));
   }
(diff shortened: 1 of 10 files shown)


src/test/java/io/stargate/sgv2/jsonapi/service/cqldriver/optvector/SubtypeOnlyFloatVectorToArrayCodecTest.java
```diff
@@ -34,8 +34,10 @@ public void shouldEncode() {
   @Test
   public void shouldDecode() {
     assertThat(decode(VECTOR_HEX_STRING)).isEqualTo(VECTOR);
-    assertThatThrownBy(() -> decode("0x")).isInstanceOf(IllegalArgumentException.class);
-    assertThatThrownBy(() -> decode(null)).isInstanceOf(IllegalArgumentException.class);
+    // Unlike the driver's FloatVectorToArrayCodec, this codec decodes null/empty
+    // input as null: missing/null-valued vectors reach decode() this way
+    assertThat(decode("0x")).isNull();
+    assertThat(decode(null)).isNull();
```
Hazel-Datastax, line 40: it was never executed — junit4 `@Test` with no `junit-vintage-engine` on the classpath, so the platform discovered 0 tests in it. with junit5 it ran for the first time. 

the assertion is just stale: #1817 (fixed in #1826) intentionally changed empty input to return `null` instead of throwing, and the test never got updated.

erichare (approved):

src/main/java/io/stargate/sgv2/jsonapi/metrics/CommandFeatures.java
```diff
@@ -76,15 +76,17 @@ public boolean isEmpty() {
   }
 
   /**
-   * Generates Micrometer Tags representing the features in this instance. Each feature is
-   * represented as a tag with its name and a value of {@code true}.
+   * Generates Micrometer Tags representing the features in this instance. Every {@link
+   * CommandFeature} is emitted as a tag, with value {@code true} if used and {@code false}
+   * otherwise: Prometheus requires all meters with the same name to have the same tag keys, so the
```
amorton: It's not Prometheus. 

Prometheus is the server that scrapes the metrics, are you thinking about micrometer the Metrics library we use ?

erichare: yeah, micrometer. but interestingly, micrometer itself attributes that behavior to Prometheus https://github.com/micrometer-metrics/micrometer/blob/v1.17.0/implementations/micrometer-registry-prometheus-simpleclient/src/main/java/io/micrometer/prometheus/PrometheusMeterRegistry.java#L593-L596

i guess we're using the Prometheus registry backend in micrometer? https://github.com/stargate/data-api/blob/7badbc0347334d81ea47623453f978c2dff4b326/pom.xml#L141-L144

src/main/java/io/stargate/sgv2/jsonapi/metrics/CommandFeatures.java
```diff
@@ -76,15 +76,17 @@ public boolean isEmpty() {
...
    *
-   * @return A {@link Tags} object containing the tags for each feature.
+   * @return A {@link Tags} object containing a tag for each feature.
    */
   public Tags getTags() {
     return Tags.of(
-        commandFeatures.stream()
-            .map(f -> Tag.of(f.getTagName(), String.valueOf(true)))
+        EnumSet.allOf(CommandFeature.class).stream()
+            .map(f -> Tag.of(f.getTagName(), String.valueOf(commandFeatures.contains(f))))
```
Hazel-Datastax: Key set must stay constant per meter name — Micrometer checks with a plain `equals` on tag keys: [PrometheusMeterRegistry#L589-L595](https://github.com/micrometer-metrics/micrometer/blob/v1.14.7/implementations/micrometer-registry-prometheus-simpleclient/src/main/java/io/micrometer/prometheus/PrometheusMeterRegistry.java#L589-L595)

This fixes stabilize that set: `CommandFeatures` now always emits all 7 keys (value true/false) instead of only used ones; 

Same for `TenantRequestMetricsTagProvider` below:  It falls back to `unknown` instead of throwing (which was dropping the `tenant` key).

Not a Micrometer change — the check is identical in 1.14.7 and 1.17.0. Quarkus
added the throw: [3.38.1](https://github.com/quarkusio/quarkus/blob/3.38.1/extensions/micrometer/runtime/src/main/java/io/quarkus/micrometer/runtime/export/PrometheusMeterRegistryProducer.java#L26-L29) vs [3.30.8 — no such call](https://github.com/quarkusio/quarkus/blob/3.30.8/extensions/micrometer/runtime/src/main/java/io/quarkus/micrometer/runtime/export/PrometheusMeterRegistryProducer.java)

Before that it was a silent no-op ([L1240](https://github.com/micrometer-metrics/micrometer/blob/v1.14.7/micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java#L1240-L1244)), so these metrics were being dropped since #2058, not just now.

src/main/java/io/stargate/sgv2/jsonapi/metrics/TenantRequestMetricsTagProvider.java
```diff
@@ -36,8 +36,17 @@ public TenantRequestMetricsTagProvider(
...
+    // Prometheus requires all meters with the same name to have the same tag keys. If this
+    // contributor throws (e.g. no active request scope for early-rejected requests), Quarkus
+    // records the HTTP metric without our tags, poisoning the meter's key set — so always
+    // resolve to a fallback value instead of failing.
+    String tenantValue;
+    try {
+      tenantValue = requestContext.tenant().toString();
+    } catch (RuntimeException e) {
+      tenantValue = UNKNOWN_VALUE;
+    }
```
Hazel-Datastax, line 61: Not adding this would fail IT — confirmed from the actual failing run.

A request to an unmatched route (e.g. `/unknown/...`) hits
`TenantRequestMetricsTagProvider.contribute()` before our request scope is active, so `requestContext.tenant()` throws `ContextNotActiveException`. Quarkus doesn't drop the metric — it records `http_server_requests_seconds` **without** the `tenant` tag. Any normal request after that (which does have `tenant`) now registers a different key set for the same meter name → `IllegalArgumentException`.

Seen in the CI [log](https://github.com/stargate/data-api/actions/runs/31440073152/job/93622786369) before this fix:
> existing meter 'http_server_requests_seconds' keys `[method, module, outcome, status, uri]`
> attempting to register keys `[method, module, outcome, status, tenant, uri]`


The `try/catch` just makes sure `tenant` is *always* present (falling back to `unknown`), so the key set never depends on whether resolving the tenant succeeded.

src/main/java/io/stargate/sgv2/jsonapi/metrics/TenantRequestMetricsTagProvider.java
```diff
@@ -36,8 +36,17 @@ public TenantRequestMetricsTagProvider(
...
   public Tags contribute(Context context) {
-    // resolve tenant
-    Tag tenantTag = Tag.of(config.tenantTag(), requestContext.tenant().toString());
+    // Prometheus requires all meters with the same name to have the same tag keys. If this
+    // contributor throws (e.g. no active request scope for early-rejected requests), Quarkus
+    // records the HTTP metric without our tags, poisoning the meter's key set — so always
+    // resolve to a fallback value instead of failing.
+    String tenantValue;
+    try {
+      tenantValue = requestContext.tenant().toString();
```
amorton, line 52: this try-catch block is unnecessary, the request context always has a tenant and toString() does nothing.  The code in the consturctor creates a tenant, and the comments on the tenant() function say it will return non-null. 

and catching the RuntimeError is too wide, it covers every possible non-checked exception and swallows it. This is not great in Java because it can stop shutdowns and hide real problems.

src/main/java/io/stargate/sgv2/jsonapi/metrics/CommandFeatures.java
```diff
@@ -76,15 +76,17 @@ public boolean isEmpty() {
...
+   * key set must not vary with which features a command used.
    *
-   * @return A {@link Tags} object containing the tags for each feature.
+   * @return A {@link Tags} object containing a tag for each feature.
    */
   public Tags getTags() {
     return Tags.of(
-        commandFeatures.stream()
-            .map(f -> Tag.of(f.getTagName(), String.valueOf(true)))
+        EnumSet.allOf(CommandFeature.class).stream()
```
amorton: nit: we could store the Enum of all vaules as static

amorton (changes requested):
couple of tweaks to reduce unneeded AI code. I will make the tweaks

src/main/java/io/stargate/sgv2/jsonapi/metrics/TenantRequestMetricsTagProvider.java
amorton: it does not need such a wide error check to get this effect, catching and hiding more errors that you are expecting is an anti-pattern

amorton (approved):

PR #2542, opened 2026-08-12
FIX: #2539 check CqlSession has metadata

### LINKED ISSUE #2539: Ensure New CqlSessions have driver metadata loaded.
When a new session is created the Java driver immediately tries to read the system schema here 

https://github.com/apache/cassandra-java-driver/blob/4.x/core/src/main/java/com/datastax/oss/driver/internal/core/metadata/schema/queries/CassandraSchemaQueries.java#L110

If there is an error while refreshing the schema is is handled here 

https://github.com/apache/cassandra-java-driver/blob/4.x/core/src/main/java/com/datastax/oss/driver/internal/core/session/DefaultSession.java#L456

This results in the driver not having metadata, but the completing the session creation as normal.

Errors such as a Authorization error reading from a schema table, or inconsistent Authentication can result in this code path being followed. 

The result is the API gets a session from the driver, but the session does not have metadata on it. So any request for keyspace / table or collection may fail because from the API it looks like there are no keyspaces. 

The fix for this will be: 
1. When creating a session check that we have `system.local` table , prob do this in the `CqlSessionFactory` class. 
2. If this is not present return a SERVER family error that we could not connect the session. We need to make it clear what the fail is, but dont give out the names of the keyspaces we can see as this is a potential information leak. 

Below is some testing in Python that we can always see this table, checked against a local running HCD and production Astra DB (in the example below): 

```
In [1]: from cassandra.cluster import Cluster
   ...: from cassandra.auth import PlainTextAuthProvider
   ...: 
   ...: cluster = Cluster(
   ...:     cloud={"secure_connect_bundle": "scb-smoketest-prod-aws-us-east-2.zip"},
   ...:     auth_provider=PlainTextAuthProvider("token", "AstraCS:SawrYIm..."),
   ...: )
   ...: session = cluster.connect()
   ...: 
   ...: meta = cluster.metadata
   ...: print(meta.cluster_name)
   ...: print(meta.partitioner)
   ...: print(list(meta.keyspaces))
cndb
org.apache.cassandra.dht.Murmur3Partitioner
['datastax_sla', 'data_endpoint_auth', 'system_auth', 'system_schema', 'system', 'system_traces', 'default_keyspace', 'system_views', 'system_virtual_schema']

In [2]: for name in sorted(cluster.metadata.keyspaces["system"].tables):
   ...:     print(name)
   ...: 
IndexInfo
available_ranges
available_ranges_v2
batches
built_views
compaction_history
local
paxos
peer_events
peer_events_v2
peers
peers_v2
prepared_statements
repairs
size_estimates
sstable_activity
sstable_activity_v2
table_estimates
transferred_ranges
transferred_ranges_v2
view_builds_in_progress
 
```

amorton:
Add check when creating a new CqlSession that it has Metadata so we know what keyspaces/tables in the DB.

See comments in CqlSessionFactory
Fixes #2539 

**Checklist**
- [ ] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/CqlSessionFactory.java
```diff
@@ -213,11 +217,38 @@ public CompletionStage<CqlSession> apply(Tenant tenant, CqlCredentials credentia
...
+              // or if some coordinators do and some do not validate the credentials.
+              // In Java Driver see CassandraSchemaQueries.executeOnAdminExecutor() and
+              // DefaultSession.initialSchemaRefresh()
+              // NOTE: while throwing the error should prevent the session from getting into the
+              // CqlSessionCache
+              // using ExceptionFlags.UNRELIABLE_DB_SESSION tells the CommandProcessor we want to
+              // evict the session
+              // when from cache when the request is over.
+              if (cqlSession.getMetadata().getKeyspace("system").isEmpty()) {
+                throw DatabaseException.Code.FAILED_TO_READ_METADATA.get(
```
erichare: We called .buildAsync() already right, does that mean we should explicitly close the session before throwing the exception here? Any consequence that the session remains "open" (if in fact it does?)

erichare (approved):
@amorton approved but left one comment, not 100% sure if its a concern or not.

Hazel-Datastax (approved):

src/main/java/io/stargate/sgv2/jsonapi/service/cqldriver/CqlSessionFactory.java
amorton: good catch, should close in that situation

PR #2545, opened 2026-08-13
check http status when checking content type

### LINKED ISSUE #2544: RERANKING_PROVIDER_UNEXPECTED_RESPONSE returned when Reranking provider returns non 2XX response
The `RerankingProviderResponseValidation` is used to check the response from reranking provider is the expected content type. 

Currently if the response is a non 2XX error the content type is plain text, not the expected JSON, and this throws a `RERANKING_PROVIDER_UNEXPECTED_RESPONSE` error saying the content type was wrong. 

It needs to check the http status the same way the `EmbeddingProviderResponseValidation` does 

```java
    // only validate for successful responses, errors may return non-JSON content,
    // e.g. a HTTP 401 may just have "Unauthorized" in the response body
    if (responseContext.getStatusInfo().getFamily() != Response.Status.Family.SUCCESSFUL) {
      return;
    }
```

Bonus points - this two classes are essentially the same, the only difference is the error they throw. Refactor so there is a single instance and the fix only needs to be in one place.

amorton:
check http status when checking content type

and refactor to have all the content type filter in one class, the only difference was the error code.

**What this PR does**:

Fixes #2544 

**Checklist**
- [ ] Changes manually tested


src/main/java/io/stargate/sgv2/jsonapi/service/provider/ProviderContentTypeFilter.java
```diff
@@ -51,27 +61,32 @@ public void filter(ClientRequestContext requestContext, ClientResponseContext re
 
     // Throw error if there is no response body
     if (!responseContext.hasEntity()) {
-      throw EmbeddingProviderException.Code.EMBEDDING_PROVIDER_UNEXPECTED_RESPONSE.get(
+      throw unexpectedResponseCode.get(
           Map.of("errorMessage", "No response body from the embedding provider"));
```
erichare: Would we be throwing an error that says "embedding provider" even when the error is from the reranker?

erichare (approved):
Found only one thing that might be an issue, just wording. LGTM otherwise!

src/main/java/io/stargate/sgv2/jsonapi/service/provider/ProviderContentTypeFilter.java
amorton: yeah good catch, will tweak the message to just say "model provider" the code will be different

PR #2554, opened 2026-08-19
fix: disable provenance attestations to remove unknown/unknown manife…

clun:
## Problem

When inspecting `stargateio/data-api:v1.0.49` (and earlier releases) on Docker Hub, the image index contains **four** manifests instead of the expected two:

| Platform | Expected? |
|---|---|
| `linux/amd64` | ✅ |
| `linux/arm64` | ✅ |
| `unknown/unknown` | ❌ |
| `unknown/unknown` | ❌ |

The two spurious `unknown/unknown` entries are **provenance attestation manifests** automatically generated by Docker BuildKit. This causes problems for tooling and runtimes that iterate over the manifest list and do not expect non-platform entries.

The same issue was previously observed and fixed in the **Reaper** image by disabling provenance and SBOM attestations at build time (`provenance: false`, `sbom: false`).

## Root cause

`docker/setup-buildx-action` enables BuildKit by default. Starting with BuildKit 0.11, multi-platform builds automatically attach two OCI artefacts to the image index:

1. A **provenance attestation** (SLSA) — recorded as `unknown/unknown` with `mediaType: application/vnd.in-toto+json`
2. An **SBOM attestation** — also recorded as `unknown/unknown`

Neither is requested explicitly in our workflows; they are injected by the BuildKit daemon unless explicitly suppressed.

## Fix

Add `provenance: false` to every `docker/setup-buildx-action` step. This tells BuildKit not to generate or attach attestation manifests, leaving the image index with only the intended platform-specific entries.

Three steps were updated:

- `.github/workflows/release.yaml` — `publish-image-dockerhub` job
- `.github/workflows/release.yaml` — `publish-image-ecr` job
- `.github/workflows/docker-image-publish.yaml` — `main` job

```yaml
- name: Set up Docker Buildx
  uses: docker/setup-buildx-action@v4
  with:
    provenance: false   # prevents unknown/unknown attestation manifests
```

## Verification

After merging, inspect the next published image with:

```bash
docker buildx imagetools inspect stargateio/data-api:<tag>
```

The manifest list should contain exactly two entries: `linux/amd64` and `linux/arm64`.

### DIFF
--- a/.github/workflows/docker-image-publish.yaml
+++ b/.github/workflows/docker-image-publish.yaml
@@ -52,6 +52,8 @@ jobs:
 
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v4
+        with:
+          provenance: false
 
       - name: Configure AWS credentials via OIDC
         uses: aws-actions/configure-aws-credentials@v6
--- a/.github/workflows/docker-image-publish.yaml
+++ b/.github/workflows/docker-image-publish.yaml
@@ -52,8 +52,6 @@ jobs:
 
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v4
-        with:
-          provenance: false
 
       - name: Configure AWS credentials via OIDC
         uses: aws-actions/configure-aws-credentials@v6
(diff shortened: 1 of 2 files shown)


erichare (commented):
@clun I might be wrong, but im not sure provenance is a parameter of the setup-buildx-action... https://github.com/docker/setup-buildx-action/blob/v4/action.yml i dont see it here. That in mind, i'm wondering if we need to instead put on the build / push steps:
```
  env:
    BUILDX_NO_DEFAULT_ATTESTATIONS: 1
```

See: https://docs.docker.com/build/metadata/attestations/ . let me know if you agree and happy to push changes to your PR if so

Hazel-Datastax (approved):
LGTM

PR #2561, opened 2026-08-21
FIX #2562: send "dimensions" to Nvidia embedding NIM for variable-dimension models

### LINKED ISSUE #2562: Vectorize with nvidia/llama-3.2-nv-embedqa-1b-v2 returns 2048-dim vector when collection dimension is set (e.g. 512)
Reported by @hemidactylus while testing Data API v1.0.51 on dev.

Create a collection with an explicit dimension in the allowed range, then vectorize:

```json
{"createCollection": {"name": "new_nvidia_setdimension", "options": {"vector": {
  "dimension": 512, "metric": "cosine",
  "service": {"provider": "nvidia", "modelName": "nvidia/llama-3.2-nv-embedqa-1b-v2", "parameters": {}}}}}}
```
```json
{"insertOne": {"document": {"$vectorize": "bla"}}}
```

```
EMBEDDING_PROVIDER_UNEXPECTED_RESPONSE: Embedding provider 'nvidia' did not return expected embedding length. Expect: '512'. Actual: '2048'.
```

Stefano's diagnosis: 2048 is the model's `defaultValue` in `findEmbeddingProviders`, which suggests the dimension parameter does not make its way to the call to the embedding service. Also, allowed dimensions being a finite enumeration is almost unique (only AWS Titan does the same, and is rarely tested).

Confirmed: `NvidiaEmbeddingRequest` has no `dimensions` field, so the NIM always returns its default 2048.

erichare:
## Summary

`NvidiaEmbeddingRequest` never carried `dimensions`, so the Collection/Table dimension never reached the NIM. For variable-dimension models (`nvidia/llama-3.2-nv-embedqa-1b-v2`: 384/512/768/1024/2048, default 2048) the NIM always returned 2048-dim vectors and vectorize failed:

```
EMBEDDING_PROVIDER_UNEXPECTED_RESPONSE: Embedding provider 'nvidia' did not return expected embedding length. Expect: '512'. Actual: '2048'.
```

## Changes

- `NvidiaEmbeddingRequest.dimensions`, serialized only when non-zero.  The NIM rejects `dimensions: 0` with HTTP 400, so it must be omitted (not zeroed) for fixed-dimension models.

## Verification

- Dev GPU-plane NIM, direct calls: no `dimensions` → 2048; `512` → 512; `384` → 384; `0` → HTTP 400.
- `./mvnw test -Dtest=NvidiaEmbeddingClientTest,EmbeddingProviderErrorMessageTest,EmbeddingGatewayClientTest,OpenAiEmbeddingClientTest` → 19 tests, 0 failures.

### DIFF
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/EmbeddingProvider.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/EmbeddingProvider.java
@@ -148,8 +148,7 @@ public abstract class EmbeddingProvider extends ProviderBase {
 
   /**
    * Helper method that has logic wrt whether an Nvidia model accepts {@code "dimensions"} parameter
-   * or not: only models without a fixed {@code vector-dimension} (i.e. with a {@code
-   * vectorDimension} parameter, like {@code nvidia/llama-3.2-nv-embedqa-1b-v2}) do.
+   * or not.
    */
   protected static boolean acceptsNvidiaDimensions(
       EmbeddingProvidersConfig.EmbeddingProviderConfig.ModelConfig modelConfig) {
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/EmbeddingProvider.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/EmbeddingProvider.java
@@ -152,7 +152,8 @@ public abstract class EmbeddingProvider extends ProviderBase {
    */
   protected static boolean acceptsNvidiaDimensions(
       EmbeddingProvidersConfig.EmbeddingProviderConfig.ModelConfig modelConfig) {
-    return modelConfig.vectorDimension().filter(dimension -> dimension > 0).isEmpty();
+    return modelConfig.parameters().stream()
+        .anyMatch(parameter -> parameter.name().equals("vectorDimension"));
   }
 
   /**
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/EmbeddingProvider.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/EmbeddingProvider.java
@@ -146,10 +146,7 @@ public abstract class EmbeddingProvider extends ProviderBase {
     return modelName.endsWith("titan-embed-text-v2:0");
   }
 
-  /**
-   * Helper method that has logic wrt whether an Nvidia model accepts {@code "dimensions"} parameter
-   * or not.
-   */
+  /** {@code NV-Embed-QA} and {@code nvidia/nv-embedqa-e5-v5} do not accept {@code dimensions}. */
   protected static boolean acceptsNvidiaDimensions(
       EmbeddingProvidersConfig.EmbeddingProviderConfig.ModelConfig modelConfig) {
     return modelConfig.parameters().stream()
--- a/src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/EmbeddingProvider.java
+++ b/src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/EmbeddingProvider.java
@@ -147,6 +147,16 @@ public abstract class EmbeddingProvider extends ProviderBase {
   }
 
   /**
+   * Helper method that has logic wrt whether an Nvidia model accepts {@code "dimensions"} parameter
+   * or not: only models without a fixed {@code vector-dimension} (i.e. with a {@code
+   * vectorDimension} parameter, like {@code nvidia/llama-3.2-nv-embedqa-1b-v2}) do.
+   */
+  protected static boolean acceptsNvidiaDimensions(
+      EmbeddingProvidersConfig.EmbeddingProviderConfig.ModelConfig modelConfig) {
+    return modelConfig.vectorDimension().filter(dimension -> dimension > 0).isEmpty();
+  }
+
+  /**
    * Replace parameters in a messageTemplate string with values from a map: placeholders are of form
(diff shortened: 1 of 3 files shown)


Hazel-Datastax (commented):
The fix is not generalized enough  - it can only fix nvidia's problem but not others. I think the key problem is this line https://github.com/stargate/data-api/blob/main/src/main/java/io/stargate/sgv2/jsonapi/service/resolver/VectorizeConfigValidator.java#L355 and specifically `model.vectorDimension().get() != 0`

it should't go inside this if block and the range dimension should be handled here: https://github.com/stargate/data-api/blob/main/src/main/java/io/stargate/sgv2/jsonapi/service/resolver/VectorizeConfigValidator.java#L372 and add some tests here: https://github.com/stargate/data-api/blob/main/src/test/java/io/stargate/sgv2/jsonapi/api/v1/CreateCollectionIntegrationTest.java#L734

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/EmbeddingProvider.java
```diff
@@ -146,6 +146,16 @@ protected static boolean acceptsTitanAIDimensions(String modelName) {
...
   }
 
+  /**
+   * Returns whether an Nvidia model accepts {@code dimensions}; {@code NV-Embed-QA} and {@code
+   * nvidia/nv-embedqa-e5-v5} do not.
+   */
+  protected static boolean acceptsNvidiaDimensions(
+      EmbeddingProvidersConfig.EmbeddingProviderConfig.ModelConfig modelConfig) {
+    return modelConfig.parameters().stream()
+        .anyMatch(parameter -> parameter.name().equals("vectorDimension"));
```
Hazel-Datastax, line 156: Could we explain in java doc why `parameter -> parameter.name().equals("vectorDimension")` could decide whether models accept dimensions or not

Hazel-Datastax (approved):

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/EmbeddingProvider.java
erichare: Updated the comment!

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/EmbeddingProvider.java
```diff
@@ -146,6 +146,16 @@ protected static boolean acceptsTitanAIDimensions(String modelName) {
     return modelName.endsWith("titan-embed-text-v2:0");
   }
 
+  /**
+   * Returns whether an Nvidia model accepts {@code dimensions}; {@code NV-Embed-QA} and {@code
+   * nvidia/nv-embedqa-e5-v5} do not.
+   */
+  protected static boolean acceptsNvidiaDimensions(
```
amorton, line 153: this is a base class for all models, why is the name nvidia ?

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/NvidiaEmbeddingProvider.java
```diff
@@ -38,12 +39,13 @@ public NvidiaEmbeddingProvider(
...
       int dimension,
       Map<String, Object> vectorizeServiceParameters) {
+    // "dimensions" is only sent for variable-dimension models, 0 omits it from the request
     super(
         ModelProvider.NVIDIA,
         providerConfig,
         modelConfig,
         serviceConfig,
-        dimension,
+        acceptsNvidiaDimensions(modelConfig) ? dimension : 0,
```
amorton, line 48: -1 for dimension = 0 being a magic setting, where is the code that does this ?

amorton (commented):
-1 

I dont think this should have been merged. It added nvidia specific features to a base class, and relies on a magic number for the dimensions

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/EmbeddingProvider.java
erichare: See the rest of the file for other examples. This PR just followed the convention of, e.g., `acceptsOpenAIDimensions` which predates this PR.

src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/NvidiaEmbeddingProvider.java
erichare: Fixed dimension models reject the dimensions field entirely, passing 0 is needed because of: https://github.com/stargate/data-api/blob/8644cea2447f18c3012af85b0567fee291a192ed/src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/NvidiaEmbeddingProvider.java#L145-L154
so Jackson will omit it when the value is 0, as intended.

I agree its not super readable though, would be better if we'd be able to pass null. but right now its a primitive

erichare: and we can update that, but again that was following pre-existing convention, e.g.:  https://github.com/stargate/data-api/blob/8644cea2447f18c3012af85b0567fee291a192ed/src/main/java/io/stargate/sgv2/jsonapi/service/embedding/operation/OpenAIEmbeddingProvider.java#L150-L153

amorton: yes, and that was bad code.

amorton:
This will break the embedding gateway unless we rebuild it, which we do not want to do. 

Will revert it

PR #2566, opened 2026-08-25
feat: reranking concurrency gate, total deadline, and 512-passage nemotron batches

erichare:
## What this does

PepsiCo needs 1000 simultaneous `findAndRerank` requests to succeed without errors (latency may grow). Today each request fans out `ceil(passages/10)` concurrent rerank POSTs with no concurrency limit anywhere, so a 1000-request burst is ~10,000 concurrent calls; the reranking NIM's internal queue blows past the read timeout and requests mass-fail with `RERANKING_PROVIDER_TIMEOUT`.

Queuing alone converts errors into latency but does not create throughput, so this PR combines three things:

### 1. Reranking concurrency gate (per provider+model, per pod)
- **Per-pod bulkhead**: at most `max-concurrent-calls` (default 32) rerank calls in flight across all requests; excess calls park in a bounded FIFO queue (`max-queued-calls`, default 1000). New non-blocking `RerankingConcurrencyGate` + `@ApplicationScoped` `RerankingConcurrencyGateRegistry` (providers are constructed per request, so shared state must live in a singleton).
- **Queue overflow fails fast** with the new `RERANKING_PROVIDER_OVERLOADED` error, without ever calling the provider — and it is deliberately never retried (retrying a saturation rejection deepens the overload).
- **Per-request fan-out cap**: `max-concurrent-batches` (default 8) via `Uni.join().usingConcurrencyOf()`.
- **Total deadline**: `total-timeout-millis` (default 30000) covers queue wait + all batches + retries. Expiry surfaces `RERANKING_PROVIDER_TIMEOUT`; parked work is abandoned before it reaches the provider. The embedding-gateway path inherits the gate and deadline through the `RerankingProvider` base class — previously the gRPC path had no client-side time bound at all.
- **Metrics** per provider+model: `rerank.all.queue.depth`, `rerank.all.inflight.calls` (gauges), `rerank.all.queue.wait.duration` (timer, parked grants), `rerank.all.queue.rejected.count` (counter).

### 2. nemotron rerank model with 512-passage batches
`nvidia/llama-3.2-nemoretriever-500m-rerank-v2` (`max-batch-size: 512`, non-default) collapses a typical findAndRerank from 10 rerank calls to 1 — the lever that makes the 1000-burst arithmetically feasible (PepsiCo's own 2-pod nemotron benchmark absorbed an 1800-request burst at 0% errors).

### 3. Config plumbing + handoff docs
- New properties flow through `RequestProperties`, the yaml schema, and `embedding_gateway.proto` (`optional` fields 7–10). Unset or out-of-range gateway values fall back to the Data API defaults, never zero, so the Data API can deploy before EGW.
- `CONFIGURATION.md` gains a Reranking configuration section (there was none).
- `docs/reranking-nim-capacity-recommendations.md`: gpu-helm-charts changes (nemotron NIM deployment, ingress route, `NIM_SERVER_MAX_QUEUE_SIZE` / `NIM_SERVER_REQUEST_TIMEOUT_S`, replica sizing formula).
- `docs/reranking-egw-coordination.md`: EGW proto mirror + serving nemotron with batch 512 for Astra.

## Behavior changes to note
- Enabled by default with generous limits: steady-state traffic never queues (well below 32 in-flight/pod); only true bursts change behavior, which is the point.
- Requests with >8 batches (>80 passages on the batch-10 model) serialize past the fan-out cap — bounded latency increase, no errors.
- New terminal error code `RERANKING_PROVIDER_OVERLOADED` (clients should treat as retry-later).
- The EGW path gains a client-side total deadline (intentional gap closure).

## Test plan
- [x] `RerankingConcurrencyGateTest` (9): cap enforcement incl. multi-threaded contention, FIFO order, no barging, overflow fail-fast (work never invoked), cancel-while-parked frees the slot, release on failure/cancellation, meter tracking
- [x] `RerankingProviderTest` (+6): fan-out cap with correct aggregation, shared gate across provider instances, deadline-while-parked (provider never called, queue drained), deadline mid-call releases permit, overflow surfaced to caller, OVERLOADED never retried / TIMEOUT still retried
- [x] `RerankingProvidersConfigTest` / `...OverrideTest` / `RerankingProviderConfigProducerTest`: yaml defaults, per-deployment overrides, EGW proto pass-through + fallback semantics (incl. explicit `max_queued_calls=0` honored)
- [x] `AllErrorCodesLoadTest` covers the new error template
- [x] Full unit suite: 3487 tests, 0 failures
- [ ] Integration test run (DSE via colima) — `FindRerankingProvidersIntegrationTest` expectations updated for the new model entry
- [ ] Burst load test on the dev GPU plane (rerank-loadtest, 1000-request burst): expect queueing not rejection at defaults, `OVERLOADED` only beyond the configured envelope
- [ ] Confirm nemotron ingress route + model id against the gpu-helm-charts deployment before cutover (route is projected; documented in the capacity recommendations doc)

### DIFF
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -81,3 +81,25 @@ Other Quarkus properties that are specifically relevant for the service:
 | Property                        | Type      | Default | Description                                                                                                         |
 |---------------------------------|-----------|---------|---------------------------------------------------------------------------------------------------------------------|
 | `stargate.feature.flags.tables` | `boolean` | `true` (enabled by default)   | Setting it to `true` enables Tables functionality; `false` disables; leaving as `null` uses the default (enabled).|
+
+## Reranking configuration
+*Configuration for reranking providers and the reranking concurrency gate, defined by [RerankingProvidersConfig.java](src/main/java/io/stargate/sgv2/jsonapi/service/reranking/configuration/RerankingProvidersConfig.java). Providers and models are loaded from `reranking-providers-config.yaml` (overridable with the `RERANKING_CONFIG_PATH` env var or the `RERANKING_CONFIG_RESOURCE` system property); when the embedding gateway is enabled the provider/model list is served by the gateway instead.*
+
+Each model under `stargate.jsonapi.reranking.providers.<provider>.models[]` supports the following `properties`:
+
+| Property                 | Type     | Default | Description                                                                                                                                                                                                            |
+|--------------------------|----------|---------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `at-most-retries`        | `int`    | `3`     | Maximum attempts per batch call before failing (1 request + 2 retries).                                                                                                                                                |
+| `initial-back-off-millis`| `int`    | `100`   | Initial retry delay, doubling up to `max-back-off-millis`.                                                                                                                                                             |
+| `read-timeout-millis`    | `int`    | `5000`  | HTTP read timeout for a single batch call.                                                                                                                                                                             |
+| `max-back-off-millis`    | `int`    | `500`   | Maximum retry delay.                                                                                                                                                                                                   |
+| `jitter`                 | `double` | `0.5`   | Random variation applied to retry delays.                                                                                                                                                                              |
+| `max-batch-size`         | `int`    | (none)  | Maximum passages per reranking call; a rerank request fans out into `ceil(passages / max-batch-size)` calls.                                                                                                            |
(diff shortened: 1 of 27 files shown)


erichare:
Closing — this bundles too much for one review. Splitting into smaller PRs: (1) the nemotron 512-passage model entry, (2) a total reranking deadline, (3) the concurrency gate. Branch feat/reranker-concurrency-gate is retained as the reference implementation.

PR #2578, opened 2026-09-03
Fix: #2030 - IllegalArgumentException when using findAndRerank without specifying rerankOn

sl-at-ibm:
**What this PR does**:

Until now, there are conditions where a call to `findAndRerank`  without supplying `rerankOn` results in a naked "unexpected server error" / "illegal argument exception". This PR replaces those with a clean error belonging to the `errors.yaml` taxonomy.

**Which issue(s) this PR fixes**:
Fixes #2030.

**Checklist**
- [x] Changes manually tested
- [x] Automated Tests added/updated
- [n/a] Documentation added/updated
- [?] CLA Signed: [DataStax CLA](https://cla.datastax.com/)

PR #2579, opened 2026-09-08
Issue 2577: enable lexical-only findAndRerank (i.e. no-vector/ize one-leg retrieval path)

### LINKED ISSUE #2577: IllegalArgumentException / UNEXPECTED_SERVER_ERROR in FindAndRerank when omitting $vectorize in $hybrid
A collection with lexical, rerank and vectorize support still may return a "naked" `Unexpected server error / IllegalArgumentException` when a findAndRerank command does not provide a $vectorize (rather only $lexical) in its sort clause and rerankQuery is provided.

Expectation: an error message that does not expose an internal exception, rather wraps it and rejects it gracefully at command-validation stage. (see current error below)

_Verified on 2026-09-03 on a prod Astra Data API on kube 189267260851 (AWS us-east-2)_

## Reproduction

_Note_: below, "bad query" means not that this does not succeed (which is the correct thing): it means that **the returned error looks like a unhandled internal "IllegalArgumentException" rather than a clean validation error**. When omitting `$vectorize` from the sort, but also skipping `rerankQuery`, the error is a good, clear validation error as it should be (==> `MISSING_RERANK_QUERY_TEXT`).

Create collection:

```
{
    "createCollection": {
        "name": "c",
        "options": {
            "vector": {
                "service": {
                    "provider": "nvidia",
                    "modelName": "nvidia/nv-embedqa-e5-v5"
                }
            }
        }
    }
}
```

Bad query 1:

```
{
    "findAndRerank": {
        "filter": {
            "$lexical": {
                "$match": "bla"
            }
        },
        "sort": {
            "$hybrid": {
                "$lexical": "ble"
            }
        },
        "options": {
            "rerankOn": "$lexical",
            "rerankQuery": "blo"
        }
    }
}
```

Bad query 2:

```
{
    "findAndRerank": {
        "filter": {
            "$lexical": {
                "$match": "bla"
            }
        },
        "sort": {
            "$hybrid": {
                "$lexical": "ble"
            }
        },
        "options": {
            "rerankQuery": "blo"
        }
    }
}
```

Naked error the caller receives:

```
Unexpected server error: An unexpected server error occurred while processing the request. 

Error Class: IllegalArgumentException
Error Message: buildVectorRead() - no vector or vectorize

Review the Error Message before retrying this request. (UNEXPECTED_SERVER_ERROR)
```

sl-at-ibm:
**What this PR does**:

Address #2577 _under the assumption that symmetry is preferrable_ (between vector/ize-only and lexical-only paths for f.A.R.R.)

In other words, rather than better dressing the error mentioned in the issue (no vector/ize provided => illegal arg exception), it accepts a findAndRerank with the lexical part only in $hybrid, executes that leg of the retrieval, and submits those results to the reranker.

_Note_: this PR is kept intentionally clear of fixing the absence of "rerankOn" (these two errors are close to each other). This is to avoid overlapping with #2578  which deals with that issue. If that other ever gets merged, it might be a good idea to add a test to this one for the case with lexical-only and _no rerankOn_.

Fixes #2577 

**Checklist**
- [x] Changes manually tested
- [n/a] Documentation added/updated
- [n/a] CLA Signed: [DataStax CLA](https://cla.datastax.com/)


amorton (changes requested):
-1 - the collection must have a vector to use hybrid. We are defining hybrid as vector + optional other legs.

sl-at-ibm:
> -1 - the collection must have a vector to use hybrid. We are defining hybrid as vector + optional other legs.

ACK, makes sense. Thanks for taking the time to review this one.
In light of this tenet (hybrid must have a vector leg), I am about to close this one and have rather decided to work from scratch on [this other PR](https://github.com/stargate/data-api/pull/2584), that improves on the error messages (and the validation logic and priorities) by respecting the must-have-vector/ize tenet.

amorton:
closing per discussions above

PR #2584, opened 2026-09-21
Fix: #2577 - IllegalArgumentException when using findAndRerank without specifying $vector/$vectorize

### LINKED ISSUE #2577: IllegalArgumentException / UNEXPECTED_SERVER_ERROR in FindAndRerank when omitting $vectorize in $hybrid

sl-at-ibm:
**What this PR does**:

Replaces a naked `IllegalArgumentException / Unexpected server error` occurring when a findAndRerank command lacks a prescription on how to run the vector leg, explicitly or implicitly (i.e. no `$vector/$vectorize` in the sort clause and no shorthand `$hybrid: <string>` form for sort).

Fixes #2577 .

_Notes_:

This PR replaces #2579 , which is going to be closed after discussion (that one approached this problem by _enabling_ a no-vector/ize findAndRerank, which was later clarified to not be in scope).

Note also that this PR does not simply replace the generic naked error with throwing the user-friendly new exception added to `errors.yaml`. Rather, it moves such validation up in the chain, so that the missing vector prescription is detected and thrown first, before "secondary" problems (such as, missing rerankOn) are detected. This, I believe, improves the experience of a developer approaching this command.

As a last point, I have added unit tests around the various findAndRerank call pattern that are supposed to trigger this error, as well as a few cases where the command should succeed. Given the intricacies of this API command, and the fact that clarifications were needed a few times around its exact behaviour, I deemed it reasonable to try and have the test be more prescriptive about the expected constraints of this command.

**Checklist**
- [x] Changes manually tested
- [n/a] Documentation added/updated
- [n/a] CLA Signed: [DataStax CLA](https://cla.datastax.com/)


src/main/java/io/stargate/sgv2/jsonapi/exception/RequestException.java
```diff
@@ -56,6 +56,7 @@ public enum Code implements ErrorCode<RequestException> {
     INVALID_CREATE_COLLECTION_FIELD,
     INVALID_RERANK_OVERRIDE,
     MISSING_RERANK_QUERY_TEXT,
+    MISSING_VECTOR_OR_VECTORIZE_IN_HYBRID_SORT,
```
amorton, line 59: this is an error in the sort clause, we should put the error in that class. Will do a longer review later today

